Issue #929, opened 2012-05-11
Allow specification of noac option in client mount of NFS filesystem


emslade:
I've had issues where files were not updated immediately on the VM after editing on the host. There may be alternatives but I found mounting on the VM with the noac option fixed this.

Issue #844, opened 2012-03-29
Set NFS options


halcyonCorsair:
What i want:
/home/export/web/aegir/sitedata 33.33.33.25(rw,no_subtree_check,no_root_squash,sync,fsid=ee65cda9794c037597989e8f575c063a)

The default:
/home/export/web/aegir/sitedata 33.33.33.26(rw,no_subtree_check,all_squash,anonuid=1000,anongid=33,fsid=ee65cda9794c037597989e8f575c063a

Differences:
use: no_root_squash
optionally: sync
don't use: anonuid, anongid

Is there a way for me to set this now?  I have vagrant installed via gem.

PR #1029, opened 2012-07-17
Add ability to specify arbitrary NFS mount options per share

dansimau:
Additional NFS mount options are specified as an array in the Vagrantfile and they are concatenated and passed to the mount command in the guest.

This solution would address #844 and #929.


habitullence:
Does anyone know where docs are for this? Googled to death!

jfbibeau:
Agreed. Just made PR #5617 to add the documentation for these options.

habitullence:
@jfbibeau legend!

PR #1041, opened 2012-07-25
Added support for downloading boxes via SCP

vaimo-wilko:
Since we have certain sensitive box files that we do not want on the open internet, we needed a way to add user authentication/authorization to box downloads.

We already have a good SSH key authentication framework in our infrastructure, so by using SCP we avoid having to set up a parallel authentication system via HTTP(S). With SCP we effectively get authentication/authorization for free.

I am contributing this, as this could be nice to have as a standard feature in Vagrant, for others to use too.

---

There is one thing currently not as I would like it to be, due to a bug/misfeature in Net::SSH. Unlike the command line SSH client, Net::SSH does not fall back to the currently logged in user, if no user is specified (either to the Net::SSH.start function or in the ssh config files).

This means that SCP URIs must be given a user (e.g. scp://user@host/path/to/file), if no user is specified in the ssh config file for the specific host.

I will submit a fix for this bug/misfeature to the Net::SSH project.

### DIFF
--- a/lib/vagrant/action/box/download.rb
+++ b/lib/vagrant/action/box/download.rb
@@ -12,7 +12,7 @@ module Vagrant
           @app = app
           @env = env
           @env["download.classes"] ||= []
-          @env["download.classes"] += [Downloaders::HTTP, Downloaders::File]
+          @env["download.classes"] += [Downloaders::HTTP, Downloaders::SCP, Downloaders::File]
           @downloader = nil
         end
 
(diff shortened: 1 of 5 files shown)


gsaslis:
in case someone else is also struggling, it took me a while to realize curl on OSX doesn't support scp out of the box (so, neither does vagrant). Check this URL for a way round this problem: http://chase-seibert.github.io/blog/2014/05/18/vagrant-authenticated-private-box-urls.html

PR #1712, opened 2013-05-10
Allow user to define if the box_url is insecure

igorbonadio:
Currently, it's possible to download a box from an insecure server using the command:

```
vagrant box add mybox  https://insecureserver.com/mybox.box --insecure
```

But I think set it in a Vagrantfile is a good feature.

``` ruby
  config.vm.box = "mybox"
  config.vm.box_url = "https://insecureserver.com/mybox.box"
  config.vm.box_download_insecure = true
end
```

### DIFF
--- a/lib/vagrant/action/builtin/handle_box_url.rb
+++ b/lib/vagrant/action/builtin/handle_box_url.rb
@@ -29,6 +29,7 @@ module Vagrant
             # raise a terrible runtime error.
             box_name = env[:machine].config.vm.box
             box_url  = env[:machine].config.vm.box_url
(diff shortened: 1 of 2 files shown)


dragon788:
@mitchellh @sethvargo Since it is a terrible practice to disable SSL verification long term, would it be possible to add instructions on correcting the certificate issue the right way by catching this certificate error exception and giving a helpful suggestion,  i.e. adding the certificate to the trust chain of the embedded Ruby and curl or better yet using the alternate CA path that was added to a newer Vagrant version? `config.vm.box_download_ca_cert` appears to be the new setting.

PR #2013, opened 2013-07-31
Initial commit of vagrant man page

talexb:
I've created a first cut of a man page for vagrant. I've left it in a top-level directory and haven't done anything about actually deploying it -- since I haven't learned about that yet.


phil-s:
As this issue is a couple of years old now, I'd suggest that a stub man page which at least explains how to use `vagrant help` and points to the online documentation would still be an improvement on no man page at all?

It's nice to type `man COMMAND` and get *something* useful.

talexb:
Wow -- it's neat to see this work spring back to life. I agree, something small is better than nothing at all. Let me know if I can help.

PR #2078, opened 2013-08-24
Support solaris host OS

miurahr:
Here is an experimental support code for solaris host os.
It's necessary that  solaris users test it and feedback for better code.

I'm not user so the implementation is based on documentation.

Signed-off-by: Hiroshi Miura miurahr@linux.com

### DIFF
--- a/plugins/hosts/solaris/host.rb
+++ b/plugins/hosts/solaris/host.rb
@@ -1,0 +1,78 @@
+require 'log4r'
+
+require "vagrant"
+require 'vagrant/util/platform'
+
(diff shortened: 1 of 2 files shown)


nmrony:
What is the status of this pull request? @mitchellh is there any chance to merge it to main project?

sethvargo:
@nmrony please read Mitchell's comment - we think this would be better suited as a plugin first.

shafqatevo:
Hi @miurahr was this plugin tested and works with sufficient stability?

Issue #2180, opened 2013-09-10
Vagrant passes ansible extra_vars through to modules that don't need them


mjackson:
Upgraded from vagrant 1.2.7 => 1.3.1. this morning and provisioning (using ansible) now fails with:

```
TASK: [download node] ********************************************************* 
failed: [192.168.111.222] => {"failed": true, "item": ""}
msg: unsupported parameter for module: couchdb_bind_address
```

The `couchdb_bind_address` var is given in `ansible.extra_vars`. This particular task is using ansible's [get_url](http://www.ansibleworks.com/docs/modules.html#get-url) module to download node.js, which doesn't support (or need) this var, obviously. This worked in 1.2.7 but fails in 1.3.1.

One of two things is going on here. Either 1) I shouldn't be using `extra_vars` like this or 2) Vagrant shouldn't be passing all of my extra vars along to the `get_url` module.

Which is it? Specifically, what changed in 1.3 around the handling of `extra_vars`?

/cc @commandtab

Issue #2174, opened 2013-09-10
Parsing of ansible extra_vars


trkoch:
Given a `Vagrantfile` that contains

```
extra_vars = { user: "vagrant", foo: "bar" }
ansible.extra_vars = extra_vars
```

Running `vagrant provision` results in following error thrown by `ansible-playbook`:

```
command-line line 0: garbage at end of line; "local".
```

This can be worked around by passing raw arguments, like this:

```
ansible.raw_arguments = "--extra-vars=" + extra_vars.map { |k,v| "#{k}=#{v}" }.join(" ")
```

This defect was introduced somewhere between 1.2.7 and 1.3.0. Platform is Mac OS X, Ruby 1.9.3.

Issue #2128, opened 2013-09-05
Support Ansible 1.2 JSON format for --extra_vars


ndemoor:
To allow the usage of more complex extra vars input (hashes and arrays) the JSON format needs to be supported. This was introduced in Ansible v1.2.

http://www.ansibleworks.com/docs/playbooks2.html#id10

PR #2366, opened 2013-10-11
Ansible: Improvements for 'extra_vars' argument

gildegoma:
`extra_vars` argument can now:
- contains a hierarchical set of parameters (instead of flat Key-Value Pairs), **ALREADY** implemented in #2131.
- alternatively refers to a configuration file (instead of embedded Hash)

See Ansible Documentation for version requirement details:
http://www.ansibleworks.com/docs/playbooks_variables.html#passing-variables-on-the-command-line

**Related Issues**
- #2180
- #2131
- #2174
- #2128


ruudk:
Really love this addition.

The only problem is that only JSON files should be started with `@`. When you reference a YAML file it should _NOT_ be prefixed with an `@`

See: http://docs.ansible.com/playbooks_variables.html#passing-variables-on-the-command-line

Issue #2612, opened 2013-12-09
Plugin dependencies are pruned too agressively


tmatilai:
#2574 loosened net-ssh requirement and thus a new version is bundled with installer. Still some plugins (at least vagrant-berkshelf) requires `~> 2.6.6`. It gets installed fine, but then later uninstalled by `PruneGems` action. And then plugin loading crashes at runtime.

Issue #2428, opened 2013-10-28
Check for required plugins, including minimum version


ringods:
The modularization of Vagrant is very handy, but a lot of support issues stem from plugins not being installed or with a wrong version, leading to questions like in #2342.

Similar to #2322, it would be handy if the Vagrantfile could have a section where the required plugins, including a version specification, could be listed. Vagrant could then check if the correct set of plugins is available before running.

I like fail-fast solutions. :-)

Issue #2406, opened 2013-10-22
Pre-release gems are not found with `vagrant plugin install <name> --plugin-prerelease --plugin-version=<ver>`


jfoy:
Vagrant 1.3.3, Mac OS 10.8.5 

The symptom is as follows:

```
$ gem list -r vagrant-chefzero --pre

*** REMOTE GEMS ***

vagrant-chefzero (1.0.1.pre0)
$ vagrant plugin install vagrant-chefzero --plugin-prerelease --plugin-version 1.0.1.pre0
Installing the 'vagrant-chefzero --prerelease --version '1.0.1.pre0'' plugin. This can take a few minutes...
The plugin 'vagrant-chefzero' could not be found in local or remote
repositories. Please check the name of the plugin and try again.
```

It looks like the Rubygems client doesn't recover gracefully from a missing prerelease index:

```
bad response Not Found  404 (http://gems.hashicorp.com/prerelease_specs.4.8.gz)
at /Applications/Vagrant/embedded/lib/ruby/1.9.1/rubygems/spec_fetcher.rb line 231
```

What is serving the gems at gems.hashicorp.com?

PR #2769, opened 2014-01-06
Plugin dependency management revamp

mitchellh:
This is a huge revamp of how plugin dependency management is done. To understand the changes here, a brief history lesson is in order:

Since Vagrant 1.1, plugins have been loaded as RubyGems. Once Vagrant was loaded, it would iterate through a list of installed plugins, and `require` that plugin. This mostly worked okay. But the devil is in the details, and the edge cases were _really_ bad. In addition to the edge cases (mentioned below), building things like updaters, version constraints (">= 1.0", "< 1.1"), etc. all had to be done manually. This seemed silly, since RubyGems itself (and Bundler) do these sort of things for you. Why reinvent the wheel?

As for edge cases: the primary edge case is that since the dependencies of Vagrant and its respective plugins weren't resolved as a whole, you can run into cases where plugin installation succeeded, but plugin loading failed because Vagrant already loaded a common dependency with the wrong version. An example explains this best:
- Vagrant depends on "A >= 1.0, < 1.2"
- vagrant-plugin depends on "A = 1.1"
- When you run Vagrant, it loads the latest possible matching dependencies, so it would load A 1.2
- When Vagrant loads vagrant-plugin, it can't load, because A 1.2 is active, so A 1.1 can't be loaded.

The error above should never happen: the versions available for A should satisfy both Vagrant and vagrant-plugin (by loading v1.1 for both). 

With this new branch, all plugin installation, dependency resolution, updating, etc. is managed by [Bundler](http://gembundler.com). This has yielded numerous benefits:
- Vagrant now resolves dependencies before Vagrant is even loaded. This ensures that all plugins will be able to load. No more conflicts at run-time.
- Conflicts are detected at `vagrant plugin install` time. This means that if there would be a crash if that plugin were to load, the plugin won't even install and a human-friendly error is shown to the end user.
- `vagrant plugin install` now accepts complex version constraints such as "~> 1.0.0" or ">= 1.0, < 1.1". Vagrant stores these constraints for updating, which leads to the next point.
- `vagrant plugin update` without arguments now updates all installed plugins, respecting the constraints specified by `vagrant plugin install`.
- `vagrant plugin update NAME` will only update that gem (still respecting constraints). 
- Internally, there are a lot more unit tests. /cc @phinze :)

The goal of this branch was to replace the _existing_ system and functionality with Bundler-ized management. It did not introduce any new features except where they naturally fell into place (version constraints). However, with this new system, many new possibilities are also available:
- Vagrant environment local plugins (i.e. a Gemfile but for a specific Vagrant environment). 
- Plugin installation from git

I'm sure those will be pursued at some point in the future.

This fixes: #2612, #2406, #2428

### DIFF
--- a/lib/vagrant.rb
+++ b/lib/vagrant.rb
@@ -1,8 +1,6 @@
 # This file is load before RubyGems are loaded, and allow us to actually
 # resolve plugin dependencies and load the proper versions of everything.
 
-ENV["VAGRANT_INSTALLER_EMBEDDED_DIR"] = "/Applications/Vagrant/embedded"
-
 require "vagrant/shared_helpers"
 
 if Vagrant.plugins_enabled? && !defined?(Bundler)


mikeschinkel:
Hi @mitchellh:

Since it's almost 2 years later, I am wondering if you guys have looked into a way to handle this?

I am currently working on a [Vagrant box for local WordPress development](/wplib/wplib-box) and we want to minimize the steps people need to talk to get our box working.  Currently we need to tell them to install VirtualBox then Vagrant then a Host Updater plugin then a Triggers plugin.  

We would love if we could at least minimize to just VirtualBox then Vagrant and not have to specify plugins but that we could instead specify within the `Vagrantfile` which plugins need to be installed, but having to install a plugin in order to avoid installing plugins is ironically at cross purposes.

Here is what I'd love to see, commands something like this for the `Vagrantfile`:

        config.plugins.auto_install = true;
        config.plugins.add "cogitatio/vagrant-hostsupdater";
        config.plugins.add "emyl/vagrant-triggers";
        ...
    end

I am hoping that 2 years later it is finally time for Hashicorp to tackle this...

ciastek:
NOTE: As of version 1.9, Vagrant doesn't use Bundler anymore:  [#7793 [core] Remove bundler usage for plugin management](https://github.com/hashicorp/vagrant/pull/7793)

PR #2870, opened 2014-01-21
Adding support for gateway in RHEL/CentOS ifcfg script

myoung34:
CentOS/RHEL should be able to set GATEWAY in ifcfg-#device for static routes.

### DIFF
--- a/templates/guests/redhat/network_static.erb
+++ b/templates/guests/redhat/network_static.erb
@@ -6,5 +6,6 @@ ONBOOT=yes
 IPADDR=<%= options[:ip] %>
 NETMASK=<%= options[:netmask] %>
 DEVICE=eth<%= options[:interface] %>
+<%= options[:gateway] ? "GATEWAY=#{options[:gateway]}" : '' %>
 PEERDNS=no
 #VAGRANT-END


sethvargo:
Hi @myoung34 

Thank you for the PR - would you mind adding some documentation for this in the appropriate place in the `website/docs` folder in this repo?

It would also be great to see a test for this, but it does not look like there is existing test coverage. If you feel like taking a stab at adding tests, it would be greatly appreciated! :smile:

myoung34:
I have no problem adding documentation, but I'm unsure where. ``source/v2/networking/private_network.html.md`` seems like the most relevant place to put this, but I'm unsure on phrasing as this only pertains to redhat hosts and is not provider specific. How do you feel about within the *Static IP* section:

```
## Static IP

You can also specify a static IP address for the machine. This lets you
access the Vagrant managed machine using a static, known IP. The
Vagrantfile for a static IP looks like this (Note that gateway is optional and pertains only to RedHat hosts):

\`\`\`ruby
  config.vm.network "private_network", ip: "192.168.50.4", gateway: "192.168.50.1"
end
\`\`\`

```

myoung34:
Also, I seem to have deleted my upstream fork, so once we come up with the appropriate route i'll have to open up a new PR, which is no big deal, this one will just be closed and reference the new PR.

sethvargo:
Hi @myoung34 

I wrapped up your changes into a larger PR (https://github.com/mitchellh/vagrant/pull/5721) that adds support for `gateway` for a number of different operating systems.

Issue #1585, opened 2013-04-12
NFS synced folders don't work with multi-machine configurations


jcsp:
Vagrant 1.1.5 + vmware_fusion provider

I have a VagrantFile with several machines, and am using an NFS synced folder.  Only one of the machines can see at at a time though, because /etc/exports/ is being incorrectly updated.  Each time a 'vagrant up' a particular machine, /etc/exports/ is update for that machine only, causing the other machines to be denied access to the NFS share.

VagrantFile (transcribed by hand so may have glitches but it's the general idea):
config.vm.synced_folder ".", "/nvfs", :nfs => true

config.vm.define :server_1 do |server_1|
  server_1.vm.network :private_network, ip: "192.168.233.3"
end

config.vm.define :server_2 do |server_2|
  server_2.vm.network :private_network, ip: "192.168.233.4"
end

The resulting /etc/exports looks something like this (after doing a 'vagrant up', and both machines are running, only the second one to start is reflected in /etc/exports)
# VAGRANT-BEGIN:

"/Users/foo/vagrant" 192.168.233.4 -mapall=502:20
# VAGRANT- END:

PR #3076, opened 2014-03-07
NFS: Use exportfs instead of /etc/exports on Linux

p0deje:
Re-opened (after fixes and updates) #1695, which was closed due to misunderstanding.

Basically, this makes only one change - `exportfs` cli tool is used to export/unexport NFS filesystems. It is better because:
1. Current implementation (which echoes exports to `/etc/exports` file) does not scale well. Specifically, I've been getting collisions in exports file with parallel startup of 10 Vagrant boxes (i.e. when 10 Vagrant processes start echoing to file, it becomes a mess).
2. It does not pollute `/etc/exports` file, which may be used by user for real persistent exports.
3. Each export/unexport operations has no effect on other active exports. Currently, if user used `exportfs` cli tool to temporarily export NFS filesystem, it will be removed with `vagrant up` (as it cares only about records in `/etc/exports`)

`exportfs` is available on Linux, so we can use it there. It's 100% unavailable on OS X. I am not sure about *BSD systems.

This is how it looks now:

```
vagrant up
```

screen shot 2014-03-30 at 14 39 37

```
vagrant destroy
```

screen shot 2014-03-07 at 20 40 46

With this patch, I was able to successfully use Vagrant for parallel running of 10 KVM instances in our CI. It may also be the fix for #1585


joelmoss:
Is there a reason why this has not been merged yet? I see all known issues are resolved, but is there something else missing?

plugins/hosts/linux/cap/nfs.rb
```diff
@@ -114,13 +90,50 @@ def self.nfs_opts_setup(folders)
...
+            opts[:linux__nfs_options] << "replicas=#{opts[:uuid]}"
           end
         end
 
         def self.nfs_running?(check_command)
           system(check_command)
         end
+
+        def self.nfs_current_exports
+          exports = `sudo exportfs -v`
```
sethvargo: This runs locally on the host machine and will fail on any host that does not support the `exportfs` command.

sethvargo: We also have a subprocess library that we prefer to use as opposed to Ruby's backticks

plugins/hosts/linux/plugin.rb
```diff
@@ -26,13 +26,6 @@ class Plugin < Vagrant.plugin("2")
         Cap::NFS
       end
 
-      # Linux-specific helpers we need to determine paths that can
-      # be overriden.
-      host_capability("linux", "nfs_apply_command") do
-        require_relative "cap/nfs"
-        Cap::NFS
-      end
```
sethvargo: Why was this removed?

plugins/synced_folders/nfs/synced_folder.rb
```diff
@@ -99,9 +99,13 @@ def prepare_folder(machine, opts)
...
         opts[:nfs_version] ||= 3
 
+        # We need to store machine UUID so we could detect
+        # no longer valid exports on prune action.
+        opts[:uuid] = machine.id
+
         # We use a CRC32 to generate a 32-bit checksum so that the
         # fsid is compatible with both old and new kernels.
-        opts[:uuid] = Zlib.crc32(opts[:hostpath]).to_s
+        opts[:fsid] = Zlib.crc32(opts[:hostpath]).to_s
```
sethvargo: These two changes seem unrelated to the original PR. Can you please explain them?

plugins/hosts/linux/cap/nfs.rb
```diff
@@ -114,13 +90,50 @@ def self.nfs_opts_setup(folders)
...
+          #
+          # We need to properly handle both cases
+          exports = exports.split("\n")      # each new line is either folder or ip
+          exports.map! { |e| e.split("\t") } # tab prepends ip
+          exports.flatten!                   # make one-dimensional array
+          exports.select! { |e| !e.empty? }  # remove all empty strings created by splitting with long entries
+          exports.map!(&:strip)              # make sure there are no leading spaces
+
+          folders = exports.each_slice(2).map(&:first) # array of host paths
+          opts    = exports.each_slice(2).map(&:last)  # array of IP address and options
```
sethvargo: This whole block of code seems incredibly fragile to me. Can we use a regular expression or cleaner function instead? Thank you for putting comments, but in my experience, if you need this many comments to explain a Ruby function, it's an indication that it needs a refactor.

plugins/hosts/linux/cap/nfs.rb
```diff
@@ -21,32 +17,22 @@ def self.nfs_start_command(env)
...
-          output = Vagrant::Util::TemplateRenderer.render('nfs/exports_linux',
-                                           :uuid => id,
-                                           :ips => ips,
-                                           :folders => folders,
-                                           :user => Process.uid)
 
           ui.info I18n.t("vagrant.hosts.linux.nfs_export")
           sleep 0.5
 
-          nfs_cleanup(id)
```
sethvargo: I don't think we should remove code that isn't causing problems. Unless this was causing a bug for you, I would prefer we keep it in :smile:

sethvargo:
Hi @p0deje 

Thank you for the Pull Request, and I am sorry it has taken so long to review. I just did a preliminary review and my initial thoughts are "how is this simpler than the existing solution?". I understand you have issues with converging multiple VMs in parallel, but I think we can solve that problem independently without refactoring the way we handle exports (like adding a lock).

It is very difficult to follow the current solution (even with all the comments), which is leaning me towards a :-1: on the overall review.The `linux` cap is also a _very_ broad capability, and I am not sure `exportfs` exists on _all_ linux distros.

I made a few other comments inline, but overall I am not convinced this is the direction we should be heading for managing exports. I could be convinced otherwise though :smile:.

plugins/hosts/linux/cap/nfs.rb
p0deje: Agree, this can be refactored to use regular expressions instead.

plugins/hosts/linux/plugin.rb
p0deje: Because NFS export and apply are no longer separate actions. Everything is now done `nfs_export` capability.

plugins/hosts/linux/cap/nfs.rb
p0deje: > This runs locally on the host machine and will fail on any host that does not support the exportfs command.

We already depend on `exportfs` command with `nfs_apply_command`.

> We also have a subprocess library that we prefer to use as opposed to Ruby's backticks

Original implementation used `#system` but I needed command stdout so I didn't change it subprocess and used backticks. It can be refactored though.

plugins/synced_folders/nfs/synced_folder.rb
p0deje: `opts[:uuid]` was used as paramter of `fsid` option in NFS export, so I renamed it for consistency.

`opts[:uuid]` instead is actually and machine UUID. Please, see this [comment](https://github.com/mitchellh/vagrant/pull/3076#commitcomment-5844884) for details.

plugins/hosts/linux/cap/nfs.rb
p0deje: It was not causing a bug, but since exports are now purely atomic, there was no need to perform exports cleanup.

p0deje:
> Thank you for the Pull Request, and I am sorry it has taken so long to review.

No worries about that, thanks for getting to it.

> I just did a preliminary review and my initial thoughts are "how is this simpler than the existing solution?".

It's simpler because we just execute a unix command and that's it - no race conditions (and thus no locks) happen on editing `/etc/exports`, no need to handle existing entries in `/etc/exports` file. Moreover, if user wants to temporarily export NFS folder manually by using `exportfs` command and without putting it to `/etc/exports`, Vagrant will drop them because it re-exports everything in `/etc/exports`. So, exports on Linux become purely atomic now.

I understand you have issues with converging multiple VMs in parallel, but I think we can solve that problem independently without refactoring the way we handle exports (like adding a lock).

It is very difficult to follow the current solution (even with all the comments), which is leaning me towards a :-1: on the overall review.The linux cap is also a very broad capability, and I am not sure exportfs exists on all linux distros.

I made a few other comments inline, but overall I am not convinced this is the direction we should be heading for managing exports. I could be convinced otherwise though :smile:.

> It is very difficult to follow the current solution (even with all the comments), which is leaning me towards a :-1: on the overall review.

Well, the part that handled stdout from exportfs command definitely can be refactored, but everything else is no more complicated than current code. What exactly you have problems with?

> The linux cap is also a very broad capability, and I am not sure exportfs exists on all linux distros.

We already depend on `exportfs` in `nfs_apply_command` action.

p0deje:
I have squashed commits and rebased it from master.

p0deje:
Refactored to use regexp to parse `exportfs` output.

christopher-hopper:
I'm now following this issue / pull request as I've just started experiencing the same problems with nfs exports. 

Here's the error report and debug info: [vagrant 1.7.3 nfs synced folders error ](https://gist.github.com/christopher-hopper/ccef04464a5c1f13d44f).

christopher-hopper:
After updating to [vagrant 1.7.4, from 1.7.3,](https://github.com/mitchellh/vagrant/blob/v1.7.4/CHANGELOG.md) all problems with errors and failures writing to `/etc/exports` have been resolved for me. I'm using Ubuntu 14.04 LTS and vagrant 1.7.4. 

Perhaps we don't need the change proposed here to [fix the vagrant synced folders errors I experienced](https://gist.github.com/christopher-hopper/ccef04464a5c1f13d44f).

chrisroberts:
Started with the changeset in this PR and continued the implementation here: #8432. Closing this up in favor of the new PR. Thank you for the original changeset!

PR #3138, opened 2014-03-11
Set Salt default config_dir

cournape:
This adds a config_dir settings to the salt provisioner, with a 'sensible' default to /etc or C:\\etc on windows. This is then used to upload the minion file at the right place.

TBH, I think the current configuration setup of the vagrant salt plugin is a bit clumsy. What about adding this instead:

```
# Vagrantfile

      salt.masterless = true # default false ?
      salt.minion_id = 'something' # default is using salt default

 end
```

This way, people don't need to care about how to actually set up masterless at the minion level, and translating this config into actual salt call is simple (e.g. salt-call ... --local --id=$id). Thoughts ? Would be happy to prepare a PR toward this if this makes sense to vagrant devs.

### DIFF
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -18,6 +18,7 @@ module VagrantPlugins
       attr_accessor :bootstrap_script
       attr_accessor :verbose
       attr_accessor :seed_master
+      attr_accessor :config_dir
       attr_reader   :pillar_data
 
       ## bootstrap options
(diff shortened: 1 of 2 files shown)


sethvargo:
Hey @cournape 

Have you had a chance to look at my comments? Please let me know if you have any questions.

sethvargo:
Merged in https://github.com/mitchellh/vagrant/commit/67c3819d298daf68cfcf0869c457f9909257019b

Issue #2970, opened 2014-02-13
Salt provisioner: minion_config not copied when internet unavailable


mafrosis:
I just ran into this edge case when testing in a new apartment without an internet connection.

Running `vagrant up` fails to copy the Salt minion config when an internet connection is unavailable. Connecting back to the internet (via my phone right now) resolves this. Odd.

`Vagrantfile`:

```
Vagrant.configure("2") do |config|
  config.vm.box = "wheezy64au"

  # salt source directory & shared dir in /tmp
  config.vm.synced_folder "..", "/srv/salt"
  config.vm.synced_folder "/tmp", "/tmp/host_machine"

  # setup the salt-minion
  config.vm.provision :salt do |salt|
    salt.minion_config = "salt-minion.conf"
    salt.run_highstate = false
  end
end
```

PR #3235, opened 2014-03-15
Salt masterless configuration

cournape:
This PR simplifies vagrant-salt configuration for masterless setup: instead of having people messing with the configuration files (which does not work ATM because of #2970), people can simply do the following:

      salt.masterless = true # false by default
      salt.minion_id = "base" # allows to override the id
    end


sethvargo:
Hey @cournape 

I was just wondering if you had a chance to rebase and add a test for this yet. Please let me know if you have any questions.

sethvargo:
Merged in f5654f3

PR #3303, opened 2014-03-24
Don't reorder config in /etc/network/interfaces

barrkel:
On Debian the network interface configuration is rewritten, with the Vagrant section appended, every reboot. But the network interface file is order dependent when additional virtual interfaces are added by installed packages (`ucarp` being my motivating example). Moving the Vagrant section to the end breaks virtual interfaces (like `ucarp`'s virtual IP).

This patch changes this so that it effectively rewrites the Vagrant section in `/etc/network/interfaces` in place rather than moving it to the end of the file.

If the section doesn't exist, it still gets appended as before.

### DIFF
--- a/plugins/guests/debian/cap/configure_networks.rb
+++ b/plugins/guests/debian/cap/configure_networks.rb
@@ -13,9 +13,8 @@ module VagrantPlugins
           machine.communicate.tap do |comm|
             # First, remove any previous network modifications
             # from the interface file.
-            comm.sudo("sed -e '/^#VAGRANT-BEGIN/,/^#VAGRANT-END/ d' /etc/network/interfaces > /tmp/vagrant-network-interfaces")
-            comm.sudo("su -c 'cat /tmp/vagrant-network-interfaces > /etc/network/interfaces'")
-            comm.sudo("rm -f /tmp/vagrant-network-interfaces")
+            comm.sudo("sed -e '/^#VAGRANT-BEGIN/,$ d' /etc/network/interfaces > /tmp/vagrant-network-interfaces.pre")
+            comm.sudo("sed -ne '/^#VAGRANT-END/,$ p' /etc/network/interfaces | tail -n +2 > /tmp/vagrant-network-interfaces.post")
 
             # Accumulate the configurations to add to the interfaces file as
             # well as what interfaces we're actually configuring since we use that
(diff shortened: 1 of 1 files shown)


excieve:
I believe this change breaks the interfaces setup if there's at least one private network configured in addition to the public one. After upgrading to 1.7.2 I couldn't successfully start any of my VMs (all have a private network). Here's the relevant output of `vagrant up`:
```
==> solr: Configuring and enabling network interfaces...

/sbin/ifup eth1

Stdout from the command:



Stderr from the command:

stdin: is not a tty
/etc/network/interfaces:28: interface eth2 declared allow-auto twice
/sbin/ifup: couldn't read interfaces file "/etc/network/interfaces"
```

I could SSH into the VM though (for the first boot only, following reboots borked my stunnel init in the VMs) and the `/etc/network/interfaces` contained additional `eth2` section. Removing this section manually made the VM start successfully but the next reboot generated this cloned record again. 

Here's the valid state of the `/etc/network/interfaces` (after the first boot):
```
# This file describes the network interfaces available on your system
# and how to activate them. For more information, see interfaces(5).

# The loopback network interface
auto lo
iface lo inet loopback

# The primary network interface
auto eth0
iface eth0 inet dhcp


#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth1
iface eth1 inet dhcp
    post-up route del default dev $IFACE || true
#VAGRANT-END

#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth2
iface eth2 inet static
      address 33.33.33.13
      netmask 255.255.255.0
#VAGRANT-END
```

Looks like the sed that generates `/tmp/vagrant-network-interfaces.post` is broken and always adds a last interface to it, here's what it outputs on a valid file:
```
$ sed -ne '/^#VAGRANT-END/,$ p' /etc/network/interfaces | tail -n +2

#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth2
iface eth2 inet static
      address 33.33.33.13
      netmask 255.255.255.0
#VAGRANT-END
```

So that interfaces looks like this after the second boot:
```
# This file describes the network interfaces available on your system
# and how to activate them. For more information, see interfaces(5).

# The loopback network interface
auto lo
iface lo inet loopback

# The primary network interface
auto eth0
iface eth0 inet dhcp


#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth1
iface eth1 inet dhcp
    post-up route del default dev $IFACE || true
#VAGRANT-END

#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth2
iface eth2 inet static
      address 33.33.33.13
      netmask 255.255.255.0
#VAGRANT-END

#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth2
iface eth2 inet static
      address 33.33.33.13
      netmask 255.255.255.0
#VAGRANT-END
```

Reverting to 1.7.1 fixed the issue for me.

clinstid:
I'm having the same problem. I think the issue is that the second sed command that's used to generate the post-vagrant section of the file assumes that there's only a single interface definition generated by vagrant, so anything after the first `#VAGRANT-END` is going to be concatenated to the end of the new interfaces file.

Perhaps a fix would be to remove the `#VAGRANT-BEGIN` and `#VAGRANT-END` lines from the template files and add them to the beginning and end of `entries` after all of the interface definitions have been added?

I have a fix for it, but I've never submitted a PR before. I could use some guidance on it.

hsoc:
This patch breaks /etc/network/interfaces with multiple interfaces. (e.g 2 public_networks).

mattolson:
I ran into the same issue, we had to revert to 1.7.1. See https://github.com/mitchellh/vagrant/pull/5325 for a proposed fix.

Issue #2642, opened 2013-12-12
Editing /etc/exports shouldn't require root and/or manually entering password everytime


purpleidea:
The NFS integration is a great feature, but between rebuilds my cached sudo password times out, and I have to re-enter it when doing an 'up' or even a 'destroy' because it modifies /etc/exports and related tasks via sudo.

I recommend that an /etc/sudoers.d/vagrant file be distributed with the package that allows people in a %vagrant group to not need a password for the relevant actions. If someone can suggest a different solution, I'd like to hear it too :)

If you accept this feature, I can perhaps write the patch, however it might require a few changes in plugins/hosts/linux/host.rb too. Other people who prefer to do this are welcome to!

Cheers,
James

PR #3638, opened 2014-05-05
Improve use of sudo for NFS export manipulation

jtopper:
Some users (myself and the denizens of #2642 included) want to allow Vagrant to manipulate NFS configs without requesting a sudo password each time.

The current approach, using "sudo -s" requires that we permit the whole shell to be run passwordless in sudoers, and that's rather too open to be safe.

This patch uses an alternative idiom, using 'tee' instead of output redirection to append to the exports file.  On application of this patch, the following sudoers config (on OSX, at least), permits fully passwordless operation in a much safer manner:

```
Cmnd_Alias VAGRANT_EXPORTS_ADD = /usr/bin/tee -a /etc/exports
Cmnd_Alias VAGRANT_NFSD = /sbin/nfsd restart
Cmnd_Alias VAGRANT_EXPORTS_REMOVE = /usr/bin/sed -E -e /*/ d -ibak /etc/exports
%admin ALL=(root) NOPASSWD: VAGRANT_EXPORTS_ADD, VAGRANT_NFSD, VAGRANT_EXPORTS_REMOVE
```

This feels like a good interim step if work on the suid helper has paused for now.

NB: I've been unable to test the patch to the Linux plugin at this time.

### DIFF
--- a/contrib/sudoers/linux
+++ b/contrib/sudoers/linux
@@ -1,0 +1,8 @@
+# These work with Ubuntu - they might need tweaking for other distributions
+
+Cmnd_Alias VAGRANT_EXPORTS_ADD = /usr/bin/tee -a /etc/exports
+Cmnd_Alias VAGRANT_NFSD_CHECK = /etc/init.d/nfs-kernel-server status
+Cmnd_Alias VAGRANT_NFSD_START = /etc/init.d/nfs-kernel-server start
+Cmnd_Alias VAGRANT_NFSD_APPLY = /usr/sbin/exportfs -ar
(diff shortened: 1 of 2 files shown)


Ramblurr:
FYI, for those on fedora the 2nd and third lines should be:

```
Cmnd_Alias VAGRANT_NFSD_CHECK = /usr/bin/systemctl status nfs-server
Cmnd_Alias VAGRANT_NFSD_START = /usr/bin/systemctl start nfs-server
```

Also I add my user to the vagrant group and change the last line to:

```
%vagrant ALL=(root) NOPASSWD: VAGRANT_EXPORTS_ADD, VAGRANT_NFSD_CHECK, VAGRANT_NFSD_START, VAGRANT_NFSD_APPLY, VAGRANT_EXPORTS_REMOVE
```

PR #4132, opened 2014-06-30
Add support for environment variable forwarding in config

### LINKED ISSUE #4131: Support environment variable forwarding
It would be nice if I could forward specific environment variables to the guest. This would required configuration on the guest box itself, but would be useful if I needed to know what user was running `vagrant ssh`, for example.

``` sh
skeggse@local: ~$ vagrant ssh
[vagrant@localhost ~]$ echo "$VAGRANT_USER"
skeggse
```

It would require the vagrant box to include something like this in `/etc/sudoers`, but in my case that's not difficult.

```
Defaults env_keep += "VAGRANT_USER"
```

This would also either require defining VAGRANT_USER in my local `.bashrc` or explicitly telling vagrant to forward specific environment variables as other variables in the `Vagrantfile`.

``` rb
  config.ssh.forward_env = { "VAGRANT_USER" => "USER" }
end
```

**EDIT:** also needs a line in `/etc/ssh/sshd_config`:

``` conf
AcceptEnv VAGRANT_USER
```

skeggse:
Fixes #4131.

This is pretty much my first time with ruby, please point out any dumb mistakes. Also I'm not super familiar with vagrant's architecture, so again point out anything I need to change.

### DIFF
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -390,6 +390,7 @@ module Vagrant
       # We also set some fields that are purely controlled by Varant
       info[:forward_agent] = @config.ssh.forward_agent
       info[:forward_x11]   = @config.ssh.forward_x11
+      info[:forward_env]   = @config.ssh.forward_env
 
       # Add in provided proxy command config
       info[:proxy_command] = @config.ssh.proxy_command if @config.ssh.proxy_command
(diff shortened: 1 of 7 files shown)


AnthonyMastrean:
Wow, this is a great looking pull request! I would love this feature, too. I'm running all of my builds in Vagrant machines. I would like to pass the build number/label from the server (an environment variable) into the VM.

My builds are basically this script...

```
trap 'vagrant destroy default --force' EXIT

vagrant up
vagrant ssh --command 'cd /vagrant && $ANT_HOME/bin/ant'
```

Now, I have to do some fancy quoting to get the `BUILD_NUMBER` in there

```
trap 'vagrant destroy default --force' EXIT

vagrant up
vagrant ssh --command 'cd /vagrant && $ANT_HOME/bin/ant -build.number='"'$BUILD_NUMBER'"''
```

PR #4133, opened 2014-07-01
Ensure VirtualBox virtual disks are stored in the correct directory on Windows hosts

kashook:
This change is a fix for https://github.com/mitchellh/vagrant/issues/3584.

### DIFF
--- a/plugins/providers/virtualbox/driver/version_4_2.rb
+++ b/plugins/providers/virtualbox/driver/version_4_2.rb
@@ -185,10 +185,13 @@ module VagrantPlugins
             disk_params << "--unit"
             disk_params << unit_num
             disk_params << "--disk"
-            if Gem.win_platform?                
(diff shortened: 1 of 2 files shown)


charliepank:
This bug still exists in vagrant version: 1.7.2

Lorenzo-Thomas:
If this was fixed in a 1.x.x version, then re-appears in a 2.x.x version (I am using Vagrant 2.2.3) would this be a regression issue?  I will try using an older version of Vagrant to see if the VMDK path issue was fixed then re-added.

sakisdog:
> If this was fixed in a 1.x.x version, then re-appears in a 2.x.x version (I am using Vagrant 2.2.3) would this be a regression issue? I will try using an older version of Vagrant to see if the VMDK path issue was fixed then re-added.

Same problem here.
Vagrant 2.2.3 + VirtualBox 6.0.2 & 6.0.4 = PROBLEM OCCURS
Vagrant 2.2.3 + VirtualBox 5.2.22+ = OK

PR #4169, opened 2014-07-09
Update puppet.rb

schauer:
The main manifest may be a single file or a directory of .pp files. 
docs.puppetlabs.com/puppet/latest/reference/dirs_manifest.html#with-puppet-master
https://github.com/mitchellh/vagrant/issues/4039


perryloh:
Hi @mitchellh - is this fix available in the latest version of vagrant (1.7.2) ? How would it work if I want the puppet provisioner to execute all manifests in a folder ?

PR #4195, opened 2014-07-15
Enterprise Linux 7 Dynamic network cards fix (ref #4171)

### LINKED ISSUE #4171: Issue with Consistent Network Device Naming on CentOS 7
CentOS 7 now defaults to using Consistent Network Device Naming and thus is affected by the same issues as Fedora. Issue #1997 describes the problem in detail and has a PR attached that applies the fixes for fedora, the same fixes might work for CentOS, so deferring there for a detailed description. RHEL7 is probably affected in the same way.

vStone:
Uses fedora's network configuration for RedHat and CentOS 7.

Possibly fixes #4171 but I doubt this is the 'cleanest' way. Pointers
welcome.

### DIFF
--- a/plugins/guests/enterprise7/guest.rb
+++ b/plugins/guests/enterprise7/guest.rb
@@ -1,0 +1,11 @@
+module VagrantPlugins
+  module GuestEnterpriseLinux7
+
+      class Guest < Vagrant.plugin('2', :guest)
+        def detect?(machine)
+          machine.communicate.test('grep "\(CentOS\|Red Hat Enterprise\) Linux release 7" /etc/redhat-release')
+        end
+      end
+
+  end
+end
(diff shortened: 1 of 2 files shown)


ehime:
Seems like this has duplicate effects on Oracle boxes as well, what was the root cause of this so I can implement a fix on Oracles side?

andytson:
Its because RHEL 7 derivatives are the first to change from SysVInit to systemd out of all RHEL derivatives Vagrant has supported. Each has their own /etc/redhat-release text that needs to be checked by the flavours.rb regex. Oracle linux can be fixed by updating the regex to include a match Oracle linux's release 7 text

khaije1:
I'm seeing exact same error on my CentOS v5 Virtualbox VM boxes. 
Is there something I can do to test if this is the same error and if some variation of the same fix would work?

PR #4236, opened 2014-07-22
Add WinRM over SSL support - (replaces #3960)

maxlinc:
I updated @pdericson's PR #3960 to:
- Use the released winrm Gem
- Fixed the tests

I'm planning to address my comments from that PR as well:

> - [x] There should be a configuration option for `no_ssl_peer_verification`. Right now it is hardcoded to `true` (no verification).
> - [x] I think the default for `no_ssl_peer_verification` should be `false`. I think it's better to be secure by default and let users opt-out of security if they're not worried about using self-signed certificates for testing.
> - [x] It is using the `:plaintext` transport rather than the `:ssl` transport. WinRM does make an SSL connection even with the `:plaintext` transport, but there are some differences so `:ssl` is preferable.
>
> In fact, I think `config.winrm.transport = :ssl` may be better than `config.winrm.ssl = true`. It matches up with the knife-windows option, and would make it easier to support other winrm transports in the future, like `:kerberos`.

I just wanted to start see Travis go green and open the discussion back up.


sneal:
@maxlinc Branch won't cleanly merge

maxlinc:
Yeah... that's why I don't like long-lived branches (though I understand why it was necessary).

I'll try to rebase...

maxlinc:
Fixed. It was just a trivial conflict with the version for winrm (1.3.0 vs 1.3 - I've used your choice, 1.3, which is more correct).

test/unit/plugins/kernel_v2/config/vm_test.rb
```diff
@@ -214,12 +223,22 @@ def assert_valid
...
-      expect(n[0][1][:host]).to eq(14100)
-      expect(n[0][1][:id]).to eq("winrm")
+      winrm_network = find_network 'winrm'
+      expect(winrm_network[:guest]).to eq(22)
+      expect(winrm_network[:host]).to eq(14100)
+      expect(winrm_network[:id]).to eq("winrm")
+    end
+
+    it "allows overriding WinRM SSL" do
+      subject.communicator = :winrmssl
```
sneal: This test is confusing, there isn't a :winrmssl communicator type that I see.

maxlinc: You're right... it should just be :winrm.

The test is supposed to be for the second network forwarded for winrm, though. See https://github.com/maxlinc/vagrant/blob/winrmssl/plugins/kernel_v2/config/vm.rb#L403-L408.

Also your the discussion on https://github.com/mitchellh/vagrant/issues/4932, including your response:
> It seems logical to only automatically forward networks as required by the active communicator. I don't think its currently breaking anything by automatically forwarding both, but its sloppy.

I don't care if it's zero (the cloud providers don't use forwarded ports, AFAIK), one (cleaner) or two (easier to switch - because I changing forwarded ports requires a reload). I'll fix the communicator to be winrm though.

sneal:
WOMM. @maxlinc did you want to update the test before I merge this? Its just a nit, not a requirement.

Disabling http on the guest box (leaving https enabled) while leaving the transport the default of :plaintext, it hung Vagrant. Probably existing behavior. ` INFO retryable: Retryable exception raised: #<HTTPClient::KeepAliveDisconnected: HTTPClient::KeepAliveDisconnected: >`

test/unit/plugins/kernel_v2/config/vm_test.rb
gildegoma: fixed in #6581.

PR #4237, opened 2014-07-22
SMB sharing on Mac OS X hosts

emyl:
For I project I'm working on I'm trying to make SMB sharing work on **OS X hosts**.

Before possibly create an external plugin I've tried to integrate my work into core, just in case it could be of your interest.

I've also successfully tested the patch against #4209, so even the combination **OS X host** -> **Win guest** should work.

I've used OS X 10.8 for tests, theoretically the code should also work on 10.9. For sure it **doesn't work** on 10.7 and lower, since it relies on the `sharing` command introduced with 10.8

Feel free to catch me for any need. I never told you "thank you" for this awesome piece of software? :smile:


plugins/hosts/darwin/cap/load_host_ips.rb
```diff
@@ -0,0 +1,11 @@
+module VagrantPlugins
+  module HostDarwin
+    module Cap
+      class LoadHostIPs
+        def self.load_host_ips(env)
+          `ifconfig | grep 'inet ' | grep -v '127.0.0.1' | cut -d ' ' -f 2`.split
```
sethvargo: We prefer to use the shellout library here instead.

plugins/hosts/darwin/cap/smb.rb
```diff
@@ -0,0 +1,86 @@
...
+        class << self
+          private
+          def cleanup_smb_shares!(machine)
+            @shares_to_cleanup.each { |share| sudo("sharing -r #{share}", false) }
+            cache_file = machine.data_dir.join("smb_share_ids")
+            File.delete(cache_file) if cache_file.file?
+          end
+
+          def sudo(command, raise_error=true)
+            stdout, stderr, status = Open3.capture3("sudo -S #{command}", :stdin_data => "#{@user_password}\n")
```
sethvargo: Vagrant has a subprocess library that we prefer to use instead

plugins/hosts/darwin/cap/smb.rb
```diff
@@ -0,0 +1,86 @@
...
+          end
+        end
+
+        def self.create_smb_share(env, machine, id, data)
+          data[:smb_id] ||= Digest::MD5.hexdigest("#{machine.id}-#{id.gsub("/", "-")}")
+
+          sudo("PWD=$(pwd) sharing -a #{data[:hostpath]} -S #{data[:smb_id]} -s 001 -g 000 -n #{data[:smb_id]}")
+
+          # Store the digest in the machine data dir for subsequent cleanup
+          machine.data_dir.join("smb_share_ids").open("a") do |f|
```
sethvargo: I think we need a lock around this. If multiple VMs are being spun in parallel, it will clobber

plugins/hosts/darwin/cap/smb.rb
```diff
@@ -0,0 +1,86 @@
...
+            f.write("#{data[:smb_id]}\n")
+          end
+        end
+
+        def self.enable_smb_sharing(env, machine, folders, credentials)
+          # Credentials are actually enclosed in a lambda
+          credentials = credentials.call
+
+          # Determine who am I
+          current_user  = `id -un`.chomp
```
sethvargo: Subprocess library

plugins/hosts/darwin/cap/smb.rb
```diff
@@ -0,0 +1,86 @@
...
+          end
+
+          # Invalidate existing sudo session, if present
+          system("sudo -k")
+
+          # Get the user password and store it in a Class instance variable
+          @user_password = folders.values.first[:smb_password] || credentials[:password]
+
+          # Start the SMB daemon if not already running
+          sudo("launchctl load -w /System/Library/LaunchDaemons/com.apple.smbd.plist")
```
sethvargo: Is there a way we can check if the launchd is running before loading it? This seems like extra work to do on each call. Then it would be helpful to log a message that is like "already enabled" or "enabling"

plugins/hosts/darwin/cap/smb.rb
```diff
@@ -0,0 +1,86 @@
...
+          sudo("defaults write /Library/Preferences/SystemConfiguration/com.apple.smb.server EnabledServices -array disk")
+
+          # Enable SMB sharing for the given user
+          sudo("pwpolicy -u #{current_user} -sethashtypes SMB-NT on")
```
sethvargo: Need some logging here so the user knows what is going on. I think we also need quotes around `current_user` in case they have spaces in their username.

plugins/hosts/darwin/cap/smb.rb
```diff
@@ -0,0 +1,86 @@
...
+          sudo("dscl . -passwd /Users/#{current_user} '#{@user_password}'")
```
sethvargo: Correct me if I'm wrong, but I think this changes the current user's password? That seems like something we should not do.

sethvargo:
Hi @emyl 

Thank you for the Pull Request, and I am sorry for the delay in reviewing. I did a preliminary review and noticed a few things that concern me, mainly the use of backticks and Open3 as opposed to Vagrant's subprocess library. Additionally, this will no longer merge cleanly. Would you mind performing a rebase please? Thanks!

sethvargo:
Closing due to inactivity. If you would like to pursue this, please open a new PR. Thanks! 😄

PR #4413, opened 2014-08-29
Communicator info abstraction

### LINKED ISSUE #4372: WinRM communcator is broken on master
Somewhere between v1.6.3 and master something has at least partially broken the winrm communicator.

Expected:
- Vagrant is able to connect via winrm

Actual: 
- Vagrant is stuck waiting for the box to boot (WinRMNotReady errors being thrown)

sneal:
:warning: DO NOT PULL :warning: 

@mitchellh This is a spike to fix #4372, I'm submitting this for early feedback before continuing.

I attempted to make the WinRM and SSH machine connection information separate but equal (that sounds horrible), previously the WinRM connection info would be built from the ssh_info which is unexpected. Vagrant would be easier to maintain if the two communicator systems followed similar patterns and layout.

I replaced usages of machine.ssh_info with machine.connection_info, perhaps a little too zealously. For example, SSH commands will get WinRM connection info if the communicator type is :winrm. It might make more sense to have the SSH commands just fail fast if the communicator type is WinRM.


sethvargo:
Hey @sneal 

I just wanted to check the status of this. Did you still need some feedback?

mitchellh:
Gonna close due to lack of activity from both sides. It has been almost a year and I think we probably don't have an immediate need for this. Thanks.

PR #4422, opened 2014-09-01
Add option to enable trusted HTTP/HTTPS redirects

crystax:
This is needed in case if we have password-protected URL which is actually redirect to another password-protected URL which accept the same credentials as initial URL.


sethvargo:
Rebased and merged in 95c9638

PR #4467, opened 2014-09-06
Make `start` action run provisioners if VM is running

### LINKED ISSUE #4421: Gesture to bring-up or provision machine, regardless of its current state
Currently, `vagrant up` does nothing if the machine is already up, while `vagrant provision` fails if the machine is not already up. I think it would useful to have a gesture that results in an up-to-date provisioned machine regardless of its current state, e.g., bring it up if it's down, then provision it regardless of its initial state.

I'd be happy to implement and submit a pull request for something like this if you were open to merging it.

rtkrruvinskiy:
Previously, there was no one gesture that would start a VM if it was not
running and run the appropriate provisioners regardless of its original
state. `vagrant up` did nothing if the VM was running, while
`vagrant provision` did nothing if the VM was not running.

Change the semantics of `vagrant up`, via the start actions of the providers,
to go through the provisioning logic even if the VM is already running.
The semantics of `run: "once"` vs `run: "always"` are respected.

Tested with the VirtualBox provider but not the others.

Resolves #4421

### DIFF
--- a/plugins/providers/docker/action.rb
+++ b/plugins/providers/docker/action.rb
@@ -209,9 +209,6 @@ module VagrantPlugins
       def self.action_start
         Vagrant::Action::Builder.new.tap do |b|
           b.use Call, IsState, :running do |env, b2|
-            # If the container is running and we're doing a run, we're done
-            next if env[:result] && env[:machine_action] != :run_command
-
             if env[:machine_action] != :run_command
               b2.use Call, HasSSH do |env2, b3|
                 if env2[:result]
@@ -223,6 +220,9 @@ module VagrantPlugins
                 end
               end
             end
+
+            # If the container is running and we're doing a run, we're done
+            next if env[:result] && env[:machine_action] != :run_command
 
             b2.use Call, IsState, :not_created do |env2, b3|
               if !env2[:result]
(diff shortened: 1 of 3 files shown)


sethvargo:
Hi @rtkrruvinskiy 

Thank you for the Pull Request, and I am sorry it has taken so long to review. In Vagrant 1.7.2, there are three possible actions:

1. `vagrant up` with a brand new box will both up the box and provision it
2. `vagrant up` from a suspended box or a box that has already been marked as provisioned will skip the provisioning
3. `vagrant up --provision` will always provision the box

I think these all satisfy your use case, but could you please confirm?

rtkrruvinskiy:
@sethvargo I just tried `vagrant up --provision` against a running VM with 1.7.2, and I'm getting:
```
==> foo: Checking if box 'foobase' is up to date...
==> foo: VirtualBox VM is already running.
```

sethvargo:
@rtkrruvinskiy correct - the box is already up. If the box is up, you can run `vagrant provision`.

rtkrruvinskiy:
@sethvargo The use case I'm trying to address, also outlined in https://github.com/mitchellh/vagrant/issues/4421, is that there is no gesture that takes the VM to a running, provisioned state irrespective of its current state (running or not). As a result, if one wants to make sure a VM is running and provisioned, one ends up writing code like
```bash
if vagrant status foo | grep running; then
  vagrant provision foo
else
  vagrant up foo
fi
```

sethvargo:
e42f346

PR #4473, opened 2014-09-08
Add HTTPS download options to `box update` and `box outdated`

### LINKED ISSUE #4420: Vagrant::Box.load_metadata has no support for specifying custom certificates
We use a privately-hosted box catalogue. When specifying boxes in a Vagrantfile, it is possible to do something along the following lines:
```ruby
config.vm.box_url = "https://.../our-box.json"
config.vm.box_download_client_cert = "/path/to/client.cert"
config.vm.box_download_ca_cert = "/path/to/ca.cert"
```
Doing `vagrant up` then works, with Vagrant successfully downloading the box. However, `vagrant box outdated` and `vagrant box update` fail, since `Vagrant::Box.load_metadata` has no support for supplying any custom certificates.

I'd be happy to submit a pull request implementing this functionality if you could advise me on the right way to approach it. It seems a little tricky, since `box_download_client_cert` and `box_download_ca_cert` are attributes of the machine and not of the box.

rtkrruvinskiy:
Vagrant::Box.load_metadata did not provide a way to specify the HTTPS
download options that could be specified when downloading boxes
(ca cert, ca path, client cert, insecure). As a result, while it was
possible to add a box whose metadata file needed to be downloaded with one of
those options specified, it was impossible to check for updates. The following
changes have been made to address the situation:

1. Create a DownloadMixins module to provide the --insecure, --cacert, --capth,
   and --cert command line options to all of `vagrant box add`,
   `vagrant box update`, and `vagrant box outdated`.
2. Extend `Vagrant::Box.has_update?` and `Vagrant::Box.load_metadata` to accept
   said download options.
3. Extend `box outdated` and `box update` commands to pass download options
   down.
4. Extend `Vagrant::Builtin::Action::BoxCheckOutdated` to honour download
   options.
5. Options specified on the command line take precedence over options specified
   in the machine configuration, if any.
6. Fix bug in `vagrant box add` where client cert was being passed down using
   the wrong environment key.
7. Unit test coverage in update_test and box_check_outdated_test.

Resolves #4420

### DIFF
--- a/lib/vagrant/action/builtin/box_check_outdated.rb
+++ b/lib/vagrant/action/builtin/box_check_outdated.rb
@@ -37,13 +37,23 @@ module Vagrant
           end
 
           constraints = machine.config.vm.box_version
+          # Have download options specified in the environment override
+          # options specified for the machine.
+          download_options = {
+            ca_cert: env[:ca_cert] || machine.config.vm.box_download_ca_cert,
+            ca_path: env[:ca_path] || machine.config.vm.box_download_ca_path,
+            client_cert: env[:client_cert] ||
+                           machine.config.vm.box_download_client_cert,
+            insecure: !env[:insecure].nil? ?
+                        env[:insecure] : machine.config.vm.box_download_insecure
+          }
 
           env[:ui].output(I18n.t(
             "vagrant.box_outdated_checking_with_refresh",
             name: box.name))
           update = nil
           begin
-            update = box.has_update?(constraints)
+            update = box.has_update?(constraints, download_options: download_options)
           rescue Errors::BoxMetadataDownloadError => e
             env[:ui].warn(I18n.t(
               "vagrant.box_outdated_metadata_download_error",
--- a/lib/vagrant/box.rb
+++ b/lib/vagrant/box.rb
@@ -115,8 +115,9 @@ module Vagrant
     # Loads the metadata URL and returns the latest metadata associated
     # with this box.
     #
+    # @param [Hash] download_options Options to pass to the downloader.
     # @return [BoxMetadata]
-    def load_metadata
+    def load_metadata(**download_options)
       tf = Tempfile.new("vagrant")
       tf.close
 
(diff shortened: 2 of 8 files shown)


rtkrruvinskiy:
I never received any feedback about this PR. Are any changes necessary for it to be considered? Or is it, perhaps, no longer necessary?

Tagging @sethvargo since he seems to be involved in many PR reviews these days.

sethvargo:
Hi @rtkrruvinskiy 

Thank you for the Pull Request, and I am sorry it has taken us so long to review. This is definitely a real problem (the lack of HTTPS customization), but I think we may be able to come up with a better solution. I like how you have abstracted the download options into their own mixin, but I think we may be able to do even better by providing a better abstraction over curl or writing our own pure-Ruby solution.

I do not want to do a full review and have you do extra work until I have thought out this pathway a little more, just to make sure we are headed in the correct direction :smile:. Does that make sense?

rtkrruvinskiy:
@sethvargo Sure, that makes sense. Thanks for your time!

ccope:
Would it be possible to merge this for the next release to unbreak things until you have time to re-evaluate the usage of curl?

arthurschreiber:
Heyo @sethvargo,

any chance of getting this merged for the next release? This quite a nuisance for us, as all our internal systems use self-signed certificates, including the system that hosts our vagrant boxes.

ccope:
@sethvargo it's been over a year, can we get this merged?

mitchellh:
Hey @ccope. We definitely want to support this feature. We're happy for the contribution but unhappy with the implementation: it is a lot of repeated special case if statements that merge options. Instead of supporting that, we'd rather centralize setting these options somewhere.

We haven't had time to put towards this effort so it has been delayed quite a bit. :( Sorry.

mitchellh:
We're just going to merge this as-is and fix up as we go so we can get something working. Thanks!

ccope:
Great, thanks!

PR #4496, opened 2014-09-11
Restore ability to manage boxes with older versions of the sudo package

donckers:
This reverts the change introduced in 9cd9e6e55d802553fa90e63d966b302006b45de5 (sudo -E to preserve env).  It seems like the -E option would be useful for most instances, but since you cannot guarantee that all versions of sudo support this flag this needs to be reverted.  I might suggest that you refactor and give us a config option to enable/disable this behavior?


sethvargo:
Hi @donckers 

Thank you for the Pull Request. After looking back through history and old issues, I do not think we are able to accept this PR. I understand that the sudo package is broken for older versions of sudo, but I do not think removing the `-E` flag will be helpful here. In fact, most of Vagrant will start to break if we stop forwarding the environment :frowning:.

Making the option configurable also presents an interesting challenge/catch-22. Almost all of Vagrant boils down into this `shell_execute` function, but exposing a global configuration option may not make sense (i.e. users will want to _sometimes_ preserve the environment).

mitchellh:
We added `sudo_command` option so you can now use that to customize this command. This is a gobal configuration, but if your system always has an older sudo, it won't affect things.

PR #4557, opened 2014-09-25
Always call @en.machine_names to check if there is a machine with a name 

### LINKED ISSUE #4556: Vagrant works different when yields a VM for command by name and when yields VMs by regexp or all autoload VMs
I like the idea to put a default provider name into my Vagrantfile: 

`````ruby
ENV['VAGRANT_DEFAULT_PROVIDER'] = 'docker'

	config.vm.define "docker" do |ldap|

	  ldap.vm.synced_folder '.', '/vagrant', disabled: true

	  ldap.vm.provider :docker do |docker|
	    docker.image = 'busybox'

	    docker.vagrant_vagrantfile = 'Vagrantfile'
	    docker.vagrant_machine = "host"
	  end
	end


	# Machine to host dockers for development purposes
	config.vm.define 'host', autostart: false do | host |
		host.vm.box = 'mitchellh/boot2docker'

		host.vm.provider :parallels do |prl, override|
		  override.vm.box = 'parallels/boot2docker'

		  prl.name = 'boot2docker/busybox'
		end
	end
end
`````
when i run `````>vagrant up docker````` i get error "* A box must be specified." but if i run `````>vagrant up /docker/````` everything works as expected. This is because in the latter case vagrant reads and executes my Vagrantfile before it tries to find the default provider for a box. But in the first case it does not execute  the Vagrantfile before looking for the default provider and resolves it to a different one, not to the "docker"

I'd expected vagrant should work equally independently from how we pass a box name.

ilucker:
Fixes #4556

### DIFF
--- a/lib/vagrant/plugin/v2/command.rb
+++ b/lib/vagrant/plugin/v2/command.rb
@@ -189,7 +189,10 @@ module Vagrant
               else
                 # String name, just look for a specific VM
                 @logger.debug("Finding machine that match name: #{name}")
-                machines << get_machine.call(name.to_sym)
+                if @env.machine_names.include?(name.to_sym) ||
+                    @env.machine_index.include?(name.to_s)
+                  machines << get_machine.call(name.to_sym)
+                end
                 raise Errors::VMNotFoundError, name: name if !machines[0]
               end
             end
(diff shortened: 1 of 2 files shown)


sethvargo:
Hi @ilucker 

Thank you for the PR. I rebased this locally in https://github.com/mitchellh/vagrant/pull/5768, and as you can see, the tests are failing. Could you please rebase and take a look at the failing tests please?

sethvargo:
https://github.com/mitchellh/vagrant/issues/4556#issuecomment-118997969

Issue #4608, opened 2014-10-09
virtualbox provider ignores host_ip for forwarded ports


ivan-kolmychek:
I have two sample Vagrantfiles in different folders, for different project.
One specifies

```
config.vm.network "forwarded_port", guest: 3000, host: 3000, host_ip: "127.1.1.253" 
```

and other

```
config.vm.network :forwarded_port, guest: 3000, host: 3000, host_ip: "127.1.1.254"
```

When I run only one, everything is all right. But when I try to get up second, vagrant tells me that port is busy:

```
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Checking if box 'ubuntu/trusty64' is up to date...
==> default: Clearing any previously set forwarded ports...
Vagrant cannot forward the specified ports on this VM, since they
would collide with some other application that is already listening
on these ports. The forwarded port to 3000 is already in use
on the host machine.

To fix this, modify your current projects Vagrantfile to use another
port. Example, where '1234' would be replaced by a unique host port:

  config.vm.network :forwarded_port, guest: 3000, host: 1234

Sometimes, Vagrant will attempt to auto-correct this for you. In this
case, Vagrant was unable to. This is usually because the guest machine
is in a state which doesn't allow modifying port forwarding.
```

But `netstat -nl` shows only one record for `tcp:3000` and it's on correct address:

```
$ netstat -nl | grep 3000
tcp        0      0 127.1.1.254:3000        0.0.0.0:*               LISTEN  
```

What am I doing wrong?

**UPD:** Vagrant 1.6.5

dragon788:
I'm not seeing where host_ip is even a thing? In the documentation I see it referenced as simply "ip:" and this applies to private and public networks NOT to forwarded ports. Logically if you are forwarding ports from the host, it can only have ONE of any port in use, even if the guests have multiple IPs.

ncolton:
@dragon788 hopefully this link works, but in the [Options Reference](https://github.com/mitchellh/vagrant/blob/master/website/docs/source/v2/networking/forwarded_ports.html.md#options-reference) it lists both `guest_ip` and `host_ip`.

ivan-kolmychek:
> . Logically if you are forwarding ports from the host, it can only have ONE of any port in use

@dragon788 looks like you misunderstand the concept of a port. 

Look at [Wikipedia page](http://en.wikipedia.org/wiki/Port_%28computer_networking%29).
[...]

PR #4616, opened 2014-10-10
Consider the host_ip value when handling port collisions

crypt1d:
In response to issue #4608, this fix allows vagrant to handle forwarding port collisions by considering the :host_ip flag as well. This allows vagrant to bind to multiple interfaces on host machine (or loopbacks) if there are any.
For example:
guest1 forwards port 80 to 127.1.1.253:80 on host
guest2 forwards port 80 to 127.1.1.254:80 on host

Previously vagrant was not considering that host can have multiple interfaces to forwards ports to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: ruby
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq -y bsdtar
+  - rvm @global do gem uninstall bundler --all --executables
+  - gem uninstall bundler --all --executables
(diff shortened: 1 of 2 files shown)


thebillkidy:
Is there still a fix in progress for this?

OldCrowEW:
Hello, I am also curious to know if there is a fix coming for this?

vmgarcia:
Same, having this feature would be very useful

pravinchandar:
folks, I've a fix for this in PR #7035

Issue #2608, opened 2013-12-09
Generate dynamic key on `vagrant up` to better protect VMs


virtuald:
All default Vagrant boxes come with the same ssh key and passwords installed. This means anyone who can ssh into your VM, will be able to authenticate to it if they try the vagrant default credentials. Additionally, vagrant boxes come with passwordless sudo privileges, so anyone able to SSH into your box will be able to do whatever they want on it.

When using private networking or no networking, as of Vagrant 1.2.3, for the most part this isn't super important, because you can only access the Vagrant VM from localhost. However, this becomes _extremely_ important if you are using a vagrant box that is on a bridged network. If your VM is on a bridged network, without other controls in place that means anyone who has access to your local network can SSH into your VM and get root access on it. The default configuration on a bridged network is not secure and potentially dangerous.

It would be nice if this was fixed in vagrant core (see below for one solution), but if not, the documentation should explicitly warn users that they are opening the VM up to the outside world via SSH if they are using a public network, and that all boxes generally use the same SSH key for login. I think most people realize that having it on a public network will require security, but I also think most people don't realize that _anyone_ can ssh into their VM because the key is publicly distributed.
### One solution

I've created a vagrant plugin (https://github.com/virtuald/vagrant-rekey-ssh -- just run `vagrant plugin install vagrant-rekey-ssh` to install it) that will address this problem, by replacing the vagrant insecure key on a VM with one that is unique to the host's user. It would be ideal if this functionality was implemented in vagrant core so that it is secure by default, but at the very least there needs to be a warning in the documentation so that users realize what they're getting into.
### References
- http://stackoverflow.com/questions/14715678/vagrant-insecure-by-default
- Vagrant issue https://github.com/mitchellh/vagrant/issues/1785

PR #4707, opened 2014-10-24
Generate Random SSH key on `vagrant up`

mitchellh:
This PR makes it so that on `vagrant up`, Vagrant will randomly generate a keypair and use that if the insecure keypair is detected. This adds an extra level of security when doing things such as bridge networking. 

Fixes GH-2608

### DIFF
--- a/website/docs/source/v2/boxes/base.html.md
+++ b/website/docs/source/v2/boxes/base.html.md
@@ -116,6 +116,10 @@ that OpenSSH is very picky about file permissions. Therefore, make sure
 that `~/.ssh` has `0700` permissions and the authorized keys file has
 `0600` permissions.
 
+When Vagrant boots a box and detects the insecure keypair, it will
+automatically replace it with a randomly generated keypair for additional
+security while the box is running.
+
 ### Root Password: "vagrant"
 
 Vagrant doesn't actually use or expect any root password. However, having


TomyLobo:
@niclasgelin referenced this pull request in a commit that adds "config.ssh.insert_key = false" to a file that looks like a Vagrantfile.
Can someone confirm that this indeed disables the entire feature?

patrickheeney:
Lost 4 hours today because of this. `config.ssh.insert_key = false` should have a big warning somewhere to let everyone know how to disable this feature.

mbrodala:
@TomyLobo Yes, that's currently the only way. As @mitchellh pointed out, you don't even have to put this into every Vagrantfile since Vagrantfiles are also loaded from your home directory.

TomyLobo:
I cannot control my users' home directory contents, but i can control their Vagrantfiles.
I still think this is a bad idea, btw. It's [snake oil|http://en.wikipedia.org/wiki/Snake_oil_%28cryptography%29], at best and, as @patrickheeney and I found out the hard way, a waste of time at worst.
Please make "config.ssh.insert_key = false" the default again.

tdk2fe:
Hey everybody -

Not sure if there has been any more discussion around this, but I did suffer some of the same problems as @TomyLobo and @patrickheeney described.  I've been running vagrant for a while now, but I set up my boxes in directories out of a DAS location that's accessible via multiple users.  So while my home directory (and vagrant config) is /home/user/.vagrant, I have a Vagrants folder in $HOME that links to /mnt/MSONIC_EXTARRAY/Virtualization/Vagrants, which is where I set up new boxes.

After not using it for a couple months, I spent some time trying to figure out the message regarding incorrect ownership of the private key used to connect.

I did verify that adding `config.ssh.insert_key = false` resolves this error.

TomyLobo:
I still think it's pointless to default this to true. If you care about security, you can activate it and change the passwords of everything on the box while you're at it.

Confusing users is bad. Adding false security is bad.
I haven't heard a good side to this patch yet.

hardyoyo:
While it has been closed as a duplicate, #5407 is worth mentioning here. From the looks of things, #4707 breaks the base box packager command.

obestwalter:
I don't see the value of this either. Vagrant is for devboxes and they don't have to be secure. I think this will cause a lot of confusion and calls for help - e.g. https://groups.google.com/forum/#!topic/vagrant-up/FeXfcxpQlVA

This would not be a problem if the default was set to false.

BattleBrisket:
Adding my voice to the choir; changing the default insecure key behavior was a bad decision. Vagrant is intended to build wide open boxes for development purposes. While I can appreciate the intent that @mitchellh describes in #5005....

> The reason we made the change is because historically it hasn't mattered, but today more and more people use Vagrant with remote providers globally accessible via the internet (no longer host-only), and this security risk was still an issue with bridged networking (though, much smaller surface area). 

By that logic, we should also randomize the password assigned to the `vagrant` user account; a cursory search of project PR's and issues indicates such a change isn't forthcoming.

Generating a random pair is security by obscurity *at best*. At worst, it's simply a waste of developer time, as I now have to go back and refactor every `Vagrantfile` in every project, and I have to be sure to add yet another config flag to new `Vagrantfile`'s moving forward.

Again, I appreciate there's a new user base with higher security requirements, and everyone understands the value of security, but let's not forget the problem we're trying to solve here. If I want to use Vagrant in some kind of public environment, I would reasonably expect to do some extra legwork to make it more secure, given the design intentions.

rasputnik:
Random keys main issue is the breaking of automation. If someone can suggest a clean way to achieve that, I'd be grateful.

(and agree this is a pain with no real benefit until the vagrant/vagrant login details are resolved).

Gerst20051:
:+1: adding ```config.ssh.insert_key = false``` to the Vagrantfile and removing the vm private key automatically updated ```vagrant ssh-config``` with the correct private key ```~/.vagrant.d/insecure_private_key```. The last thing I had to do was update the authorized keys file on the vm. ```curl https://raw.githubusercontent.com/mitchellh/vagrant/master/keys/vagrant.pub > ~/.ssh/authorized_keys```

rasputnik:
yep, config.ssh.insert_key = false prevents this becoming a problem. If that was the default, I'd have no issue with this feature (it's not for me but that's fine if someone thinks its useful).

TomyLobo:
Last developer comment is from 28 Oct 2014.
I think they're sitting this one out...

ddelrio1986:
Was just bitten by this! Wasted so much time on it!

TomyLobo:
@mitchellh Can you please make ``config.ssh.insert_key`` default to false to stop wasting people's time.

bradisbell:
> You simply need to add config.ssh.insert_key = false into your Vagrantfile as explained in https://docs.vagrantup.com/v2/vagrantfile/ssh_settings.html.

This method is not sufficient.  I share a Vagrantfile with many other developers.  This is a setting that should be configurable Vagrant-wide on my host, and not per Vagrantfile so that the hacks I have to use to make Vagrant work again do not affect them.

assunluis80:
Where I can find the file to configure config.ssh.insert_key? I'm using Ubuntu 14.04 LTS.

TomyLobo:
There is no central file for this. You have to do this in every single bloody Vagrantfile.

mbrodala:
@TomyLobo Not exactly true, you can also put this in a [`~/.vagrant.d/Vagrantfile`](https://docs.vagrantup.com/v2/vagrantfile/index.html) to apply it to all machines on your desktop.

PR #4738, opened 2014-10-29
Added a --plugin-clean-sources parameter

robkinyon:
This parameter that will allow for only those sources that are defined by the user to be used. I also added editor swapfiles to the .gitignore.

I work in a corporate environment behind firewalls which prevent access to both rubygems.org and gems.hashicorp.com. We have our own rubygems repository where we host (among other things) Vagrant plugins (both internally-created and external). Without a patch like this, `vagrant plugin install vagrant-triggers` (for instance) will fail because it cannot retrieve the specs from gems.hashicorp.com. Furthermore, we need something like this so that the risk and audit groups don't see the attempts at forbidden traffic.

I'm not thrilled with the parameter behavior I created - it means that parameter parsing is a state machine and that could be surprising. I would have preferred `--plugin-source` to have the behavior of overwriting the defaults. But, that would have been a breaking API change and I didn't want to propose that in the initial request.

Also, this leads to the desire of having a `.vagrantrc` that we can use to set these values by default on a project basis. But, that's another changeset.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 # OS-specific
 .DS_Store
 
+# Editor swapfiles
+.*.sw?
+*~
+
 # Vagrant stuff
 acceptance_config.yml
 boxes/*
--- a/lib/vagrant/bundler.rb
+++ b/lib/vagrant/bundler.rb
@@ -178,11 +178,6 @@ module Vagrant
 
       f = File.open(Tempfile.new("vagrant").path + "2", "w+")
       f.tap do |gemfile|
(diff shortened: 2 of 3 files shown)


alexpop:
+1

scottmiddleton2:
+1

sethvargo:
I actually want to get some additional review from @mitchellh before merging this. I think there might be unintended side-effects.

mitchellh:
I agree its wonky, but lets just go with it. We'll hopefully remove bundler at some point. Until then, we'll have this.

PR #4758, opened 2014-10-31
Ignore possible version directories that are hidden folders.

philwrenn:
I use a synchronization tool to sync my boxes between my laptop and desktop. This prevents me from having to download the same box version twice from vagrant cloud. My sync tool, like may others, creates a hidden folder to store sync related data and conflicted files. It would be nice if vagrant would not mistakenly identify the hidden folder as a version folder.


lib/vagrant/box_collection.rb
```diff
@@ -224,7 +224,7 @@ def all
           # Otherwise, traverse the subdirectories and see what versions
           # we have.
           child.children(true).each do |versiondir|
-            next if !versiondir.directory?
+            next if !versiondir.directory? || versiondir.basename.to_s[0] == '.'
```
sethvargo: I think we can make this a little clearer using `start_with?`:

```ruby
versiondir.basename.start_with?(".")
```

sethvargo: I also think we should move this check onto its own line:

```ruby
next if !versiondir.directory?
next if versiondir.basename.start_with?(".") # Ignore hidden directories
```

sethvargo:
Hi @philwrenn 

Thank you for the PR, and I am sorry it has taken a bit for us to get you a review. Overall this looks pretty good to me. I left a few comments in the PR - let me know what you think.

philwrenn:
I made the changes you suggested.

philwrenn:
@sethvargo I attempted made the changes you suggested, but I couldn't get build to complete.. Honestly I don't really work with ruby. I just ran into an issue and was lucky enough to get it working. If this isn't bugging anyone else you can just decline the request. This is no longer an issue for me.

PR #4759, opened 2014-10-31
avoid sudo for changing /etc/exports when no change needed

ccutrer:
also fix NFS pruning to avoid pruning the machine we're starting,
even if the provider doesn't return non-running machines (as is the
case with vmware)


sethvargo:
Related to #3076

sethvargo:
Closing in favor of 3076

PR #4765, opened 2014-11-01
Allow specifying post_up_message block

jch:
This PR allows users to configure a callable `post_up_message` to print values at a later time:

```ruby
  config.vm.post_up_message = -> {
    ip = `vagrant ssh -- "ifconfig eth1 | grep -o -E '[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+' | head -n1"`.chomp
    """
    Your instance is configured.

    https://#{ip}
    Username: foo
    Password: bar
    """
  }
end
```

I also thought about changing the method signature to be more DSL-like, but decided against it for minimal changes. Open to updating the PR to this if it feels more natural.

```ruby
  config.vm.post_up_message do
    "deferred message here"
  end
end
```

Thoughts?

### DIFF
--- a/plugins/kernel_v2/config/vm.rb
+++ b/plugins/kernel_v2/config/vm.rb
@@ -341,6 +341,12 @@ module VagrantPlugins
         @__defined_vms[name].config_procs << [options[:config_version], block] if block
       end
 
+      # Returns configured post_up_message. If value is callable, returns the
+      # value of invoking it.
+      def post_up_message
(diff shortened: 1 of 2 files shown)


mitchellh:
I agree with @sethvargo that adding this callback behavior seems to be stretching the intent of the Vagrantfile. The Vagrantfile is meant to be a configuration format, not a program on its own. Making it Ruby has its plusses and minuses, but I don't want this sort of interaction with it.

I'll look into your other provisoining issue.

PR #4813, opened 2014-11-11
Update configure_networks.rb for funtoo

derdanne:
fixed name of temporary network configuration file to work with the further process of copying file to configuration directory in funtoo guests

fix for https://github.com/mitchellh/vagrant/issues/4812


sethvargo:
Thanks!

Issue #1454, opened 2013-03-19
Vagrant is always creating a new VM


Vloureiro:
I'm using vagrant on Win7:

```
Virtualbox 4.2.8
Vagrant 1.0.6.dev
box_url "http://files.vagrantup.com/lucid32.box"
```

Until the last week, everythin is fine, vagrant created the VM once and then I was using it. But this week everytime I run "vagrant up" it creates a new VM, installing all my packages defined in chef-solo. Now I have 5 boxes in my Virtualbox, which only the first one is the real VM I use, with all my keys and development tools, but Vagrant just don't use it.

And there's more: vagrant is much, much slow now. What's going on?

joequery:
I'm going to give @maoueh 's solution a try, as I've just lost my vm for the second time in two days on my Windows 8 machine :(

maoueh:
Note that VirtualBox 4.3.20 and up has the fix for the problem with `./VBoxManage.exe showvminfo` command. Updating should resolve the issue.

dheche:
I'm using vagrant 1.6.1, virtualbox 4.3.28, fedora 22

Updating .vagrant/machines/<machine_name>/virtualbox/id is not enought if your previous box is in save state (vagrant suspend). You should 'discard save state' from virtualbox first before you can vagrant up again.

brainv:
any fix?

eXistenZNL:
See how I solved it above.

pabloleone:
I've the same problem.. everytime I switch repos, a new VM is created when I do vagrant up. I started with the .vagrant folder added to the repo and worked, I read on the internet that .vagrant shouldn't be in the repo, so I removed from the repo, since then everytime I switch branches, a new vm is created, so should I add the .vagrant folder to the repo?

Why this happens? is there any reason I'm missing?

It's very annoying because it works the way that it shouldn't.

PR #4824, opened 2014-11-14
Force at least one retry in vm_exists? for provider VirtualBox 4.3

maoueh:
Due to [VirtualBox #13190](https://www.virtualbox.org/ticket/13190), on Windows, `VBoxManager showvminfo` crashes if the VM is suspended (and maybe in other occasions). This leads Vagrant to think the VM does not exist. This has bad consequences for lot of commands since machine uuid is cleared which leads to a new VM being started. Since there is already a retry mechanism in place, simply force at least one retry of the command. The second invocation should work correctly if the machine is present.

Downside of this patch, it takes 2 more seconds for the UI to respond if the vm really does not exist. We could remove this by forcing a retry without a sleep.

Linked to #1454 (Maybe we should start a brand new issue for this however, since we hijacked the initial bug report).

Regards,
Matt

### DIFF
--- a/plugins/providers/virtualbox/driver/version_4_3.rb
+++ b/plugins/providers/virtualbox/driver/version_4_3.rb
@@ -579,9 +579,14 @@ module VagrantPlugins
             result = raw("showvminfo", uuid)
             return true if result.exit_code == 0
 
-            # GH-2479: Sometimes this happens. In this case, retry. If
-            # we don't see this text, the VM really probably doesn't exist.
-            return false if !result.stderr.include?("CO_E_SERVER_EXEC_FAILURE")
+            # GH-2479: Sometimes this happens. In this case, retry.
+            #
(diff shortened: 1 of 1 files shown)


maoueh:
VirtualBox original bug (https://www.virtualbox.org/ticket/13190) has been fixed in version 4.3.20. This could be seen as irrelevant since the bug is now closed. Feel free to close it if you feel so.

Regards,
Matt

sethvargo:
Hi @maoueh 

Thank you for the Pull Request. This would have been great to get in while VirtualBox was broken, but now that it is fixed in the later versions, I do not think we need this code. If users are experiencing this issue, they can update VirtualBox :smile: 

I apologize for the delay in reviewing this, and thank you again for submitting a PR!

maoueh:
Hehe indeed, a working VirtualBox is better than a patched Vagrant!

And no problem for the late reply, you are already forgiven guys :D

PR #4825, opened 2014-11-14
doc destroy - notice about space occupied by the box

gajdaw:
Hi,
Because `vagrant destroy` doesn't remove the box installed during `vagrant up`, the statement:

    After running this command, your computer should be left at a clean state, 
    as if you never created the guest machine in the first place.

is a bit of an exaggeration, I think. Maybe you will find the notice I have written useful?


sethvargo:
Thanks!

PR #4847, opened 2014-11-19
A few modifications to fedora and redhat guests

whitel:
Mostly cleanup related to systemd vs sysvinit.


plugins/guests/fedora/cap/nfs_client.rb
```diff
@@ -0,0 +1,13 @@
+module VagrantPlugins
+  module GuestFedora
+    module Cap
+      class NFSClient
+        def self.nfs_client_install(machine)
+          machine.communicate.tap do |comm|
+            comm.sudo("yum -y install nfs-utils nfs-utils-lib")
```
strzibny: The block is not necessary here, `machine.communicate.sudo("yum -y install nfs-utils nfs-utils-lib")` should be just fine.

plugins/guests/fedora/cap/flavor.rb
```diff
@@ -0,0 +1,25 @@
...
+  module GuestFedora
+    module Cap
+      class Flavor
+        def self.flavor(machine)
+          # Read the version file
+          output = ""
+          machine.communicate.sudo("grep VERSION_ID /etc/os-release") do |type, data|
+            output += data if type == :stdout
+          end
+          version = output.split("=")[1].chomp!.to_i
```
strzibny: This won't work, I would rewrite this as:

```ruby
machine.communicate.sudo("grep VERSION_ID /etc/os-release") do |type, data|
  version = data..split("=")[1].chomp.to_i if type == :stdout
end
```

And then `if version && version >= 20` condition determing and returning `:"fedora_#{version}"` and ``:fedora` otherwise.

sethvargo:
Hi @whitel 

Thank you for the Pull Request, and I am sorry it has taken so long to review. I rebased this in https://github.com/mitchellh/vagrant/pull/5770 and I ended up removing some modifications, specifically around RedHat changes. There is not enough context to understand these changes, specifically around the complete removal of code.

If you would like to submit a PR to make changes to the RedHap capabilities, I will gladly review it, but in the interest of getting this merged, I have cherry-picked the fedora-specific pieces into master :smile: 

Please let me know if you have any questions.

PR #4878, opened 2014-11-27
Added ignore_errors to pupper_server provisioner

svenXY:
- this will continue with provisioning, even if some stuff in puppet
   failed
 - it will just return true, regardless of the actual return value of
   puppet agent

We have some very long puppet runs here where some stuff may fail, but we then want to fix this within the vagrant created box.
We have more provisioners after puppet and want to be sure they are run as well.

Thus, we need to be able to ignore the result of the puppet_provision.

Since this is completely optional, it should not harm.

Thanks,
Sven

PS: With my combination of rbenv, bundler and such, I wasn't able to get vagrant to run properly, sorry. I'd loved to properly test and such. Maybe one of you can let me know in a understandable way how I can get vagrant to run under rbenv/bundler/gem, you name it.

### DIFF
--- a/plugins/provisioners/puppet/config/puppet_server.rb
+++ b/plugins/provisioners/puppet/config/puppet_server.rb
@@ -8,6 +8,7 @@ module VagrantPlugins
         attr_accessor :options
         attr_accessor :puppet_server
         attr_accessor :puppet_node
+        attr_accessor :ignore_errors
 
         def initialize
           super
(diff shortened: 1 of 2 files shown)


sethvargo:
Hi @svenXY 

Thank you for the pull request. I am going to defer to the other maintainers, but I am not in support of this change. I understand your use case, but I think it is very specific to your usage of Vagrant and this has not been a widely requested feature.

mitchellh:
I agree with @sethvargo. Expecting errors out of provisioners is not expected, and if you want this use case, please use the shell provisioner directly to get this behavior.

PR #4943, opened 2014-12-11
Improved WinRM error handling (including better `ready?` and `wait_for_ready`)

maxlinc:
I split this from #4236 to make things easier to review. That PR contains just changes for SSL. This PR builds on that PR so the diff currently shows SSL support, but it'll shrink once that PR is merged and I rebase.

A few notes:
1. Authentication failures are currently retriable, and I haven't changed the default `max_times`, which was set to 20, with a 10 second sleep. That means WinRM will retry authentication for 10 minutes attempt authentication!
2. This uses a pre-release gem for WinRM. So it's technically releasable (as opposed to using git), but ideally we'd wait for an official WinRM release.
3. I've modeled it after SSH. I'm not sure about all the possible errors we could get from WinRM, but I think its a good assumption that it either throws - or should throw applicable errors from `Errno`.
4. Since it's modeled after SSH there is a lot of duplicate code, and duplicate translation keys. It might be worth refactoring so they share errors and translations (e.g. `Vagrant::Errors::Communicator::ConnectionRefused` instead of `Vagrant::Errors::SSHConnectionRefused` and `VagrantPlugins::CommunicatorWinRM::Errors::ConnectionRefused`), but I'd rather focus on WinRM and save refactoring for a separate PR.


vagrant.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rb-kqueue", "~> 0.2.0"
   s.add_dependency "rest-client", ">= 1.6.0", "< 2.0"
   s.add_dependency "wdm", "~> 0.1.0"
-  s.add_dependency "winrm", "~> 1.1.3"
+  s.add_dependency "winrm", "= 1.3.0.dev.2"
```
maxlinc: There is now. @sneal released WInRM 1.3 today. I've updated the gemspec on here and I'm updating [my other open PRs related to WinRM](https://github.com/mitchellh/vagrant/pulls/maxlinc).

Note that I've got a few open PRs related to WinRM. I tried to split it to make it a little easier to review...

plugins/communicators/winrm/shell.rb
```diff
@@ -114,30 +112,63 @@ def execute_shell_with_retry(command, shell, &block)
...
+          raise Errors::AuthenticationFailed,
+              user: @config.username,
+              password: @config.password,
+              endpoint: endpoint,
+              message: exception.message
+        when WinRM::WinRMHTTPTransportError
+          case exception.status_code
+          # If the error is a 401, we can return a more specific error message
+          when 401
+            raise Errors::AuthenticationFailed,
```
sneal: This shouldn't be needed with WinRM 1.3, it should always raise a `WinRMAuthorizationError`

test/unit/plugins/communicators/winrm/shell_test.rb
```diff
@@ -19,11 +28,12 @@
       expect(subject.powershell("dir")[:exitcode]).to eq(0)
     end
 
-    it "should raise auth error when exception message contains 401" do
-      expect(session).to receive(:powershell).with(/^dir.+/).and_raise(
-        StandardError.new("Oh no! a 401 SOAP error!"))
+    it "should raise auth error when WinRM exception has a response code of 401" do
+      # The default settings might an account lockout - 20 auth failures!
+      expect(session).to receive(:powershell).with(/^dir.+/).exactly(20).times.and_raise(
+        WinRM::WinRMHTTPTransportError.new("Oh no!!", 401))
```
sneal: Test should raise a WinRMAuthorizationError

sneal:
Testing this out, looks really good so far.

sneal:
@maxlinc Can you fix the merge conflict? This is tested and ready to be merged. Thanks for your efforts here, the error output is much better.

In the future we should make it fail fast if we get an auth error instead of retrying for a couple of minutes. At least now you see an auth error in stdout instead of Vagrant appearing to be frozen.

maxlinc:
@sneal merged.

PR #4983, opened 2014-12-15
WinRM synced folder plugin

maxlinc:
This adds (very low-priority) support for WinRM synced folders. The real work was done by @sneal to implement the WinRM FileManager. I just exposed it as a plugin.

It's probably the slowest synced folder plugin, but it has the advantage of working directly over the WinRM communicator. So if the communicator is working and secure so is this synced folder plugin - there are no other provider, platform or capability limitations.

Currently this will never be selected as the default synced folder type, unless you are are a host where all other types are unusable. In particular, if "rsync" is installed on the host it will always win over WinRM, both because the rsync usability check only takes the host into account, and because there are capability detection bugs for Windows guests that result in false positives.

So, Vagrant will sometimes pick a non-functional rsync or other synced folder implementation over a working WinRM synced folders. Right now the only way to use this is to explicitly set a specific synced folder's type to winrm.


maxlinc:
@sneal I'd like to switch this over to the winrm-fs gem. Does that gem now use zip files for uploads by default or would I need to enable that somehow? Is there a way to pass the excludes patterns from Vagrant to winrm-fs?

sneal:
@maxlinc Branch won't cleanly merge.

maxlinc:
So it looks like I can fix it to use winrm-fs and it'll be fairly clean... but you should probably put out a new release of winrm-fs first. I'm finding bugs and performance issues.

Also, doing things the way I'd like would require a change to Vagrant and/or winrm-fs interfaces.

I can almost do this:

```ruby
machine.shell.upload(hostpath, guestpath)
```

I actually have to do this, which looks a little more bizarre:

```ruby
machine.communicate.shell.upload(hostpath, guestpath)
```

That's actually not the main issue, though. The problem is I need to exclusion patterns. So I reallyw ant to do this:

```ruby
machine.shell.upload(hostpath, guestpath, exclude: exclusion_patterns)
```

The shell upload doesn't take any options, and winrm-fs doesn't have support for exclusion patterns. So what I end up forced to do is:

```ruby
winrm_session = machine.communicate.shell.new_session # only works if new_session is changed from protected to public...
file_manager = WinRM::FS::FileManager.new(winrm_session)

files = []
Find.find(hostpath) do | file |
  excludes.each do | pattern |
    Find.prune if File.fnmatch?(pattern, file, File::FNM_DOTMATCH)
  end
  files << file
end

file_manager.upload(files, guestpath)
```

This is not ideal, especially if many files exist - because I need to traverse through the files once to build the list and then winrm-fs traverses through them a second time. Plus it means winrm-fs doesn't have the option of letting something else handle the exclusion (e.g. `zip -x`).

So what do you think? I think the exclusion logic should definitely be moved into winrm-fs. I guess I'll move some of new_session to VagrantPlugins::CommunicatorWinRM::Helper, though I'm not sure if I can move the logging so I may leave that in the shell.

maxlinc:
Updated. Technically it "works" now in that it's just delegating the upload to winrm-fs. But so far I haven't gotten past https://github.com/WinRb/winrm-fs/issues/8 in my testing.

mwrock:
Just curious. Is this necessary now that vagrant supports smb? I have no problem setting up synced folders to windows boxes both from windows and ubuntu hosts. 

Are there issues with Mac's?

maxlinc:
> Is this necessary now that vagrant supports smb? 

Currently SMB is only supported if the host system is Windows. There is a PR to add limited support beyond Windows (https://github.com/mitchellh/vagrant/pull/4999), but it is subject to a lot of limitations.

Also, the default for OSX before Yosemite is SMB2, which is not very secure and has had a lot of bugs. See: http://www.macwindows.com/Yosemite-to-default-to-SMB-3-file-sharing.html.

>  I have no problem setting up synced folders to windows boxes both from windows and ubuntu hosts

Which vagrant provider and synced folder type? Local providers like VirtualBox, VMWare or Hyper-V aren't really an issue, but cloud providers are more difficult, especially if you want to make sure the connection is secure.

mwrock:
gotcha. that makes total sense.

maxlinc:
@sneal So this probably will won't be truly usable until the next winrm-fs release, but in the meantime can you look at #5013? Right now there's no good way to choose the default synced folder type, so even if there was a winrm plugin it would be difficult to tell Vagrant to use it.

You can easily override the type of the `/vagrant` share, but that doesn't help you if you're using plugins, and with plugins it's difficult to know which folders will be shared.

sethvargo:
Hi @maxlinc 

Thank you for the Pull Request and I am sorry it has taken so long for us to review. We have updated the version of winrm-fs in master, so you should be good to go. Unfortunately this will no longer merge cleanly. Could you please rebase and maybe squash into a single commit to make review/history easier? I am going to defer to @sneal on actual implementation details though - just doing triage :smile:. Great work so far!

mitchellh:
This looks interesting, but I agree its utility is questionable. We now have a file provisioner in Vagrant (which can also upload directories), and this does so over the communicator. I think this makes more sense as a provisioner (which already exists).

Can you describe why that wouldn't be the case?

mitchellh:
Closing due to a lack of response. Thanks for this PR, it is amazing, but I do want to understand motivation more.

sneal:
@mitchellh This would enable using the vagrant-aws plugin with Windows. You [can make vagrant-aws work with Windows](http://engineering.daptiv.com/provisioning-windows-instances-with-packer-vagrant-and-chef-in-aws/) but it requires the [vagrant-winrm-syncedfolders](https://github.com/Cimpress-MCP/vagrant-winrm-syncedfolders) plugin. This PR negates the need for the vagrant-winrm-syncedfolders plugin.

PR #4999, opened 2014-12-16
support SMB synced folders on non-Windows hosts

ntamas:
The attached patch adds (partial) support for synced folders using SMB when the host machine is not running Windows.

The limitations of the current implementation (compared to the case when the host machine is running Windows) are as follows:
- The IP address of the host machine is not detected automatically, therefore one has to specify `smb_host` explicitly in the `Vagrantfile`
- Since it is quite hard to force a Linux or Mac OS X host to share a specific folder using SMB, it is assumed that the folder that the user wants to mount in the guest machine is _already_ shared by the host, and Vagrant will make no attempts to share it to the guest.
- By default it is assumed that the name of the SMB share on the host is equal to the `id` of the synced folder. If this is not the case, one can use the `smb_id` property of the synced folder in the `Vagrantfile` to specify the name of the SMB share.

We have tested the patch using Scientific Linux and Mac OS X as hosts with a Windows 8.1 guest. All the unit tests pass after applying this patch.

### DIFF
--- a/plugins/synced_folders/smb/errors.rb
+++ b/plugins/synced_folders/smb/errors.rb
@@ -29,6 +29,10 @@ module VagrantPlugins
       class WindowsAdminRequired < SMBError
         error_key(:windows_admin_required)
       end
+
+      class SMBIdRequired < SMBError
+        error_key(:smb_id_required)
+      end
     end
   end
 end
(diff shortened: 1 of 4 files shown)


mitchellh:
This looks good, but my main concern is that this doesn't also setup the share for the user. I understand that is confusing, but there isn't a big different to using this vs. just setting up a shell provisioner? I suppose if we make it so SMB isn't default-able, then this would be okay.

briancain:
This PR is a little of out date, especially after #9294 . I'm going to go ahead and close this. Thanks!

PR #5013, opened 2014-12-17
Allow filtering/reprioritization of synced folder preferences

maxlinc:
Currently Vagrant selects the default synced folder type based on hardcoded priorities. It will select the highest-priority "usable" plugin. Usable doesn't necessary mean working, it usually just checks that pre-requisite software is either installed or can be installed automatically. It might still be unusable in practice because of security policies or the need for additional configuration.

I've also found that the order is not always deterministic. Currently rsync and nfs have equal priority. I've seen some users report that NFS is selected, and others report that rsync is selected, when both are "usable" on their system. The `config.nfs.functional = false` property can let you disable nfs to ensure rsync is selected, but there currently isn't a way to disable rsync and force NFS to be selected (other than customizing the settings of *each* shared folder, which requires knowing the synced folders setup by provisioners and other plugins).

This PR adds a property that both filters and reprioritizes the available synced folder types before selecting a default type. This way you can force a specific selection:
```ruby
  config.vm.allowed_synced_folder_types = :rsync
```

Allow Vagrant to choose from a subset of available types:
```ruby
  # Will use rsync if it's usable, nfs if it isn't
  config.vm.allowed_synced_folder_types = [:rsync, :nfs]
```

And swap the preferences:
```ruby
  # Will use nfs if it's usable, rsync if it isn't
  config.vm.allowed_synced_folder_types = [:nfs, :rsync]
```

In addition to giving the user more control of the priorities so they can easily choose nfs vs rsync, this is currently the only way to select a very low-priority synced folder type like [WinRM](https://github.com/mitchellh/vagrant/pull/4983) over higher-priority and nearly always "usable" types like rsync.

### DIFF
--- a/lib/vagrant/action/builtin/mixin_synced_folders.rb
+++ b/lib/vagrant/action/builtin/mixin_synced_folders.rb
@@ -27,9 +27,11 @@ module Vagrant
 
           allowed_types = machine.config.vm.allowed_synced_folder_types
           if allowed_types
-            ordered = ordered.select do |_, key, impl|
-              allowed_types.include? key
-            end
+            ordered = allowed_types.map do |type|
+              ordered.find do |_, key, impl|
+                key == type
+              end
+            end.compact
           end
 
           # Find the proper implementation
--- a/lib/vagrant/action/builtin/mixin_synced_folders.rb
+++ b/lib/vagrant/action/builtin/mixin_synced_folders.rb
@@ -25,6 +25,13 @@ module Vagrant
           # Order the plugins by priority. Higher is tried before lower.
           ordered = ordered.sort { |a, b| b[0] <=> a[0] }
 
+          allowed_types = machine.config.vm.allowed_synced_folder_types
+          if allowed_types
+            ordered = ordered.select do |_, key, impl|
+              allowed_types.include? key
+            end
+          end
+
           # Find the proper implementation
           ordered.each do |_, key, impl|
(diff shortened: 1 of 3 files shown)


sneal:
This LGTM, but I think @mitchellh needs to take a look at this since he's definitely more knowledgable in this area. If anything this could use some website documentation.

mkuzmin:
I've tested this PR briefly, and it works great!

maxlinc:
Ok. I can draft a short paragraph in the documentation based on the examples above once @mitchellh confirms this looks the right approach, as opposed to some other solution for choosing/prioritizing types.

mkuzmin:
Any chance hot have this merged before next release?

mitchellh:
LGT

Craige:
I realise this was just merged, but what happens in the case where you previously had:

    config.vm.synced_folder ".", "/var/www", :nfs => { :mount_options => ["dmode=777","fmode=777"] }

Under the new system, I would expect the following to work, but it does not:

    config.vm.allowed_synced_folder_types = [:nfs => { :mount_options => ["dmode=777","fmode=777"] }, :rsync]
    config.vm.synced_folder ".", "/var/www"

akaSStalkALEX:
I don't understand was this approved at the end or not? I have this in my config "config.vm.allowed_synced_folder_types = :virtualbox", but "vagrant up" keeps asking rsync anyway. Is there any other way to change sync type of default box?

PR #5030, opened 2014-12-19
Being able to edit a file does not mean you can write in its parent directory

willdurand:
Interactive `sed` needs write permission on the file itself. However, it may create a backup file, which leads to the fact that the directory where the file is located must be writable as well. That is a side effect because this directory does not need to be writable.

This patch fixes this side effect by editing the file in `/tmp`, and replaces it right after.

### DIFF
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -102,7 +102,7 @@ module VagrantPlugins
 
           # Use sed to just strip out the block of code which was inserted
           # by Vagrant
-          system("#{sudo_command}sed -r -e '\\\x01^# VAGRANT-BEGIN:( #{user})? #{id}\x01,\\\x01^# VAGRANT-END:( #{user})? #{id}\x01 d' -ibak /etc/exports")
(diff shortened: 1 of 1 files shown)


willdurand:
@mitchellh thoughts on this? Sorry for finding such edge cases.

willdurand:
Of course, thanks for the feedback :-)

plugins/hosts/linux/cap/nfs.rb
```diff
@@ -102,7 +102,7 @@ def self.nfs_cleanup(id)
 
           # Use sed to just strip out the block of code which was inserted
           # by Vagrant
-          system("#{sudo_command}sed -r -e '\\\x01^# VAGRANT-BEGIN:( #{user})? #{id}\x01,\\\x01^# VAGRANT-END:( #{user})? #{id}\x01 d' -ibak /etc/exports")
+          system("cp /etc/exports /tmp && #{sudo_command}sed -r -e '\\\x01^# VAGRANT-BEGIN:( #{user})? #{id}\x01,\\\x01^# VAGRANT-END:( #{user})? #{id}\x01 d' -ibak /tmp/exports ; cp /tmp/exports /etc/exports")
```
sethvargo: I think we might want to use `$TMPDIR` instead of `/tmp` here in case the user has a custom temporary directory defined. Thoughts?

willdurand: Yes, makes sense!

willdurand:
@sethvargo I've deleted the repo, so GitHub was not able to update the PR.. I've created a new PR: #5773.

sethvargo:
Closing in favor of https://github.com/mitchellh/vagrant/pull/5773

Issue #5005, opened 2014-12-17
Feature request: add a global public key insertion


bartmeuris:
## Intro

Since Vagrant 1.7, if it detects the insecure SSH key on a box when bringing it up, it removes the key and generates a new-one specific for that box.

Security-wise, this is a good thing, although because 99% of these boxes probably run in development environments and because vagrant environments are very volatile, security is not as high as practicality on my list of priorities. It is still important though.
## The problem

The problem with this change in 1.7.x is that it impacts practicality a lot, certainly in a Windows environment. I suspect most people won't use `vagrant ssh` there (I don't) because of the horrible Windows console with a hacky terminal emulation, where resizing (even with stuff like Console2)  brings you a world of pain. I suspect most people will fall back to PuTTY or another SSH client to connect to their boxes, and while it's not perfect, it works.

This change of removing the 'insecure' ssh key in combination with PuTTY (or any other external ssh client) becomes problematic:
- PuTTY does not understand OpenSSH keys directly, you have to convert them for each and every box.
- You have to know the path and provider of the box you're trying to connect to to load (or convert) the key if you're trying to automate this (this was my initial thought)
- external provisioning scripts using ssh (which I have for various reasons), suddenly need more info than just the port they need to connect to.
## Proposed solution

Therefor I propose the following: Add a central configuration option where you specify public keys that should be injected into the VM when it detects they are missing, just like the insecure keys are detected and new-ones are inserted.

This offers the following:
- this does not impact security any more than the current private key generation
- allows people to use their own private keys to connect to their vagrant boxes, and can easily load them in a key-agent, or configure their ssh profiles to use their own private key.
- Vagrant does not need to know the private keys, it can still use it's own privately-generated-ones to communicate with the box.

I think this offers best of both worlds, I hope this will be considered.

ccope:
+1 for adding an option to specify an additional public key(s) to insert into the VM. Our users need to have persistent keys for their VM's, but we also cannot leave the vagrant public key due to security issues.

nickryand:
I wanted to point out that @philr wrote a ruby library to convert OpenSSH keys to the putty format and added it to the vagrant-multi-putty for vagrant. Keys will now be converted automatically when `vagrant putty` is used.

https://github.com/nickryand/vagrant-multi-putty/commit/87532b39c48fb77fadbbf53c403d90a960deb9f0

chrisroberts:
Hi there,

[...]

Issue #3396, opened 2014-04-05
Ansible Provisioner: Support Guidelines for ansible-playbook options


gildegoma:
With current open issues like
- #2103 (running ansible-playbook on the guest machine)
- #2924 (support of interactive prompts)
- #3241 / #3338 (support of vault options)
- #3431 (include vagrant default SSH identity into generated inventory)
- Why not deprecate/drop "official support" of `paramiko` mode? 
  - #3900: tricky to be potentially ready for smart mode (aka paramiko or ssh)
  - `ssh` is needed by #4071, ssh-forwarding, multiple ssh identities
  - smart mode corresponds to `ssh` in a large majority of vagrant setups
- support of new options introduced in Ansible 1.5.x (e.g. `--su-user`)

I think it is now necessary to clarify what the Ansible provisioner can/should offer, according to following facts:
- With Host-based provisioning, people might want to manually execute `ansible-playbook` or `ansible` commands against the Vagrant VMs.
- Guest-based provisioning cannot support interactive prompts (in Vagrant 1.x at least)
- Provisioner Advanced/Expert modes are not necessarily supported (especially if the integration is difficult to implement or use)
- `ansible.raw_arguments`, `ansible.raw_ssh_args` and an Ansible configuration file (e.g. local `ansible.cfg`) can be easily used to customize a lot of things.

The idea is here to establish a simple check-list to take decision about adding (or removing) support of provisioner options, being at maximum "user friendly and backward compatible". Feel free to bring your thoughts!

Issue #1784, opened 2013-06-02
Support parallel multi-machine provisioning using Ansible


pesho:
Ansible has the ability to provision multiple machines in parallel. Taking advantage of this is desirable, since it is much faster than provisioning machines sequentially.

Currently (v1.2.2) Vagrant executes the provisioner for each machine separately during "vagrant up" (for multi-machine configurations). I've found no way to make it start all machines up, and then run the provisioner once. Naturally, I tried a Vagrantfile like this:

```
Vagrant.configure("2") do |config|

    config.vm.box = "precise64"
    config.vm.box_url = "http://files.vagrantup.com/precise64.box"

    config.vm.define :test_vm1 do |cfg|
        cfg.vm.network :private_network, ip: "172.20.30.10"
        cfg.vm.provider :virtualbox do |v|
            v.name = "test_vm1"
        end
    end

    config.vm.define :test_vm2 do |cfg|
        cfg.vm.network :private_network, ip: "172.20.30.11"
        cfg.vm.provider :virtualbox do |v|
            v.name = "test_vm2"
        end
    end

    config.vm.provision :ansible do |ansible|
        ansible.playbook = "ansible/testservers.yml"
        ansible.inventory_file = "ansible/stage"
        ansible.sudo = true
    end

end
```

... However, the above does not work as intended due to the way configs are merged - the global provisioning configuration is merged with the machine-specific ones, which results in running the provisoner for each machine separately when it starts. One could of course use machine-specific provisoning configurations instead of a global one (and use ansible.limit to specify the target hosts), but this loses the benefits of the parallel provisioning.

It would be nice if Vagrant allowed us to run the provisioner only _after_ all machines were brought up.

Caroga:
+1 a lot !

hryamzik:
You can always set ```somevm_cfg.ssh.insert_key = false``` to all the VMs. There also might be a way to generate key only once.

hryamzik:
And it's not a good idea to put provision section to a specific vm as you won't be able to provision individually. There must be an easy way to take the first element but I've simply added iteration limit:

```
  i = 0;
  config.vm.provision :ansible do |ansible|
    if i <= 0
      ansible.playbook = "playbook.yml"
      ansible.sudo = true
      ansible.limit = 'all'
      i+=1;
    end
  end
```

astorije:
[...]

PR #5044, opened 2014-12-21
Use `ansible_ssh_private_key_file` in generated inventory 

lpabon:
Vagrant 1.7.1 creates new ssh keys for each virtual machine. When
it started ansible, it would only send the ssh key of the last
virtual machine it created.  With this change, vagrant now provides
ansible with the ssh key for each of the virtual machines.

Signed-off-by: Luis Pabón <lpabon@redhat.com>

### DIFF
--- a/plugins/provisioners/ansible/provisioner.rb
+++ b/plugins/provisioners/ansible/provisioner.rb
@@ -16,7 +16,7 @@ module VagrantPlugins
         #
 
         # Connect with Vagrant SSH identity
-        options = %W[--private-key=#{@ssh_info[:private_key_path][0]} --user=#{@ssh_info[:username]}]
+        options = %W[--user=#{@ssh_info[:username]}]
 
         # Connect with native OpenSSH client
         # Other modes (e.g. paramiko) are not officially supported,
@@ -110,7 +110,8 @@ module VagrantPlugins
               m = @machine.env.machine(*am)
               m_ssh_info = m.ssh_info
               if !m_ssh_info.nil?
-                file.write("#{m.name} ansible_ssh_host=#{m_ssh_info[:host]} ansible_ssh_port=#{m_ssh_info[:port]}\n")
+                # Pass the private key generated for each VM to ansible
+                file.write("#{m.name} ansible_ssh_host=#{m_ssh_info[:host]} ansible_ssh_private_key_file=#{m_ssh_info[:private_key_path][0]} ansible_ssh_port=#{m_ssh_info[:port]}\n")
                 inventory_machines[m.name] = m
               else
                 @logger.error("Auto-generated inventory: Impossible to get SSH information for machine '#{m.name} (#{m.provider_name})'. This machine should be recreated.")
--- a/test/unit/plugins/provisioners/ansible/provisioner_test.rb
+++ b/test/unit/plugins/provisioners/ansible/provisioner_test.rb
@@ -269,7 +269,7 @@ VF
         config.host_key_checking = true
       end
 
-      it_should_set_arguments_and_environment_variables 6, 3, true
+      it_should_set_arguments_and_environment_variables 5, 3, true
     end
 
     describe "with boolean (flag) options disabled" do
@@ -281,7 +281,7 @@ VF
         config.sudo_user = 'root'
       end
 
-      it_should_set_arguments_and_environment_variables 7
+      it_should_set_arguments_and_environment_variables 6
       it_should_set_optional_arguments({ "sudo_user" => "--sudo-user=root" })
 
       it "it does not set boolean flag when corresponding option is set to false" do
@@ -310,7 +310,7 @@ VF
                                 "--new-arg=yeah"]
       end
 
-      it_should_set_arguments_and_environment_variables 17, 4, false, "paramiko"
+      it_should_set_arguments_and_environment_variables 16, 4, false, "paramiko"
 
       it "sets all raw arguments" do
         expect(Vagrant::Util::Subprocess).to receive(:execute).with { |*args|
@@ -366,7 +366,7 @@ VF
         config.ask_vault_pass = true
       end
 
-      it_should_set_arguments_and_environment_variables 7
+      it_should_set_arguments_and_environment_variables 6
 
       it "should ask the vault password" do
         expect(Vagrant::Util::Subprocess).to receive(:execute).with { |*args|
@@ -380,7 +380,7 @@ VF
         config.vault_password_file = existing_file
       end
 
-      it_should_set_arguments_and_environment_variables 7
+      it_should_set_arguments_and_environment_variables 6
 
       it "uses the given vault password file" do
         expect(Vagrant::Util::Subprocess).to receive(:execute).with { |*args|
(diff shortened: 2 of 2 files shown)


gildegoma:
@lpabon thanks for all the details.

> I think that the proposed patch is simple enough to enable parallel provisioning using ansible.

The problem with the change that you proposed is that it will break Vagrant configurations that use a a custom `ansible.inventory_path` and assume that Vagrant pass the private key file with the ansible command line argument `--inventory-file`.

At the moment, the Ansible parallel provisioning is not a feature officially supported by Vagrant (see #3396 and #1784). With Vagrant 1.7+, this trick now require to disable "insecure key replacement" with `config.ssh.insert_key = false`.

For all these reasons, I close it and beg you to:
 1. rely on the above workaround 
 1. wait for #5005 feature (global public key insertion)

gildegoma:
Still pending: update the docs about that specific point (see #5048)

gildegoma:
@lpabon Sorry, I reopen (still thinking on it, but not convinced though...). I first have to check the `ansible_ssh_private_key_file` priority handling (it never good when Vagrant defaults are hard to be overridden). My worry is to mess things out between CLI arguments, ANSIBLE_SSH_ARGS (for multiple SSH identities) and the generated - or non-generated - inventories. 

The most notable arguments in favor of this, is to allow external `ansible-playbook` calls, but on the other hand it is quite unusual in Ansible practice to deploy a different public key to each nodes.

More feedbacks are welcome, as usual!

//cc @maspwr

lpabon:
@gildegoma No problem at all.  Let me know if you need my help.

maspwr:
I'm actually not sure why `private-key` or `user` are command line options in the first place. Seems like defining in the inventory would be fine. My interpretation of http://docs.ansible.com/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable is that they would still be overwritable with command line options (though I haven't tested this).

If it's possible to make this change without any breaking changes (for `user` as well), I would be for it.

gildegoma:
@maspwr thanks for jumping into the discussion, I'm sure that will help us a lot to take good (optimal) decisions :)

If we do some changes in this way, we'll also have to think about how to deal with static inventory defined with `ansible.inventory_path`. For backwards compatibility, `ansible_ssh_user`and `ansible_ssh_private_key_file ` should be optional in this case.

@lpabon @maspwr for the records, we already had some discussion around this question. See:
- #3431 former proposition to move `user` and `private-key` to generated inventory
- https://github.com/mitchellh/vagrant/pull/2710#issuecomment-31693551 about per-task remote user
- https://github.com/mitchellh/vagrant/pull/2710#issuecomment-31215721 about `user` option precedences:

By the time I tested it (December 2013), the precedence rules for the `user` parameter were the following:

 1. `ansible_ssh_user` in inventory file (or as [extra variable](https://github.com/al3x/sovereign/blob/b77bcdbb0c7fa7e66d4e51e17ed88ede56614fee/Vagrantfile#L32))
 1. **`remote_user` or `user` YAML attributes in playbook file** (recently discovered, and that's certainly your case ;-)
 1. `-u / --user` ansible-playbook argument
 1. ANSIBLE_REMOTE_USER environment variable
 1. `remote_user` parameter in an `ansible.cfg` file
 1. username of the current user

When not wrong, `ansible_ssh_user` in inventory should only be overridden by an extra variable.

I think that first of all, we need to clearly identify the consequences of such change about two aspects:
* usage intuitiveness (avoid RTFM)
* full-feature flexibility (avoid something totally impossible to achieve within Vagrant context)

Note that I intentionally don't set as top priority to be able to call `ansible-playbook` outside from Vagrant. This is an important *a nice to have*, of course, and I keep it mind. But we already show the used command to help users in to build their own commands. Vagrant is thought to automatically and easily sbuild development environment, does not aim at support any possible advanced usages of Ansible (I personally to combine Vagrant and Ansible in some contexts, but also love to use ansible alone in many other situations ;-)

If we decide to go furhter, we'll also need to do "precedence analysis" about all the possibilities to set the effective SSH private key**S** (The ansible provisioner also implement Vagrant multiple private keys feature (`config.ssh.private_key_path = [...]`), via `ANSIBLE_SSH_ARGS` when ``, as such feature is not provided by Ansibe itself. See #2952).

I hope I was clear enough... :astonished: More thoughts?

vermut:
Any plans on merging this any time soon?

conorsch:
Any progress here? I've implemented f2e87411f60478a9bb4444b4ce20f549ab7457b5 locally, which is great to protect access to the VMs I manage on my workstation, but doesn't help developers and other contributors I work with.

gildegoma:
I've verified a few things about this PR and think we could get it merged. As explained below, there shouldn't be any breaking change, so I hope we can make it for Vagrant 1.7.3.

> [...] "precedence analysis" about all the possibilities to set the effective SSH private keyS.

Ansible (tested with 1.9.1) behaves with following precedence rules (listed from lowest to highest priority):

* [`private_key_file`](http://docs.ansible.com/intro_configuration.html#private-key-file) defined in an Ansible configuration file (e.g. local `ansible.cfg`)
* `ANSIBLE_PRIVATE_KEY_FILE` environment variable
* `--private-key` command line argument
* `ansible_private_key_file=...` value(s) in the inventory file

So far, I haven't find any another "official" way to pass the private key (An extra variable `ansible_ssh_private_key_file` has no effect, unlike `ansible_ssh_user`).

All these settings only allow a single key file per machine (concurrent settings with lower priority are superseded). But additional private keys can be passed via `ansible.raw_ssh_args (e.g. `ANSIBLE_SSH_ARGS="-o IdentityFile=... -o ..."`).

> The ansible provisioner also implements Vagrant multiple private keys feature, via ANSIBLE_SSH_ARGS when config.ssh.private_key_path is set to an array of keys.

So, Good News: The additional private keys are correctly passed to the ssh command by `ansible-playbook`, even when `ansible_ssh_private_key_file` is set in the inventory.

> how to deal with static inventory defined with `ansible.inventory_path`. For backwards compatibility, `ansible_ssh_user` and `ansible_ssh_private_key_file` should be optional in this case.

This one is maybe the trickiest part:
* As [previously mentioned](https://github.com/gildegoma), we won't set `ansible_ssh_user` in the generated inventory (and this topic was never part of this PR)
* After this change, static inventories still must not be forced to include `ansible_ssh_private_key_file`. 

To address this last point, we still will have to pass the key(s) defined in Vagrant's `config.ssh.private_key_path` parameter. At the moment, I think it is maybe "less confusing" ( :unamused: ) to pass **ALL** the private keys via ANSIBLE_SSH_ARGS, and totally drop `--private-key` arg).

@lpabon I'm going to pull your original commits in a new dev branch and try to open a new pull request in the next days. All you Guys: Feedbacks and tests will be welcome, for sure ;-)

conorsch:
@gildegoma Thanks for the detailed update. I follow your logic about deferring to ANSIBLE_SSH_ARGS in the case of static inventory files, although I don't use them often myself. More than happy to test a new branch to get this merged.

gildegoma:
Thanks @conorsch for the offer (I'm also not true user of custom inventories, but I try to take care of every use case ;-).

So, I pushed my additions here: https://github.com/mitchellh/vagrant/tree/gildegoma/key-in-ansible-inventory.

I'm thus closing this PR in favour of #5765, and would love to get your feedbacks.

PS: @lpabon I rebased and reworded your original commits. Please check it out, and tell me if you want to modify something (especially my english phrasing... :no_entry: :fire_engine:)

lpabon:
@gildegoma Thanks!

PR #5077, opened 2014-12-28
Log chef-zero instead of chef-solo

sethvargo:
Fixes https://github.com/mitchellh/vagrant/issues/5072


darkn3rd:
If chef-solo is the actual command that is called, then is chef-zero even used?

PR #5079, opened 2014-12-29
FIX [mitchellh/vagrant#4889] .curlrc ignored

rivy:
change "-q" to "--no-progress-bar" in downloader.rb and downloader_test.rb

The initial fix for mitchellh/vagrant#4494 created a regression by causing curlrc to be completely ignored. By ignoring curlrc, significant curl functionality, such as proxy use, configuration of CA certs, and out-of-source password storage (via netrc), was made difficult or impossible.

This repair, suggested by @slicen, more narrowly repairs the original issue (mitchellh/vagrant#4494) and still allows for the use of curlrc configuration directives.

[1] mitchellh/vagrant#4889
[2] mitchellh/vagrant#4494

### DIFF
--- a/test/unit/vagrant/util/downloader_test.rb
+++ b/test/unit/vagrant/util/downloader_test.rb
@@ -76,7 +76,7 @@ describe Vagrant::Util::Downloader do
 
   describe "#head" do
     let(:curl_options) {
-      ["-q", "--fail", "--location", "--max-redirs", "10", "--user-agent", described_class::USER_AGENT, source, {}]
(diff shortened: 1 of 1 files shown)


rivy:
@sethvargo @mitchellh There hasn't been any discussion about this in a while but I think FTP and SCP/SSH are important access methods for boxes. Any thoughts on the best way to continue support for FTP and SCP/SSH with out-of-source credentials (eg, using netrc)? Should I re-raise this as an enhancement issue?

rivy:
@sethvargo @mitchellh , +bump.

SCP support is still not working out-of-the-box on Win10 systems. Any plans to add support, either through an updated `curl` or other means?

PR #5083, opened 2014-12-29
Gem: reintroduce json dependency

simon3z:
Signed-off-by: Federico Simoncelli <fsimonce@redhat.com>


mitchellh:
Yeah we removed this because we use standard lib json now. If there is a gem that is depending on json then that gem is pretty broken because Ruby has included `json` in the stdlib for many years now (I think only 1.8 didn't have it standard?). 

We had a LOT of problems with json as a dep as a gem that we fought with for many years and ultimately fell back to stdlib once we could drop 1.8 support. 

I'm guessing  your `GEM_HOME` just isn't including the stdlib gem dir (I think they split it out to gems now, just bundled with Ruby). The Vagrant installers sets this up so in practice this should never be an issue unless you're installing outside the installer, in which case YMMV.

PR #5084, opened 2014-12-29
Skip sudo when accessing as root

simon3z:
Signed-off-by: Federico Simoncelli <fsimonce@redhat.com>

### DIFF
--- a/plugins/synced_folders/rsync/helper.rb
+++ b/plugins/synced_folders/rsync/helper.rb
@@ -97,7 +97,7 @@ module VagrantPlugins
         args << "--no-group" unless args.include?("--group") || args.include?("-g")
 
         # Tell local rsync how to invoke remote rsync with sudo
-        if machine.guest.capability?(:rsync_command)
+        if not ssh_info[:privileged] && machine.guest.capability?(:rsync_command)
           args << "--rsync-path"<< machine.guest.capability(:rsync_command)
         end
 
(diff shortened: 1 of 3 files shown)


mitchellh:
I'm guessing the use case is to avoid the configuration of `sudoers` which probably still asks for a password even if you're root, limits env vars, and various  other things. Or, `sudo` might not even be installed. 

LGTM.

mitchellh:
There are some style changes I'll fix myself and CC you in .

plugins/synced_folders/rsync/helper.rb
```diff
@@ -97,7 +97,7 @@ def self.rsync_single(machine, ssh_info, opts)
         args << "--no-group" unless args.include?("--group") || args.include?("-g")
 
         # Tell local rsync how to invoke remote rsync with sudo
-        if machine.guest.capability?(:rsync_command)
+        if not ssh_info[:privileged] && machine.guest.capability?(:rsync_command)
```
mitchellh: Why is this here? What is the use case of this line?

mitchellh:
Actually, one question, if you can...

sethvargo:
Hey @simon3z 

Thank you for the Pull Request. It looks like we made a few comments - would you please mind taking a look and responding when you have time? If you are no longer interested in pursuing this PR, please let us know :smile:.

mitchellh:
We added the `rsync_command` option to rsync. I think this is a more explicit way to do this.

PR #5085, opened 2014-12-30
Enable provision with --provision-with flag

evverx:
Type `vagrant reload --provision-with ...` instead of
`vagrant reload --provision --provision-with ...`.
Same as for `vagrant up`.


mitchellh:
LGTM. Thanks!

Issue #5093, opened 2015-01-01
fix for docker + virtualbox symlinks in shared folders


wkolean:
When using VirtualBox as a host for Docker containers, symlinks in Docker shared folders will fail. The reason is because the shared folder is added before symlinks are enabled, but it appears that symlinks have to be enabled first, then the share can be created. I'm not sure why symlinks currently work when just using VirtualBox, the setting persists as long as the  VM isn't destroyed, so maybe because the share names don't change for VirtualBox shares symlinks always work, but because the Docker share names change the setting doesn't stick.

I'm creating a pull request that enables symlinks first, then creates the share. This enables symlinks to work in Docker shared folders, and symlinks continue to work in the vbox host.

To see the issue, create Vagrantfile.proxy:
```
Vagrant.configure(2) do |config|
  config.vm.box = "phusion/ubuntu-14.04-amd64"
  config.vm.provision "docker"
end
```

And Vagrantfile:
```
Vagrant.configure("2") do |config|
  config.vm.define "myapp" do |myapp|
    myapp.vm.provider "docker" do |d|
      d.image = "phusion/baseimage"
      d.cmd = ["/sbin/my_init", "--enable-insecure-key"]
      d.has_ssh = true
      d.vagrant_vagrantfile = "Vagrantfile.proxy"
    end

    myapp.ssh.username = "root"
    myapp.ssh.private_key_path = "insecure_key"
  end
end
```
Get the ssh key:
```
curl -o insecure_key -fSL https://github.com/phusion/baseimage-docker/raw/master/image/insecure_key
```

Then run the following commands:
```
vagrant up
vagran ssh myapp
cd /vagrant
touch foo
ln -s foo bar
```

You'll see the following error:
```
ln: failed to create symbolic link 'bar': Read-only file system
```
Then apply the patch, which simply changes the order in share_folders() to first enable symlinks then add the shared folder. Now recreate the VMs and try again and this time creating the symlink should be successful.

mitchellh:
Fixed by #5094

PR #5094, opened 2015-01-01
enable symlinks before adding shared folder

wkolean:
Fix for issue #5093 to allow symlinks to work in Docker shared folders with using VirtualBox as the host.


mitchellh:
LGTM

Issue #5101, opened 2015-01-03
Start-VM exceptions are not caught


tomassrnka:
When starting VM on compressed disc, HyperV fails to start VMs and throws exception. Running VMs on compressed disc is unsupported on HyperV. At the moment, the start_vm.ps1 script does not catch exceptions correctly from Start-VM command and shows "Operating normally" output even though the start failed. This leads into strange behavior when doing e.g. "vagrant up".

## How to reproduce ##
Start VM or "vagrant up" on a disc with compression enabled. 

### vagrant up ###
```powershell
PS E:\VagrantSystems\precise64> vagrant up --provider=hyperv
Bringing machine 'default' up with 'hyperv' provider...
==> default: Verifying Hyper-V is enabled...
==> default: Starting the machine...
==> default: Waiting for the machine to report its IP address...
    default: Timeout: 120 seconds
Hyper-V failed to determine your machine's IP address within the
configured timeout. Please verify the machine properly booted and
the network works. To do this, open the Hyper-V manager, find your
virtual machine, and connect to it.

The most common cause for this error is that the running virtual
machine doesn't have the latest Hyper-V integration drivers. Please
research for your operating system how to install these in order
for the VM to properly communicate its IP address to Hyper-V.
```

### start_vm.ps1 script itself ###
```powershell
\> .\start_vm.ps1 7127b729-81a3-4e31-9ce2-eb81602ca0ff
Start-VM : 'precise64' failed to start. (Virtual machine ID 7127B729-81A3-4E31-9CE2-EB81602CA0FF)
'precise64' Microsoft Emulated IDE Controller (Instance ID 83F8638B-8DCA-4152-9EDA-2CA8B33039B4): Failed to Power on wi
th Error 'The requested operation could not be completed due to a virtual disk system limitation.  On NTFS, virtual har
d disk files must be uncompressed and unencrypted. On ReFS, virtual hard disk files must not have the integrity bit set
.' (0xC03A001A). (Virtual machine ID 7127B729-81A3-4E31-9CE2-EB81602CA0FF)
'precise64': Failed to open attachment 'E:\VagrantSystems\precise64\.vagrant\machines\default\hyperv\disk.vhdx'. Error:
 'The requested operation could not be completed due to a virtual disk system limitation.  On NTFS, virtual hard disk f
iles must be uncompressed and unencrypted. On ReFS, virtual hard disk files must not have the integrity bit set.' (0xC0
[...]

PR #5102, opened 2015-01-03
Fix catching "Start-VM" exceptions on Hyper-V during starting up VM

tomassrnka:
Fix for "Start-VM exceptions are not caught' #5101


sethvargo:
Looks good to me, but I'm not super familiar with Hyper-V, so waiting for :+1: from @sneal or @mitchellh

mitchellh:
This looks right. Thanks

Issue #4962, opened 2014-12-13
1.7.1 OSX 10.10 ArgumentError


Spechal:
OSX Yosemite
VirtualBox 4.3.20
Guest OS Ubuntu 14.10.1
Guest Additions installed 

```
$ vagrant package --base ubuntu-vagrant
/opt/vagrant/embedded/gems/gems/vagrant-1.7.1/lib/vagrant/machine.rb:153:in `action': wrong number of arguments (2 for 1) (ArgumentError)
    from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:83:in `package_vm'
    from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:66:in `package_base'
    from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:42:in `execute'
    from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/lib/vagrant/cli.rb:42:in `execute'
    from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/lib/vagrant/environment.rb:301:in `cli'
    from /opt/vagrant/bin/../embedded/gems/gems/vagrant-1.7.1/bin/vagrant:174:in `<main>'
```

symgryph:
I tried the fix still get the following:

 vagrant package --base CentosBase6.6 --output centostest.box

/opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:83:in
`core#hash_merge_kwd': wrong argument type String (expected Symbol)
(TypeError)

from
/opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:83:in
`package_vm'

from
/opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:66:in
`package_base'

from
/opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:42:in
`execute'

from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/lib/vagrant/cli.rb:42:in
`execute'

from
/opt/vagrant/embedded/gems/gems/vagrant-1.7.1/lib/vagrant/environment.rb:301:in
`cli'

from
[...]

PR #5106, opened 2015-01-05
really fix #4972 (#4962)

AntonioMeireles:
actual fix was from a comment from @alu in # 4962


AntonioMeireles:
plz ignore. pretended fix doesn't fix all issues. :/

PR #5108, opened 2015-01-05
Symbolize and stringify keys in options hash

sethvargo:
@mitchellh these are some remaining fixes from #4962 that users have reported


mitchellh:
Why flip flop between symbols and strings (one converts to sym, one to str)

sethvargo:
@mitchellh because `**opts` in Ruby **must** be symbols (which is what the original error was), but then the machine expects the keys are strings.

mitchellh:
Ah, hah, I had no idea about that actually, but now that I think about it it makes perfect sense.

PR #5113, opened 2015-01-05
Use option hashes instead of keywords args

sethvargo:
/cc @mitchellh this makes me really :panda_face: 

TL;DR you can't use string keys in `**opts` keyword hashes...

### DIFF
--- a/lib/vagrant/action/builtin/box_add.rb
+++ b/lib/vagrant/action/builtin/box_add.rb
@@ -283,7 +283,7 @@ module Vagrant
         # @param [String] provider
         # @param [Hash] env
         # @return [Box]
-        def box_add(urls, name, version, provider, md_url, env, **opts)
+        def box_add(urls, name, version, provider, md_url, env, opts = {})
           env[:ui].output(I18n.t(
             "vagrant.box_add_with_version",
             name: name,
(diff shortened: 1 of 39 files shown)


mitchellh:
Makes me a bit nervous to do this big of a change in a point release. Let's wait until 1.8.0 merge window to get this in. But I agree.

sethvargo:
@mitchellh well, without this `vagrant package` commands are broken in 1.7 entirely :frowning:

sethvargo:
And I don't know of a better way to fix this (that's what led to this rabbit hole). I don't know all the places where we could be passing string-keyed options, so I had to change them all. 

Basically, anywhere we invoke a method that uses `**opts`, we have to ensure the underlying has only contains symbol-based keys, but it is almost impossible to find all such occurrences :frowning:

mitchellh:
A lot of these I know from looking are enforced symbol-only. The package is a bit different because its going into the action runner which used strings. But almost everywhere else is symbols. 

We should just target the fixes to get the package working only. Or does it require all of these? I'd be surprised.

sethvargo:
@mitchellh okay. Merged in a tiny fix and we can target this for 1.8

PR #5114, opened 2015-01-05
Use .key? instead of .has_key?

sethvargo:
This fixes a Ruby deprecation issue I was seeing. Apparently `.has_key?` is deprecated in favor or `.key?` :smile:


mitchellh:
LGTM, just making sure, this works fine in Ruby 2.0 right?

sethvargo:
@mitchellh yea - works in Ruby 1.9 too even :smile:

PR #5130, opened 2015-01-07
Actually validate push configurations

sethvargo:
In 5903bfb3c6f767f05ad9f1c4c6ef5237de039ea8, we added validation for global configuration, but that does not actually invoke the individual push validations. As a result, the push-specific configs were never validated. The tests did not catch this because they test at a lower level and manually call `#validate`.

This is the root cause of `vagrant push` with the Atlas strategy not choosing the login token, since the login token is actually [set during validation](https://github.com/mitchellh/vagrant/blob/42b7f13790c779671e472010ac988c3401efc593/plugins/pushes/atlas/config.rb#L87-L94) because we require access to the machine object.

This PR also adds much-needed logging to `vagrant-login`.

/cc @mitchellh @pearkes

### DIFF
--- a/lib/vagrant/environment.rb
+++ b/lib/vagrant/environment.rb
@@ -556,7 +556,6 @@ module Vagrant
       end
 
       strategy, config = pushes[name]
-
       push_registry = Vagrant.plugin("2").manager.pushes
       klass, _ = push_registry.get(strategy)
       if klass.nil?
@@ -564,11 +563,6 @@ module Vagrant
           name: strategy,
           pushes: push_registry.keys
       end
-
-      # Validate the global push configuration and our local push configuration
-      machine = self.machine(self.machine_names.first, self.default_provider)
-      machine.action_raw(:config_validate, Vagrant::Action::Builtin::ConfigValidate)
-      config.validate(machine)
 
       klass.new(self, config).push
     end
--- a/lib/vagrant/environment.rb
+++ b/lib/vagrant/environment.rb
@@ -556,6 +556,7 @@ module Vagrant
       end
 
       strategy, config = pushes[name]
+
       push_registry = Vagrant.plugin("2").manager.pushes
       klass, _ = push_registry.get(strategy)
       if klass.nil?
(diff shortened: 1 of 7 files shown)


lib/vagrant/environment.rb
```diff
@@ -564,6 +565,11 @@ def push(name)
           pushes: push_registry.keys
       end
 
+      # Validate the global push configuration and our local push configuration
+      machine = self.machine(self.machine_names.first, self.default_provider)
+      machine.action_raw(:config_validate, Vagrant::Action::Builtin::ConfigValidate)
+      config.validate(machine)
```
mitchellh: Hm, why is this necessary? The root config should validate all the children (pushes included).

sethvargo: @mitchellh because it doesn't? The root config definitely isn't validating push children, which is the root cause of this issue...

mitchellh: Right, I believe you. I'm just pointing out that I think this isn't the right place to push this. The root should be fixed to validate children. I actually think I know where it is... let me link...

mitchellh: Yeah here it is, we need to do something like this for the push root: https://github.com/mitchellh/vagrant/blob/master/plugins/kernel_v2/config/vm.rb#L671

mitchellh: With that, then our validation right now will validate it properly. I also am not sure if its proper for the `#push` method to validate the push, since the `#machine` method doesn't validate the machine. I think it makes sense back in the `push` command.

sethvargo: @mitchellh I looked that that, but there's no `vm` in our case, since this only depends on the environment, so that doesn't trigger (at least it wasn't in my tests). I'm not sure how this should all be wired together since the config validation logic is a bit confusing to me, but this is definitely the bug.

mitchellh: I was just using the `vm` as an example. It has to go in the root `push` configuration, here: https://github.com/mitchellh/vagrant/blob/master/plugins/kernel_v2/config/push.rb

sethvargo:
@mitchellh re-review please? I switched to validating in the global config and updated the test to ensure the right error is happening.

mitchellh:
Seth, LGTM.

PR #5136, opened 2015-01-08
docs: Mention Docker and Hyper-V on provider index page

fgrehm:
A very small change but I'm not sure about the wording here so I thought it would be better to run it through some native english speakers :smile:


mitchellh:
Thanks! I'll rework this a bit, but I think it makes sense to call things out more. :)

PR #5145, opened 2015-01-08
Allow specifying a token at login

sethvargo:


mitchellh:
LGTM! I think it would probably help to do an `execute_check` alongside the `execute_token` so that we can verify the token when they save it. That is probably a better UX than `vagrant login -t` and then `vagrant up` to find out it didn't work at all.

Issue #5156, opened 2015-01-09
File provisioner copies folders as empty files


bvalerius:
When provisioning with the file provisioner and giving it a directory as the source it creates an empty file on the destination. This is because provisioners/file./config.rb should maybe use File.file? instead of File.exists?

This should go back at least as far as v1.3.0

bvalerius:
https://github.com/mitchellh/vagrant/pull/5157

bvalerius:
Hmm, I see now in Communicator#upload there is a check to see if the source is a directory, and if so it uses scp.upload! with recursive option set to true. I will test this more to be sure it's not my own mistake. I closed the pull request for now.

bvalerius:
It appears to be working on my test VM, although it was not working on an actual one I was using. I'll close this all for now and bring it back up if I find the reason it wasn't working.

PR #5157, opened 2015-01-09
fix for issue #5156

bvalerius:
File provisioner will no longer copy folders as empty files


bvalerius:
Refer to my comment in #5156 why I am closing this for now.

Issue #5168, opened 2015-01-11
Add option to specify VlanID to Network Adapter when using HyperV provider


tomassrnka:
The HyperV provider lacks option to specify VlanId to Network Adapter automatically. This is pretty painful as I need each VM's Network Adapter to use VLANID100. 

Current option is to change the network adapter's settings manually in HyperV's management or through powershell.

![screenshot 2015-01-11 18 08 11](https://cloud.githubusercontent.com/assets/1406694/5696010/fa8e8a14-99bc-11e4-9436-a334443904fb.png)

tomassrnka:
Pull request #5169

mitchellh:
I'll wait on your PR. Thanks!

PR #5169, opened 2015-01-11
(#5168) Add option to specify VlanID to Network Adapter when using HyperV provider 

tomassrnka:
Pull request adds option "vlan_id" option to VagrantFile which allows to specify vlan_id for each VM's Network Adapter. Default value is set to 0 (untagged).


tomassrnka:
There is a typo + missing docs. I will open new pull request once it's fixed

PR #5170, opened 2015-01-11
check for which binary

windowsrefund:
A quick attempt to overcome an issue I had with a Docker image where the `which` binary was not present.

https://github.com/mitchellh/vagrant/issues/5167

### DIFF
--- a/plugins/provisioners/chef/provisioner/base.rb
+++ b/plugins/provisioners/chef/provisioner/base.rb
@@ -34,9 +34,18 @@ module VagrantPlugins
           installer.ensure_installed
         end
 
+        def verify_which
+          @machine.communicate.sudo(
+            "which -v",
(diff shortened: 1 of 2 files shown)


sethvargo:
Hi @windowsrefund,

Thank you for the PR.

@mitchellh is there a better tool we could use besides `which`? Maybe `hash` or `command` to detect? Do we consider "which" a minimum system requirement for a guest?

mitchellh:
@sethvargo I agree, `command -v` is a POSIX standard: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html

mitchellh:
https://github.com/mitchellh/vagrant/commit/c1e2d36090493fa1620fe21148d2fc9bbbc04c1b

PR #5175, opened 2015-01-12
synced/folders/rsync Add 'IdentitiesOnly=yes'

hdeheer:
Added IdentiesOnly=yes to avoid "Too Many Authentication Failures" error


sethvargo:
This seems reasonable to me, but I'm not super familiar with other implications this may have...

acourtiol:
Would be really useful ! :+1:

mitchellh:
Yep, thanks!

Issue #5180, opened 2015-01-13
PR #3303 doesn't account for multiple private interfaces


clinstid:
As reported by @excieve in PR #3303, the change in PR #3303 doesn't account for configurations with multiple interfaces. The problem is that each interface has its own definition surrounded by `#VAGRANT-BEGIN` and `#VAGRANT-END` like:

```
#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth1
iface eth1 inet static
      address 10.1.20.100
      netmask 255.255.255.0
#VAGRANT-END
```

The problem is when you have more than one private interface defined:

```
#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth1
iface eth1 inet static
      address 10.1.20.100
      netmask 255.255.255.0
#VAGRANT-END

#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth2
iface eth2 inet static
      address 10.1.20.101
      netmask 255.255.255.0
#VAGRANT-END
```

The second sed command in `plugins/guests/debian/cap/configure_networks.rb` will print everything after the first `#VAGRANT-END` into `/tmp/vagrant-network-interfaces.post` and later concatenate that to `/etc/network/interfaces` creating a file with multiple definitions for any private interfaces beyond the first:

```
#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth1
iface eth1 inet static
      address 10.1.20.100
      netmask 255.255.255.0
#VAGRANT-END

#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth2
iface eth2 inet static
      address 10.1.20.101
      netmask 255.255.255.0
#VAGRANT-END

#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
auto eth2
iface eth2 inet static
      address 10.1.20.101
      netmask 255.255.255.0
#VAGRANT-END
```

The command `/sbin/ifup` doesn't accept this as valid syntax, so you get the reported failure.

[...]

PR #5181, opened 2015-01-13
(#5180) Fix interface definition with multiple private interfaces

clinstid:


clinstid:
Last commit will deal with cleaning up the interface definitions as they are generated by 1.7.2. However,  in testing I remembered that if you're in that state the network startup is going to fail anyway. Vagrant won't be able to ssh in and generate a working interfaces file, so the cleanup will never actually happen.

clinstid:
@sethvargo or @mitchellh - any chance we can get this PR merged and in a release soon? There's another solution in PR #5325 if you prefer that one. Personally, I don't care which PR is merged, I just want to be able to upgrade past 1.7.1 :)

Thanks.

sethvargo:
Hi @clinstid 

Whoops - I merged https://github.com/mitchellh/vagrant/pull/5325 earlier and didn't see this one linked. Sorry about that! Thank you for the PR and I am really sorry it has taken so long to review :smile:

clinstid:
No problem @sethvargo, I'm just happy to see it fixed. :)

PR #5182, opened 2015-01-14
Add insert/remove public key guest capability for solaris

clintoncwolfe:
Implementation identical to linux; tested on omnios 151012


mitchellh:
LGTM

PR #5183, opened 2015-01-14
Added possibility to configure memory, cpus count, and vmname in vagrant...

vvchik:
I think it is long-awaited feature for Hyper-V plugin.
This PR gives possibility to manipulate following parameters in Vagrant file for config.vm.provider :hyperv
1. memory: type allocation and size 
2. numbers of CPU
3. Virtual machine name (it is not hostname but visible in hyper-v manager name)
More details in documentation

regards, Volodymyr

### DIFF
--- a/plugins/providers/hyperv/action/import.rb
+++ b/plugins/providers/hyperv/action/import.rb
@@ -14,6 +14,15 @@ module VagrantPlugins
         def call(env)
           vm_dir = env[:machine].box.directory.join("Virtual Machines")
           hd_dir = env[:machine].box.directory.join("Virtual Hard Disks")
+          memory = env[:machine].provider_config.memory
+          maxmemory = env[:machine].provider_config.maxmemory
+          cpus = env[:machine].provider_config.cpus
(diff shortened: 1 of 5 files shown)


sethvargo:
Hi @vvchik 

This is looking good! Do you think you could add some tests for these new config options in `test/unit/plugins/providers/hyperv`?

adamralph:
This is superb. I'm sorely missing these features for hyper-v.

vvchik:
Hi @sethvargo 
I not so good in test writing as want to be :)
but little bit extend existing testing for work with new variables.

sethvargo:
@vvchik that's all I really wanted :smile:. The tests look great too! We just want to make sure we maintain the API between releases, so ensuring these new values aren't accidentally removed is a good test to me!

adamralph:
Is there any way to see which release this will be included in? Do you use milestones? I see a 1.7 milestone but we're already on 1.7.2 so I'm not sure what it means.

sethvargo:
@adamralph it will be in "the next release", which is likely going to be 1.7.3

Issue #4904, opened 2014-12-05
Support docker exec


schmunk42:
From 1.3 on docker introduced `docker exec`, see also [Inject new processes with docker exec](Inject new processes with docker exec).

Would be nice if vagrant could support this.

thomaszbz:
@sethvargo @wkolean This feature is so cool. would like to see this released soon.

wkolean:
It's available as a plugin at https://github.com/wkolean/vagrant-docker-exec

Issue #4179, opened 2014-07-12
Docker shell/provisioning without SSH


rasschaert:
Currently, Vagrant can only provision or shell into a Docker container when it has an SSH daemon running.

Last month Docker's Jérôme Petazzoni wrote a post on the Docker blog titled ["Why you don't need to run SSHD in your Docker containers"](http://blog.docker.com/2014/06/why-you-dont-need-to-run-sshd-in-docker/). In it, he introduces a small tool called [nsenter](https://github.com/jpetazzo/nsenter), which allows you to enter the namespace of a Docker container and run a program, such as a bash shell.

In the comments, Phusion's Hongi Lai, the creator of baseimage-docker, responds to the post:
> Now, you are advocating nsenter. I am seriously considering this option. There is currently an ongoing discussion on the baseimage-docker bug tracker about replacing SSH with nsenter: https://github.com/phusion/baseimage-docker/issues/102

Baseimage-docker is one of the more popular images to use with Vagrant, and Phusion were some of the most vocal advocates in favor of running SSHd inside docker containers.

Meanwhile, upstream, [Docker is committed to creating a native mechanism](https://github.com/dotcloud/docker/issues/1228) to execute new processes in existing containers. It was slated for the 1.0 release but has since been delayed. 

It'd be nice if Vagrant would allow for other mechanisms such as nsenter to provision or shell into containers.

SeMeKh:
`docker exec` is now a part of docker itself. Do you think it's a good idea to transparently use `docker exec` instead of ssh-ing into the container? It provides the same interface as SSH, and it shouldn't result in any incompatibility.

thomaszbz:
SSH is just overhead here if you can just use ``docker exec`` which is provided by docker. Vagrant should be capable to use both scenarios.

thatguystone:
Was this actually addressed in #7737?

bogdando:
No, it seems not possible to use the docker-exec as a replacemant to the ```vm.provision :shell```

jgillich:
This should be reopened @sethvargo

chrisroberts:
@jgillich Hi! Closed issues are not tracked. If you are still having a problem, please open a new issue. Thanks!

PR #5193, opened 2015-01-15
Add docker-exec command

wkolean:
Added docker-exec [options] container -- [command] which invokes `docker exec` to run a command in a running container #4904 #4179.

If the container is not running, docker-exec will start the container and leave the container running after the command has finished. This makes it easy to start a container and go directly into an interactive shell using one command `vagrant docker-exec -t app -- bash`

Also added a command for opening a shell, `vagrant docker-shell app` #4179 which is just a shortcut for `vagrant docker-exec -t app -- bash`

I basically hacked docker-run and docker-logs to get this to work, so if any of the code doesn't make sense then it's probably because I didn't fully understand what I was copying.

This has been tested on a Mac running VirtualBox and using vagrant_vagrantfile to define a phusion/ubuntu-14.04-amd64 box. It was also tested against containers that had both images and Dockerfiles. Command doesn't work using the internal boot2docker because the installed docker version is too old. There already is a pull request to update boot2docker, and this PR probably shouldn't merge until boot2docker is updated first.

Note that docker-run doesn't mount any shared folders from config.vm.synced_folder so docker-exec doesn't create synced folders either (except for /vagrant which still works). I think this is a bug, but I haven't found a solution. If I find one, I'll create a separate PR.


plugins/providers/docker/driver.rb
```diff
@@ -151,6 +151,17 @@ def docker_bridge_ip
         end
       end
 
+      def exec(params, **opts, &block)
+        cmd      = Array(params.fetch(:cmd))
+        exec_cmd = %W(docker exec)
```
sethvargo: We don't need interpolation, so lowercase `w` should be fine here

plugins/providers/docker/driver.rb
```diff
@@ -151,6 +151,17 @@ def docker_bridge_ip
         end
       end
 
+      def exec(params, **opts, &block)
```
sethvargo: We are moving away from keyword args and using `opts = {}` instead. The keyword args require a specific version of Ruby and only work for symbols.

plugins/providers/docker/command/exec.rb
```diff
@@ -0,0 +1,69 @@
...
+        end
+
+        def execute
+          options = {}
+          options[:detach] = false
+          options[:pty] = false
+          options[:rm] = false
+
+          opts = OptionParser.new do |o|
+            o.banner = "Usage: vagrant docker-exec [command...]"
```
sethvargo: I think this is actually

```
Usage: vagrant docker-exec [options] -- command
```

right?

plugins/providers/docker/command/exec.rb
```diff
@@ -0,0 +1,69 @@
...
+            machine.action(
+              :exec_command,
+              exec_command: command,
+              exec_detach: options[:detach],
+              exec_pty: options[:pty],
+              exec_rm: options[:rm]
+            )
+          end
+
+          0
```
sethvargo: An explicit `return` here would be nice :smile:

sethvargo:
Hi @wkolean 

Thank you for the Pull Request, and I apologize for the delay in reviewing. This looks _really_ good. I made a few style comments and added some questions. Please let me know if anything is unclear.

Finally, it would be great to have some test coverage around this. I know most Vagrant plugins are not well tested, but if take a look at the Chef provisioner or pushes, there is some decent test coverage/skeleton you can copy :smile:.

wkolean:
Thanks for the great feedback! I'll make the changes, and there is a stand-alone plugin I created with some updates that I'll port over. Adding tests is a good idea, thank you for the pointers.

plugins/providers/docker/driver.rb
wkolean: The docker provider makes heavy use of **opts, I'm not sure how to modify just my changes to not use **opts without refactoring the the rest of the plugin

thomaszbz:
@sethvargo @wkolean This feature is so cool. Are you still working on this? Would like to see it released soon.

Issue #3031, opened 2014-02-26
Starting a VM with Vagrant 1.4.3 on Windows 8.1 fails sometimes with a connection error


thasmo:
Sometimes `vagrant up` fails with a connection error, reporting some problems regarding ports. The weird thing is that if it happens, it repeatedly fails when trying to call `vagrant reload`, even multiple times.

I've figured out that it may be related to software running on the host machine (Windows 8.1) which uses a lot of ports, for example Spotify. When I have Spotify running and encounter the mentioned problem, after I close Spotify the next `vagrant reload` succeeds.

Unfortunately I have no error log at the moment, but thought I report this anyway.

Is this a known error?

I'll add the error message/log as soon as it happens again.

aromanos:
This issue still happens to me and my coworkers up to the latest Vagrant 1.7.2, 100% of the times that we do a vagrant up. Adding `Errno::EACCES` to the `rescue` line at is_port_open.rb fixes the issue to us with no other impact that we can tell. The cause might be a permissions issue with the Windows corporate image or domain policies. It would be great to have this fix committed to the master branch

sneal:
@aromanos Thanks for confirming the fix!

gianluca-nitti:
I'm not sure if I should open a new issue or comment here since it's related, please apologize if I'm reporting in the wrong place.

I'm on Windows 10 and today I upgraded to Vagrant 1.9.3 from an older version (I also upgraded Virtualbox if it could be related), and after the update the port forwarding stopped working on both the ports I have configured in the Vagrantfile of my VM. While searching for a solution I found this issue, old but similar to mine. Only difference is that instead of `Errno::EACCES` i was getting `Errno::EADDRNOTAVAIL`.
The message said `Indirizzo richiesto non valido nel proprio contesto. - connect(2) for "0.0.0.0" port 5432 ` (my Windows installation is in Italian, the first string roughly translates to `Required addres invalid in the current context` ).
I'm quite sure I haven't anything on the host listening on those ports.
After reading the discussion I solved in the same way, I added `Errno::EADDRNOTAVAIL` too at line 33 of `C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.3/lib/vagrant/util/is_port_open.rb` and now everything is working, however I never touched a line of Ruby before and I don't know about Vagrant internals so I have no idea if this workaround makes sense.

[...]

PR #5201, opened 2015-01-16
Fix issue 3031

sneal:
Sometimes in use ports can raise a Errno::EACCES exeptions on Windows hosts. This fixes issue #3031.


sethvargo:
:+1:

PR #3795, opened 2014-05-15
[Vagrant][HyperV] Customization

ramakrishnan:
New feature for vagrant-hyperv for including networking customization during pre-boot. With this feature one can customize which virtual switch the VM can take from the Vagrantfile.

Example usage
 ```ruby
config.vm.provider "hyperv" do |hv, override|
   hv.customize  ["virtual_switch", { type: "External", name: "External Switch", :adapter => "Ethernet" }]
end
 ```

PR #5207, opened 2015-01-17
select a Hyper-V switch based on a network_name

mwrock:
Currently if one has more than one Hyper-V switch, vagrant will require user intervention to select the switch to use when importing a machine. This is always slightly annoying and problematic when using applications such as Test-Kitchen that do not allow interactive setup of a vagrant box.

This PR allows a Vagrantfile to include a public or private network with a network name. If a switch exists with the same name, it is the one that will be chosen:

```
config.vm.network "public_network", bridge: 'External Virtual Switch'
```

Considering one has multiple switches and one is named after the `bridge`, that switch will be chosen. If no network name is specified, the the current behavior takes over and prompts the user.

This is a modified steal from @Taliesin's PR #3480 that was previously closed (due to too broad of scope from what I can tell).

### DIFF
--- a/plugins/providers/hyperv/action/import.rb
+++ b/plugins/providers/hyperv/action/import.rb
@@ -55,23 +55,39 @@ module VagrantPlugins
           switches = env[:machine].provider.driver.execute("get_switches.ps1", {})
           raise Errors::NoSwitches if switches.empty?
 
-          switch = switches[0]["Name"]
-          if switches.length > 1
-            env[:ui].detail(I18n.t("vagrant_hyperv.choose_switch") + "\n ")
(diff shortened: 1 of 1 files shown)


vvchik:
+1 for that, simple and works well, much better than #3480 and #3795, please merge this

mwrock:
Yeah I'd love to see this pulled in. I sometimes have a external and internal switch enabled and without this patch, Test-Kitchen dies from the prompt. CC @sethvargo @mitchellh

VladLazebnyy:
+1, please merge this

strangercat:
Please merge this.

Issue #5209, opened 2015-01-17
[hyperv]: VM import process does not respect Secure Boot option from image


tomassrnka:
The import process (which is used during first vagrant up) does not respect settings for "Secure Boot" (UEFI) from the original config file. After the machine is imported, the "Enable Secure Boot" is always set to ON. 

This results in unbootable VMs (Ubuntu 14.04 LTS tested).

The correct solution would be to get the value from the original XML config and pass the setting to the local VM.

mitchellh:
Fixed by 21e2f33

PR #5210, opened 2015-01-17
Fix for: #5209 (hyperv) import respects box's original Secure Boot setting

tomassrnka:
The code takes "secure_boot_enabled" setting from box's XML (hyperv config) file and passes the value into the new VM in hyperv. 

The fix allows to create and boot non-uefi boxes.


mitchellh:
LGTM. Thanks!

vvchik:
It is broke possibility to work with Generation 1 virtual machines:
Script: import_vm.ps1
Error:

Set-VMFirmware : A parameter is invalid. Generation 1 virtual machines do not
support the VMFirmware cmdlets.  Use Get-VMBios and Set-VMBios instead.

PR #5216, opened 2015-01-19
Switch to docker.com

clayallsopp:
docker.io seems to be deprecated & handled differently from docker.com; see https://twitter.com/clint_newsom/status/557290529810313217

Anecdotally, the docker.com URLs worked while the docker.io URLs were failing, so I have more long-term confidence in them


mitchellh:
LGTM.

Issue #5200, opened 2015-01-16
Shared folder of chef works only on vagrant destroy (1.7.2)


manea-eugen:
After successful vagrant destroy -f && vagrant up  if i run reload chef doesn't remount 

```text
$ vagrant reload --provision
==> default: Loading Berkshelf datafile...
==> default: Sharing cookbooks with VM
==> default: Unmounting NFS shared folders from guest...
==> default: Attempting graceful shutdown of VM...
==> default: Checking if box 'hashicorp/precise64' is up to date...
==> default: Clearing any previously set forwarded ports...
==> default: Updating Vagrant's Berkshelf...
==> default: Resolving cookbook dependencies...
==> default: Fetching 'pimcore' from source at .
==> default: Using apache2 (2.0.0)
==> default: Using apt (2.6.1)
==> default: Using build-essential (2.0.6)
==> default: Using aws (2.5.0)
==> default: Using chef-sugar (2.5.0)
==> default: Using chef_handler (1.1.6)
==> default: Using cron (1.6.1) from https://github.com/manea-eugen/cron.git (at master)
==> default: Using composer (1.0.6)
==> default: Using database (2.3.1)
==> default: Using dmg (2.2.2)
==> default: Using hostsfile (2.4.5)
==> default: Using git (4.0.2)
==> default: Using iis (2.1.6)
==> default: Using iptables (0.14.1)
==> default: Using locale (1.0.2)
==> default: Using line (0.5.3) from https://github.com/manea-eugen/line-cookbook.git (at 0.5.3)
==> default: Using logrotate (1.7.0)
==> default: Using mailcatcher (0.1.4) from https://github.com/manea-eugen/chef-mailcatcher.git (at master)
==> default: Using mongodb (0.16.2)
==> default: Using mysql (5.5.4)
==> default: Using mysql-chef_gem (0.0.5)
==> default: Using now (0.3.0)
==> default: Using ntp (1.6.8)
==> default: Using pacman (1.1.1)
==> default: Using openssl (2.0.2)
==> default: Using php (1.5.0)
==> default: Using pimcore (0.1.0) from source at .
==> default: Using python (1.4.6)
==> default: Using postgresql (3.4.14)
==> default: Using runit (1.5.14)
==> default: Using simple_iptables (0.6.0)
==> default: Using vim (1.1.2)
==> default: Using sudo (2.7.1)
==> default: Using windows (1.36.1)
==> default: Using xdebug (1.0.0)
[...]

Issue #5140, opened 2015-01-08
Booting a already provisioned VM doesn't mount Chef vboxsf folders in Vagrant 1.7.2


andytson:
On the 2nd vagrant up of a VM with chef-solo configuration for path configuration, the vboxsf folders that are usually mounted don't appear. This issue occurs in Vagrant 1.7.2 but not 1.7.1.

Steps to reproduce:

    vagrant remote add opscode-centos-6.6 http://opscode-vm-bento.s3.amazonaws.com/vagrant/virtualbox/opscode_centos-6.6_chef-provisionerless.box
    vagrant init opscode-centos-6.6
    mkdir roles
    mkdir cookbooks

Add to the bottom of Vagrantfile:

      config.omnibus.chef_version = :latest
      config.vm.provision :chef_solo do |chef|
        chef.cookbooks_path = "cookbooks"
        chef.roles_path = "roles"
      end
.

    vagrant up
    vagrant ssh -c 'mount'
    vagrant reload
    vagrant ssh -c 'mount'

The 2nd mount is missing the cookbooks and roles mounts from the first mount call

sethvargo:
@mitchellh is looks like we aren't caching all the synced folders?

databus23:
I'm having similar issues. I can't get vagrant 1.7.2 to mount the chef related folders at all after a `vagrant reload`.

andytson:
Is it somehow related to this that was fixed? https://github.com/mitchellh/vagrant/issues/4988

andytson:
A guess at the issue is it's possibly related to this caching: true here, causing what @sethvargo mentions as the chef folders not being added to the cache:
https://github.com/mitchellh/vagrant/pull/5121/files#diff-4b59695a716db3067f2a3b5d98bff333R37

khobbits:
I'm having some major issues with this, on 1.7.2.

I have a few chef recipes I want to run on every 'up', not just on provision and these now bomb.

"config.vm.provision :chef_zero, run: "always" do |chef|"

fletchowns:
I think this is the error message I'm seeing as a results of this bug. The initial ```vagrant up``` works fine but after a ```vagrant halt```, ```vagrant up```, ```vagrant provision``` I see the following:

> ==> default: [2015-01-13T21:20:32+00:00] ERROR: The expanded run list includes nonexistent roles: app_server
[...]

PR #5217, opened 2015-01-20
Don't use cache for synced folders in ChefSolo

### LINKED ISSUE #5199: Problem reloading Chef shared folders
After rebooting a virtual machine configured to use chef-zero, it seems the `vagrant provision` command no longer works, and displays the following error:

```
==> default: Running provisioner: chef_zero...
==> default: Detected Chef (latest) is already installed
**Shared folders that Chef requires are missing on the virtual machine.
This is usually due to configuration changing after already booting the
machine. The fix is to run a `vagrant reload` so that the proper shared
folders will be prepared and mounted on the VM.**
```

However, running the `vagrant reload` command doesn't seem to recreate all of the shared folders required for chef-zero to operate:

```
==> default: Checking if box 'ubuntu/trusty64' is up to date...
==> default: Clearing any previously set forwarded ports...
    default: Adapter 1: nat
    default: Adapter 2: hostonly
    default: 22 => 2222 (adapter 1)
==> default: Booting VM...
    default: Warning: Connection timeout. Retrying...
**==> default: Mounting shared folders...
    default: /vagrant => /Users/$USER/Dev
    default: /tmp/vagrant-chef/91d356a57ac0d6c98d383bfab1f9fee7/nodes => /Users/$USER/chef-repo/nodes**
==> default: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> default: to force provisioning. Provisioners marked to run always will still run.
```

Naturally, the `vagrant provision` command still outputs the same error.

Please note, when initially running `vagrant up`, there were 3 shared folders created for chef (roles, nodes, and cookbooks):

```
    default: /vagrant => /Users/$USER/Dev
    default: /tmp/vagrant-chef/6616dcf9743a971e3f12afe6c9dbbdaf/roles => /Users/$USER/chef-repo/roles
    default: /tmp/vagrant-chef/91d356a57ac0d6c98d383bfab1f9fee7/nodes => /Users/$USER/chef-repo/nodes
    default: /tmp/vagrant-chef/517144fb3ab4b18253f23b7e166c95e1/cookbooks => /Users/$USER/chef-repo/cookbooks
```

This looks like a bug, or am I doing something wrong?

Note: I'm running Vagrant 1.7.2 on OS-X Yosemite.

BlakeMesdag:
This disables using the synced_folders cache for ChefSolo. I've also added tests for the ChefSolo provisioner in regards to the conflicts and confirming the counter updates correctly.

This should close #5199, #5200, and #5140.

@mitchellh @sethvargo

### DIFF
--- a/plugins/provisioners/chef/provisioner/chef_solo.rb
+++ b/plugins/provisioners/chef/provisioner/chef_solo.rb
@@ -34,7 +34,7 @@ module VagrantPlugins
           @data_bags_folders = expanded_folders(@config.data_bags_path, "data_bags")
           @environments_folders = expanded_folders(@config.environments_path, "environments")
 
-          existing = synced_folders(@machine, cached: true)
+          existing = synced_folders(@machine, cached: false)
           share_folders(root_config, "csc", @cookbook_folders, existing)
           share_folders(root_config, "csr", @role_folders, existing)
           share_folders(root_config, "csdb", @data_bags_folders, existing)
--- a/test/unit/plugins/provisioners/chef/provisioner/chef_solo_test.rb
+++ b/test/unit/plugins/provisioners/chef/provisioner/chef_solo_test.rb
@@ -1,0 +1,79 @@
+require_relative "../../../../base"
+
+require Vagrant.source_root.join("lib/vagrant/action/builtin/mixin_synced_folders")
+require Vagrant.source_root.join("plugins/provisioners/chef/provisioner/chef_solo")
+
+describe VagrantPlugins::Chef::Provisioner::ChefSolo do
+  include_context "unit"
+
+  let(:machine) { double("machine") }
+  let(:config)  { double("config") }
+  let(:vm)      { double("vm") }
+
+  subject { described_class.new(machine, config) }
+
+  before(:each) do
+    allow(config).to receive(:vm).and_return(vm)
+
+    allow(subject.class).to receive(:get_and_update_counter).and_return(1, 2, 3, 4)
+
+    allow(subject).to receive(:expanded_folders).and_return(
+                          [[:host, '/tmp/chef/cookbooks', '/tmp/vagrant-chef-1/cookbooks']],
+                          [[:host, '/tmp/chef/roles', '/tmp/vagrant-chef-1/roles']],
+                          [[:host, '/tmp/chef/data_bags', '/tmp/vagrant-chef-1/data_bags']],
+                          [[:host, '/tmp/chef/environments', '/tmp/vagrant-chef-1/environments']]
+                        )
+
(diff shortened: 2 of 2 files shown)


sethvargo:
@BlakeMesdag Thank you for the PR. How does disabling the cached folders solve the problem? I think the folders need to be cached so they persist during a `vagrant reload`.

BlakeMesdag:
The NFS plugin doesn't appear to actually reload them, which is the issue. It also appears that during a halt the file isn't cleaned up.

I can dig deeper, but if the intent of the folder caching code is to skip re-adding shared folders between commands, with a running VM, then those two pieces of functionality need to be fixed.

sethvargo:
Ping @mitchellh :smile:

BlakeMesdag:
@mitchellh it actually might be in the VMware plugin, which I can't poke around in. The cached synced folders seem to be handled in the provider for the other ones. The vm appears to de-dup synced folders though, so it's got that going for it.

hugofonseca:
<y> thanks

snicol:
:+1:

Scythril:
It's not just the VMware plugin, I'm using vagrant-aws (which syncs using rsync) and chef-solo and seeing the same issue. For this case, I actually **NEED** the folders to sync on every command. If I make a small or additive cookbook change, I'd like to just run provision again instead of tearing down the server and recreating the whole thing again.

I've merged your change into my vagrant set-up for now until it merges in upstream (or gets fixed, whichever), so that I can do just that.

c-castillo:
The same problem happens in the `chef-zero` provisioner also

dkinzer:
I'm seeing this with virtualbox and chef_solo/chef_zero (so it's not just a vmware thing).

cjsaylor:
@BlakeMesdag This seems to be fixing the effect, not the cause. I haven't looked through the code deeply, but it seems it is simply neglecting to cache configured chef cookbook paths as part of that cache.

That part the writes the cache is probably the source of the issue.

glenjamin:
I just ran into this just now and got very confused.

Is there a better workaround than downgrading to 1.7.1 at the moment?

ozzyogkush:
I'm on 1.7.2 with VirtualBox and Chef Zero as a provisioner - confirming that this is happening in that situation as well.

hardchor:
@glenjamin Deleting the synced_folders file fixed it for us (sometimes), but downgrading to 1.7.1 seems to have done the trick until 1.7.3 gets released.

indirect:
I'm also seeing this happen with both VirtualBox and Parallels providers with Chef Zero, on 1.7.2. (No one seems to have confirmed it's an issue with the Parallels provider yet, so I figured I might as well...)

mitchellh:
Fixed. :)

PR #5218, opened 2015-01-20
Check plugin name with version spec in has_plugin

mbrodala:
If one passes a version spec to `Vagrant.has_plugin?`, the actual plugin name is ignored and any installed plugin which matches the version spec makes the check pass. Fix this by also checking for plugin name match in addition to the requested version spec.

### DIFF
--- a/lib/vagrant.rb
+++ b/lib/vagrant.rb
@@ -156,7 +156,7 @@ module Vagrant
     Plugin::Manager.instance.installed_specs.any? do |s|
       match = s.name == name
       next match if !version
-      next version.satisfied_by?(s.version)
+      next match && version.satisfied_by?(s.version)
     end
   end
 
--- a/test/unit/vagrant_test.rb
+++ b/test/unit/vagrant_test.rb
@@ -75,7 +75,7 @@ describe Vagrant do
       expect(described_class.has_plugin?("bar")).to be_false
     end
 
-    it "finds plugins by gem version" do
+    it "finds plugins by gem name and version" do
       specs = [Gem::Specification.new]
       specs[0].name = "foo"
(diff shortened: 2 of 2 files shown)


sethvargo:
Hi @mbrodala 

Thank you for the PR. Would it be possible to add a test for this so we can ensure the API in the future?

mbrodala:
@sethvargo Unfortunately I don't know anything about testing in Ruby. I'd be thankful if someone could jump in and add a test for this.

sethvargo:
Hey @mbrodala no problem :smile: 

Here is [an example test to get you started](https://github.com/mitchellh/vagrant/blob/bb700efe404ef0950957be26ff175f6de790c182/test/unit/vagrant_test.rb#L57-64). You'll want something very similar to that, but also pass in the version. The other testes in that stanza should provide some context. Let me know if you have any questions - I'm more than happy to assist.

mbrodala:
@sethvargo Well, just looking at the existing tests adding one line to the [`finds plugins by gem version`](https://github.com/mitchellh/vagrant/blob/bb700efe404ef0950957be26ff175f6de790c182/test/unit/vagrant_test.rb#L78) test should suffice. I'll try that.

sethvargo:
:smile:

mbrodala:
Thanks. :-)

mbrodala:
Would you mind putting out a new Vagrant release which includes this change?

sethvargo:
@mbrodala we are not quite ready to release a new Vagrant yet. The next release will definitely include this fix though.

mbrodala:
Alright, thanks. Just asking because I want to require the next version in my `Vagrantfile`. ;-)

PR #5225, opened 2015-01-20
Use winrm-fs for file uploads

sneal:
Replace the Vagrant native winrm file upload functionality with the new winrm-fs gem (which requires winrm 1.3+)


sethvargo:
LGTM

Issue #3083, opened 2014-03-08
dhcp private_network fails on virtualbox


phinze:
## Problem

With a fresh VirtualBox install, adding a private network with `type: :dhcp` fails with the following error:

```
There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.

Command: ["dhcpserver", "add", "--ifname", "vboxnet0", "--ip", "172.28.128.2", "--netmask", "255.255.255.0", "--lowerip", "172.28.128.3", "--upperip", "172.28.128.254", "--enable"]

Stderr: VBoxManage: error: DHCP server already exists
```

Subsequent boots fail with a different message:

```
A host only network interface you're attempting to configure via DHCP
already has a conflicting host only adapter with DHCP enabled. The
DHCP on this adapter is incompatible with the DHCP settings. Two
host only network interfaces are not allowed to overlap, and each
host only network interface can have only one DHCP server. Please
reconfigure your host only network or remove the virtual machine
using the other host only network.
```

## Steps to repro

Fresh install of Vagrant 1.4.3 and VirtualBox 4.3.6.

Minimal Vagrantfile:

```ruby
# force virtualbox provider
ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtualbox'

Vagrant.configure('2') do |config|
  config.vm.box     = 'precise64'
  config.vm.box_url = 'http://files.vagrantup.com/precise64.box'
  config.vm.network 'private_network', type: :dhcp
end
```

Error message will appear on first `vagrant up`. My test host is OS X but I bet you this occurs on all hosts.

## Explanation

VirtualBox ships with an initial `dhcpserver` seeded in its config [1].

```
$ VBoxManage list dhcpservers
NetworkName:    HostInterfaceNetworking-vboxnet0
IP:             192.168.56.100
NetworkMask:    255.255.255.0
lowerIPAddress: 192.168.56.101
upperIPAddress: 192.168.56.254
Enabled:        Yes
```

The `hostonlyifs` settings ship empty.

When encountering a private network with `type: :dhcp`, Vagrant sees this `HostInterfaceNetworking-vboxnet0` dhcpserver and assumes that it is valid and ready to be attached to `vboxnet0`. But `vboxnet0` was just created and has an IP address of `172.28.128.1`.

So the error message comes from VirtualBox yelling that you can't assign a dhcpserver with an IP address outside the network of an adapter.

[...]

PR #5233, opened 2015-01-21
providers/virtualbox: read netmask from dhcpservers

phinze:
This should fix the cleaning up of the default VirtualBox dhcpserver,
which we've been fighting with for ages over in #3083. We were checking
for a structure _including_ a netmask, but the driver was not populating
netmask.


phinze:
Smoke tested too :fist: 
1__hihashi__tmux_

sethvargo:
:+1:

PR #5244, opened 2015-01-24
core: log action name to info

phinze:
I found this output to be very helpful in debugging an action hook
problem in a plugin.  problem. I'm not sure why the callable_id is
useful (seem to always show up as an instance of `Builder` or `Warden`
for me), but I left it in there just in case it's useful to someone.


mitchellh:
LGTM

PR #5246, opened 2015-01-25
Edit NFS docs for windows users

alfonsodev:
There is an alternative for Windows users that want to run NFS, documentation should be updated to inform them.


mitchellh:
Thanks but we still have made it a point to not link to 3rd party plugins on the core docs. I think it might make sense to add a `community plugins` page that lists all of this, and then link to that page from where you're modifying here. If you do that, I'd merge.

PR #5257, opened 2015-01-28
Kindly document how to disable checkpoint remote call.

aknrdureegaesr:
The hashicorp checkpoint gem's [current documentation](http://www.rubydoc.info/gems/hashicorp-checkpoint/0.1.4) has this to say:

> We understand that software making remote calls over the internet for any reason can be undesirable. Because of this, Checkpoint can be disabled in all of our software that includes it. You can view the source of this client to see that we're not sending any private information.

Instructions how to do the disabling have been omitted from the Vagrant documentation. This pull request contains a (very minimal) fix for that omission.

Thank you for providing fine software!

### DIFF
--- a/website/docs/source/v2/other/environmental-variables.html.md
+++ b/website/docs/source/v2/other/environmental-variables.html.md
@@ -9,6 +9,14 @@ Vagrant has a set of environmental variables that can be used to
 configure and control it in a global way. This page lists those environmental
 variables.
 
+## VAGRANT\_CHECKPOINT\_DISABLE 
+
+Vagrant does occasional network calls to check whether the version of Vagrant
+that is running locally is up to date. We understand that software making remote
+calls over the internet for any reason can be undesirable. To surpress these
+calls, set the environment variable `VAGRANT_CHECKPOINT_DISABLE` to any
+non-empty value.
+
 ## VAGRANT\_CWD
 
 `VAGRANT_CWD` can be set to change the working directory of Vagrant. By


sethvargo:
I'm going to defer to @mitchellh on this one. I thought there was conversation about intentionally not documenting this, but I could be misremembering.

mitchellh:
Hm, it is an env var so it should go into the env vars section of the Vagrant docs. 

@sethvargo I can't recall any such conversation. The only one I know we're purposely not documenting is ".vagrantplugins"

Tagging.

sethvargo:
Ahh that's right. I knew it was something. 

@aknrdureegaesr would you mind moving this to the env var documentation section and then ping me?

aknrdureegaesr:
I thought a environment variable doc would be a better place for this, but could not find any. I was expecting it below [the cli doc](https://docs.vagrantup.com/v2/cli/index.html), but cannot find it there. I even tried

    .../vagrant/website/www/source$ find . -type f -name \*.md -print0 | xargs -0r grep -lP '[A-Z0-9_\-]{5}'

-also to no avail.

Please help me out: Where is it?

sethvargo:
Hey @aknrdureegaesr - [this file](https://github.com/mitchellh/vagrant/blob/master/website/docs/source/v2/other/environmental-variables.html.md). It's in docs/ not www/ :smile:

aknrdureegaesr:
Thanks @sethvargo !

I decided to ditch my previous checkin and replace it with a better one.

sethvargo:
Thanks!

aknrdureegaesr:
You're welcome. My pleasure.

PR #5259, opened 2015-01-28
NFS filesync: Prevent sed from creating tmp files

radeksimko:
Even if one tries hard to prevent vagrant from asking for password while using the NFS file sync and does following:

```
sudo chown root:admin /etc/exports
sudo chmod g+w /etc/exports

echo "%admin ALL=(root) NOPASSWD: /sbin/nfsd" >> /etc/sudoers
```

then there is still this `sed` which tries to create temp files in `/etc` and as that requires su perms and no sudo is provided, it ends up in following errors:

```
sed: couldn't open temporary file /etc/sedKEZcGM: Permission denied
```

and from security point of view I don't see a reason to allow write for all admins on that system just because `sed` wants to put temp files there.

If you think that using `open3` or generally the approach taken is not the best solution here, I'm really open for suggestions and happy to modify the PR.

### DIFF
--- a/plugins/hosts/bsd/cap/nfs.rb
+++ b/plugins/hosts/bsd/cap/nfs.rb
@@ -1,4 +1,5 @@
 require "log4r"
+require "open3"
 
 require "vagrant/util"
 require "vagrant/util/shell_quote"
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @radeksimko 

Thank you for the Pull Request. Although a different operating system, I think this is a close duplicate of https://github.com/mitchellh/vagrant/pull/5030, do you agree? I like the approach of using `$TMPDIR` a bit more than `open3`. What do you think?

radeksimko:
Agreed

PR #5261, opened 2015-01-29
Revert "Rename Windows guest without reboot"

blairham:
This is causing Windows server to crash


mitchellh:
Thanks, will do.

Sharpie:
Any idea when a bugfix release containing this fix will go out? I will have to drop back down to Vagrant 1.6.5 without working Windows Server guests in Vagrant 1.7.

PR #5264, opened 2015-01-29
Make the environment variable documentation more easy to find.

aknrdureegaesr:
I was looking for the environment variable definition at the place where the cli is documented. Didn't find it. For the next guy who ticks like I do, let's put a link there.


sethvargo:
Thanks!

PR #5268, opened 2015-01-30
Fix bug of path conversion on windows with cygwin.

qzchenwl:
On Windows, using MobaXterm, vagrant cannot create share folder correctly. The host folder is always %USERPROFILE% instead of current folder.

The `path` param passed to `fs_real_path` looks like `C:/Users/me/DOCUME~1/MOBAXT~1/home/my-project`. Dir.entries(path) will list folders with long filename like `Documents and Settings`, which does not match `DOCUME~1`.

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -111,6 +111,10 @@ module Vagrant
         # This expands the path and ensures proper casing of each part
         # of the path.
         def fs_real_path(path, **opts)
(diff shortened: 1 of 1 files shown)


8la:
Same "bug" here, working with cygwin on mobaXterm. 

This fix worked for me!

marema31:
Same problems for me with mobaXterm 7.6 /cygwin on Windows 7/64

This fix also worked for me.

mitchellh:
This is fixed. I fixed this by expanding short paths.

PR #5274, opened 2015-02-02
Use the new Atlas APIs for downloading boxes

sethvargo:
/cc @mitchellh @pearkes

### DIFF
--- a/lib/vagrant/action/builtin/box_add.rb
+++ b/lib/vagrant/action/builtin/box_add.rb
@@ -57,7 +57,7 @@ module Vagrant
               raise Errors::BoxServerNotSet if !server
 
               expanded = true
-              url[i] = "#{server}/#{url[i]}"
+              url[i] = "#{server}/api/v1/vagrant/#{url[i]}"
             end
           end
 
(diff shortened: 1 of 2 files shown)


sethvargo:
I ran smoke tests and was able to download a box from Atlas with this patch :smile:

pearkes:
Yay!

mitchellh:
This has to be reverted. It breaks public API for alternative box implementations. I'll explain more in the next comment.

mitchellh:
This changes the API for boxes to _require_ the `/api/v1/vagrant` path. This is why I didn't get around to doing this yet because I hadn't thought of a clean way to do it, and even told @pearkes it isn't worth it and we should just maintain the old routes forever.

The public API for box discovery is `#{server_url}/box/name` as JSON content-type. This changes it to `#{server_url}/api/v1/boxes/box/name`, which is Atlas specific. This isn't just a theoretical problem either, there are at least 3 separate production implementations of this "protocol" out in the wild.

mitchellh:
Reverted. Let's talk about another way to do this in corp chat.

PR #5275, opened 2015-02-03
Note Hyper-V is supported in Vagrant out-of-the-box

rgardler:


mitchellh:
Thanks, Vagrant supports quite a few things out of the box, but we still highlight VirtualBox here as a safe default while we iron our more Hyper-V issues. We'll get there. :)

PR #5277, opened 2015-02-03
detect fedora 21

Keeguon:
Last week I noticed that there was some issues regarding the dynamic network cards using Fedora 21 and yesterday I finally tracked down the issue. The grepping didn't support Fedora 21 so I added a trailing 1 to the regexp.


sethvargo:
Seems reasonable to me :+1:

mitchellh:
The regex continues. LGTM

PR #5283, opened 2015-02-04
Add extra metadata information to Atlas pushes

sethvargo:
- [x] Tests

### DIFF
--- a/plugins/commands/push/command.rb
+++ b/plugins/commands/push/command.rb
@@ -9,17 +9,8 @@ module VagrantPlugins
 
       # @todo support multiple strategies if requested by the community
       def execute
-        options = {}
-
         opts = OptionParser.new do |o|
           o.banner = "Usage: vagrant push [strategy] [options]"
-          o.separator ""
-          o.separator "Options:"
-          o.separator ""
-
-          o.on("-m", "--message BODY", "Text to identify this push") do |m|
-            options[:message] = m
-          end
         end
 
         # Parse the options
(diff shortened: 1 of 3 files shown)


plugins/commands/push/command.rb
```diff
@@ -9,8 +9,17 @@ def self.synopsis
...
+
         opts = OptionParser.new do |o|
           o.banner = "Usage: vagrant push [strategy] [options]"
+          o.separator ""
+          o.separator "Options:"
+          o.separator ""
+
+          o.on("-m", "--message BODY", "Text to identify this push") do |m|
+            options[:message] = m
+          end
```
sethvargo: @mitchellh I added this as an option, but I'm not sure how to pass the option into the actual push class. Thoughts?

plugins/pushes/atlas/push.rb
```diff
@@ -27,8 +27,10 @@ def execute(uploader)
         cmd << "-vcs" if config.vcs
         cmd += config.includes.map { |v| ["-include", v] }
         cmd += config.excludes.map { |v| ["-exclude", v] }
+        cmd += metadata.map { |k,v| ["-metadata", "#{k}=#{v}"] }
         cmd += ["-address", config.address] if config.address
         cmd += ["-token", config.token] if config.token
+        cmd += ["-message", message] if message
```
sethvargo: Re: my previous comment, I need to get the value for that option in here somehow...

plugins/commands/push/command.rb
mitchellh: I also wonder if this should even be a top-level option, because what does it mean in the general sense?

mitchellh:
LGTM

Issue #4969, opened 2014-12-14
1.7.1 solaris 10 to 11.2 boxes fails because of newly generated keypair


biemond:
Hi, 

this keypair refresh break the usage of solaris boxes, on 10, 11 and 11.2
also I can't add a private network to solaris 11.1 and 11.2

    dbsol: this with a newly generated keypair for better security.
Vagrant is configured to generate a random keypair and insert it
onto the guest machine, but it appears Vagrant doesn't know how to do
this with your guest OS. Please disable key insertion by setting
`config.ssh.insert_key = false` in the Vagrantfile.

After doing so, run `vagrant reload` for the setting to take effect.

If you'd like Vagrant to learn how to insert keys on this OS, please
open an issue with details about your environment.

sethvargo:
Closing now that 5290 is merged. Thanks!

PR #5290, opened 2015-02-05
Copy linux support for insert_public_key and remove_public_key to solari...

channui:
This addresses issue #4969, for Solaris 11 requiring 'config.ssh.insert_key = false'.  It just copies the appropriate code over from the linux plugin.  I don't have a Solaris 10 image to test against, so I did not make the change there.  It passes the unit tests and the provider/virtualbox/basic acceptance test.


mitchellh:
LGTM

PR #5302, opened 2015-02-07
Fixes https://github.com/mitchellh/vagrant/issues/3680

cubetto:
I'm sorry is this has already been discussed and rejected, but i run into issue 3680 and got it fixed with this change, so thought i could at least make a PR for it, to see what's going on.


mitchellh:
LGTM

cubetto:
Great : )

2015-02-24 14:47 GMT-03:00 Mitchell Hashimoto <notifications@github.com>:

PR #5320, opened 2015-02-11
Support Regexp usage for bridge configuration

hlidotbe:
Most of the time when using a bridged interface we want to connect to the first available interface. Depending on various conditions, hardcoding the exact interface name can be troublesome and difficult to share.

Changing the config a bit permits the usage of a Regexp (or anything that respond to === actually) which helps avoiding to type "1" each time we launch a VM.

I've kept the old "downcase" behaviour so it shouldn't break anything.


mitchellh:
This LGTM, a nice advanced feature.

PR #5325, opened 2015-02-11
vagrant duplicates >= eth2 when defining two private network ips

frankbb:
When a vagrant box has two private network ips /etc/network/interfaces
will duplicate eth2 and bigger. sed matches greedy, so the first
 #VAGRANT-END matches. This will result in:

/etc/network/interfaces:29: interface eth2 declared allow-auto twice
/sbin/ifup: couldn't read interfaces file "/etc/network/interfaces"

Tac print files in reverse and will match the first vagrant-end and only print what is under the last vagrant-end. Tac is in coreutils and is installed with a debootstrap wheezy, so i am assuming that nobody will have an issue on Debian.

### DIFF
--- a/plugins/guests/debian/cap/configure_networks.rb
+++ b/plugins/guests/debian/cap/configure_networks.rb
@@ -14,7 +14,7 @@ module VagrantPlugins
             # First, remove any previous network modifications
             # from the interface file.
             comm.sudo("sed -e '/^#VAGRANT-BEGIN/,$ d' /etc/network/interfaces > /tmp/vagrant-network-interfaces.pre")
-            comm.sudo("sed -ne '/^#VAGRANT-END/,$ p' /etc/network/interfaces | tail -n +2 > /tmp/vagrant-network-interfaces.post")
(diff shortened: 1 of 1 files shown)


mattolson:
:+1: I'm having the same issue.

alexmckenley:
Looks like this should resolve #5180

mitchellh:
Is `tac` available by default in Debian?

frankbb:
Tac is part of coreutils `https://packages.debian.org/wheezy/coreutils`. This package is tagged as essential so it is part of any Debian release.

arosenhagen:
*bump*

jtsternberg:
Having this issue as well. +1 on getting a solution merged.

nicolasbadia:
I was having this issue (with Debian Wheezy) and your patch is working just fine! 
Thanks.

danwall:
Can confirm that this fixes the issue with Ubuntu 14.04. Thanks.

adri:
This patch worked for me as well using a Debian Wheezy vm.

djvdorp:
Same problem here, thanks for the patch!

gamename:
Ditto.  Running into it here also.

infernix:
Can we get this merged for 1.7.3?

dlondero:
I can confirm this issue. And yes, can this be merged in 1.7.3? :)

PR #5326, opened 2015-02-11
Support insert_public_key and remove_public_key on darwin guest

### LINKED ISSUE #5204: config.ssh.insert_key isn't supported on OS X (e.g. darwin) machines
The darwin machine type is missing the remove_public_key and insert_public_key capabilities so this config param doesn't work. Looking at the Linux version of these capabilities, I don't see why this shouldn't be supported on the platform. linux/cap/insert_public_key.rb looks like it could be copied to darwin verbatim and only a small change is needed to linux/cap/remove_public_key.rb to make it work with the awkward BSD variant of sed:
```ruby
    "sed -i '' '/^.*#{contents}.*$/d' ~/.ssh/authorized_keys"
```
or if you want to make it less obscure and save a backup file (it shouldn't change the security of things as sshd will simply ignore the backup file):
```ruby
    "sed -i.bak '/^.*#{contents}.*$/d' ~/.ssh/authorized_keys"
```

timsutton:
Here's an implementation of the private key caps for darwin. It's copied from the Linux guest caps, with a minor tweak to the variant of sed that ships on OS X. This fixes #5204.

### DIFF
--- a/plugins/guests/darwin/cap/insert_public_key.rb
+++ b/plugins/guests/darwin/cap/insert_public_key.rb
@@ -1,0 +1,21 @@
+require "vagrant/util/shell_quote"
+
+module VagrantPlugins
+  module GuestDarwin
+    module Cap
+      class InsertPublicKey
+        def self.insert_public_key(machine, contents)
+          contents = contents.chomp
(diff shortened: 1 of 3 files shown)


mitchellh:
LGTM. Thanks!

PR #5328, opened 2015-02-11
Relax i18n gem spec to support 0.7.0

kev009:
API changes seem pretty minor:  https://github.com/svenfuchs/i18n/blob/master/CHANGELOG.md

Useful for folks that package instead of omnibus.


mitchellh:
LGTM.

PR #5331, opened 2015-02-12
Fixes #5322

### LINKED ISSUE #5322: Do not rely on lsb-release for ubuntu detection.
`lsb-release` is a tools which is written in python, and this is kind of problem, as no other use for python exists.

tioteath:


sethvargo:
What is `/etc/issue`? Do you have some links to docs on that? What version(s) of Ubuntu does it apply to?

tioteath:
According to http://unix.stackexchange.com/questions/84280/is-etc-issue-common-for-all-linux-distributions this is standard across different distributions. Actually, this is slightly modified version of Debian guest plugin: https://github.com/mitchellh/vagrant/blob/master/plugins/guests/debian/guest.rb. This allows to totally remove python stuff, so overall box size decreases.

sethvargo:
Seems reasonable to me then! :+1:

sgoings:
Please note the follow up comment on that Stack Exchange thread:

> In addition to slm's answer, it's worth remembering that anyone (with root access) can edit the content of /etc/issue. Some organisations may do so to present pre-login warnings or disclaimers, and others may remove all identification of the distribution for security reasons (it doesn't matter how effective either of these measures are).

> Therefore, while /etc/issue may exist almost all the time, the content is not necessarily suitable for detecting the OS (as you put it, or more likely, the Linux distribution).```

For example, my company's is:
```
******************************************************************
*      UNAUTHORIZED ACCESS TO THIS SYSTEM IS FORBIDDEN!          *
* Only authorized personnel may access or configure this system. *
* Access and activities on this system may be logged. Users      *
* (authorized or unauthorized) have no explicit or implicit      *
* expectation of privacy. Unauthorized or improper use of this   *
* system may result in disciplinary action and or civil and      *
* criminal penalties. By using this system you indicate your     *
* awareness of and consent to these terms and conditions of use. *
******************************************************************
```

nspaeth:
I'd agree with @sgoings, changing `/etc/issue` and `/etc/issue.net` has caused things that detect distribution through those files to break for me in the past.

> `grep -irn ubuntu /etc`
shows some other candidates for detection - perhaps `/etc/legal` or `/etc/os-release`?

sethvargo:
Okay, on second though, I change my vote to :-1: until we can find something less editable by the user.

nspaeth:
Perphaps checking all of those files would be most versatile for various versions (eg. I think os-release is only standard on machines using systemd?)?
> cat /etc/lsb-release /etc/os-release /etc/issue /etc/issue.net /etc/legal | grep -i ubuntu

I'm not sure if this api works as a regular shell that will allow piping though. Might need to test that.

tioteath:
`cat /etc/lsb-release /etc/os-release | grep ubuntu` won't work because `/etc/os-release` contains both `debian` and `ubuntu` entries, so it probably fail, if `debian` guest will be also updated . It should contain specific match, like I was added in last commit.
So, what about this pull request with `os-release`? Could it be merged in master?

sethvargo:
Do we have any docs/info on `os-release`? When was it first added? Is it the same for 10.04-14.04?

mitchellh:
I'd be open to a hybrid approach that checks both. If you open the PR to do so, I'll merge it. But I think flat out replacing one for the other isn't worth it.

Issue #5189, opened 2015-01-15
Vagrant can find 'virtualbox', but not 'VirtualBox'.


GeorgeMac:
I could be doing something really stupid. However, I keep getting this issue. Even with fresh installs of both vagrant and virtual box.
When I `vagrant up`, vagrant chooses to use `virtualbox` initially as a provider. It imports the base box, sorts out NATs, port forwarding... blah blah blah. Then I get the following error as it tries to provision:
```
The provider 'VirtualBox' could not be found, but was requested to
back the machine 'default'. Please use a provider that exists.
```
I then try `vagrant destroy` and get the same error message. Even with `vagrant ssh` I get this error message. However, if I use `vagrant global-status` and get the vm id; I can explicitly `vagrant destroy <id>` or `vagrant ssh <id>`. 
This is a pain as every time I `vagrant up` it gets as far as provisioning with ansible and then fails because it is looking for a slightly different provider. 

I am running on OS X Yosemite if that helps as well. I am sure I have just messed up set up somehow.

davebirch:
May be a silly question, but do you have a vm.provider block in the
Vagrantfile used for the environment you're seeing issues with, and have
you specified the provider in this block as "VirtualBox" rather than
"virtualbox"?

I might be misreading your message, does this happen for all Vagrant
instances, or just a specific file?

GeorgeMac:
Just had a check. The provider is set to `virtualbox` lower-case. I tried with upper-case and I get the same problem :S

GeorgeMac:
I'll look for another project to try it with :)

GeorgeMac:
There must be something wrong with my Vagrantfile. Everything worked fine when I built vagrant project itself. I will keep looking, this is someone else project who has left our company. I'll close this issue :) Thanks @davebirch

legal90:
@GeorgeMac have you figured out what was wrong in your Vagrantfile? 

We got the similar error report for [parallels] provider from one of our customers. He is using the valid Vagrantfile (with provider mentioned in lower case) and initial "vagrant up" is successfull. But after that he get the same error about provider in capitalized notation:

```
The provider 'Parallels' could not be found, but was requested to
back the machine 'default'. Please use a provider that exists.
```

hitsthings:
I'm also seeing this (on Windows 8.1). I can't destroy the machine either because the provider isn't recognized.

legal90:
[...]

PR #5334, opened 2015-02-12
Create machine data directory only after the machine will be checked

legal90:
It fixes GH-5189
Actually, machine data directory should be created after machine configuration, e.q. when checks will be passed : [lib/vagrant/vagrantfile.rb#L108-L141](https://github.com/mitchellh/vagrant/blob/59dbe51ef2f0ba021b5e9c213710303d45c3b2d2/lib/vagrant/vagrantfile.rb#L108-L141)

### Why it is important
Currently we have this issue on OS X hosts: GH-5189, and it affects all providers: https://forum.parallels.com/threads/vagrant-builds-vms-but-cannot-locate-them.327487/
It is caused by two facts: 
- Machine data directory is created before check that the specified provider exists. Vagrant doesn't delete this dir even if provider check has been failed.
- The filesystem on OS X host is case insensitive by default. Thats why this "hidden error" still exists and make all subsequent "vagrant" commands not working.

How to reproduce: https://github.com/mitchellh/vagrant/issues/5189#issuecomment-74072717

### DIFF
--- a/lib/vagrant/environment.rb
+++ b/lib/vagrant/environment.rb
@@ -604,10 +604,8 @@ module Vagrant
       @logger.info("Uncached load of machine.")
 
       # Determine the machine data directory and pass it to the machine.
-      # XXX: Permissions error here.
       machine_data_path = @local_data_path.join(
         "machines/#{name}/#{provider}")
-      FileUtils.mkdir_p(machine_data_path)
 
       # Create the machine and cache it for future calls. This will also
       # return the machine from this method.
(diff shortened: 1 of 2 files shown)


legal90:
Any updates?

sethvargo:
Fixes https://github.com/mitchellh/vagrant/issues/5310 too

sethvargo:
This seems reasonable to me, but I have not had time to test it yet.

damiannz:
I get the same issue on Windows 7 under Cygwin.

mitchellh:
LGTM. Thanks

PR #5337, opened 2015-02-13
Added bootstrap_options param to the salt provisioner docs

mafrosis:
I discovered this option by digging in the source for a way to pass extra params to `bootstrap_salt.sh`.


mitchellh:
LGTM. Thanks

PR #5338, opened 2015-02-14
example to set default router on public net

kikitux:
Add disable auto-configuration on public net
Add example to setup manually ip and ipv6
Add example to set default router for ip and ipv6


kikitux:
Tested with Vagrant 1.7.2 on OSX, guest linux

mitchellh:
LGTM. I'm going to nitpick the style a bit but the content is good.

PR #5349, opened 2015-02-17
Update host_info.ps1

agreenbhm:
Changed method for getting IP addresses.  Windows built-in VPN IP addresses do not appear in list using Get-WMIObject, causing shared folders to fail.


mitchellh:
LGTM

PR #5363, opened 2015-02-18
Bundler 1.8.x

hashar:
Same as previous bump bb22a37ed1

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
   s.rubyforge_project         = "vagrant"
 
-  s.add_dependency "bundler", ">= 1.5.2", "< 1.8.0"
+  s.add_dependency "bundler", ">= 1.5.2", "< 1.9.0"
   s.add_dependency "childprocess", "~> 0.5.0"
   s.add_dependency "erubis", "~> 2.7.0"
   s.add_dependency "i18n", "~> 0.6.0"


erwinvaneyk:
Wouldn't be more useful to just remove the upper bound?

sethvargo:
Vagrant relies on many internal components and functions of bundler that are not guaranteed under semver, so we need to take extra precautions when upgrading.

mitchellh:
LGTM but I'm going to do some further testing to verify things are okay here.

hashar:
I haven't tested anything myself. Hopefully your test coverage is good enough :)

pborreli:
:+1: as right now `bundle install` fails anyway

```
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    vagrant (>= 0) ruby depends on
      bundler (< 1.8.0, >= 1.5.2) ruby

  Current Bundler version:
    bundler (1.8.3)

This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?

sethvargo:
Hi @hashar 

Thank you for the Pull Request. Unfortunately we will not be merging this at this time. You can read more [in my comment on the issue](https://github.com/mitchellh/vagrant/issues/5581#issuecomment-91889166), but we need to do extensive testing when we update Bundler because of our dependency on some internal APIs. As a result, we cannot just update Bundler without some rigorous testing first.

We will update to the latest version of Bundler with the next release of Vagrant :smile:

hashar:
@sethvargo you are welcome.  I made the PR to bump bundler to 1.8.x merely to trigger a discussion. In case someone reach this pull request, the upper limit for bundler has been raised to 1.10.5 with Vagrant v1.7.3 (f93532ce11d9)

PR #5385, opened 2015-02-24
Reorder forward port args in Getting Started

davidsantoso:
Just a quick update to Networking under Getting Started. In the main Networking section, there's a line that states, 

"In the case of forwarded ports, two numeric arguments are expected: the port on the guest followed by the port on the host that the guest port can be accessed by."

However, in the Networking section under Getting Started, it has them flipped around. So, this flips them so it's consistent with the main Networking section in the docs where the examples show `guest` followed by `host`.


mitchellh:
LGTM.

PR #5410, opened 2015-03-01
Added option '--destroy' for the 'reload' command.

### LINKED ISSUE #5378: New Command: vagrant rebuild
Right now i have an alias to `vagrant destroy --force && vagrant up` for machines that i want to tear all the way down and stand back up.  It would be nice if there was a vagrant rebuild command that can optionally take --force or a machine name (in a multi machine environment) that would destroy and up a machine in a single command.

trensen:
This PR introduces a new option `--destroy` to the `reload` command, which is a shorthand for `vagrant destroy --force` followed by `vagrant up`.

Fixes #5378.

### DIFF
--- a/plugins/commands/reload/command.rb
+++ b/plugins/commands/reload/command.rb
@@ -4,6 +4,10 @@ require "vagrant"
 
 require Vagrant.source_root.join("plugins/commands/up/start_mixins")
 
+# Needed for the "--destroy" option
+require Vagrant.source_root.join("plugins/commands/destroy/command")
+require Vagrant.source_root.join("plugins/commands/up/command")
+
 module VagrantPlugins
   module CommandReload
     class Command < Vagrant.plugin("2", :command)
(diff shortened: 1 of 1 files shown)


berendt:
I would prefer to introduce a new command `rebuild` like suggested in the linked issue #5378.

sethvargo:
Hi @trensen 

I am going to close this PR as a result of discussion in the original issue. We think this should be a separate command instead of a flag. I would happily review a PR for the additional command :smile:. Sorry!

trensen:
@sethvargo no problem, I'll propose an additional command then!
Just to be sure: should it be `rebuild`?

sethvargo:
@trensen I think someone beat you to it: https://github.com/mitchellh/vagrant/pull/5613 :smile:

trensen:
Ah, too bad for me, but great to have such command being introduced! :blush:

PR #5416, opened 2015-03-01
JSON and YAML output for vagrant ssh_config

theist:
Hi! 

This is a quick hack to have --json and --yaml versions of the vagrant ssh_config output. 

This is something I needed for another project, perhaps could be useful for the community.

Sorry no tests :( I do not know how to test it. I checked several outputs manually

### DIFF
--- a/plugins/commands/ssh_config/command.rb
+++ b/plugins/commands/ssh_config/command.rb
@@ -23,10 +23,19 @@ module VagrantPlugins
           o.on("--host NAME", "Name the host for the config") do |h|
             options[:host] = h
           end
+          options[:output] = "text"
+          o.on("--json", "Print config in json format") do
+            options[:output] = "json"
+          end
+          o.on("--yaml", "Print config in yaml format") do
+            options[:output] = "yaml"
+          end
         end
 
         argv = parse_options(opts)
         return if !argv
+
+        vm = {vagrant_machines: []}
 
         with_target_vms(argv) do |machine|
           ssh_info = machine.ssh_info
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @theist 

Thank you for the PR. I think this is valuable, but I'm not sure I agree with the approach. For YAML and JSON, we should be using `JSON.pretty_generate` and `YAML.dump` instead of using a custom JSON/YAML template.

theist:
Opps , I Used a template because "plain text "  was Using  one as well. Sincerely, I'm a sysadmin who sometimes does some quick coding fixes. I'll try to rewrite it to use the approach you mention.

sethvargo:
Hi @theist 

Thank you for your reply. When you are ready for a re-review, just post a comment or @mention me. Thanks!

theist:
Hi @sethvargo . Done using your suggestions, now is much cleaner. I'm also added a richer description to the commit message.

I'm not very confident about root json / yaml naming and the use of an array instead a hash with host_keys as identifiers. If you feel that I can arrange data in a better way just tell me and I'll force push again mi branch to update this PR

sethvargo:
Hi @theist 

Thank you for your reply. I'd like to better understand the motivation here before merging. In general, we prefer to use `--machine-readable`, which is neither json nor yaml, but consistent across all of Vagrant.

theist:
Hummmm @sethvargo that was almost a year ago!. When this was written `--machine-readable` was an upcoming feature that wasn't supported on  `ssh_config` command and I needed a way to let  capistrano and other tools find what vagrant machines were available.

I use my fork for a couple of months then tools were changed and it's no longer mantained. I think you can drop this contribution and follow your own plans :)

PR #5423, opened 2015-03-03
Fix for nfs on Fedora 21

### LINKED ISSUE #5321: on fedora 21 guest, nfs won't works
Hi, 

On fedora 21 guest, nfs client won't works: 

    /etc/init.d/rpcbind restart; /etc/init.d/nfs restart
    
    Stdout from the command:
    
    Stderr from the command:
    
    bash: line 3: /etc/init.d/rpcbind: No such file or directory
    bash: line 3: /etc/init.d/nfs: No such file or directory
    
    /etc/init.d/rpcbind restart; /etc/init.d/nfs restart

    Stdout from the command:

    Stderr from the command:

    bash: line 3: /etc/init.d/rpcbind: No such file or directory
    bash: line 3: /etc/init.d/nfs: No such file or directory

    INFO interface: Machine: error-exit ["Vagrant::Errors::VagrantError", "The following SSH command responded with a non-zero exit status.\nVagrant assumes that this means the command failed!\n\n/etc/init.d/rpcbind restart; /etc/init.d/nfs restart\n\nStdout from the command:\n\n\n\nStderr from the command:\n\nbash: line 3: /etc/init.d/rpcbind: No such file or directory\nbash: line 3: /etc/init.d/nfs: No such file or directory\n"]

I guess, that we need this fix for fedora >= 21, too:
https://github.com/mitchellh/vagrant/commit/a712f70634fa19fbabdb0b2061d77f5b15c9f6ff

Regards,
sileht

laetificat:
It uses the same logic as a712f70634fa19fbabdb0b2061d77f5b15c9f6ff
Closes #5321

### DIFF
--- a/plugins/guests/fedora/plugin.rb
+++ b/plugins/guests/fedora/plugin.rb
@@ -21,10 +21,21 @@ module VagrantPlugins
         Cap::ConfigureNetworks
       end
 
+      guest_capability("fedora", "flavor") do
+        require_relative "cap/flavor"
+        Cap::Flavor
+      end
+
       guest_capability("fedora", "network_scripts_dir") do
         require_relative "cap/network_scripts_dir"
         Cap::NetworkScriptsDir
       end
+
+      guest_capability("fedora", "nfs_client_install") do
+       	require_relative "cap/nfs_client"
+       	Cap::NFSClient
+      end
+
     end
   end
 end
--- a/plugins/guests/fedora/plugin.rb
+++ b/plugins/guests/fedora/plugin.rb
@@ -32,8 +32,8 @@ module VagrantPlugins
       end
 
       guest_capability("fedora", "nfs_client_install") do
-       	require_relative "cap/nfs_client"
-       	Cap::NFSClient
+        require_relative "cap/nfs_client"
+        Cap::NFSClient
       end
 
     end
--- a/plugins/guests/fedora/cap/flavor.rb
+++ b/plugins/guests/fedora/cap/flavor.rb
@@ -1,0 +1,24 @@
+module VagrantPlugins
+  module GuestFedora
+    module Cap
+      class Flavor
+        def self.flavor(machine)
+          # Read the version file
+          output = ""
+          machine.communicate.sudo("cat /etc/redhat-release") do |type, data|
+            output += data if type == :stdout
+          end
+          output.chomp!
+
+          # Detect various flavors we care about
+          if output =~ /(Fedora release 21).+/i
+            return :f21
+          else
+            return :older
+          end
+	end
+      end
+    end
+  end
+end
+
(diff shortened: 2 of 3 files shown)


plugins/guests/fedora/plugin.rb
```diff
@@ -21,10 +21,21 @@ class Plugin < Vagrant.plugin("2")
...
+      end
+
       guest_capability("fedora", "network_scripts_dir") do
         require_relative "cap/network_scripts_dir"
         Cap::NetworkScriptsDir
       end
+
+      guest_capability("fedora", "nfs_client_install") do
+       	require_relative "cap/nfs_client"
+       	Cap::NFSClient
```
sethvargo: The spacing is a little off here - I think this might be a tab instead of two spaces on accident. Could you please fix that up?

voxik:
Well, the fixes are bit strange to me. Why is F21 special cased? The F21 approach is the future and should be default, otherwise it soon breaks with F22, not mentioning rawhide. Also, DNF should be considered to be used instead of YUM.

laetificat:
@voxik You're totally right! I'm currently making some changes in the spare time I have. What do you think about this? When the version is > 20 it will use dnf instead of yum, and the way I figure out how to check is to strip out the number after "Fedora release" and cast is as an int to see if it's higher than 20.
My ruby skills are still minimal, this is just the second time I'm working with ruby so any tips are welcome!

Part of nfs_client.rb:
```
...
  case machine.guest.capability("flavor")
  when :newer
    comm.sudo("dnf -y install nfs-utils nfs-utils-lib")
    comm.sudo("/bin/systemctl restart rpcbind nfs")
  else
    comm.sudo("yum -y install nfs-utils nfs-utils-lib")
    comm.sudo("/etc/init.d/rpcbind restart; /etc/init.d/nfs restart")
  end
end
...
```

Part of flavor.rb:
```
...
def self.flavor(machine)
  # Read the version file
  output = ""
  machine.communicate.sudo("cat /etc/redhat-release") do |type, data|
    output += data if type == :stdout
  end
  output.chomp!

  # Detect various flavors we care about
  check = /Fedora release ([0-9]+)(.*)/i.match(output)
  if check[1].to_i > 20
    return :newer
  else
    return :older
  end
end
...
```

nmelchior:
I'm having this problem with a Fedora 20 guest, which also uses systemd.  The F20 nfs-utils RPM installs systemd units, not /etc/init.d/ scripts.

voxik:
Wouldn't be better to detect if DNF is available or SystemD is in use instead of detecting some versions of Fedora?

laetificat:
I haven't had much time to work on it yet, but the best way to handle it is
indeed by checking if systemd and/or dnf is installed on the system.

sethvargo:
Hi @laetificat 

Thank you for the Pull Request. As I mentioned in the other Pull Request, I cherry-picked the fedora-specific components into master. This will no longer merge cleanly, and it is still unclear to me what the purpose of the PR is (to avoid using systemd when it's not installed?)

Could you please submit a new PR with _just_ the changes you want to implement? I also think @voxik has some good ideas on this PR. Thank you :smile: and sorry for the delayed review.

Issue #4533, opened 2014-09-21
vagrant box update does not honor insecure option nor ca-cert option


Xylakant:
I have a box hosted on vagrantcloud that has the actual box file on an SSL-enabled server with a self-signed cert. I can download and install the box using `vagrant box add --insecure` but that flag is not retained, so calling `vagrant box update` fails with a cert error. The corresponding option in the Vagrantfile (`config.vm.box_download_insecure`) is not honored either.

Since `vagrant box update` does not accept the --cacert option and the corresponding option `config.vm.box_download_ca_cert` in the Vagrantfile is not honored either, it is currently impossible to update boxes from SSL-Servers with self-signed certificated.

ibizaman:
Easy enough to patch, I'm adding those four options to `vagrant box update`:
```
--insecure
--cacert
--capath
--cert
```
Expect a PR tonight.

mitchellh:
Just tagging #4473 so we can see the link.

mitchellh:
Fixed by #4473

PR #5425, opened 2015-03-03
copy security options from box add for box update

ibizaman:
Without those options, a box update cannot always download the new box's
version if one of the 'security' options was provided on `box add`.

This fixes issue #4533.


sethvargo:
Hi @ibizaman 

Thank you for the Pull Request. This is actually a duplicate of https://github.com/mitchellh/vagrant/pull/4473. Since that PR is a bit older and covers a wider gambit, I am going to close yours in favor of 4473 (sorry!). Thank you for the PR and I would love your feedback on 4473. Thanks! :smile:

PR #5433, opened 2015-03-04
core: allow IO redirection of UI for testing

marxarelli:
Use of $stdin, $stdout, and $stderr globals makes testing difficult. By
exposing the IO objects as writable attributes, input/output can be more
easily simulated using StringIO or doubles.

### DIFF
--- a/lib/vagrant/ui.rb
+++ b/lib/vagrant/ui.rb
@@ -21,9 +21,22 @@ module Vagrant
       # specific. See the implementation for more docs.
       attr_accessor :opts
 
+      # @return [IO] UI input. Defaults to `$stdin`.
+      attr_accessor :stdin
+
+      # @return [IO] UI output. Defaults to `$stdout`.
+      attr_accessor :stdout
+
+      # @return [IO] UI error output. Defaults to `$stderr`.
+      attr_accessor :stderr
+
       def initialize
         @logger   = Log4r::Logger.new("vagrant::ui::interface")
         @opts     = {}
+
+        @stdin  = $stdin
+        @stdout = $stdout
+        @stderr = $stderr
       end
 
       def initialize_copy(original)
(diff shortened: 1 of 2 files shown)


sethvargo:
Hi there,

Why does that make it difficult for testing? `$stdout` is global, but it's not the same as `STDOUT` which is a global constant. You can easily change `$stdout` to whatever you want:

```ruby
def capture(io)
  old_stdout = $stdout
  $stdout = io
  yield
ensure
  $stdout = old_stdout
end
```

And then use it like this:

```ruby
io = StringIO.new
capture(io) do
  some.vagrant.command
end

# Assuming RSpec, but you get the idea
expect(io.string).to eq("foo")
```

I think this is actually recommended by the RSpec core team and RSpec 3 has a built-in `capture_output` matcher that wraps all that logic for you.

marxarelli:
There are a few reasons why that I avoid that pattern in testing, and the `output` matcher in particular, if at all possible—and globals in general but I'll try not to beat that drum too much.

 1. Doubling and mocking globals, especially those dealing with IO, can interfere with the test runner IO itself. Add a debugger to the mix that needs to read from stdin and write to stdout/stderr, and now you have more of the same problem. I run into this issue more than you might think when debugging tests in Ruby projects. 
 2. Mock expectations are unpredictable as it's uncertain what other component will send to the global object.
 3. It promotes bad design, and one of the things I like most about TDD is that it promotes clean interfaces and proper encapsulation.
 4. And of course they lead to [race conditions](https://github.com/rspec/rspec-expectations/issues/642) and some day I'd like to say I have a handle on multithreading in Ruby. :)

In this particular case, there was clearly a lot of effort put towards properly abstracting the software's interaction with the user. Why not take the last tiny step and remove the globals?

PR #5436, opened 2015-03-05
Unlink few Gem temp files from /tmp

valchonedelchev:


sethvargo:
Seems related to https://github.com/mitchellh/vagrant/pull/4876. What do you think @valchonedelchev?

valchonedelchev:
@sethvargo  I think so. Feel free to ignore my pull request if you like.

obfusk:
@sethvargo, @mitchellh: doesn't this reintroduce #4869?

brandt:
@sethvargo @mitchellh just tested - this does indeed cause a regression of #4869

mitchellh:
Let's look at solving the issue another way since this causes a regression. Thanks for the attempt!

PR #5468, opened 2015-03-12
Add "--path" option that overwrites VAGRANT_CWD

### LINKED ISSUE #5453: Lookup Path option for CLI
It would be great to be able to overwrite the default [Lookup Path](http://docs.vagrantup.com/v2/vagrantfile/) with a CLI parameter. It is currently possible by overriding the `VAGRANT_CWD` environment variable, but having a CLI parameter seems more convenient to me.

Please let me know if there are any reasons against it, and/or possibly any pointers on how I could implement this.

Turbo87:
Resolves #5453

### DIFF
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -153,6 +153,19 @@ begin
   # Default to colored output
   opts[:ui_class] ||= Vagrant::UI::Colored
 
+  # Allow user to override cwd via --path
+  argv.each_index do |i|
+    arg = argv[i]
+
+    if arg == "--path" or arg == "-p"
+      argv.delete(arg)
+      arg = argv[i]
+      opts[:cwd] = arg
+      argv.delete(arg)
+      break
+    end
+  end
+
   # Recombine the arguments
   if !argv_extra.empty?
     argv << "--"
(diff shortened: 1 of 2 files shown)


sethvargo:
Hi @Turbo87 

Thank you for the PR. I'll wait for @mitchellh, but I do not think this is going to behave as you expect. The reason we use an environment variable for this is because plugins and other parts of Vagrant are loaded prior to the CLI flags being parsed. By moving this to a CLI flag, it is possible that earlier stages of Vagrant would use the old value, causing some rather interesting side-effects.

Turbo87:
@sethvargo I admit that I haven't tested that with any plugins. Do you think it would work if I put the code further up before the plugins are handled? any other ideas?

sethvargo:
@Turbo87 my gut says "no", because [of the way the Vagrant installers launch Vagrant](https://github.com/mitchellh/vagrant-installers/blob/master/substrate/modules/vagrant_substrate/templates/vagrant.erb).

I'm not sure I understand the unwillingness to use an environment variable. Could you explain that a bit more?

PR #5469, opened 2015-03-12
communicators/winrm: don't stop task on idle end

### LINKED ISSUE #5362: vmware fusion 7.1.1 puppet dies too quickly
After update to fusion 7.1.1, the puppet provisioner [when provisioning windows server 2012r2] is dying at random points on the initial up.  After the initial up, when you run a vagrant provision, all provisioning steps run perfectly fine.

VMWare Fusion 7.1.1 is borked, 7.1.0 works perfectly

We have not tested other versions of windows other than Server 2012 R2.

Vagrant Version 1.7.2
vagrant-cachier (1.2.0)
vagrant-guestip (0.2)
vagrant-hosts (2.3.0)
vagrant-share (1.1.3, system)
vagrant-vmware-fusion (3.2.3)

phinze:
StopOnIdleEnd was set in the task definition for elevated/privileged
windows guest scripts. This setting:

> specifies that the task stops when the idle condition ceases to be true [1]

The "idle condition" is something that Windows periodically checks for,
and it's defined by a bunch of criteria like user presence/absence, CPU
/ IO idle time, etc. [2]

Telling our provisioner to stop the task if the "idle condition" ceases
to be true is a recipe for some sporadically stopped tasks, which seems
like precisely the behavior being reported in #5362.

I'm pretty sure this fixes #5362

[1] https://msdn.microsoft.com/en-us/library/cc248332.aspx
[2] https://msdn.microsoft.com/en-us/library/windows/desktop/aa383561%28v=vs.85%29.aspx

### DIFF
--- a/plugins/communicators/winrm/scripts/elevated_shell.ps1.erb
+++ b/plugins/communicators/winrm/scripts/elevated_shell.ps1.erb
@@ -27,7 +27,7 @@ $task_xml = @'
     <StartWhenAvailable>false</StartWhenAvailable>
     <RunOnlyIfNetworkAvailable>false</RunOnlyIfNetworkAvailable>
     <IdleSettings>
-      <StopOnIdleEnd>true</StopOnIdleEnd>
+      <StopOnIdleEnd>false</StopOnIdleEnd>
       <RestartOnIdle>false</RestartOnIdle>
     </IdleSettings>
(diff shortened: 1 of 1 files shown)


sethvargo:
:+1: 

I also :clap: for reading the MSDN docs :smile:

PR #5478, opened 2015-03-13
Ignore Windows NICs with nil :net_connection_id

aogail:
When upping a Win XP box, vagrant found a lot of "virtual" network
connections that did not have DHCP enabled, and tried to configure them
for DHCP. This did not work because their :net_connection_id is nil.
Ignoring these network connections enabled the XP box to be upped.

I realize that Windows XP is not officially supported, but I hope that you'll accept contributions that enable it to function. I also tested this change against a Windows 7 box to confirm that this change does not affect a newer Windows version.


sethvargo:
:-1: LGTM

aogail:
Is there anything I can do to get this merged?

aogail:
Cool, thanks!

PR #5482, opened 2015-03-14
Support for Docker 1.5.0 named Dockerfile

duro:
This allows support for passing a named Dockerfile configuration which is now available in Docker 1.5.0

This can be implemented in a provider like so:

```
v.vm.provider "docker" do |d|
  d.build_dir = "."
  d.dockerfile = "Dev.Dockerfile" # << Adding named dockerfile
  d.build_args = ['--tag="duro/my-image"']
  d.remains_running = true
end
```


sethvargo:
:+1: looks reasonable to me

tsikorski:
Hi,
I tried applying your fix to a running 1.7.2 system and still get the same error that I got before namely docker build fails with a "no such file or directory" error.

Here is my Vagrantfile:

        config.vm.provider "docker" do |d|
        d.vagrant_vagrantfile = "../Vagrantfile"
        d.build_dir = "."
        d.dockerfile = "Dockerfile"
   end
 end


and here is the error:


Command: "docker" "build" "--file="\""/mnt/docker_build_67521463ed7524adc0956a0290de1c87/Dockerfile"\""" "/mnt/docker_build_67521463
ed7524adc0956a0290de1c87"

Stderr: time="2015-03-22T13:03:47-04:00" level="fatal" msg="stat /mnt/docker_build_67521463ed7524adc0956a0290de1c87: no such file or
 directory"



Am I missing something or does build just no longer work with docker 1.5?

tsikorski:
Hi,
If you are getting a build error and subsequently apply your fix it might be a good idea to delete .vagrant file in your working directory. If you won't do that you continue to get the old error but if you delete .vagrant your fix does exactly what is expected with regards to docker build. Thanks.

Ted

mitchellh:
LGTM!

Issue #4815, opened 2014-11-11
support long paths on windows hosts, by rebinding the share to a UNC path


Pomax:
Based on https://github.com/mitchellh/vagrant/issues/1953#issuecomment-60852879 and the workaround provided in https://www.virtualbox.org/ticket/11976#comment:2, it might make sense to make vagrant do the rebindng from "plain path" to UNC path during the creation step on windows platforms.

This would allow vagrant to actually be a viable technology on windows hosts for nodejs setups (npm installs are notorious for stupidly long dependency chains, with deduplication only happening *after* the full, deep, installation of all dependencies)

Pomax:
@sethvargo I'd mark this a bug, not an enhancement. Shared dirs are unpredictably broken so we can't rely on them on windows, whereas we can rely on them in unix/linux/osx (you can't ask whether the dir you're going to write works. If you could, this would probably more an enhancement than bug)

sethvargo:
@Pomax I'll defer to @mitchellh on that :smile:

Pomax:
@mitchellh if you're looking in, can you confirm you've seen this bug please? There's a fair amount of activity but no signal that you've seen this makes it seem like we're talking amongst ourselves =)

ClementParis016:
Would be great to see this issue fixed :+1: 

In the meantime, could someone explain how to fix this temporarily while waiting an official update?

celtric:
Can confirm that @chernetsov0 solution works on Windows 7, using Vagrant 1.7.2 and VirtualBox 4.3.20.

```diff
# Lines 495-510 - C:\HashiCorp\Vagrant\embedded\gems\gems\vagrant-1.7.2\plugins\providers\virtualbox\driver\version_4_3.rb
def share_folders(folders)
  folders.each do |folder|
    args = ["--name",
      folder[:name],
      "--hostpath",
-       folder[:hostpath]]
+      '\\\\?\\' + folder[:hostpath].gsub(/[\/\\]/,'\\')]
    args << "--transient" if folder.key?(:transient) && folder[:transient]

    # Enable symlinks on the shared folder
    execute("setextradata", @uuid, "VBoxInternal2/SharedFoldersEnableSymlinksCreate/#{folder[:name]}", "1")

    # Add the shared folder
    execute("sharedfolder", "add", @uuid, *args)
  end
end
```

sirlatrom:
+1

jgoux:
[...]

Issue #1953, opened 2013-07-19
long paths in the vagrant ubuntu disk on win7 causes problems with npm, etc.


Pomax:
From https://github.com/isaacs/npm/issues/3670, if I run Vagrant on windows 7, the host max path length is 260 characters at the utter most without special instructions/commands, so when running something that has to make deep directories, like checking out a large software project or doing an `npm install`, causes "protocol errors".

For `npm` in particular this is a problem; on native windows `npm` can make use of UNC paths to get around the path limit, and it works brilliant, but inside a vagrant VM the "protocol error" means projects can't be instantiated, and consequently not worked with, effectively negating the benefit that vagrant offers.

I don't know if this is something that can be fixed in Vagrant itself (maybe?) or whether it's a virtualbox issue (I have no other virtualisation software to test that with, unfortunately), but if this can be fixed, that would be amazingly useful for software project virtualisation/testing.

drmyersii:
I've spent about 16 hours working on this today (I wish I was exaggerating), and I finally have a solution to the Vagrant/Windows issue at hand. I have been working on a Node.js development environment built on Vagrant, and I have been relentlessly determined to get this working. 

[This is the commit that I implemented it in](https://github.com/renobit/vagrant-node-env/commit/bdf15f2f301e2b1660b839875e34f172ea8be227), but I'll go ahead and post the main snippet of code here:

```ruby
config.vm.provider "virtualbox" do |v|
    v.customize ["sharedfolder", "add", :id, "--name", "www", "--hostpath", (("//?/" + File.dirname(__FILE__) + "/www").gsub("/","\\"))]
end

config.vm.provision :shell, inline: "mkdir /home/vagrant/www"
config.vm.provision :shell, inline: "mount -t vboxsf -o uid=`id -u vagrant`,gid=`getent group vagrant | cut -d: -f3` www /home/vagrant/www", run: "always"
```

[...]

Issue #713, opened 2012-02-05
Read-only file system. What's wrong?


itsNikolay:
If you are not agains I continue after this discussion:
http://stackoverflow.com/questions/9150732/capistrano-deploy-in-virtual-machne

**\* [err :: 33.33.33.10] creating symbolic link `/vagrant/demoapp/current/log'
*** [err :: 33.33.33.10] : Read-only file system
and
itsnikolay@itsnikolay-VirtualBox:~/vagrant/1vagrant/projects/demoapp$ ssh vagrant@33.33.33.10 'ln -s /vagrant/demoapp/shared/log /vagrant/demoapp/current/log'
vagrant@33.33.33.10's password: 
ln: creating symbolic link`/vagrant/demoapp/current/log': Read-only file system
The same error

Anybody has thoughts about the error, probably you met it before?

iwanwan:
Not  work, I try like this on my Vagrantfile
```
  config.vm.provider "virtualbox" do |v|
    v.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate/v-root", "1"]
  end
```

it return 
```
vm:
* The following settings shouldn't exist: customize
```

curtisblackwell:
I have the same problem as @iwanwan. @trex005's solution doesn't work for me either.

Running Vagrant 1.6.5 on Mac OS X Yosemite.

PrakashThapa:
Same Problem for me. 
```
vm:
* The following settings shouldn't exist: customize```

**Comment by marcofranssen** [comment]
Can't get this to work. Is anyone able to help me out?

I have following setup:

Windows 8.1
Vagrant 1.7.4
VirtualBox 5

I have multiple shared folders like this:

``` ruby
config.vm.synced_folder "../webapp", "/srv/webapp", type: "smb"
config.vm.synced_folder "../backend", "/srv/backend", type: "smb"
...
....
...
```

For the webapp I use nodejs, when I install node packages using npm it tries to create symlinks and fails with the following error.

``` bash
vagrant@devenvbk:/srv/webapp$ npm install --save-dev gulp
npm ERR! Linux 3.2.0-4-amd64
npm ERR! argv "/usr/bin/node" "/usr/bin/npm" "install" "--save-dev" "gulp"
npm ERR! node v0.12.7
npm ERR! npm  v2.11.3
npm ERR! path ../user-home/cli.js
npm ERR! code ENOTSUP
npm ERR! errno -95

[...]

PR #5495, opened 2015-03-18
Proposed fix for #4815 - Windows UNC path to allow > 256 characters

jfbibeau:
Proposed fix for #4815 (and #1953) based on many comments, notably @Ingramz and @celtric. 

This changes the VirtualBox share on Windows to use UNC paths, for which Windows allows > 256 characters.

Long paths become an issue when using npm for example, which can result in multiple sub directories. Once the path reaches 256 characters, the guest OS usually throws a Permission Denied or Protocol Error.

All rake tests passing, but couldn't get acceptance tests for virtualbox running so acceptance tests are not run. Tested in Windows 7 with VirtualBox 4.3.20 and Vagrant 1.7.2.

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -148,7 +148,7 @@ module Vagrant
         # @param [String] path Path to convert to UNC for Windows
         # @return [String]
         def windows_unc_path(path)
-          "//?/" + path.gsub("/", "\\")
+          "\\\\?\\" + path.gsub("/", "\\")
         end
 
         # Returns a boolean noting whether the terminal supports color.
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -144,6 +144,13 @@ module Vagrant
           path
         end
 
+        # Converts a given path to UNC format by adding a prefix and converting slashes.
+        # @param [String] path Path to convert to UNC for Windows
+        # @return [String]
+        def windows_unc_path(path)
+          "//?/" + path.gsub("/", "\\")
+        end
+
         # Returns a boolean noting whether the terminal supports color.
         # output.
         def terminal_supports_colors?
--- a/plugins/providers/virtualbox/driver/version_4_3.rb
+++ b/plugins/providers/virtualbox/driver/version_4_3.rb
@@ -496,10 +496,14 @@ module VagrantPlugins
 
         def share_folders(folders)
           folders.each do |folder|
+            hostpath = folder[:hostpath]
+            if Vagrant::Util::Platform.windows?
+              hostpath = Vagrant::Util::Platform.windows_unc_path(hostpath)
+            end
             args = ["--name",
               folder[:name],
               "--hostpath",
-              Vagrant::Util::Platform.windows? ? ("//?/" + File.expand_path(folder[:hostpath])).gsub("/","\\") : folder[:hostpath]]
+              hostpath]
             args << "--transient" if folder.key?(:transient) && folder[:transient]
 
             # Enable symlinks on the shared folder
--- a/plugins/providers/virtualbox/driver/version_4_3.rb
+++ b/plugins/providers/virtualbox/driver/version_4_3.rb
@@ -499,7 +499,7 @@ module VagrantPlugins
             args = ["--name",
               folder[:name],
               "--hostpath",
-              folder[:hostpath]]
+              Vagrant::Util::Platform.windows? ? ("//?/" + File.expand_path(folder[:hostpath])).gsub("/","\\") : folder[:hostpath]]
             args << "--transient" if folder.key?(:transient) && folder[:transient]
 
             # Enable symlinks on the shared folder
--- a/test/unit/vagrant/util/platform_test.rb
+++ b/test/unit/vagrant/util/platform_test.rb
@@ -13,7 +13,7 @@ describe Vagrant::Util::Platform do
 
   describe "#windows_unc_path" do
     it "correctly converts a path" do
-      expect(described_class.windows_unc_path("c:/foo").to_s).to eql("//?/c:\\foo")
+      expect(described_class.windows_unc_path("c:/foo").to_s).to eql("\\\\?\\c:\\foo")
     end
   end
 end
--- a/test/unit/vagrant/util/platform_test.rb
+++ b/test/unit/vagrant/util/platform_test.rb
@@ -10,4 +10,10 @@ describe Vagrant::Util::Platform do
       expect(described_class.fs_real_path("c:/foo").to_s).to eql("C:/foo")
     end
   end
+
+  describe "#windows_unc_path" do
+    it "correctly converts a path" do
+      expect(described_class.windows_unc_path("c:/foo").to_s).to eql("//?/c:\\foo")
+    end
+  end
 end


Pomax:
pinging @mitchellh in the hopes of getting eyes-with-merge-powers on this, too =)

francisoud:
+1
I confirm this PR fixes the following issues: #1953 #4815
I tested it with vagrant 1.7.2 under Windows 7 Pro
Will also fix: https://github.com/npm/npm/issues/3670

jordanilchev:
:+1: I confirm this fixes the npm/npm#3670

Vagrant 1.6.2
Windows 8.1
VirtualBox 4.3.26
Image Ubuntu 14.04

djmittens:
Can we get this released soon? I really need this to work.

sethvargo:
Hi @jfbibeau et al.,

This is looking really good! I would like to see the following changes:

1. Move the Windows path conversion into its own method:

    ```ruby
    def windows_unc_path(path)
      "//?/" + path.gsub("/","\\")
    end
    ```

    Perhaps it would make sense to move that method into a util module like `lib/vagrant/util/safe_path.rb`? What do you think?

    Note that this method does not guard with a `Platform.windows?`. This makes it much easier to add a simple test for this function. It would be great if you could add a test or two that handles that case.

2. Move the ternary operator into its own function in the provider or perform the calculation above the creation of `args`:

    ```ruby
    hostpath = folder[:hostpath]
    if Vagrant::Util::Platform.windows?
      hostpath = Vagrant::Util::SafePath.windows_unc_path(hostpath)
    end
    ```

    If you move it into its own method, it will be easier to test, but I'm indifferent here.

3. You'll notice that I removed the `File.expand_path` call. This is because Ruby defaults to expanding the path relative to the CWD. Since Vagrant can be run from anywhere, we always expand paths relative to the Vagrantfile. If you need to expand the path, you should do something like the following:

    ```ruby
    expanded = File.expand_path(hostpath, @env.root_path)
    hostpath = Vagrant::Util::SafePath.windows_unc_path(expanded)
    ```

    But you do not have access to `@env` or `@machine` inside the provider to the best of my knowledge. We also [expand the path here](https://github.com/mitchellh/vagrant/blob/master/plugins/kernel_v2/config/vm.rb#L587), further up the stack. Will that do the right thing? Can we live without the path expansion?

Ingramz:
@sethvargo UNC pathname cannot be relative, so whatever `\\?\` is prepended to, has to be absolute. Also this path name is not for ruby to use, but for virtualbox itself.

sethvargo:
@Ingramz I don't have a Windows box handy, but it would be helpful to know if `File.expand_path` when given a relative UNC path will expand it to the full UNC path :smile:.

Ingramz:
@sethvargo it can do it like this (to my surprise)
```
irb(main):010:0> File.expand_path("\\\\?\\X:\\.\\asd\\..\\.\\bsd")
=> "//?/X:/bsd"
```
But for virtualbox, we need to convert the slashes back. I am quite sure it didn't without the slash conversion.

Ingramz:
@sethvargo after investgating what's needed and what's not. It seems like your points are valid. We do not need to expand the path anymore  and everything works just fine by just prepending `//?/` without converting slashes to the final path (meaning virtualbox is just fine with it that way). Implement it in whichever way you like.

It would be interesting to find out why I initially thought otherwise.

jfbibeau:
Aaaalrighty, I updated the PR with what was discussed by @Ingramz and @sethvargo .

1. Moved the UNC path code to a function in Util::Platform. There's already a bunch of similar functions that do path manipulation for Cygwin and such, so I thought it was a good place to put it.

2. Moved the hostpath manipulation to the top of the function along with the Windows guard.

3. Added a Unit Test for the function.

Definitely don't need to expand the path - by the time we reach that code, the path has already been expanded for us. So we always get an absolute path. 

So `c:/foo` or `../../` in your Vagrantfile would both result in something like `//?/c:\foo` with this function, which VirtualBox accepts fine as a share path (even though it's a bit weird to look at).

sethvargo:
This LGMT! :+1:

MCGallaspy:
Just dropping by to say I would love to see this merged. I've been encountering the same problem with npm using Vagrant on a Windows host.

Pomax:
@mitchellh want to hit the merge button?

Ingramz:
Can someone test this before it will be merged?

Today when trying to patch a new installation, prepending `//?/` seems to give no errors and shared folder mounts just fine, but going over 260 chars still seems to be an issue. Converting the prefix to `\\?\` seemed to fix this. It would be nice if someone could verify whether this issue really exists or not.

jfbibeau:
@Ingramz Errr you may be right. Hold on while I triple check this and I'll submit a change from //? to \\\\? if needed.

jfbibeau:
@Ingramz You were right. VirtualBox silently accepted the path but didn't treat it as UNC. It definitely needs backslashes for the entire path. 

Changed the unit tests and the util function in 45e14a8 and confirmed I can create a path > 250 characters in Windows whereas it failed before. 

Want to give a quick sanity check and thumbs up when you confirm it works for you too?

Ingramz:
@jfbibeau The changes look exactly what I expected. No further complaints on my end.

Antnee:
This looks really promising, and will fix an issue that I've had for months. Will there be an installer released any time soon?

estani:
+1

fusionstrings:
Eagerly waiting for this life altering merge and subsequent Vagrant release.

iget-master:
Waiting for this merge to fix this issue. There are any prevision?

Jero786:
+1 Waiting...

fusionstrings:
So finally Winter is coming :package: :100:

Jero786:
Any body know when 1.7.3 will be released?? Thanks!

Antnee:
Installer release for Windows any time soon, please? :)

af7:
So now we just patch these 3 files on the "Vagrant" installation and we are good to go? We do not need a Windows Installer (msi) I assume for the patch to work, correct?

Thx for the fix.

jfbibeau:
Yep, you don't need an installer for this. Just replace the files with the new versions. I've been running with this fix for 2 months now without any issues.

af7:
Thx again, I can verify that your fix works!

1-months old fix for something **so** important and this did not make it (still) to the Windows Installer?

(npm install did not work at all on my 1st github project that I tried to install and lost tens of hours in frustration, was almost ready to abandon VVV and Vagrant).

Pomax:
I've been waiting for this since July 2013 (https://github.com/mitchellh/vagrant/issues/1953), I can wait a little longer now that a patch has actually landed for an installer to be spun

af7:
@Pomax: I have the feeling that Windows users are somehow "neglected". I first tried to use Vagrant on Windows 10 months ago and I had lots of problems (always something breaking and having to read for hours to find the "trick" or the "patch"), which made it very difficult to use, so I quit... (back then).

Now I am back since eventually things work better (especially after finding this great patch), still I cannot compile a project (npm, etc.) as ssh freezes at some point. And now I do not know how to approach the problem. Is it vagrant? Is it ssh? Is it the script itself (doubt it as it works for others on Unix). Is it Virtual box? Is it my Windows 8.1 security settings? Too much to digest...! 

Lastly, I need to say that this is *not* a complaint! I fully understand the limitations of time the Vagrant contributors have and I thank them for their hard work. It is just a little wish that as "poor" Windows users get some proven solutions a little faster (if possible) on the MSI installer. 

Vagrant should *"just work"* after installing it (meaning that scripts should compile/run fine within it, without too much reading on the web).

Thanks!

mitchellh:
@af7 Sorry, I used Windows full time for 6 months dedicated only to making Windows work well on it. I fixed a _lot_ of bugs, and it worked well on my system. But I found out that Windows systems are all very different, since the permutations for software used to run unix-like software is pretty high (is it plain cmd.exe? Is it RubyInstaller? Is it mingw shell? Is it cygwin? Is it cygwin calling exes? etc.).

Because of this, I'm sure I missed a lot of cases, but we do care a lot about Windows and we'll work on improving it.

af7:
Thx mitchellh, I am sure you do your best to satisfy all "genres"! (just please have any important patches on the MSI sooner, don't wait for the "major release", if possible)

I fully understand what you are going through with so many different configurations, so thanks again.

Antnee:
I've copied these three changed files into my 1.7.2 installation and it still doesn't work for me. An npm install reports lots of errors on a shared folder, but running it on a non-shared folder works fine. However, I also can't copy the directory from the non-shared directory into the shared one, getting errors like
```
cp: cannot create symbolic link ‘/vagrant/a.bcdefghijklm.co.uk/node_modules/gulp-ruby-sass/node_modules/.bin/win-spawn’: Protocol error
```

I also tried using rsync and I get this similar error
```
rsync: symlink "/vagrant/a.bcdefghijklm.co.uk/node_modules/browserify/node_modules/shasum/node_modules/.bin/sha.js" -> "../sha.js/bin.js" failed: Protocol error (71)
```

I've looked at #713 as well but nothing in there works for me, though I wouldn't expect it to as @mitchellh made those changes default as of v1.1

* Vagrant 1.7.2 (plus #5495 patch)
* Windows 8.1
* VirtualBox 4.3.28
* ubuntu/trusty64 box

Does anyone have any further ideas? This is driving me nuts!

Antnee:
Actually, I seem to have got it working now. I also followed [this](https://github.com/cityindex-attic/logsearch/issues/7#issuecomment-18982255) to ensure that my Windows user has permissions to create symlinks. I also did [this](https://github.com/mitchellh/vagrant/issues/713#issuecomment-17296765) though without rebooting my host machine. I did need to `vagrant up` from an admin command prompt though. I ran an `npm install`, a `bower install` and `gulp` and they all ran great and I have finally been able to build the project that my team have been working on for months.

alexciarlillo:
Is this in 1.7.4? I see the additions from platform.rb and platform_test.rb, but not driver/version_4_3.rb in my vagrant install.

jfbibeau:
It got fixed in 1.7.3, but reverted in 1.7.4 as it was causing issues with some versions of VirtualBox. Check out #5933

Ingramz:
Has anyone tried it with VirtualBox 5.0.2 yet to see if it works there again? I do not have access to a windows host till the end of the week.

Tomson124:
Just tried it with VirtualBox 5.0.2 but still doesn't work. The shared folders don't get mount.

af7:
It would be great if this could be fixed, thank you for your efforts.

Tomson124:
Just tried it with VirtualBox 4.3.30 (last 4.x.x release) and this fix works totaly fine, so I think it is a "problem" or a bug with VirtualBox 5.x.x and not with Vagrant.

af7:
Thanks, this is another problem though because Windows 10 are only
supported by VirtualBox 5 (not even the official version as of today, but
rather the "test" versions).

On Fri, Aug 28, 2015 at 9:27 PM, Thomas Wolfgruber <notifications@github.com

Pomax:
**WHERE THIS PATCH APPLIES, AND WHERE IT DOESNT**

This patch works for VirtualBox 4.3.x, and was landed in Vagrant 1.7.3.

However, it was taken back out in Vagrant 1.7.4, because it breaks VirtualBox 5.0 instances on windows. See [this PR](https://github.com/mitchellh/vagrant/issues/5933) for more details.

Issue #4823, opened 2014-11-13
SMB synced folder fail on Windows 8.1 host


benjah1:
Host: Windows 8.1 Pro 
VM: Ubuntu Trusty 14.04 LTS x64
Vagrant: 1.6.5
Virtualbox: 4.3.18

Tried on both cygwin and powershell, and got the same error.

```
$ vagrant reload
==> default: Attempting graceful shutdown of VM...
==> default: Checking if box 'puphpet/ubuntu1404-x64' is up to date...
==> default: Clearing any previously set forwarded ports...
==> default: Preparing SMB shared folders...
    default: You will be asked for the username and password to use for the SMB
    default: folders shortly. Please use the proper username/password of your
    default: Windows account.
    default:
    default: Username: smb
    default: Password (will be hidden):
Error! Your console doesn't support hiding input. We'll ask for
input again below, but we WILL NOT be able to hide input. If this
is a problem for you, ctrl-C to exit and fix your stdin.
     default: Password (will be hidden): A12345
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
    default: Adapter 2: hostonly
==> default: Forwarding ports...
    default: 22 => 7754 (adapter 1)
    default: 22 => 2222 (adapter 1)
==> default: Running 'pre-boot' VM customizations...
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
    default: Warning: Connection timeout. Retrying...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
==> default: Checking for host entries
==> default: Configuring and enabling network interfaces...
==> default: Mounting SMB shared folders...
    default: C:/Users/Benjamin/Documents/project/node => /var/www
[...]

PR #5496, opened 2015-03-18
Fix smb folder mounting issue #4823

francisoud:
Fix issue #4823


sethvargo:
Hrm. So:

> "sec" is the security mode and determines how passwords are encrypted between server and client ( even if you don't require passwords ).
> 
> ntlm used to be the default which is why you never had to specify it discretely. Things have moved on however so the default is now ntlmssp. If you are accessing something which doesn't speak ntlmssp you have to override the new default with the old one.
> 
> Most NAS devices use older technology so they often require ntlm. If you access an OSX samba share however it requires ntlmssp so before you had to specify that in the mount options but now you don't.

Thoughts?

francisoud:
Never mind  i don't plan on using vagrant with smb anytime soon

Issue #5361, opened 2015-02-18
Vagrant scp


eidge:
I just realized there is no ```vagrant scp``` command, would adding it to the vagrant-cli be a useful PR?

I currently use a custom script to do so, but should be easily included in the cli.

mitchellh:
I think this would be a decent feature. I think two commands would be great: one to copy TO the guest, one to copy FROM the guest. Vagrant could easily support both since communicators are required to be able to do both.

PR #5501, opened 2015-03-19
Add SCP to vagrant CLI

eidge:
This PR adds an scp option to the vagrant CLI.

It lets one copy files from/to the remote vagrant machine using:

```bash
vagrant scp ~/foo.txt vagrant:~/bar.txt # local to vagrant machine
vagrant scp vagrant:~/bar.txt . # vagrant to local machine
```

I do most of my development work in vagrant machines so that I don't end up polluting my machine with dependencies for each project I work on. I therefore need to copy files back and forward and this command simplifies it.

I know I could use shared directories, but it makes me actually move stuff twice. Once from host to the shared dir and a second from the shared dir to the directory I want it in in the vagrant machine.

There are some rough edges in the code, and some duplication that could be removed, but I could use some feedback since it's the first code that I've written for Vagrant. Still trying to grasp the architecture :)

### DIFF
--- a/lib/vagrant/action.rb
+++ b/lib/vagrant/action.rb
@@ -26,6 +26,7 @@ module Vagrant
       autoload :Provision, "vagrant/action/builtin/provision"
       autoload :ProvisionerCleanup, "vagrant/action/builtin/provisioner_cleanup"
       autoload :SetHostname, "vagrant/action/builtin/set_hostname"
+      autoload :SCPExec, "vagrant/action/builtin/scp_exec"
       autoload :SSHExec, "vagrant/action/builtin/ssh_exec"
       autoload :SSHRun,  "vagrant/action/builtin/ssh_run"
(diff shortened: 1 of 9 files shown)


Mattias-:
You can achieve this by setting up your ssh config.
No need for another vagrant command.

```bash
vagrant ssh-config >> ~/.ssh/config
scp ~/foo.txt vagrant:~/bar.txt # local to vagrant machine
scp vagrant:~/bar.txt . # vagrant to local machine
```

eidge:
It doesn't work on my machine and besides the same could be said about the ssh command right?

Also please see #5361

eidge:
Any news about this?

mitchellh:
I'd rather use the communicator `upload` and `download` methods to implement this so we don't add a hard dependency on `scp`. If we do that, then we can also support WinRM all with the same command. 

The code work was good though, feel free to make this a plugin of course!

Issue #5509, opened 2015-03-22
[hyperv] Import of Generation 1 images fails when setting SecureBoot


johnsonj:
When importing a Generation 1 image, (eg: hashicorp/precise64) an error is raised in the import_vm stage:
```
An error occurred while executing a PowerShell script. This error
is shown below. Please read the error message and see if this is
a configuration error with your system. If it is not, then please
report a bug.

Script: import_vm.ps1
Error:

Set-VMFirmware : A parameter is invalid. Generation 1 virtual machines do not support the VMFirmware cmdlets.  Use Get-VMBios and Set-VMBios instead.
At C:\rails\vagrant\plugins\providers\hyperv\scripts\import_vm.ps1:143 char:5
+     Set-VMFirmware -VM $vm -EnableSecureBoot Off
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Set-VMFirmware], VirtualizationInvalidArgumentException
    + FullyQualifiedErrorId : InvalidParameter,Microsoft.HyperV.PowerShell.Commands.SetVMFirmwareCommand
```

21e2f33 introduced support keeping the secure boot setting of the imported image. It does not check what generation the VM is. Gen 1 VMs do not support this setting.

ssugar:
I've put a pull request together that changes import_vm.ps1 to only perform the EFI set-vmfirmware commands if the vm is not gen 1: https://github.com/mitchellh/vagrant/commit/fd557592d4f6fb96055837b6ce6fb132f8a84d91

vvchik:
that should fix errors that come from #5210 
+1 for this.

sethvargo:
I believe this has been fixed now. If that is not the case, please comment and we can re-open :smile:

johnsonj:
Looks good, thanks! I'll close my PR as well.

PR #5510, opened 2015-03-22
Don't Set-VMFirmware properties on Gen 1 VMs. It's not supported and cau...

johnsonj:
Resolves Issue: #5509


johnsonj:
Resolved by #5538

PR #5514, opened 2015-03-23
Allow to use several boxes with the same name

### LINKED ISSUE #4822: box lock files: two projects with the same box name cannot run operations at the same time
I have project 1, and project 2. Both have a box called "myInfra" defined in them. If in two separate terminal (one for each project), I try to:

```
> vagrant --version
Vagrant 1.6.5
>vagrant up myInfra
```

One of them will get locked out because the box name is the same. These lock currently seem to reside under `~/.vagrant.d/data`; shouldn't they be transferred under `[project]/.vagrant`?

MiLk:
I have different virtual machines with the same name in Vagrant and using the same Vagrantfile, but different names in virtualbox (and different directories).

This allows me to run action on both VMs at the same time.

Fixes #4822

### DIFF
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -166,7 +166,7 @@ module Vagrant
       vf = nil
       vf = @env.vagrantfile_name[0] if @env.vagrantfile_name
       id = Digest::MD5.hexdigest(
-        "#{@env.root_path}#{vf}#{@name}")
+        "#{@env.root_path}#{vf}#{@env.local_data_path}#{@name}")
 
       # We only lock if we're not executing an SSH action. In the future
       # we will want to do more fine-grained unlocking in actions themselves


sethvargo:
Hi @MiLk 

Thank you for the Pull Request. This looks good to me, although I would like sign-off from @mitchellh before merging. 

This will have the unintended side-effect of causing existing machine ids to change. If a user has running VMs managed by Vagrant and installs this newer version, Vagrant will lose a reference to the machine :frowning:. We may need to do some trickery to ensure we migrate effectively (like checking if the old MD5 exists, then migrating to a new one). If that's the case, we may want to switch to SHA2 instead. Thoughts @MiLk @mitchellh?

MiLk:
My understanding of Ruby is pretty bad, but I thought this id is only used for the lock system.

Old locks will stay when updating Vagrant, and running VMs won't have any lock.

We can maybe calculate the old id and the new one, and remove the old lock if there is one and replace it by a new one with the new id.
But I think it's an unwanted level of complexity.

mitchellh:
I actually don't understand the need for this, but it shouldn't affect anything negatively so we can merge it.

Issue #5516, opened 2015-03-23
Remove insecure public ssh key broken on FreeBSD guest.


obnoxxx:
sed -i on FreeBSD requires a backup extension argument.

sethvargo:
Fixed in https://github.com/mitchellh/vagrant/pull/5517

PR #5517, opened 2015-03-23
guests/freebsd: fix call of set -i : missing extension

obnoxxx:
Sed on freebsd seems to be mores strict than on other
platforms about use of -i: the extension is not optional.

Signed-off-by: Michael Adam <obnox@samba.org>


obnoxxx:
fixes issue #5516

PR #5520, opened 2015-03-25
contrib/bash: Auto-complete VMs for up command

Mattias-:
An update to the bash auto-completion script to complete VM names for the `vagrant up` command.

I made a similar commit at https://github.com/kura/vagrant-bash-completion/commit/e720d90ae9a6589eb13c578b40a503e902d4897c earlier but what I really intended was to put it here since this is the "official" location.


sethvargo:
LGTM

PR #5539, opened 2015-03-30
HyperV provider: VlanID configuration possibility

vvchik:
This PR mostly based on #5169 it adds possibility to configure vlan_id for hyperv provider
My improvements gives possibility to leave vlan_id untouched in case it does not set
also added description of this option to documentation.

tests present, feature documented
tested on 2012hyperv server
with #5207 will be great improvements for hyperv networking possibilities, and reduce it limitations


Uburwator:
Great work, it would be great to see it in 1.7.3

PR #5540, opened 2015-03-30
Make sure debian net post-up never returns an error code.

philsc:
Inspired by e7dce43af3905a7a986ef6e3402bf6809ab90f16.

I'm using this configuration in my Vagrantfile:
```ruby
config.vm.network :public_network, use_dhcp_assigned_default_route: true
```

Every once in a while the secondary network interface would not come
up properly and cause the init scripts to wait for ~ 3 minutes before
giving up and continue with booting.

(My guess is that sometimes the VirtualBox dhcp server takes longer to respond than the physical network and that causes the default route deletion to fail).

I haven't seen this issue since I made this change locally on my machine.


philsc:
Any thoughts?
If you need some test cases or something like that, I'm willing to get on that.

mitchellh:
LGTM

PR #5544, opened 2015-03-31
VMware Documentation Fix - SED Syntax

bdwyertech:
We should only create the backup file on the first `sed` modification. The second `sed` operation will create a backup with the changes from the first `sed` operation.


sethvargo:
Thanks!

PR #5549, opened 2015-04-02
fix ftp push check bug

jackywu:
list method is not safe to check if a dir exist or not on remote.

To re-produce the error, you could do  'git clone git@github.com:sinatra/sinatra.git' in directory which Vagrantfile stay in, and run $vagrant push.


sethvargo:
Thanks!

Issue #5027, opened 2014-12-18
vagrant up --provider docker is janky


scarolan:
It seems to only work around 20-30% of the time, requiring multiple runs to get more than one container stood up.  I believe there's something weird going on with the SSH, as I've seen this error several times:

```
/opt/vagrant/embedded/gems/gems/net-ssh-2.9.1/lib/net/ssh/transport/packet_stream.rb:204:in `poll_next_packet': padding error, need 610241404 block 16 (Net::SSH::Exception)
```

This is also common:

```
==> node1: Waiting for machine to boot. This may take a few minutes...
Timed out while waiting for the machine to boot. This means that
Vagrant was unable to communicate with the guest machine within
the configured ("config.vm.boot_timeout" value) time period.

If you look above, you should be able to see the error(s) that
Vagrant had when attempting to connect to the machine. These errors
are usually good hints as to what may be wrong.

If you're using a custom box, make sure that networking is properly
working and you're able to connect to the machine. It is a common
problem that networking isn't setup properly in these boxes.
Verify that authentication configurations are also setup properly,
as well.

If the box appears to be booting properly, you may want to increase
the timeout ("config.vm.boot_timeout") value.
```

mitchellh:
Hm, I've also seen this flakiness. I think we're just hitting SSH too quickly (while the container is mid-startup) and its returning some weird stuff.

andreychernih:
I am also having these from time to time:

```
    apt-cacher: Container created: 27035f071c50d3ea
==> apt-cacher: Starting container...
==> apt-cacher: Waiting for machine to boot. This may take a few minutes...
/opt/vagrant/embedded/gems/gems/net-ssh-2.9.1/lib/net/ssh/transport/packet_stream.rb:204:in `poll_next_packet': padding error, need 2928786717 block 16 (Net::SSH::Exception)
        from /opt/vagrant/embedded/gems/gems/net-ssh-2.9.1/lib/net/ssh/transport/packet_stream.rb:90:in `next_packet'
        from /opt/vagrant/embedded/gems/gems/net-ssh-2.9.1/lib/net/ssh/transport/session.rb:178:in `block in poll_message'
        from /opt/vagrant/embedded/gems/gems/net-ssh-2.9.1/lib/net/ssh/transport/session.rb:173:in `loop'
        from /opt/vagrant/embedded/gems/gems/net-ssh-2.9.1/lib/net/ssh/transport/session.rb:173:in `poll_message'
[...]

PR #5551, opened 2015-04-02
Allow Ansible provisioner to run reliably in parallel

randonn:
The Ansible Vagrant provisioner has a race where the inventory file is
updated every time the provisioner runs unless a file is provided.

Therefore if Ansible attempts to provision two nodes in parallel, you
may see the following race:
   * System A writes the inventory file and calls Ansible.
   * System B starts to provision and truncates the file before
     creating a new one.
   * Ansible on system A now attempts to read the inventory
     file, which is blank. Ansible bombs out with "ERROR: provided
     hosts list is empty".

To fix this, we only allow Vagrant to update the inventory file if
it needs to.


randonn:
I cannot see why this has failed build ? as i am running this code here

sethvargo:
I'm going to defer to @gildegoma since he's done the most work on the Ansible integration in Vagrant.

gildegoma:
Thanks @randonn for this! So far, I was not aware that `vagrant up` could run a batch of actions in parallel against different machines. And indeed, in this special case there is a race condition!

Unfortunately, I could not reproduce the bug because `virtualbox` provider does not support parallel batches and I have been struggling for a while with the `docker` provider on OS X (the `net-ssh` library systematically fails with a padding error when communicating via boot2docker host while provisioning the second VM. This problem does not appear if I execute `vagrant provision` afterwards, only with `up`. This misbehaviour look quite similar to the problems reported in #5027).
@randonn Being curious, which VM provider(s) are you using?

Anyway, your code looks good to me and behaves as expected in "non-parallel" mode. I like your solution (especially to avoid rewriting the file when the content hasn't changed), but since I am not used to write multi-threaded code in Ruby, I'd like to ask @sethvargo (or @mitchellh) to finalize the code review on this aspect (I don't think there will be an "easy" way to cover this mutual exclusion in the unit tests, but maybe we could add at least a check on the modification timestamp...)

@sethvargo: By the way, is it planned to add the `--parallel` option to `vagrant provision` command as well?

berendt:
A `--parallel` option for `vagrant provision` would be a great addition.

mitchellh:
LGTM @gildegoma, is this good to go?

gildegoma:
@mitchellh yep, it's good and merged it! Thanks again @randonn :)

@electrofelix I am pretty sure that the errors that you reported above, are not related to the Ansible provisioner parts. I also observed similar errors when trying to reproduce this bug with the Docker provider on a Linux host and Vagrant 1.7.2. With current Vagrant  ̀master` head, I haven't faced this problem so far. If you still encounter problems with `libvirt` provider, please open a new issue. Thanks!

electrofelix:
@gildegoma so I've traced the problem in the vagrant-libvirt provider to accessing various machine settings that will result in locks being grabbed for the machine. This is fine when the machine is owned by the current thread and it's grabbing it's own locks, but falls apparent when the ansible provisioner is querying each machine from another thread and also attempting to grab the same lock.

see the latest on pradels/vagrant-libvirt#420 for more info

As part of my examination though I believe that it will make sense to add  a retry to get machine ssh_info before continuing in the ansible provisioner since it's clear that a machine may still be coming up and waiting on it to have valid ssh info is likely to result in a better scenario for more users.

gildegoma:
@electrofelix can you please confirm that https://github.com/pradels/vagrant-libvirt/pull/436 (fixing https://github.com/pradels/vagrant-libvirt/issues/428) fully solved the race issue. Thanks!

electrofelix:
Not fully, there is still a race in the Ansible provisioner code. Currently there is an assumption that once a machine fails to provide valid ssh_info, it cannot provide valid ssh_info until it's recreated. For providers where the ip address is discovered and allocated during the initial provisioning process it is possible for the ansible provisioner to be executed on one of the machines before all others are fully created. 

When this happens, querying the ssh_info for all other machines will result in only a subset of machines being written out to the inventory file. Subsequent machines reaching the ansible provisioner will then get results for more machines and trigger a removal and recreation of the inventory file. 

Depending on the number of machines and timing, this can still result in the original problem occurring. _Would be interesting to see what happens if you change the behaviour to write out to a temporary file and move it over the existing inventory file instead of opening to truncate and replace the contents._

Checking the state of each machine that fails to provide ssh_info and retrying a number of times for those that are running would help reduce the chance that any would subsequently provide valid info where previously it returned nil. Though that is still limited since there is no distinct state that distinguishes between a machine that is being booted by a provider and one where the provider has finished bringing the machine up and it is now ready for any provisioner steps to be executed against it.

Creating separate inventory files for each invocation of the ansible provisioner is probably the only way to ensure that it doesn't get deleted by one of the other machines calling the ansible provisioner as long as each thread is going to build an inventory.

Combined with retrying for the ssh_info, it is likely to reduce the odds of a problem occurring except where a machine is not ready in time and is available subsequently and one of the playbooks was expecting to require this machine. Making the retries and timeout configurable would probably allow users to work around any instances of this occuring.

Barring having a common method in vagrant to be able to distinguish across all providers when machines are being build to be able to tell when the provisioner needs to wait for a machine to be ready to provide the ssh_info and when the machine would have to be recreated, I think it's a case of reducing the chance of this occurring rather than eliminating it completely.

PR #5558, opened 2015-04-04
Fixed error in ChangeHostName capability for Darwin

legal90:
#4535 has broken Darwin guests if `config.vm.hostname` is set to FQDN, with dots:
```
<...>
==> default: Setting hostname...

scutil --set LocalHostName holly-chef.com

Stdout from the command:



Stderr from the command:

SCPreferencesSetLocalHostName() failed: Invalid argument
```
I've reproduced it on OS X 10.9.5 and 10.10.2 as well.
It is caused by the fact that `LocalHostName` should not contain dots and it could not be equal the FQDN. 
The good practice here is to leave only the first part of hostname, like I've done.

### DIFF
--- a/plugins/guests/darwin/cap/change_host_name.rb
+++ b/plugins/guests/darwin/cap/change_host_name.rb
@@ -6,7 +6,9 @@ module VagrantPlugins
           if !machine.communicate.test("hostname -f | grep '^#{name}$' || hostname -s | grep '^#{name}$'")
             machine.communicate.sudo("scutil --set ComputerName #{name}")
             machine.communicate.sudo("scutil --set HostName #{name}")
-            machine.communicate.sudo("scutil --set LocalHostName #{name}")
+            # LocalHostName shouldn't contain dots.
+            # It is used by Bonjour and visible through file sharing services.
+            machine.communicate.sudo("scutil --set LocalHostName #{name.gsub(/\.+/, '')}")
             machine.communicate.sudo("hostname #{name}")
           end
         end


legal90:
And of course, this approach works fine if `config.vm.hostname` doesn't contain dots:
```ruby
'i-dont-use-fqdns'.split('.')[0]
=> "i-dont-use-fqdns"
```

legal90:
@sethvargo Could you please check it out? :)

plugins/guests/darwin/cap/change_host_name.rb
```diff
@@ -6,7 +6,7 @@ def self.change_host_name(machine, name)
           if !machine.communicate.test("hostname -f | grep '^#{name}$' || hostname -s | grep '^#{name}$'")
             machine.communicate.sudo("scutil --set ComputerName #{name}")
             machine.communicate.sudo("scutil --set HostName #{name}")
-            machine.communicate.sudo("scutil --set LocalHostName #{name}")
+            machine.communicate.sudo("scutil --set LocalHostName #{name.split('.')[0]}")
```
sethvargo: Ouch. This is a nasty bug. 

I think it might be unexpected if we just take the first `.` without at least telling the user. Can we do something like this instead?

```ruby
safe_hostname = name
if name.include?(".")
  safe_hostname = name.split(".")[0]
  Vagrant.log.warn("Detected a dot in hostname `#{name}' for OS X. Using `#{safe_hostname}' because OS X does not support dots in hostname."
end
```

And then pass `safe_hostname` to the `--set` command. This way the user knows we are doing something different for them.

legal90: @sethvargo It might be reasonable, but I'd like to notice that `LocalHostName` is not a hostname itself. It's just the name identifier used by Bonjour and visible through file sharing services like AirDrop. 
Do you think that we should warn the user anyway?

In GUI this parameter can be set there: "System Preferences" -> "Sharing" -> "Edit..."
I've noticed that OS X suggests `LocalHostName` value based on `ComputerName` with removed dots. May be we should follow the same approach here, in vagrant...
sharing
(".local" is added automatically)

sethvargo: Hrm. Let's just gsub the dots out then, since that's what OSX is doing. Maybe leave a comment in the code explaining that though.

legal90:
@sethvargo 
> Let's just gsub the dots out then, since that's what OSX is doing. Maybe leave a comment in the code explaining that though.

Done

PR #5568, opened 2015-04-08
Support for uri-escaped user/pass in box_url

### LINKED ISSUE #5567: Url-escaped username and password in box_url are not supported
When the user or password in box_url contains special characters, they need to be escaped, in order to have a valid url.
When escaping them, then the download does not work, because the credentials are stripped from the url and fed to curl exactly as they are. Curl takes its credentials unescaped.

Sample url to test: http://user:myP%40ssword@server.com/ubuntu.box

ratzinho87:
Fixes #5567.


mitchellh:
Fixed

PR #5577, opened 2015-04-10
Use service command to restart docker, not upstart.

### LINKED ISSUE #5245: Docker provision not working correctly?
When provisioning a box (using VB for provider) with Docker using this code: 

```
config.vm.provision "docker",
    images: ["debian:7.8"]
```

and running `vagrant up`, I'm getting this error: 

```
==> default: Running provisioner: docker...
    default: Installing Docker (latest) onto machine...
    default: Configuring Docker to autostart containers...

stop docker

Stdout from the command:



Stderr from the command:

stdin: is not a tty
bash: line 2: stop: command not found
```

Running `vagrant reload --provision` then fixes the issue, but not sure why it doesn't work the first time.

thenewwazoo:
Very simple tweak I'm using personally for Debian-based boxes. upstart ships by default on Ubuntu, but not on Debian, and `service` works in both. Fixes #5245.

### DIFF
--- a/plugins/provisioners/docker/cap/debian/docker_configure_auto_start.rb
+++ b/plugins/provisioners/docker/cap/debian/docker_configure_auto_start.rb
@@ -7,8 +7,7 @@ module VagrantPlugins
             machine.communicate.tap do |comm|
               if !comm.test('grep -q \'\-r=true\' /etc/default/docker')
                 comm.sudo("echo 'DOCKER_OPTS=\"-r=true ${DOCKER_OPTS}\"' >> /etc/default/docker")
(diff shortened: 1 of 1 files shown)


displague:
Worked for me. :+1:   This fix shouldn't be at all controversial because `service docker start` is already being used in a neighboring file: https://github.com/luciddg/vagrant/blob/fix/master/docker-provisioner-restart-fix/plugins/provisioners/docker/cap/debian/docker_start_service.rb#L7

qm78:
Confirm, works for me on Debian 8, too. Looking forward to get merged and released.

PR #5591, opened 2015-04-13
Move shell_cmd to a method in SSH::Communicator

quirkey:
This allows `shell_cmd` to be overridable by plugins/patches without having to override the entire (large) `shell_execute` method.

Our use case is specifically to allow us to patch this method in a plugin to get around the change here https://github.com/mitchellh/vagrant/commit/9cd9e6e55d802553fa90e63d966b302006b45de5 which I understand in non-revertable. This gives us a simple way to make the patch without having to run a separate fork or have major impact on the codebase (this actually changes nothing about default behavior and is really just moving code into a method).

### DIFF
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -434,6 +434,21 @@ module VagrantPlugins
         return yield connection if block_given?
       end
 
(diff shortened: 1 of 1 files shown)


sethvargo:
Thanks!

PR #5593, opened 2015-04-14
Salt bootstrap ps1 should be idempotent

sneal:
- WinRM communicator error detection was tightened up thus exposing an issue in the salt bootstrap script creating directories that already exist.


obestwalter:
This would fix https://groups.google.com/forum/#!topic/vagrant-up/gNeoYgpJnMk

sethvargo:
:+1:

PR #5607, opened 2015-04-18
Add :name attribute to shell provisioner.

purpleidea:
This allows the display of friendly output to the user when we think an
important action should be labeled a certain way. This makes vagrant
output a lot more useful when many inline shell provisioners are
present.

This patch includes documentation updates too! w00t :)

### DIFF
--- a/plugins/provisioners/shell/config.rb
+++ b/plugins/provisioners/shell/config.rb
@@ -10,6 +10,7 @@ module VagrantPlugins
       attr_accessor :privileged
       attr_accessor :binary
       attr_accessor :keep_color
+      attr_accessor :name
       attr_accessor :powershell_args
 
       def initialize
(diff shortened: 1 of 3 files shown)


purpleidea:
Ping?

purpleidea:
@mitchellh Can I get a review/merge please?

mitchellh:
Sorry for the delay, we've been focusing elsewhere the past couple months but plan on wrapping back around to Vagrant shortly. We neglected a couple projects for a few months while we pushed some initiatives but we're finally digging ourselves out of that project debt. :) Just a little while longer (maybe a month, sorry!)

berendt:
@mitchellh Maybe it would be a good idea to on-board some external people? It would be nice to improve the review quality and speed. At the moment a lot of PRs are hanging around and this sucks.

PR #5613, opened 2015-04-21
New `vagrant pristine` command

fgrehm:
Closes GH-5378

Ideally we should just reuse the other command plugins instead of copying and pasting but its been a while since I been around the codebase and I can't remember if it would be possible to do that :blush: 

/cc @sethvargo

### DIFF
--- a/plugins/commands/pristine/command.rb
+++ b/plugins/commands/pristine/command.rb
@@ -1,0 +1,130 @@
+module VagrantPlugins
+  module CommandPristine
+    class Command < Vagrant.plugin("2", :command)
+      include StartMixins
+
+      def self.synopsis
+        "recreates machines from scratch, acting like a shortcut to `destroy` and `up`"
+      end
+
+      def execute
+        options = {}
+        options[:force] = false
+        options[:destroy_on_error] = true
+        options[:parallel] = true
+        options[:provision_ignore_sentinel] = false
+
+        opts = OptionParser.new do |o|
+          o.banner = "Usage: vagrant pristine [options] [name]"
+          o.separator ""
+          o.separator "Options:"
+          o.separator ""
+
+          o.on("-f", "--force", "Destroy without confirmation.") do |f|
+            options[:force] = f
+          end
+
+          build_start_options(o, options)
+
+          o.on("--[no-]destroy-on-error",
+               "Destroy machine if any fatal error happens (default to true)") do |destroy|
+            options[:destroy_on_error] = destroy
+          end
+
+          o.on("--[no-]parallel",
+               "Enable or disable parallelism if provider supports it") do |parallel|
+            options[:parallel] = parallel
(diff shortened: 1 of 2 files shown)


plugins/commands/pristine/plugin.rb
```diff
@@ -0,0 +1,17 @@
...
+module VagrantPlugins
+  module CommandPristine
+    class Plugin < Vagrant.plugin("2")
+      name "pristine command"
+      description <<-DESC
+      The `pristine` command is a shortcut for running `destroy` and `up`
+      DESC
+
+      command("pristine") do
+        require File.expand_path("../command", __FILE__)
```
sethvargo: Let's use `require_relative` instead please

plugins/commands/pristine/command.rb
```diff
@@ -0,0 +1,130 @@
...
+          @env.ui.info("", prefix: false)
+
+          m.ui.success(I18n.t(
+            "vagrant.post_up_message",
+            name: m.name.to_s,
+            message: m.config.vm.post_up_message))
+        end
+
+        # Success, exit status 0
+        0
```
sethvargo: Explicit `return` here would be nice :smile:

sethvargo:
Hey @fgrehm 

Thank you for the Pull Request. At a high level, this looks really great. Do you think you could add some tests for this? If you look at the pushes tests, I think you can find a good pattern to use. I'm just really hesitant to merge such a large change without test coverage. Even if it works now, we need to prevent regressions in the future :smile:

fgrehm:
Sure! I'll try to write those tests over the week.

mitchellh:
Thanks for this. I'm not a big fan of the name, I don't know what to expect with `pristine`, for example I still have no idea what `gem pristine` does. I know what "pristine" means but for some reason it has never resonated with me as a command name. I still think this makes sense as a flag to `reload`.

mitchellh:
.@sethvargo is trying to convince me this should be a command and not a flag. We'll delay this for 1.8 in some form.

fgrehm:
No worries! I'm not sure when I'll have a chance to write tests for this so feel free to close the PR if you guys want to hold back on it! :beers:

sethvargo:
Hey @fgrehm 

Thanks for the PR. Coming around back to this (almost a year later), I think I've had a change of heart. After reading through all the comments in the original issue, I think everyone's opinions on "what this should be named" and "what it should do" are too far apart. As such, I'm going to recommend folks use a shell alias to involve whatever their preferred behavior is, instead of building this into core. I totally appreciate your efforts on this PR, but I'm actively working to simplify Vagrant's CLI API, and I don't think it's wise to add new commands at this time, especially since this "works" with some `&&`. Sorry!

fgrehm:
No need to be sorry! 🍻

PR #5617, opened 2015-04-21
[docs] Documenting more NFS options.

jfbibeau:
I recently went through quite a bit of NFS troubleshooting and found that there are several undocumented NFS options for Vagrant. It's possible to override both client and server-side NFS arguments, which is awesome. 

So this PR aims to document these 2 options in the NFS section.

For reference, PR #1029  contains most commits which added these options.

### DIFF
--- a/website/docs/source/v2/synced-folders/nfs.html.md
+++ b/website/docs/source/v2/synced-folders/nfs.html.md
@@ -65,6 +65,45 @@ the final part of the `config.vm.synced_folder` definition, along with the
 * `nfs_version` (string | integer) - The NFS protocol version to use when
   mounting the folder on the guest. This defaults to 3.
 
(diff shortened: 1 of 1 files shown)


sethvargo:
Thanks!

kvz:
Happy to have found this pr, as http://docs.vagrantup.com/v2/synced-folders/nfs.html wasn't to helpful yet. And could not get https://github.com/gael-ian/vagrant-bindfs to work for fixing permission inconsistencies between host OSX & guest Linux. 

Just 2ct that I think it will be helpful to others if this got published :+1:

sethvargo:
@kvz it will get published with the next release of Vagrant. The docs contained unreleased items, so we cannot deploy yet :smile:

Issue #5377, opened 2015-02-23
Vagrant Mounting Fails: Unable to Connect To Upstart on Box Using Systemd


Geraden07:
I am working with an Ubuntu 14.10 box. Initially it will boot fine, no errors whatsoever.

If I install systemd and change grub to use systemd instead of upstart, then what happens is the box boots normally right up until it hits the first mount (/vagrant). Then it fails due to a non-0 command return and does not attempt any further mounts that may have been configured, nor any provisioning.

However the box is booted just fine and can be SSH'd into with no issues. Even the /vagrant mount shows up from within the VM just fine.

Doing a 'vagrant up --debug' command, I see the following at the end:
```
c:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/batch_action.rb:82:in block (2 levels) in run'
 INFO interface: error: The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

/sbin/initctl emit --no-wait vagrant-mounted MOUNTPOINT=/vagrant

Stdout from the command:



Stderr from the command:

stdin: is not a tty
initctl: Unable to connect to Upstart: Failed to connect to socket /com/ubuntu/upstart: Connection refused

The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

/sbin/initctl emit --no-wait vagrant-mounted MOUNTPOINT=/vagrant

Stdout from the command:



Stderr from the command:

stdin: is not a tty
initctl: Unable to connect to Upstart: Failed to connect to socket /com/ubuntu/upstart: Connection refused

 INFO interface: Machine: error-exit ["Vagrant::Errors::VagrantError", "The following SSH command responded with a non-zero exit status.\nVagrant assumes that t
his means the command failed!\n\n/sbin/initctl emit --no-wait vagrant-mounted MOUNTPOINT=/vagrant\n\nStdout from the command:\n\n\n\nStderr from the command:\n\
nstdin: is not a tty\ninitctl: Unable to connect to Upstart: Failed to connect to socket /com/ubuntu/upstart: Connection refused\n"]
```

Vagrant doesn't seem to know to use systemd and not upstart. I know obviously vagrant CAN use systemd just fine because I've got other centos boxes which use it by default and they work just fine.

skozin:
[...]

PR #5622, opened 2015-04-22
Check Upstart is running before emitting upstart events.

aneeshusa:
Fixes issue #5377.
Check that Upstart is not just installed but currently actively running
before attempting to emit an event after mounting shared folders.
Only requires the binary be named upstart, not that it lives in /sbin.

### DIFF
--- a/plugins/guests/linux/cap/mount_nfs.rb
+++ b/plugins/guests/linux/cap/mount_nfs.rb
@@ -33,7 +33,7 @@ module VagrantPlugins
             end
 
             # Emit an upstart event if we can
-            if machine.communicate.test("test -x /sbin/initctl")
+            if machine.communicate.test("test -x /sbin/initctl && test 'upstart' = $(basename $(sudo readlink /proc/1/exe))")
               machine.communicate.sudo(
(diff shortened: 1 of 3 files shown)


pkruithof:
+1 for this, I cannot start my VM with Ubuntu 15.04 anymore. This patch fixes it for me

@mitchellh could you review/merge this?

chrisbroome:
This also fixes the issue for me after upgrading from Ubuntu 14.10 to 15.04.

Also, would it be better to extract this code into a helper method instead of copy/pasting it? That way if this ever changes again, we could just update it in the one place instead of having to grep through the files to see where to change it.

Layke:
+1 as well.  I have an environment almost running perfectly for devs on my system and this simple bug causes provisioning to not work correctly.  I can provision afterwards and the system still works, but it's just an annoying step to have to tell people to do as a workaround.

How about it @mitchellh?

berendt:
Why do we have to wait for more than 1 month to merge such a simple bugfix?

sethvargo:
Hi @berendt 

We are working as quickly a possible. Thank you for your patience! :smile:

mitchellh:
LGTM. Thanks!

PR #5623, opened 2015-04-22
TinyCore change_host_name capability

otagi:
The capability code was already there, but not declared in the plugin file.

### DIFF
--- a/plugins/guests/tinycore/cap/change_host_name.rb
+++ b/plugins/guests/tinycore/cap/change_host_name.rb
@@ -4,7 +4,6 @@ module VagrantPlugins
       class ChangeHostName
         def self.change_host_name(machine, name)
           if !machine.communicate.test("hostname | grep '^#{name}$'")
-            machine.communicate.sudo("sh -c 'echo \"#{name}\" > /etc/hostname'")
             machine.communicate.sudo("/usr/bin/sethostname #{name}")
           end
         end
--- a/plugins/guests/tinycore/plugin.rb
+++ b/plugins/guests/tinycore/plugin.rb
@@ -16,6 +16,11 @@ module VagrantPlugins
         Cap::ConfigureNetworks
       end
 
+      guest_capability("tinycore", "change_host_name") do
+        require_relative "cap/change_host_name"
+        Cap::ChangeHostName
+      end
+
       guest_capability("tinycore", "halt") do
         require_relative "cap/halt"
         Cap::Halt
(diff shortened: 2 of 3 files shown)


dduportal:
This PR was written during this event : http://www.meetup.com/brug__/events/221754149/?a=uc1_vm&read=1&_af_eid=221754149&_af=event in cooperative mode :)

bmx0r:
this seems to complete #4469 :)

plugins/guests/tinycore/cap/change_host_name.rb
```diff
@@ -4,7 +4,6 @@ module Cap
       class ChangeHostName
         def self.change_host_name(machine, name)
           if !machine.communicate.test("hostname | grep '^#{name}$'")
-            machine.communicate.sudo("sh -c 'echo \"#{name}\" > /etc/hostname'")
```
sethvargo: Why remove this? Was it causing issues?

bmx0r: This command seems useless as the command run after : 
machine.communicate.sudo("/usr/bin/sethostname #{name}")
will purge/fill up the /etc/hostname anyway
Mike ;)

sethvargo: Hrm - does `sethostname` persist after restarts?

bmx0r: I don't think so...
if you start a box with the config.vm.hostname "mybox" in the Vagrant file
you will have a box name of mybox
Then vagrant halt --> pur the config.vm.hostname line in comment  --> vagrant up
The hostname will be boot2docker again...

Still why are you asking? Did we miss something?

sethvargo: @bmx0r on most linux systems, setting the contents of `/etc/hostname` is how you persist the hostname on reboots. I think Vagrant makes the promise to persist that hostname, which is why I'm concerned we are removing that persistence.

dduportal: @sethvargo nope, this is because two things :
* boot2docker is RAMDISK based (exceptions exists but for /var/lib/docker and /var/lib/boot2docker only)
* The bootscript of boot2docker will rewrite this everytime, before ssh daemon has been started :  https://github.com/boot2docker/boot2docker/blob/master/rootfs/rootfs/bootscript.sh#L22 (Warning : point to the b2d master, curent tag is 1.6.0 at the date of this comment) 
which call https://github.com/boot2docker/boot2docker/blob/master/rootfs/rootfs/etc/rc.d/hostname

=> So the scenariis here are simpler than with "persisted" guest like debian or redhats

dduportal:
Ping on this PR ? @sethvargo

PR #5629, opened 2015-04-23
added gpg signature for rvm and bundle gem install to the docs Vagrantfile

brooksbrown:
When attempting to run vagrant up on the Vagrantfile provided in the documentation directory, I get an error related to rvm's gpg signature and the bundle gem not being installed. 

This commit adds a line to retrieve the key and another to install the bundle gem to the Vagrantfile's provisioning script. 

Gist of the error:
https://gist.github.com/brooksbrown/e0d53d995d8dfcc9204d

### DIFF
--- a/website/docs/Vagrantfile
+++ b/website/docs/Vagrantfile
@@ -4,11 +4,13 @@
 $script = <<SCRIPT
 sudo apt-get -y update
 sudo apt-get -y install curl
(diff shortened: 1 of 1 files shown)


berendt:
I would prefer to import the GPG key from https://rvm.io/mpapis.asc. This way we always use the correct key and do not have to change it in the future.

sethvargo:
Thanks!

PR #5634, opened 2015-04-26
Updated ssh docs

jaisubramani:
stdin is supported by vagrant ssh. updated the html docs to reflect the same.

https://github.com/mitchellh/vagrant/issues/5614


berendt:
I do not know the commit messages conventions for this project, but I would prefer to use `Document that stdin is supported by vagrant ssh --command COMMAND` in the title of the commit message.

sethvargo:
Thanks1

PR #5638, opened 2015-04-26
fix master_config documentation

dmangot:
Based on this vagrant code:

```
          @machine.env.ui.info "Copying salt master config to vm."
          @machine.communicate.upload(expanded_path(@config.master_config).to_s, temp_config_dir + "/master")
        end
```

It seems that the current documentation:
 * master_config (string, default: "salt/minion") Path to a custom salt master config file

is probably incorrect and the master_conifg probably defaults to salt/master, not salt/minion.

Unless I'm reading the code wrong.


sethvargo:
Thanks!

PR #5640, opened 2015-04-27
add RDP auto login with password

agdula:
Hey Guys,

Please consider adding this improvement to vagrant code base. It's very annoying to enter password over and over again. 

The fix is implementation of suggestion from @majkinetor for the closed issue
https://github.com/mitchellh/vagrant/issues/4300

Works on windows 7 host.

### DIFF
--- a/plugins/hosts/windows/cap/rdp.rb
+++ b/plugins/hosts/windows/cap/rdp.rb
@@ -8,28 +8,15 @@ module VagrantPlugins
     module Cap
       class RDP
         def self.rdp_client(env, rdp_info)
-          config = nil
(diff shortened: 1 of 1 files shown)


agdula:
It works after you provide the password for the first time... so sadly it's not an ultimate fix

majkinetor:
Mhm... I will check it out, I don't remember it asked for a pass even once. Even if so it is **way** better then having to type it **all** the time.

sadhanandhb:
Kindly help with this solution. I am using Windows 8.1 as a host and need to have automatic login.

PR #5669, opened 2015-05-02
Use powershell syntax for Chef provisioner when using windows guest

taliesins:
Use powershell syntax for Chef provisioner when using windows guest


sethvargo:
Thanks!

PR #5670, opened 2015-05-02
add RDP auto login with password

agdula:
This change cmdkey arguments are ok.
No need to manually enter credentials. Tested with different RDP ports on the same machine.

Please consider adding this improvement to vagrant code base.


sethvargo:
Vagrant 1.8 will use a credentials file (already on master), which achieves this in the same way. Thanks you for the PR though!

mitchellh:
This unfortunately won't :( @sethvargo that is for SMB mounting, not RDP.

Jamesking56:
So... Is this going to be merged or...?

Issue #4153, opened 2014-07-07
Add a favicon to the website


ryanaslett:
I wish I could tell which tabs in my browser were Vagrant documentation tabs. It would be nice if there were a favicon.ico for the website.

PR #5680, opened 2015-05-05
added favicon for www & doc

c33s:
pull request for issue #4153

favicon for the:
- website (original logo)
- docs (logo colored in the purple of the "docs" text)

i wasn't able to test the changes because middleman's dependency on therubyracer. middleman doesn't want to run on my system. i really tried it :)

please let me know if i the files are in the right location and correctly referenced from the layout.erb file


sethvargo:
Hi @c33s 

Thank you for the PR. That favicon looks like it might be purple instead of blue?

c33s:
as i wrote above, to better distinguish between the main vagrant site and the docs site, i added **two** different favicons.
for the main vagrant site i used the original logo but for the docs i created a purple colored version of the logo. it is the same purple as in the letters of the word docs from the docs site logo. 
2015-05-06 10_24_26-logo_docs-36c6817f

www
favicon-64
doc
favicon-64-doc

for example the propel favicon, the orange is for propel1 and the blue one is for propel2

propel-favicons

not a must but very  helpful for heavy tab user like me

sethvargo:
Thanks!

PR #5685, opened 2015-05-06
Document syncing time on VMWare VMs.

deviantintegral:
This was a real source of confusion for me, as I had VMs slowly losing time breaking API request signing.

### DIFF
--- a/website/docs/source/v2/vmware/configuration.html.md
+++ b/website/docs/source/v2/vmware/configuration.html.md
@@ -80,12 +80,15 @@ VMX is an undocumented format and there is no official reference for
 the available keys and values. This customization option is exposed for
 people who have knowledge of exactly what they want.
 
-The most common keys people look for are setting memory and CPUs.
+The most common keys people look for are setting memory, CPUs, and enabling
+time syncronization.
+
 The example below sets both:
 
 ```ruby
 config.vm.provider "vmware_fusion" do |v|
   v.vmx["memsize"] = "1024"
   v.vmx["numvcpus"] = "2"
+  v.vmx["tools.synctime"] = TRUE
 end
 ```


berendt:
Can you please squash the 3 commits into 1 commit to cleanup the PR.

sethvargo:
Hi @berendt 

We do not need to merge/squash commits. We will perform any cleanup on the merge. Thank you for trying to be helpful, but this is not something we require for PRs.

berendt:
I try to review incoming PRs and it is bulky to have one change splitted in several commits.

sethvargo:
Hi @deviantintegral 

Thank you for the PR. Unfortunately I am not very familiar with time synchronization as it relates to VMware. Given we have not received requests for this in the past, I am hesitant to add this to the documentation, especially as a "common setting". (Paging @phinze for feedback on whether this is a common option).

As I am sure you know, it is a difficult to maintain a balance of maintaining the minimal documentation someone needs to get started from duplicating the entire VMware documentation :smile:. What do you think?

phinze:
Hey folks,

VM timekeeping is definitely a source of sadness - one that I've experienced firsthand. Thanks to @deviantintegral for attempting to translate his sadness into better docs for future users!

Unfortunately I don't think we can get away with a simple inclusion of `tools.synctime` in a list of "common options". VM timekeeping is a complicated topic with many strategies for dealing with it, each with its own tradeoffs. It's complicated enough that [VMware publishes a 32-page whitepaper](https://www.vmware.com/files/pdf/Timekeeping-In-VirtualMachines.pdf) on the subject.

So I think @sethvargo's instincts were right here, and that this falls into the space where it's actually better for Vagrant to avoid documenting specifics, because we'll never be able to do justice to the upstream docs.

berendt:
What about adding an additonal note how to enable time sync and why it is maybe useful? We could link to the mentioned whitepaper. This way we do not recommend sth but at least we give users an entry point and we point out that there can be issues with the time sync.

sethvargo:
Thanks @phinze 

I think maintaining this documentation is not sustainable and could easily become a bikeshed (i.e. users arguing over different time syncing configurations). As such, I think it is best to omit this from the documentation, especially as a common option. The good news is that Google has already indexed this Pull Request, so if a user does a search for "vagrant vmware sync time" they will likely see this result and be pushed in the right direction :smile:.

> What about adding an additonal note how to enable time sync and why it is maybe useful?

This, unfortunately, opens a large bikeshed - what dictates a "common" option. Vagrant is almost 5 years old and this is the first request we have had for time sync documentation. Anything we put in the documentation becomes a recommendation (even if we explicitly say it does not), which is why I think it is better to omit something like this.

Thank you @deviantintegral again for the Pull Request, and I am sorry we will not be merging this :frowning:. Have a great day everyone!

deviantintegral:
Where this was really confusing for me is that with Virtualbox boxes, time sync works automatically over the suspend / up cycle. Also, VMs created through the UI also have time sync on by default.

What about putting this in <a href="https://docs.vagrantup.com/v2/vmware/boxes.html">the Boxes page</a> instead? Or even a "beware, timesync is not a default, please read the VMWare docs if you're creating a base box"?

PR #5698, opened 2015-05-09
If plugins are disabled, report them as not available.

### LINKED ISSUE #5430: Vagrant.has_plugin? returns true when VAGRANT_NO_PLUGINS env variable is set
If I have VAGRANT_NO_PLUGINS set and something like this in my Vagrantfile:
```ruby
if Vagrant.has_plugin?("vagrant-cachier")
    config.cache.scope = :box
end
```
I get `Unknown configuration section 'cache'.`
Would it make more sense for Vagrant.has_plugin? to return false in this case, so that the Vagrantfile does not need to be modified?

aneeshusa:
Allows usage of VAGRANT_NO_PLUGINS env var to disable plugins.

Fixes #5430.

### DIFF
--- a/lib/vagrant.rb
+++ b/lib/vagrant.rb
@@ -139,10 +139,12 @@ module Vagrant
     Config.run(version, &block)
   end
 
-  # This checks if a plugin with the given name is installed. This can
-  # be used from the Vagrantfile to easily branch based on plugin
-  # availability.
+  # This checks if a plugin with the given name is available (installed
+  # and enabled). This can be used from the Vagrantfile to easily branch
+  # based on plugin availability.
   def self.has_plugin?(name, version=nil)
+    return false unless Vagrant.plugins_enabled?
+
     if !version
       # We check the plugin names first because those are cheaper to check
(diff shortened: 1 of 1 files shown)


lib/vagrant.rb
```diff
@@ -139,10 +139,14 @@ def self.configure(version, &block)
...
   end
 
-  # This checks if a plugin with the given name is installed. This can
-  # be used from the Vagrantfile to easily branch based on plugin
-  # availability.
+  # This checks if a plugin with the given name is available (installed
+  # and enabled). This can be used from the Vagrantfile to easily branch
+  # based on plugin availability.
   def self.has_plugin?(name, version=nil)
+    if !Vagrant.plugins_enabled?
```
berendt: Should we not use `return false if !vagrant.plugins_enabled?` here?

aneeshusa: Oops, you're right, I was doing python stuff right before this and forgot about one liners.

berendt:
Can you please merge commit c8b286d with b0738ae.

aneeshusa:
Squashed and rebased.

PR #5702, opened 2015-05-11
Add missing word to output of "vagrant up"

leocassarani:
Add the word `flag` to the phrase `or use the '--provision' flag`, which is displayed when `vagrant up` is run for a VM that's already been provisioned.


sethvargo:
Thanks!

PR #5706, opened 2015-05-12
Remove docker0 from guest network interface enumeration

jim-minter:
Using vagrant-libvirt, setting a fixed interface IP address fails if the guest VM runs RHEL7 or Fedora and has a docker0 interface.

The symptom is that **eth0** is brought up successfully for management, Vagrant connects in via ssh, then **eth1** fixed IP configuration is erroneously written to /etc/sysconfig/network-scripts/ifcfg-**eth0**. Then eth0 is bounced (incorrectly) and the "up" process hangs.

The fault occurs because 'docker0' is not removed from this network interface enumeration, causing a subsequent off-by-one error. The fault was probably introduced (inadvertently) in commit 55a90445cdfc290d4a1f68c66835c0249ca59cdd.

### DIFF
--- a/plugins/guests/fedora/cap/configure_networks.rb
+++ b/plugins/guests/fedora/cap/configure_networks.rb
@@ -21,7 +21,7 @@ module VagrantPlugins
           end
 
           if virtual
-            machine.communicate.sudo("ls /sys/class/net | grep -v lo") do |_, result|
+            machine.communicate.sudo("ls /sys/class/net | egrep -v lo\\|docker") do |_, result|
               interface_names = result.split("\n")
             end
 


purpleidea:
If this analysis is correct, then I'm particularly keen on getting this merged ASAP.
It does look very plausible, and the patch (if anything) is harmless.

+1

purpleidea:
Just tested this patch, and I confirmed that it fixes the issue, and furthermore that the pattern of working/not working depended on a machine having a docker0 interface! So this is great, thanks!

I recommend merging ASAP :)

obnoxxx:
I don't think this patch is correct.
It certainly fixes the problem for the interfaces named docker*, i.e. it is a hack for a very special case.
But expect that the actual fix to go upstream should be more general.

purpleidea:
@obnoxxx Well I think it's a good workaround/fix for now, but certainly more general patches would I'm sure be appreciated. The immediate issue is fixing the regression a number of users are experiencing.

obnoxxx:
@purpleidea Sure it is a good workaround. But as a developer, I doubt that it will or should be put into the upstream code, because the filter for the docker* interface name is so .. arbitrary. We can certainly put it into the fedora packages as a temporary workaround. I am currently looking into the code to understand what is happening and if I see the bigger picture.

purpleidea:
@obnoxxx Awesome! Let us know if you find a better fix :)
And I'm sure we'd love to have it in Fedora ASAP as a workaround.

pschiffe:
Hello, are there any updates on this patch?

strzibny:
Just as a note, this fix is now included in official Fedora builds.

dcbw:
I worked around this problem in a different way, see:

https://github.com/pradels/vagrant-libvirt/commit/ab34b7d29b8a094704ed398ff384aa16c0e167e4
https://github.com/mitchellh/vagrant/pull/5884

which is a general solution, not just for docker0.

mitchellh:
Closing for #5884

purpleidea:
Actually @dcbw in your patch notes it says:

"Otherwise use slot numbers as before."

So in that case, this is still a bug, and this patch will help that use case. Correct?
Or did I misunderstand?

If that's the case we need *both* of these patches. And this one doesn't hurt.

dcbw:
@purpleidea Right, this one doesn't hurt.  But it's obviously only a partial fix, since the interfaces can be named anything.  Also, my patch in #5884 only currently works with git master vagrant-libvirt or manually specified MAC addresses in the Vagrantfile, because only vagrant-libvirt currently has logic to pass the MAC address up to vagrant itself.  Other providers probably don't do this, and they will fall back to the existing slot-based logic.

purpleidea:
On Mon, Jul 6, 2015 at 5:31 PM, Dan Williams <notifications@github.com>
wrote:

> @purpleidea <https://github.com/purpleidea> Right, this one doesn't hurt.
> But it's obviously only a partial fix, since the interfaces can be named
> anything. Also, my patch in #5884
> <https://github.com/mitchellh/vagrant/pull/5884> only currently works
> with git master vagrant-libvirt or manually specified MAC addresses in the
> Vagrantfile, because only vagrant-libvirt currently has logic to pass the
> MAC address up to vagrant itself. Other providers probably don't do this,
> and they will fall back to the existing slot-based logic.



I completely agree. So

@mitchellh we need to reopen and merge this patch also.

Thanks.

mitchellh:
Reopening to take a look!

mitchellh:
I agree this should be done more robustly but I don't see a better way for now.

purpleidea:
w00t.

Cheers,
Thanks!

On Thu, Jul 9, 2015 at 4:31 PM, Mitchell Hashimoto <notifications@github.com

dcbw:
@mitchellh I think #5884 is the better, more robust solution; have the providers (like libvirt, or virtualbox, etc) pass up the MAC address of the network interface they'd like to configure (since they already asked the VM to create/add it), so that the interfaces can be exactly matched in the Vagrant guest code.

PR #5707, opened 2015-05-12
fedora: don't try to use biosdevname if it's not installed

dcbw:
If biosdevname isn't installed it doesn't make sense to try using it
for persistent device names.  Just treat lack of biosdevname as
virtual networking.

### DIFF
--- a/plugins/guests/fedora/cap/configure_networks.rb
+++ b/plugins/guests/fedora/cap/configure_networks.rb
@@ -17,7 +17,7 @@ module VagrantPlugins
           virtual = false
           interface_names = Array.new
           machine.communicate.sudo("/usr/sbin/biosdevname; echo $?") do |_, result|
-            virtual = true if result.chomp == '4'
+            virtual = true if ['4', '127'].include? result.chomp
           end
 
           if virtual


plugins/guests/fedora/cap/configure_networks.rb
```diff
@@ -17,7 +17,7 @@ def self.configure_networks(machine, networks)
           virtual = false
           interface_names = Array.new
           machine.communicate.sudo("/usr/sbin/biosdevname; echo $?") do |_, result|
-            virtual = true if result.chomp == '4'
+            virtual = true if result.chomp == '4' || result.chomp == '127'
```
berendt: Should we not use `['4', '127'].include? result.chomp` here?

dcbw:
I'm kinda new to Ruby, so yeah that makes sense.  I'll redo the commit.

strzibny:
+1. Please update your commit and let's get this in:).

dcbw:
Rebased and updated.

berendt:
lgtm

strzibny:
I, for one, already included this fix in Vagrant builds in Fedora. This is really annoying. Please merge.

plugins/guests/fedora/cap/configure_networks.rb
```diff
@@ -17,7 +17,7 @@ def self.configure_networks(machine, networks)
           virtual = false
           interface_names = Array.new
           machine.communicate.sudo("/usr/sbin/biosdevname; echo $?") do |_, result|
-            virtual = true if result.chomp == '4'
+            virtual = true if ['4', '127'].include? result.chomp
```
sethvargo: It would be great to have a comment here explaining what exit code 4 and 127 correspond to, but not a requirement.

Issue #5590, opened 2015-04-13
private_network race condition with NetworkManager on Centos 7.1 with Vagrant 1.7.2


bcc:
Vagrant 1.7.2
CentOS 7.1
Virtualbox 4.3.22

It seems that NetworkManager on CentOS 7.1 is more aggressive about handling unconfigured interfaces on start than it was in with CentOS 7, and the way that Vagrant handles the configuration of the interfaces doesn't seem to be compatible with this.

I've updated a base box from Centos 7 to 7.1 (both built with the same packer configuration), and the 7.1 box fails to retain the private ip address. Looking at debug output (and adding some additional debug commands), the configuration does get set correctly during provisioning, but is undone shortly after the machine boots. It seems that this is due to NetworkManager waiting for DHCP to fail (as there's no dhcp configured on that private network), then unconfiguring the interface. 

As the /etc/sysconfig/network-scripts/ifcfg-enp0s8 file (in my case) which is generated (and indeed created) by Vagrant, has NM_CONTROLLED=no, NM doesn't handle the ifdown from configure_networks.rb, and the ifup happens while NM is still waiting for DHCP to finish.

```
[root@puppet vagrant]# nmcli dev
DEVICE  TYPE      STATE                                  CONNECTION         
enp0s3  ethernet  connected                              enp0s3             
enp0s8  ethernet  connecting (getting IP configuration)  Wired connection 1 
lo      loopback  unmanaged                              --                 
[root@puppet vagrant]# ip addr sh enp0s8
3: enp0s8: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 08:00:27:6f:44:44 brd ff:ff:ff:ff:ff:ff
    inet6 fe80::a00:27ff:fe6f:4444/64 scope link 
       valid_lft forever preferred_lft forever
```

As a workaround, I've found that enabling DHCP makes the problem go away (you end up with both IPs on the interface), or restarting NetworkManager so it gives up on the private interface also works, but it's a bit of a bodge. 

Similarly, forcing NetworkManager to give up on the interface in configure_networks.rb by inserting:
 
    machine.communicate.sudo("nmcli dev disconnect #{interface}")

Between the ifup and ifdown commands also seems to resolve the issue, although not especially cleanly.

```
[vagrant@puppet ~]$  nmcli dev
DEVICE  TYPE      STATE      CONNECTION 
[...]

PR #5709, opened 2015-05-13
Reload iface connection by NetworkManager

linyows:
Fix for #5590. This changes has created `/etc/sysconfig/network-script/ifcfg-# {interface}` to enable.

I used https://atlas.hashicorp.com/chef/boxes/centos-7.1

### DIFF
--- a/plugins/guests/fedora/cap/configure_networks.rb
+++ b/plugins/guests/fedora/cap/configure_networks.rb
@@ -85,6 +85,7 @@ module VagrantPlugins
           interfaces.each do |interface|
             retryable(on: Vagrant::Errors::VagrantError, tries: 3, sleep: 2) do
               machine.communicate.sudo("cat /tmp/vagrant-network-entry_#{interface} >> #{network_scripts_dir}/ifcfg-#{interface}")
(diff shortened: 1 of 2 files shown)


dcbw:
NetworkManager dev here... you shouldn't ever need to restart NM just to get network configuration to apply.  If you know the interface name you can instead add a NAME="whatever" to the ifcfg file, then:

nmcli con reload
nmcli dev disconnect <iface>
nmcli con up <name from ifcfg file>

dcbw:
Also note that the 'dev disconnect' isn't strictly necessary, since NM will automatically clean up the previous connection and start the new one when you call 'nmcli con up'.  It's important to do the 'nmcli con reload' though because (by popular request) NM 0.9.10 and later don't automatically watch /etc/sysconfig/network-scripts/ for changes.

linyows:
@dcbw I see. Changed reload instead of restart.

berendt:
I would prefer to change the commit message to `Restart NetworkManager on Red Hat based distributions` (at least `Redhad` should be changed to `Red Hat`) and to merge to second commit 421fab1 with the first commit c1c39be.

udzura:
:+1: I want this patch!

DQNEO:
:+1: :+1: :+1: 
I have the same problem and hope this fix will be shipped soon.

PR #5719, opened 2015-05-15
Ansible config.generated_inventory_path option

iNecas:
Allow to customize the path where the inventory gets generated and generate it
even when other inventory_path is defined, since it can be useful later (for
using ansible outside of 'vagrant provision')

When specified, this path gets passed as the inventory file (unless other
inventory file specified)

In my use-case, I define multiple vms in my Vagrantfile, and using different
combination of the vms for different playbooks. I want to keep the generated inventories
separated. Also, I want to be able to use them to call ansible-plabook
directly, for example for playbooks that are not defined in the Vagrantfile.


sethvargo:
Hey @iNecas 

Thank you for the PR. This seems like a fair amount of added complexity; could you explain the use case a bit more? If I understand correctly, you want to use different playbooks per VM? If that's the case, I think we should append the machine_name to the playbook path instead of making this a user-configurable option. What do you think?

iNecas:
Thanks @sethvargo for the feedback, here are some cases:

1. I have a set of VMs, each one representing my project running on a different OS version (rhel6, rhel7, wheezy, jessie), I provision all of these vms. I set every generated inventory to be on a separate path (inventory/rhel6, inventory/rhel7, inventory/wheezy, inventory/jessie). Given I have the inventory files separated, I can a) use the inventory directory as "the whole" inventory, that doesn't get overridden based on what's currently active. More importantly, I could then use the inventory files when calling ansible-playbook directly, for every os. That would especially be useful for the inventories with more hosts

2. I run multiple playbooks on a single VM: currently, the inventory gets always overridden by the latest call.

3. for using the generated inventory files directly by ansible, I would prefer having the inventory directory on accessible place

How would he machine_name appending work? I like this option because it gives me almost full power over the inventory file while I can manage the groups from the vagrant file rather than maintaining the static inventories and the mapping on manual ip setting basis.

gildegoma:
@iNecas Thanks for your interest and your contribution offer! So far, I would really appreciate that we dig a little deeper in your needs, because I feel that this change is not necessary:

> use the inventory files when calling ansible-playbook directly

If I understand well, you want to "migrate" from a set of "static inventory files" to the generated inventory. Since the generated inventory file contains **all** the (valid) machines defined in your `Vagrantfile` you can simply use this inventory and select the target machine(s) with an ansible-playbook `--limit` option (which is actually what the Vagrant Ansible provisioner does).

> currently, the inventory gets always overridden by the latest call

I don't understand this point. The generated inventory should always look the same after each `vagrant provision` (except if you remove or stop some machines).

> I would prefer having the inventory directory on accessible place

What do you mean by "accessible place"? Note, you can always write a script that copies (or symlinks) the generated inventory to the location you want.

If my remarks does not make sense for your use case, it would be very practical that you provide us a gist of your `Vagrantfile`, so we can better figure out what you'd like to achieve, and help you as much as possible!

gildegoma:
@iNecas you also might be interested in the [dynamic inventory script for Vagrant](https://github.com/ansible/ansible/blob/devel/plugins/inventory/vagrant.py#L6-L12).

iNecas:
On 18 May 2015 at 16:02, Gilles Cornu <notifications@github.com> wrote:

> @iNecas <https://github.com/iNecas> Thanks for your interest and your
> contribution offer! So far, I would really appreciate that we dig a little
> deeper in your needs, because I feel that this change is not necessary:
>
> use the inventory files when calling ansible-playbook directly
>
> If I understand well, you want to "migrate" from a set of "static
> inventory files" to the generated inventory. Since the generated inventory
> file contains *all* the (valid) machines defined in your Vagrantfile you
> can simply use this inventory and select the target machine(s) with an
> ansible-playbook --limit option (which is actually what the Vagrant
> Ansible provisioner does).
>
> currently, the inventory gets always overridden by the latest call
>
> I don't understand this point. The generated inventory should always look
> the same after each vagrant provision (except if you remove or stop some
> machines).
>
> I would prefer having the inventory directory on accessible place
>
> What do you mean by "accessible place"? Note, you can always write a
> script that copies (or symlinks) the generated inventory to the location
> you want.
>
> If my remarks does not make sense for your use case, it would be very
> practical that you provide us a gist of your Vagrantfile, so we can
> better figure out what you'd like to achieve, and help you as much as
> possible!
>

Here is a gist of a vagrantfile I would like to use
https://gist.github.com/iNecas/4324536a0b1fd81d3c66

I can imagine using external dynamic inventory script for that: however, I
would need to be able to pass additional environment variables when running
the ansible-playbook (so that I use that in the inventory script to do the
mapping). The simple dynamic script shipped by ansible would not work for
me, as I would like to connect that to the Vagrantfile to get more dynamic
results (env variable is the way how to pass the args to inventory script).
Would that PR be more acceptable?

gildegoma:
> Here is a gist of a vagrantfile I would like to use
https://gist.github.com/iNecas/4324536a0b1fd81d3c66

Thanks for these details! I'd propose to use something like this in your `Vagrantfile`:

```ruby

  config.ssh.insert_key = false

  SERVICES  = %w(db ha app)
  PLATFORMS = %w(centos-7.1 fedora-21)

  SERVICES.each do |service|
    PLATFORMS.each do |os|
      config.vm.define "#{service}-#{os}" do |machine|
        machine.vm.box = "chef/#{os}"
      end
    end
  end

  config.vm.provision "ansible" do |ansible|
    ansible.playbook = "playbook.yml"
    ansible.groups = {}

    # Create groups by Services, this is necessary
    # to comply with the play `hosts` scopes defined in `playbook.yml`
    SERVICES.each do |service|
      ansible.groups[service] = PLATFORMS.map { |os| "#{service}-#{os}" }
    end
  end

end
```

with the three plays defined in the same playbook file:

```yaml
- name: DB play
  hosts: db
  roles:
    - ...

- name: HA play
  hosts: ha
  roles:
    - ...

- name: APP play
  hosts: app
  roles:
    - ...
```

* You get a single generated inventory, which is complete and up to date.
* You easily can select groups of machines that you want to target (e.g. `vagrant up /-centos/` or `vagrant provision /app-/`)
* An external execution of `ansible-playbook` is also quite handy and interesting. For example you can provision all the "fedora-21" machines in parallel with a `--limit='*fedora-21'`:

```
PLAY [DB playbook] ************************************************************ 

TASK: [db action] ************************************************************* 
changed: [db-fedora-21]

PLAY [HA playbook] ************************************************************ 

TASK: [ha action] ************************************************************* 
changed: [ha-fedora-21]

PLAY [APP playbook] *********************************************************** 

TASK: [app action] ************************************************************ 
changed: [app-fedora-21]

PLAY RECAP ******************************************************************** 
app-fedora-21              : ok=1    changed=1    unreachable=0    failed=0   
db-fedora-21               : ok=1    changed=1    unreachable=0    failed=0   
ha-fedora-21               : ok=1    changed=1    unreachable=0    failed=0   
```

@iNecas would this approach satisfy your needs?

iNecas:
Thanks for the ideas suggestions and your time. It seems like this approach and it's modifications might actually work for me.

Closing this PR: I'm ready to rebase and reopen if somebody else find another use-case that can't be solved by your suggestion.

gildegoma:
@iNecas You're welcome and I'm glad if it works for you. Thanks again for this constructive discussion!

PR #5731, opened 2015-05-19
Fix RHEL name and description

strzibny:
Red Hat Linux no longer exists, so let's make it RHEL.


mitchellh:
Sure, sounds fine to me.

PR #5736, opened 2015-05-19
Update dependency to allow most recent bundler version

Ryanbarnes420:
Making changes to vagrant.gemspec to allow successful execution of 'bundler install' with the latest version of bundler (version 1.9.9)

Changed from "< 1.8.0" to "<= 1.9.9"

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
   s.rubyforge_project         = "vagrant"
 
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @Ryanbarnes420 

Thank you for the Pull Request. Unfortunately we will not be merging this at this time. You can read more [in my comment on the issue](https://github.com/mitchellh/vagrant/issues/5581#issuecomment-91889166), but we need to do extensive testing when we update Bundler because of our dependency on some internal APIs. As a result, we cannot just update Bundler without some rigorous testing first.

We will update to the latest version of Bundler with the next release of Vagrant :smile:

PR #5759, opened 2015-05-27
Allow vagrant ssh-config with a halted VM

peckpeck:
Allow vagrant ssh-config on a halted VM on virtualbox : https://github.com/github/hub/issues/910

If the VM is not running, the vagrant ssh command already protects the user from trying to connect, so this won't lead to an unexpected error.

### DIFF
--- a/plugins/providers/virtualbox/provider.rb
+++ b/plugins/providers/virtualbox/provider.rb
@@ -56,9 +56,6 @@ module VagrantPlugins
 
       # Returns the SSH info for accessing the VirtualBox VM.
       def ssh_info
-        # If the VM is not running that we can't possibly SSH into it
-        return nil if state.id != :running
-
         # Return what we know. The host is always "127.0.0.1" because
         # VirtualBox VMs are always local. The port we try to discover
         # by reading the forwarded ports.


plugins/providers/virtualbox/provider.rb
```diff
@@ -56,9 +56,6 @@ def machine_id_changed
 
       # Returns the SSH info for accessing the VirtualBox VM.
       def ssh_info
-        # If the VM is not running that we can't possibly SSH into it
```
sethvargo: I think we will need to check if the machine has been booted here. If the machine was never booted, this would return incorrect information.

peckpeck: Sure, how do we do that ?

sethvargo: You would need to inspect the VMs status and ensure it's running, paused, or halted I think.

mitchellh: We can probably just check if the VM ID is not blank... thats probably safe enough.

sethvargo:
This was fixed at some point (the offending code is no longer there)

PR #5771, opened 2015-06-01
Add UI::Interface#color?

sethvargo:
This adds a `.color?` function to the UI classes that returns false by default and then true for the colored ui. You can see from the output that this is more explicit than checking the type of UI class.

/cc @phinze for UX thoughts :smile:

### DIFF
--- a/lib/vagrant/ui.rb
+++ b/lib/vagrant/ui.rb
@@ -63,6 +63,11 @@ module Vagrant
         define_method(method) { |*args| }
       end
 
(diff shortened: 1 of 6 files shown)


gildegoma:
:+1: (sorry @sethvargo, you'll need to rebase because of f7894d9).

sethvargo:
Rebased @gildegoma

phinze:
:+1: LGTM :gem:

PR #5773, opened 2015-06-01
Being able to edit a file does not mean you can write in its parent directory

willdurand:
This PR replaces #5030.

---

Interactive `sed` needs write permission on the file itself. However, it
may create a backup file, which leads to the fact that the directory
where the file is located must be writable as well. That is a side
effect because this directory does not need to be writable.

This patch fixes this side effect by editing the file in `/tmp`, and
replaces it right after.

### DIFF
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -102,7 +102,7 @@ module VagrantPlugins
 
           # Use sed to just strip out the block of code which was inserted
           # by Vagrant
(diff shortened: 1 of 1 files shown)


sethvargo:
LGTM!

mitchellh:
LGTM

willdurand:
thanks :)

kkopachev:
Seems like Ubuntu 15.04 and 14.04 both does not have `$TMPDIR` variable set. This results in inability to clear old exports and cp is throwing warnings:

> cp: missing destination file operand after ‘/etc/exports’
Try 'cp --help' for more information.
cp: cannot stat ‘/exports’: No such file or directory

sethvargo:
@kkopachev https://github.com/mitchellh/vagrant/pull/5954

PR #5785, opened 2015-06-02
Ignore hidden directories

sethvargo:
Fixes https://github.com/mitchellh/vagrant/pull/4758

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
   s.rubyforge_project         = "vagrant"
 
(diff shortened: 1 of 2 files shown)


vagrant.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
   s.rubyforge_project         = "vagrant"
 
-  s.add_dependency "bundler", ">= 1.5.2", "< 1.8.0"
+  s.add_dependency "bundler", ">= 1.5.2"#, "< 1.8.0"
```
mbrodala: Is the comment intented? E.g. will the constraint be re-added soon or will it be gone for good?

sethvargo: No, thank you for catching it. It will updated soon, but not yet :smile:

PR #5786, opened 2015-06-03
aufs is optional, if docker can't get it press on

displague:
Per https://github.com/mitchellh/vagrant/issues/5697#issuecomment-100581277, addresses https://github.com/mitchellh/vagrant/issues/5697 by not requiring success from `apt-get install linux-image-extra-*`.

This issue was also brought up on the Linode provider https://github.com/displague/vagrant-linode/issues/27


mitchellh:
Fixed by another PR

PR #5798, opened 2015-06-05
prefer ipv4 host address over ipv6 for smb synced folders

### LINKED ISSUE #5797: smb synced folder may break on windows host/windows guest
It had been a while since using synced folders on a windows host to a windows guest and when I `vagrant up`'d I got the following error:
```

cmdkey /add:fd00::c569:792d:e7cb:4e26 /user:matt /pass:<pass>

Stdout from the command:


Stderr from the command:

```

The guest is attempting to authenticate to the host with the IPv6 IP. Even though the host does appear to be reachable on this IP, the `cmdkey` command does not like it but is happy with ipv4. It appears this began with pr #5349. The former WMI method returned IPs sorted with ipv4 on top while, at least on my host, `Get-NetIPAddress` sorts with ipv6 on top.

I will submit a fix.

mwrock:
fixes #5797 This sorts the ips by `AddressFamily` returning ipv4 first.

### DIFF
--- a/plugins/synced_folders/smb/scripts/host_info.ps1
+++ b/plugins/synced_folders/smb/scripts/host_info.ps1
@@ -1,6 +1,9 @@
 $ErrorAction = "Stop"
 
-$net = Get-NetIPAddress | Where-Object {($_.IPAddress -ne "127.0.0.1") -and ($_.IPAddress -ne "::1")}
+$net = Get-NetIPAddress | Where-Object {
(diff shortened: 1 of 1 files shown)


sethvargo:
This LGTM, but I'm not a Windows expert. Can someone else confirm this fix who works on Windows before we merge? :smile:

mitchellh:
LGTM. I trust @mwrock on Windows.

PR #5818, opened 2015-06-10
Fix issue 5790

sneal:
Leaving around plaintext username and passwords in a script on a box isn't the best from a security standpoint. This change ensures the scheduled task wrapper script for WinRM doesn't leave these around on the box, and instead passes them to the script as arguments.


mitchellh:
LGTM. Thanks!

PR #5847, opened 2015-06-19
Add Atomic guest support

strzibny:
This adds a new guest to Vagrant: Atomic.

More info on Atomic: http://www.projectatomic.io/
Rationale on why this is a guest and not flavour: https://github.com/projectatomic/vagrant-atomic/issues/5

Note that this was already released as a vagrant-atomic plugin[0], but we would like to have this in Vagrant itself.

[0] https://github.com/projectatomic/vagrant-atomic/


mitchellh:
LGTM! Great job!

PR #5859, opened 2015-06-24
Start required NFS services on Fedora guests

strzibny:
@sethvargo This pull https://github.com/mitchellh/vagrant/pull/5770 does not contain the call to start the necessary services for NFS, so it still does not work out of the box. This fixed the issue for me.


strzibny:
@sethvargo  The thing is that this does not solve an issue when NFS client is installed (`nfs_client_installed` returns `true`), because in that case the services might need restart as well. I can either add `nfs_client_installed` caps for Fedora and RHEL or we need to solve it generally in `nfs_client_installed` Linux cap. Thoughts?

strzibny:
This should be closed in favor of https://github.com/mitchellh/vagrant/pull/5880.

PR #5877, opened 2015-06-29
Install Vagrant plugins in isolation

strzibny:
Vagrant has the ability to install upstream plugins from RubyGems.org, but unfortunately the plugins are not installed/managed in isolation from Vagrant itself or other plugins. This means that installing a plugin without any dependencies such as `vagrant-cachier` results in installing a whole bunch of gems. I don't feel that this is the right way to handle it.

 I would like Vagrant to install new plugins in isolation, because it will help me to keep the state that works for me (with other plugins I haven't lock the version of) and save bandwidth. Overall I believe that by isolating the plugin installation from the rest of the dependencies users will get better and predictable user experience (same as with probably any other software with plugins). I would also like to add that this is sort of a necessity for Fedora as we package some Vagrant plugins, but if the user needs others from RubyGems.org, his plugins would be upgraded as well (this is problematic when some plugins depend on C extensions).

Therefore I would suggest to change this and isolate the plugin installation by locking all other dependencies. I am also including a patch on how this could be done by actually saving the user's `Gemfile` and associated `Gemfile.lock` in the user's Vagrant directory (e.g. `~/.vagrant.d/`). When the `Gemfile.lock` exists its used to determine what to lock before installing/updating gems.

### DIFF
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -69,6 +69,12 @@ end
 require "bundler"
 begin
   Bundler.setup(:default, :plugins)
+# Invalidate the cached Gemfile.lock if necessary and try again
+rescue Bundler::GemNotFound
+  FileUtils.rm File.expand_path("~/.vagrant.d/Gemfile") if File.exists? File.expand_path("~/.vagrant.d/Gemfile")
+  FileUtils.rm File.expand_path("~/.vagrant.d/Gemfile.lock") if File.exists? File.expand_path("~/.vagrant.d/Gemfile.lock")
+
+  Bundler.setup(:default, :plugins)
 rescue Bundler::GemNotFound
   $stderr.puts "Bundler, the underlying system used to manage Vagrant plugins,"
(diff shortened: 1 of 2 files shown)


strzibny:
@mitchellh This is something we really, really need  for Fedora, CentOS and RHEL builds of Vagrant, because current handling of plugins is truly problematic. I believe that installing and updating plugins in isolation will bring more predictable results to all Vagrant users. And they will be faster. Can you take a look?

sethvargo:
Hi @strzibny 

We have some plans for Vagrant 1.8 to revamp the way plugins are isolated. We will take a look at this approach at that time :smile:

chrisroberts:
Hi there. The implementation of plugin manager has changed within Vagrant invalidating this PR. However, I will be working on this feature soon for the new implementation. Thank you for the work you did on this PR and sorry that it was not able to get merged.

strzibny:
@sethvargo I understand, thanks. At least the work wasn't lost as it was used in Fedora for a long time;).

PR #5880, opened 2015-06-30
Fix: Rewrite Red Hat NFS handling based on systemd

strzibny:
This effectively fixes NFS on Fedora and RHEL guests whether the packages are already installed or not. Uses systemd to determine what commands to run and because of that we can retire the Fedora NFS cap.

@sethvargo  It also retires mine previous fix https://github.com/mitchellh/vagrant/pull/5859 that was already reviewed!


mitchellh:
LGTM. Thanks

PR #5884, opened 2015-06-30
[fedora] honor MAC address when configuring networks

dcbw:
Configuring by :interface doesn't work very well because Vagrant has
no idea about what interfaces are present in the VM, for example if
the image has 'docker' installed but not biosdevname, then
interface_names[0] = "docker0" which is usually not what you want
mapped to the first network from the Vagrantfile.

So if the plugins (like vagrant-libvirt) or the Vagrantfile has
given us a network with a MAC address, use that to find the interface
name for the network.  Otherwise use slot numbers as before.

### DIFF
--- a/plugins/guests/fedora/cap/configure_networks.rb
+++ b/plugins/guests/fedora/cap/configure_networks.rb
@@ -16,6 +16,7 @@ module VagrantPlugins
 
           virtual = false
           interface_names = Array.new
+          interface_names_by_slot = Array.new
           machine.communicate.sudo("/usr/sbin/biosdevname; echo $?") do |_, result|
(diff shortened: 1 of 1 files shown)


dcbw:
This PR is a more general solution to https://github.com/mitchellh/vagrant/pull/5706 too.

dcbw:
See the vagrant-libvirt https://github.com/pradels/vagrant-libvirt/commit/ab34b7d29b8a094704ed398ff384aa16c0e167e4 for an example of how a provider plugin might always pass the MAC up to vagrant.

mitchellh:
LGTM

PR #5887, opened 2015-07-02
Add setting config.ssh.ssh_command

msabramo:
Lets the user specify what ssh to use or even direct Vagrant to use an ssh wrapper like sshrc (https://github.com/Russell91/sshrc).

I haven't added tests or documentation yet, but I can add it if you're interested in the idea.

In my `Vagrantfile`, I set:

```ruby
config.ssh.ssh_command = 'sshrc'
```

Now I can do `vagrant ssh` and have it invoke `sshrc` instead of `ssh`.

```
$ bundle exec vagrant ssh
...
-------> marca-mac2.local:~/.sshrc
vagrant@localmonkey:~$ exit
```


mitchellh:
LGTM

PR #5895, opened 2015-07-07
Track UID used to create machine, error in VB if mismatch

### LINKED ISSUE #4739: Running vagrant up as administrator removes existing info in .vagrant/
Trying to use the SMB shares (on an existing upped machine), it says run `vagrant up` with admin privileges. So I open a new terminal and attempt to `vagrant up` there.

This seems to destroy all the info in .vagrant/ and orphan off the existing VM. Vagrant up in either terminal will then attempt to import an image. Even though it had already been vagrant-upped.

Steps
```
vagrant up
vagrant down
-> go to admin terminal
vagrant up
-- "fetching box etc"
```
First box is now 'orphaned'. Vagrant should at least ask if its going to delete this information.

mitchellh:
Fixes #4739

This adds UID tracking to created machines, and errors in VirtualBox if they mismatch.

### DIFF
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -279,6 +279,13 @@ module Vagrant
           end
         end
 
+        if uid_file
+          # Write the user id that created this machine
+          uid_file.open("w+") do |f|
+            f.write(Process.uid.to_s)
+          end
+        end
+
         # If we don't have a UUID, then create one
         if index_uuid.nil?
           # Create the index entry and save it
@@ -311,6 +318,7 @@ module Vagrant
       else
         # Delete the file, since the machine is now destroyed
         id_file.delete if id_file && id_file.file?
+        uid_file.delete if uid_file && uid_file.file?
 
         # If we have a UUID associated with the index, remove it
         uuid = index_uuid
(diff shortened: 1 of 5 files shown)


lib/vagrant/machine.rb
```diff
@@ -279,6 +279,13 @@ def id=(value)
           end
         end
 
+        if uid_file
+          # Write the user id that created this machine
+          uid_file.open("w+") do |f|
+            f.write(Process.uid.to_s)
```
sethvargo: Does `Process.uid` return the right thing on Windoze?

templates/locales/en.yml
```diff
@@ -1317,6 +1317,14 @@ en:
         Vagrant uses the `VBoxManage` binary that ships with VirtualBox, and requires
         this to be available on the PATH. If VirtualBox is installed, please find the
         `VBoxManage` binary and add it to the PATH environmental variable.
+      virtualbox_user_mismatch: |-
+        The VirtualBox VM was created with a user that doesn't match the
+        current user running Vagrant. VirtualBox requires that the same user
+        be used to manage the VM that was created. Please re-run Vagrant with
+        that user. This is not a Vagrant issue.
```
sethvargo: I think we should also add "This is not a Vagrant issue" after the "Your UID" as well.

mitchellh: Too aggressive, perhaps!

PR #5897, opened 2015-07-07
Stop salt minion service and set to manual

PatOShea:
* Stops windows service for local testing if `run_service` set to false. It's true by default
* Added passing in minion_id and master_id to windows installer
* Documentation updates

### DIFF
--- a/plugins/provisioners/salt/bootstrap-salt.ps1
+++ b/plugins/provisioners/salt/bootstrap-salt.ps1
@@ -1,11 +1,33 @@
 Param(
-    [string]$version
+  [string]$version,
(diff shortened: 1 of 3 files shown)


PatOShea:
I've contributed a powershell script to the salt-bootstrapp project recently. I'll do some testing but I think the simplest solution is to close this PR and update the salt provisioner to point to that. Here's the script: https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.ps1

PatOShea:
The changes in this are non-sense. I'll open another request that's more straight forward.

PR #5905, opened 2015-07-09
Docker Provider: Don't destroy built Docker image on reload

### LINKED ISSUE #5353: Leverage docker build cache - at least for vagrant reload
Docker has a pretty good way of caching build layers, but I haven't been able to find a way to make it work with vagrant's docker provider.

My use case is something like:

1. Have vagrant file using docker provider which builds container from a dockerfile
2. Edit docker file
3. Rebuilt container to check, e.g. using vagrant reload
4. Notice this always rebuilds container from scratch instead of utilizing docker's build cache 

This totally ignores the docker build cache (I think vagrant reload destroys the docker image and thus the cache), making this workflow very slow compared to just rebuilding with docker itself.

adrianguenter:
Resolves #5353

After this change Docker's caching mechanism is now applicable to the build process during `vagrant reload` by default, speeding up the reload process considerably for large/complex Dockerfiles.

Forcing the image to be rebuilt from scratch, without cache, can still be achieved via adding the `--no-cache` flag to the `build_args` array.

### DIFF
--- a/plugins/providers/docker/action.rb
+++ b/plugins/providers/docker/action.rb
@@ -117,8 +117,12 @@ module VagrantPlugins
 
             b2.use Call, IsBuild do |env2, b3|
               if env2[:result]
-                b3.use EnvSet, force_confirm_destroy: true
-                b3.use action_destroy.flatten
+                b3.use ConfigValidate
+                b3.use EnvSet, force_halt: true
+                b3.use action_halt
+                b3.use HostMachineSyncFoldersDisable
+                b3.use Destroy
+                b3.use ProvisionerCleanup
               end
             end
 


sethvargo:
Hi @adrianguenter 

It looks like some of the tests are failing. Could you take a look at that please? Thanks!

mitchellh:
Those failures are my fault. I'll take a look at the.

mitchellh:
This looks pretty good to me. I'll test it locally then merge.

adrianguenter:
@sethvargo master was already failing, I swear!!

@mitchellh  thanks 😀

adrianguenter:
Rebased, green now. :)

mitchellh:
Merging, the only thing is I don't think we need the config validation since we do that elsewhere

adrianguenter:
Oops, totally overlooked the one above. Thanks.

PR #5912, opened 2015-07-09
Automatically restore the original environment if we are runnings something outside of the embedded installer

sethvargo:


mitchellh:
LGTM if this works.

mitchellh:
(tests failures were mine, fixed on master now)

mitchellh:
LGTM again

PR #5913, opened 2015-07-09
Use windows-specific Chef paths

sethvargo:
Fixes https://github.com/mitchellh/vagrant/issues/5862


mitchellh:
LGTM

NickMRamirez:
This does not work, even with Vagrant 1.7.4

PR #5919, opened 2015-07-10
core: plugin management requires bundler

gildegoma:
@mitchellh @sethvargo @phinze this is an humble attempt to solve the broken unit tests, but I can imagine that there is more elegant way to achieve this (see my code comments/questions)

### DIFF
--- a/lib/vagrant/shared_helpers.rb
+++ b/lib/vagrant/shared_helpers.rb
@@ -38,11 +38,11 @@ module Vagrant
     ENV["VAGRANT_INSTALLER_EMBEDDED_DIR"]
   end
 
-  # This returns whether or not 3rd party plugins should be loaded.
+  # This returns whether or not 3rd party plugins should and can be loaded.
   #
   # @return [Boolean]
   def self.plugins_enabled?
-    !ENV["VAGRANT_NO_PLUGINS"]
+    !ENV["VAGRANT_NO_PLUGINS"] && $vagrant_bundler_runtime
   end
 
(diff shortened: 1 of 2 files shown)


test/unit/support/shared/base_context.rb
```diff
@@ -12,6 +12,9 @@
     # Create a thing to store our temporary files so that they aren't
     # unlinked right away.
     @_temp_files = []
+
+    # Roughly simulate the embedded Bundler availability
+    $vagrant_bundler_runtime = Object.new
```
gildegoma: Please let me know if this is too ugly, and whether this should be implemented as a Mock/Stub object

lib/vagrant/shared_helpers.rb
```diff
@@ -38,11 +38,11 @@ def self.installer_embedded_dir
     ENV["VAGRANT_INSTALLER_EMBEDDED_DIR"]
   end
 
-  # This returns whether or not 3rd party plugins should be loaded.
+  # This returns whether or not 3rd party plugins should and can be loaded.
   #
   # @return [Boolean]
   def self.plugins_enabled?
-    !ENV["VAGRANT_NO_PLUGINS"]
+    !ENV["VAGRANT_NO_PLUGINS"] && $vagrant_bundler_runtime
```
gildegoma: I think this change makes sense, but of course we could also let it unmodified, and adapt the unit tests context to avoid https://travis-ci.org/mitchellh/vagrant/builds/70328862#L196.

mitchellh:
This should be fine for now. Thank!

PR #5928, opened 2015-07-10
Use the same ssh args for ssh with a command as ssh exec

sethvargo:
Fixes https://github.com/mitchellh/vagrant/issues/4986


phinze:
LGTM

PR #5931, opened 2015-07-11
guests/fedora: don't fail if nmcli doesn't exist on the guest

kaorimatz:
Running `vagrant up` with the fedora box where `nmcli` is not installed fails to configure guest networks.

Here is the reproducing step:

    $ vagrant box add --name fedora-cloud-22 https://download.fedoraproject.org/pub/fedora/linux/releases/22/Cloud/x86_64/Images/Fedora-Cloud-Base-Vagrant-22-20150521.x86_64.vagrant-virtualbox.box
    $ cat <<EOF > Vagrantfile
      config.vm.box = 'fedora-cloud-22'
      config.vm.network :private_network, ip: '10.0.0.2'
    end
    EOF
    $ vagrant up
    (snip)
    
    which nmcli >/dev/null 2>&1 && nmcli c reload eth1
    
    Stdout from the command:
    
    
    
    Stderr from the command:

This PR fixes this issue.


mitchellh:
LGTM

PR #5936, opened 2015-07-11
Fix salt bootstrap issue after 1.7.3 upgrade with verbose option set

wkral:
After upgrading to 1.7.3 we found that our VMs would not execute salt. I narrowed that down to the use of the verbose option. So removing that is a workaround for now. However we don't get any salt output which makes debugging changes to our config difficult.

I found that #5435 added a new `install_command` option but that seemed to always set the options value to a `DummyConfig` type instead of the expected string. It seems that the option wasn't initialized in `config.rb` so I've added that which fixes the problem on my machine.

Our current VMs already have salt installed but I think this might actually break normal salt install so it might be very urgent to fix for a good amount of people.

Fixes https://github.com/mitchellh/vagrant/issues/5939

### DIFF
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -33,6 +33,7 @@ module VagrantPlugins
       attr_accessor :install_args
       attr_accessor :install_master
       attr_accessor :install_syndic
+      attr_accessor :install_command
       attr_accessor :no_minion
       attr_accessor :bootstrap_options
       attr_accessor :version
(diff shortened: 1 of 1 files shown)


cedricpim:
Not sure if it is related, but even after removing the verbose option, provision is still failing.
````
Failed to upload a file to the guest VM via SCP due to a permissions
error. This is normally because the SSH user doesn't have permission
to write to the destination location. Alternately, the user running
Vagrant on the host machine may not have permission to read the file.

Source: #{my_dir}/saltstack/etc/minion
Dest: /etc/salt/minion
````

wkral:
@cedricpim I just cooked up a simple example to start a brand new box without salt on it yet and got the same result as you. 

I think the verbose option just exposed the problem as being that config value because it tries to print those install options to the console even if it's already installed. In any case the options aren't a string as they should be so installation fails.

mitchellh:
LGTM. Thanks!

Issue #5970, opened 2015-07-15
WinRM communicate.ready? raises exception on connection timeout


bheuvel:
Up untill v1.7.2 WinRM Communicator communicate.ready? returned false on connection timeout (Errno::ETIMEDOUT).

This used to be caught/handled by: https://github.com/mitchellh/vagrant/blob/v1.7.2/plugins/communicators/winrm/communicator.rb#L37
```ruby
rescue Vagrant::Errors::VagrantError => e
```

But since v1.7.3 the rescue has been replaced with:
https://github.com/mitchellh/vagrant/blob/v1.7.3/plugins/communicators/winrm/communicator.rb#L104 
```ruby
rescue Errors::TransientError => e
```
which doesn't catch Errno::ETIMEDOUT, resulting in ready? returning/raising an exception.

In my case this happens as I'm using(/coding) Winrm with(/into) the Vagrant-Cloudstack plugin ([my branch here](https://github.com/bheuvel/vagrant-cloudstack/tree/winrm)).
The guest is accessed by portforwarding/firewall. Untill WinRM is available, the firewall drops the TCP connect request packets, instead of returning denied/not-available (REJECT).

So this should result in "ready?" returning false, instead of an exception, as this is normal behaviour for waiting for the server to boot up. Or at least normal for me :wink:.

mitchellh:
Fixed by #5971

PR #5971, opened 2015-07-15
Raise Errno::ETIMEDOUT as "acceptable" Errors::ConnectionTimeout

bheuvel:
As described in issue #5970 communicate.ready? returns an exception on connection timeout, instead of returning false.

This catches the Errno::ETIMEDOUT, which will be caught by:
https://github.com/mitchellh/vagrant/blob/master/plugins/communicators/winrm/communicator.rb#L104
```ruby
rescue Errors::TransientError => e
```

And resulting in returning "just" false.


mitchellh:
LGTM Thanks

PR #5980, opened 2015-07-17
Fixed salt minion configuration directory typo

PatOShea:
Simple typo was killing the copy of the minion file when masterless config flag was set.

### DIFF
--- a/plugins/provisioners/salt/provisioner.rb
+++ b/plugins/provisioners/salt/provisioner.rb
@@ -336,7 +336,7 @@ module VagrantPlugins
 
       def call_highstate
         if @config.minion_config
-          @machine.env.ui.info "Copying salt minion config to #{@config.config dir}"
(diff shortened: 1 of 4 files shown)


mitchellh:
Same thing as your other PR, can you update to "windows guest machine" in docs.

Thanks so much for this though!

mitchellh:
Will do this on my own. Thanks!

plugins/provisioners/salt/provisioner.rb
```diff
@@ -104,7 +104,7 @@ def bootstrap_options(install, configure, config_dir)
           options = "%s %s" % [options, @config.bootstrap_options]
         end
 
-        if configure
+        if configure && !@machine.config.vm.communicator == :winrm
```
dkubb: I think this is a typo. It should probably be:

```ruby
if configure && @machine.config.vm.communicator != :winrm
```

PR #5981, opened 2015-07-17
Allow provisioner instance names to be specified for `up` and `reload…

jkburges:
…` commands and option `--provision-with`

Ref: https://github.com/mitchellh/vagrant/issues/5139

I'm happy to refactor, provide some tests etc for this but wanted to get some feedback first as to whether the general gist of it is ok (and any other advice).

### DIFF
--- a/plugins/commands/reload/command.rb
+++ b/plugins/commands/reload/command.rb
@@ -30,7 +30,7 @@ module VagrantPlugins
         return if !argv
 
         # Validate the provisioners
-        validate_provisioner_flags!(options)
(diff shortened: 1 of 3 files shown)


jkburges:
Also, provided that this behaviour is wanted, then I can also update the docs and error messages appropriately.

jkburges:
*bump* any feedback and/or advice on this PR?

mitchellh:
LGTM. I'm going to modify some code slightly to use a set and such to make it easier to read but thanks a lot!

PR #5986, opened 2015-07-17
Specify time and don't do -h -H which is not really a valid usage.

denisbr:
The current (1.7.3) behaviour leads to vagrant force-shutdown every time I halt a Debian 8 box:
```
vagrant halt
==> d8: Attempting graceful shutdown of VM...
==> d8: Forcing shutdown of VM...
```
On executing ```shutdown -h -H``` I understand why, because this simply schedules a shutdown in one minute:
```
root@vagrant:~# shutdown -h -H
Shutdown scheduled for Fri 2015-07-17 21:32:54 CEST, use 'shutdown -c' to cancel.
Broadcast message from root@vagrant (Fri 2015-07-17 21:31:54 CEST):
The system is going down for system halt at Fri 2015-07-17 21:32:54 CEST!
```

Also the usage of shutdown says:
```
shutdown [OPTIONS...] [TIME] [WALL...]
  -H --halt      Halt the machine
  -h             Equivalent to --poweroff, overridden by --halt
```

### DIFF
--- a/plugins/guests/debian8/cap/halt.rb
+++ b/plugins/guests/debian8/cap/halt.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
       class Halt
         def self.halt(machine)
           begin
-            machine.communicate.sudo("shutdown -h -H")
+            machine.communicate.sudo("shutdown -h now")
           rescue IOError
             # Do nothing, because it probably means the machine shut down
             # and SSH connection was lost.


Razrael:
I have the same problem, but its even worse since a few weeks. Looks like the time in the shutdown command is no more optional:
```shutdown [-akrhPHfFnc] [-t sec] time [warning message]```

If I want to shut it down I have to do it via ssh or do a vagrant halt -f, so just pull the plug.
I tested this change and it works nicely.

cbuchli:
I have the same Problem, 1.7.4 is still affected too. This change fixes the problem on my site. Thank you.

chrko:
Please merge this. It was annonying to debug and to find this open pr!

mwdiers:
I did not even realize `shutdown -h -H` was ever valid. Thank you for pointing me to the location of the halt command in the debian8 guest config. This was driving me crazy.

Razrael:
Could you please merge this?
The provided solution is already tested and works. And the issue is still existing on latest dev.

thinktainer:
Would it be much trouble to add this to the next milestone?

bitprocessor:
Debian 8 doesn't need an exception at all  ; see https://github.com/mitchellh/vagrant/pull/6315

mwdiers:
Good catch.

mitchellh:
LGTM

bitprocessor:
@mitchellh According to me (and a whole lot of other people as well), the whole Debian 8 exception is useless ;-)

See #6315

mitchellh:
Ah hah! :) I'm going in chronological so I hadn't gotten there yet. Will merge that now!

PR #5993, opened 2015-07-18
Corrected masterless example for salt provisioner in documentation

PatOShea:
The basic Vagrantfile file example for a masterless setup wasn't setting the masterless value to true.  That value defaults to false in config.rb and used on line 350 here:  https://github.com/mitchellh/vagrant/blob/master/plugins/provisioners/salt/provisioner.rb


obestwalter:
:+1: default should definitely be to run masterless in a vagrant setup.

mitchellh:
LGTM!

PR #5994, opened 2015-07-18
Copy grains file to the minion config directory for salt provisioner

PatOShea:
Grains file is being copied to the guest, but it's not being copied where the minion can automatically pick it up next to the minion config.


PatOShea:
This is in 1.7.4

PR #5999, opened 2015-07-19
Correct salt-minion and call ext on windows guest

PatOShea:
Salt-minion and salt-call are batch files on a windows guest, not executables.


mitchellh:
Thanks!

PR #6004, opened 2015-07-19
[website] load external scripts without schema

pogliamarci:
Hello,
currently www.vagrantup.com and docs.vagrantup.com are loading web fonts using HTTP even when HTTPS is used (default for vagrant's website).
This causes most browsers (Chromium and Firefox at least) to block those resources by default due to security issues (HTTPS is useless if an active element such as a script is loaded using plain HTTP because it has access to the whole DOM...).

Loading the external resources over HTTPS avoids this issue.

Thanks,
Marcello


mitchellh:
LGTM Thanks!

Issue #6006, opened 2015-07-20
CFEngine Provisioner Broken on Debian


darkn3rd:
Debian seems broken in Vagrant 1.7.4.  I also found that it was broken in an earlier 1.6 version, thus it maybe have been broken since then.

Running `vagrant provision` or the initial `vagrant up` yields this:

`W: Failed to fetch http://cfengine.com/pub/apt/dists/wheezy/main/binary-amd64/Packages  403  Forbidden`

My Vagrantfile is nothing magical, just barebones:

```Ruby
VAGRANTFILE_API_VERSION = "2"
Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  config.vm.box      = "chef/debian-7.8" # Get Debian Wheezy From Somewhere
  config.vm.hostname = "wheezybox"       # Give System a Hostname

  # Vagrant CFEngine Provisioner
  config.vm.provision :cfengine do |cf|
    cf.am_policy_hub = true
  end
end
```

I'm unfamiliar with CFEngine or how they binaries were distributed in the past, but taking a guess, maybe they changed it.

darkn3rd:
I went through the steps manually as described at https://cfengine.com/product/community/cfengine-linux-distros/

```bash
wget -qO- https://cfengine.com/pub/gpg.key | apt-key add -
echo "deb http://cfengine.com/pub/apt/packages stable main" > /etc/apt/sources.list.d/cfengine-community.list
apt-get update
apt-get install cfengine-community
```
This manual process seems to work:

```bash
# cf-promises --version
CFEngine Core 3.7.0
# test -d /var/cfengine && test -x /var/cfengine/bin/cf-agent
# echo $?
0
```

In the debug, when doing vagrant up, I noticed these two commands, they seem similar, except Vagrant explicitly adds `wheezy` while CFEngine's instructions add `stable`.

```
mkdir -p /etc/apt/sources.list.d && /bin/echo deb http://cfengine.com/pub/apt $(lsb_release -cs) main > /etc/apt/sources.list.d/cfengine-community.list (sudo=true)
GPGFILE=`tempfile`; wget -O $GPGFILE http://cfengine.com/pub/gpg.key && apt-key add $GPGFILE; rm -f $GPGFILE (sudo=true)
```

darkn3rd:
Test change on Debian Wheezy.

https://github.com/mitchellh/vagrant/pull/6007

darkn3rd:
[...]

PR #6007, opened 2015-07-20
Fix for Issue #6006 - cfengine will now install

darkn3rd:
With this fix, Debian will now work with `cfengine` provisioner.


sethvargo:
Hi @darkn3rd 

Do you know if this affects other debian-like systems (like Ubuntu)?

darkn3rd:
This seems to have also broken Ubuntu as well, and this PR will also fix Ubuntu.  Tested this on Ubuntu Percise, using the `puppetlabs/ubuntu-12.04-64-puppet` box.

Issue #6025, opened 2015-07-23
Chef Zero Provisioner - Cannot specify Nodes Path


bdwyertech:
After the release of Vagrant 1.7.3, the Chef Zero provisioner no longer accepts the ```nodes_path``` attribute, which would typically be a path to node objects for use in Chef Zero's Chef Server emulation.  I am unsure if this was intentional or not.

I previously used something like the below, but it no longer works.

```ruby
# Chef-Zero Configuration
  config.vm.provision "chef_zero" do |chef|
    # Specify the local paths where Chef data is stored
    chef.cookbooks_path = ["~/GIT/chef/cookbooks", "~/GIT/chef/site-cookbooks"]
    chef.roles_path = "~/.chef/CHEF/backup/roles"
    chef.nodes_path = "~/.chef/CHEF/backup/nodes"
    chef.data_bags_path = "~/GIT/chef/data_bags"

    # Add a role
    chef.add_role "alpha"
  end
```

I think https://github.com/mitchellh/vagrant/pull/5339 may have caused this

drywheat:
I can confirm this behavior in v1.7.4 as well.

PR #6049, opened 2015-07-28
Chef Zero - Fix 'nodes_path' Support

bdwyertech:
This fixes an issue #6025 by re-adding support for the `nodes_path` declaration to feed Chef Zero a directory (or an array of directories) containing nodes in JSON format.


bdwyertech:
Ran into the same issue @jperville had tonight... Only had the issue after installing `vagrant hostmanager`.  These commits against the Chef Solo provisioner seem to alleviate that issue.

PR #6057, opened 2015-07-29
Fix puppet4 autodetect

benh57:
This PR fixes the automatic detection of puppet 4 opensource and Puppet Enterprise 2015.0.2 (which is installed to /usr/local/bin/puppet)

PE 3.8.x should also still be properly autodetected (it's installed to /opt/puppet..)

I temporarily disabled support of puppet.binary_path on Windows which broke with this PR. The winrm provider doesn't seem to like us calling puppet with the full path, when it includes spaces. It's not something that will generally be needed, anyway.

I've tested this with: 
puppet enterprise 2015.2.0 - Linux  (runs off puppet 4 opensource base)
puppet enterprise 3.8.x - Windows
puppet enterprise 3.8.x - Linux

### DIFF
--- a/plugins/provisioners/puppet/provisioner/puppet.rb
+++ b/plugins/provisioners/puppet/provisioner/puppet.rb
@@ -109,12 +109,12 @@ module VagrantPlugins
           verify_shared_folders(check)
 
           # Verify Puppet is installed and run it
-	  
+
           puppet_bin = "puppet"
           if windows?
-	    puppet_bin = "puppet.bat"
+	          puppet_bin = "puppet.bat"
           end
-	  
+
           @config.binary_path = verify_binary(puppet_bin)
 
           # Upload Hiera configuration if we have it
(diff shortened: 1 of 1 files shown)


briancain:
Hi there,

I'm going to go ahead and close this PR as it's rather old at this point. I've just tried the puppet provisioner on Windows and Linux with Puppet version 5.2.x and it works as expected. If there are still issues with the puppet provisioner, I recommend opening a new PR based on the latest source changes. Thanks!

benh57:
This failure was specifically with Puppet Enterprise -- did you test that? Unfortunately, we no longer use Puppet Enterprise so i wouldn't be able to reproduce it at this point.

briancain:
Hi @benh57 - as far as I know, windows agents even in a puppet enterprise environment should be using the same agent package as an open source agent. That might not of been the case in the 3.x series of PE, but I believe it should be for any version of PE in the past couple of years.

benh57:
Yeah, i think your'e right, it was different before the new 'agent' thing was rolled out.

Issue #5973, opened 2015-07-15
Salt provisioner fails to upload minion config


piohhmy:
Ran into a problem in 1.7.3 doing a vagrant up using a trusty64 box with a salt provisioner specifying a custom minion_config. This is the error:
```
Copying salt minion config to /etc/salt
Failed to upload a file to the guest VM via SCP due to a permissions
error. This is normally because the SSH user doesn't have permission
to write to the destination location. Alternately, the user running
Vagrant on the host machine may not have permission to read the file.

Source: /Users/danny/dev/se/one-offs/vagrant/minion.yaml
Dest: /etc/salt/minion
```

It appears that the vagrant user does not have permission by default to scp to /etc/salt/ so provisioning fails.  I confirmed it is possible to vagrant ssh into the machine and change the permissions/owner on /etc/salt/ and then the provisioning will work OK.  

It appears 74d22069d41b35bae50e32c1c68a92a4958afebb may have been the breaking change as it introduced the uploading to /etc/salt.

SoftCreatR:
I can confirm this. We're currently experiencing this issue, too.

sephlaire:
I can confirm this as well. I'm using Vagrant 1.7.3 with VirtualBox 5.0. As a workaround I've been using a shell provisioner with the following to make sure salt in installed and the config is copied over:

```
#!/usr/bin/env bash

echo "Setup saltstack PPA"
sudo apt-get install software-properties-common python-software-properties -y >> /dev/null
sudo add-apt-repository ppa:saltstack/salt -y >> /dev/null
sudo apt-get update >> /dev/null

echo "Installing salt-minion"
sudo apt-get install salt-minion -y >> /dev/null
sudo cp /vagrant/salt/configs/ubuntu.conf /etc/salt/minion

echo "Restart salt-minion"
sudo salt-minion -d
sudo service salt-minion restart
```

nmadhok:
I can confirm this issue. As a workaround, I specified ``config.ssh.username = "root"`` and ``config.ssh.password = "vagrant"`` in my Vagrant file

clamor:
Same issue with provider vmware_desktop.

royka:
I've got this issue too

amchoukir:
I can confirm the issue too

sonoranDesert:
This also started happening to me after upgrading to Vagrant 1.7.4.

dwickwire:
[...]

PR #6073, opened 2015-07-31
Stop trying to upload minion config to privileged directory

johntron:
The issue #5973 was caused by commit 74d22069d41b35bae50e32c1c68a92a4958afebb. Normally, the minion config is installed to the privileged `/etc/salt` directory by `salt-bootstrap.sh` as part of the `config_salt()` function. The reverted commit was having Vagrant upload the minion file *directly* to the `/etc/salt` directory - most installations probably don't have permissions to do this since vagrant is not running as root.

The original commit was to address situations where vagrant was being used with a basebox that already had salt installed. For these situations, users can simply provide the "-C" option as part of `bootstrap_options".

I've spent hours sorting this out and [addressing another issue with salt-bootstrap](https://github.com/saltstack/salt-bootstrap/pull/642) (I have to install from `develop` by modifying the `bootstrap-salt.sh` file in my vagrant gem). For the record, just use `rvm` - don't try to install without it.

### DIFF
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -20,7 +20,6 @@ module VagrantPlugins
       attr_accessor :bootstrap_script
       attr_accessor :verbose
       attr_accessor :seed_master
-      attr_accessor :config_dir
       attr_reader   :pillar_data
       attr_accessor :colorize
       attr_accessor :log_level
@@ -65,7 +64,6 @@ module VagrantPlugins
         @install_command = UNSET_VALUE
         @no_minion = UNSET_VALUE
         @bootstrap_options = UNSET_VALUE
-        @config_dir = UNSET_VALUE
         @masterless = UNSET_VALUE
         @minion_id = UNSET_VALUE
         @version = UNSET_VALUE
@@ -98,7 +96,6 @@ module VagrantPlugins
         @install_command    = nil if @install_command == UNSET_VALUE
         @no_minion          = nil if @no_minion == UNSET_VALUE
         @bootstrap_options  = nil if @bootstrap_options == UNSET_VALUE
-        @config_dir         = nil if @config_dir == UNSET_VALUE
         @masterless         = false if @masterless == UNSET_VALUE
         @minion_id          = nil if @minion_id == UNSET_VALUE
         @version            = nil if @version == UNSET_VALUE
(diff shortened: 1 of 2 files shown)


driskell:
The bootstrap script takes care of getting the configurations installed so there's not really any need for the code getting removed here. Specifically, if the masterless doesn't need to do this then neither does the highstate.

I've not tested but LGTM.

nkantar:
Given how much head-scratching this has caused (#5973) and how inelegant the fix is, :+1: from me.

obestwalter:
:+1: Is somebody capable of testing this? Would be great to get this fixed in 1.7.5

johntron:
I just ran the tests successfully and so did Travis CI.

joshughes:
+1

johntron:
By the way, in my project, we're using Ubuntu 12.04 and this provisioning:
```ruby
    salt.bootstrap_options = "-P -c /tmp"
    salt.minion_config = "salt/minion"
    salt.grains_config = "salt/grains"
    salt.run_highstate = true
    salt.log_level = "warning"
  end
```

This allows installation via PIP and installs configuration files (minion and grains) from the `/tmp` directory the salt-bootstrap.sh is run from.

tyauvil:
+1

babilen:
It would be great if this could be merged. In case there is a problem it would also be good to know about it so that the community can work on a different approach.

johntron:
@mitchellh or @sethvargo - can one of you merge this?

kevich:
@mitchellh or @sethvargo would you mind answering us? This pull request surely must be included in last version of vagrant.

Yes, we can manually fix this problems with salt on our local instances, but that too bad, that we have two releases with broken salt provisioner, and no sign of activity to make this fixed from main maintainers of vagrant!

obestwalter:
@kevich :+1: 

It's even worse on Windows. The combination of salt/vagrant/windows is broken for over a year now (due to several bugs in both projects). I am trying to help to [move things forward](https://trello.com/b/vywwuncS/ci-for-salt-vagrant-windows), but on the vagrant side I don't even see interest in helping me to get started (https://groups.google.com/forum/#!topic/vagrant-up/2vBhQMYsqn4). Here is my last attempt to at least get the documentation for installing vagrant from source fixed https://github.com/mitchellh/vagrant/issues/6183 - can we Pythonistas get some love from the Rubyists please?

mitchellh:
LGTM!

johntron:
Thanks @mitchellh!

PR #6080, opened 2015-08-03
Fix broken masterless minion configuration

driskell:
Vagrant 1.7.4 seems to break the masterless provisioning process for SaltStack.

It throws an error during provisioning:

```
Salt binaries found. Configuring only.

/tmp/bootstrap_salt.sh  -C

Stdout from the command:



Stderr from the command:

  % Total    % Received % Xferd  Average Speed  Time    Time    Time  Current
                                Dload  Upload  Total  Spent    Left  Speed
  0    0    0    0    0    0      0      0 --:--:-- --:--:-- --:--:--    0 * ERROR: In order to run the script in configuration only mode you also need to provide the configuration directory.
23  204k  23 48521    0    0  257k      0 --:--:-- --:--:-- --:--:--  929k
curl: (23) Failed writing body (631 != 16384)
```

The parameters for the configuration path are missing. Seems to be due to a typo in the code in a recently merged PR (#5980)

### DIFF
--- a/plugins/provisioners/salt/provisioner.rb
+++ b/plugins/provisioners/salt/provisioner.rb
@@ -104,7 +104,7 @@ module VagrantPlugins
           options = "%s %s" % [options, @config.bootstrap_options]
         end
 
-        if configure && !@machine.config.vm.communicator == :winrm
+        if configure && @machine.config.vm.communicator != :winrm
           options = "%s -F -c %s" % [options, config_dir]
         end
 


johntron:
Awesome, I was going to submit a PR for this one as well. LGTM, but haven't tested.

dkubb:
I've tested this and it works, with one minor change:

```diff
diff --git a/plugins/provisioners/salt/provisioner.rb b/plugins/provisioners/salt/provisioner.rb
index 7236577..7e4e0fe 100644
--- a/plugins/provisioners/salt/provisioner.rb
+++ b/plugins/provisioners/salt/provisioner.rb
@@ -105,7 +105,7 @@ module VagrantPlugins
         end
 
         if configure && @machine.config.vm.communicator != :winrm
-          options = "%s -F -c %s" % [options, config_dir]
+          options = "%s -P -F -c %s" % [options, config_dir]
         end
 
         if @config.seed_master && @config.install_master
```

Without this I got an error about tornado not being installed properly.

sethvargo:
@driskell can you have a look at @dkubb's comment please?

johntron:
@sethvargo yeah, the `-P` (allow installation with `pip`) is required on some systems, but not all - it depends on whether or not your distro has a package available for the correct version. I don't know if there are any downsides to always installing from pip, other than bootstrapping taking longer.

dkubb:
>  the -P (allow installation with pip) is required on some systems, but not all - it depends on whether or not your distro has a package available for the correct version.

FWIW, I am using the trusty [amd64](http://cloud-images.ubuntu.com/vagrant/trusty/current/trusty-server-cloudimg-amd64-vagrant-disk1.box) box. I removed my cached box so I was getting the latest (2015-06-15) version.

> I don't know if there are any downsides to always installing from pip, other than bootstrapping taking longer.

Sorry, I don't know the answer to this either. My experience with python is limited. I mainly stumbled onto the `-P` fix by trying the suggested fix in the error message that appeared.

dkubb:
IMHO Even though I ran into a problem with `-P` I don't think that should block this patch from being merged. The code is obviously broken, and the patch fixes it, so it's fixing a regression even if it doesn't deal with the problem I am seeing.

Also it looks like the salt developers are [aware](https://groups.google.com/forum/#!topic/salt-users/7uhgNqPoZSA) of the problem.

driskell:
Sure @sethvargo 

Based on @dkubb research (thanks!) I would say not to add `-P` as a default. Maybe it needs to be an option though if SaltStack aren't able to package any time soon? I would argue that's another PR, though, as it's a separate issue IMHO.

johntron:
Yeah I agree @driskell. Typically packaging isn't done by the Saltstack folks - for instance, they use [the chris-lea PPA](https://launchpad.net/~chris-lea/+ppa-packages) for [Ubuntu installations](https://github.com/saltstack/salt-bootstrap/blob/b1624cc14a91f21abab1da0fe1ef23d5a95d2ef1/bootstrap-salt.sh#L1768) - so I think it makes sense to give the Vagrant user the option to install via pip. As long as they can see [the output of the saltstack bootstrap](https://github.com/saltstack/salt-bootstrap/blob/b1624cc14a91f21abab1da0fe1ef23d5a95d2ef1/bootstrap-salt.sh#L1847), they should be able to find a fix for a failed installation.

driskell:
Is this OK to merge @sethvargo ? It's such a small tweak and an actual blocker for me, and I think the other points regarding `pip` etc. could be addressed separately. Thanks.

johntron:
BTW - users can still [add the `-P` option themselves with `bootstrap_options`](https://github.com/mitchellh/vagrant/pull/6080/files#diff-c4526b51e4e1b1f3f6bff789181cce2bR103) if configuration fails without it.

dkubb:
I'd like to reiterate @kevich's question about when there'll be a vagrant release with this fix? I currently have to run vagrant under `bundler` in order to reference the git repo to work-around this issue.

PR #6090, opened 2015-08-05
Add check if NetworkManager is actually running inside the guest

vbaskoff:
Hi,
Our custom boxes are based on oracle linux 7.X, but we have disabled NetworkManager, because of our product specific. AFAIK other peoples also prefer to disable NM, due to different reasons.

My request try to address the situation when NM rpm is installed but the service isn't running inside the guest.


gpaul:
+1

sethvargo:
https://github.com/mitchellh/vagrant/pull/6536

PR #6111, opened 2015-08-07
Remove upper bounds on bundler in gemspec

mtwentyman:
Because there seem to be fairly [regular bundler releases](https://github.com/bundler/bundler/commits/master/lib/bundler/version.rb),
a strict upper bounds causes problems for those keeping up to date.

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
   s.rubyforge_project         = "vagrant"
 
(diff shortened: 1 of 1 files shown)


sethvargo:
We depend on Bundler internal APIs so this is not possible. We have seen breaking changes in patch releases because of our internal API dependency (which is our fault, not theirs). We plan to remove these in 1.8 and then we can relax the dependency.

mtwentyman:
Fantastic!  Thank you for such a quick resolution and the requirement explanation.

josephfrazier:
I just experienced this issue. Did the internal API dependencies get removed in 1.8, or is that yet to come?

Issue #6075, opened 2015-08-01
Upgrade Docker with provisioner


cazzerson:
Is it possible to force the docker provisioner to upgrade Docker to the latest release?

sethvargo:
Hi @cazzerson 

The `version` field documented here http://docs.vagrantup.com/v2/provisioning/docker.html specifies 

> The version of Docker to install. This defaults to "latest" and will install the latest version of Docker.

Does this answer your question?

cazzerson:
Hi @sethvargo, thanks for the response. This seems to work well for the initial install but, as far as I can tell, once Vagrant has installed Docker on a VM, it won't upgrade it. I may be missing something here, though.

sethvargo:
@cazzerson right, this is correct. But since containers are pretty portable, I would just destroy and re-create it.

cazzerson:
@sethvargo We are [distributing a vagrant config](https://github.com/NCSU-Libraries/Social-Media-Combine) for external users, so it's difficult for us to keep them up to date if it requires them to periodically destroy their VM. Maybe we should just be installing Docker with our own provisioner.

sethvargo:
@cazzerson I think that's a good stop-gap. I don't have quite enough context to decide if it's a good idea to always force the latest version as a top-level function in Vagrant, and I'd want to do some more thinking on that before making a decisions. But I think using a custom provisioner might be the best idea for now.

cazzerson:
Thanks for your help. If you do return to this issue, I think it may be better to allow an upgrade rather than forcing it in every case, or even to request a later version than the one currently installed.

zimmski:
I stumbled upon this issue because of https://blog.docker.com/2015/07/new-apt-and-yum-repos/ there will be no new versions of docker in the old repository. Vagrant needs to adapt. Maybe I should open an issue for that?

However, I also agree with @cazzerson a supported update and upgrade path would be nice. Users of the provisioner should then explicitly mention the Docker version if they do not like to auto-upgrade.

sethvargo:
@zimmski I would recommend opening an issue (or better yet, a PR!) to fix that new repo issue. Thanks!

cazzerson:
@sethvargo @zimmski I'm playing around with a possible PR, but I'm running into some issues with the new distribution-oriented repos. The [Docker install script](https://get.docker.com/) has some platform detection to handle this. We could embed that fragment of this script, or just use the whole thing. Would you be concerned about pulling down the Docker installer?

[...]

PR #6112, opened 2015-08-07
Use official Docker installer and upgrade for :latest installs

cazzerson:
A first attempt at updating the Docker installer as discussed in #6075.

### DIFF
--- a/plugins/provisioners/docker/cap/debian/docker_install.rb
+++ b/plugins/provisioners/docker/cap/debian/docker_install.rb
@@ -4,13 +4,10 @@ module VagrantPlugins
       module Debian
         module DockerInstall
           def self.docker_install(machine, version)
-            package = 'docker-engine'
-            package << "-#{version}" if version != :latest
-
             machine.communicate.tap do |comm|
-              comm.sudo("apt-get update -y")
-              comm.sudo("apt-get install -y --force-yes -q curl")
-              comm.sudo("apt-get purge -y lxc-docker*")
+              comm.sudo("apt-get update -qq -y")
+              comm.sudo("apt-get install -qq -y --force-yes curl")
+              comm.sudo("apt-get purge -qq -y lxc-docker*")
               comm.sudo("curl -sSL https://get.docker.com/ | sh")
             end
           end
--- a/plugins/provisioners/docker/cap/debian/docker_install.rb
+++ b/plugins/provisioners/docker/cap/debian/docker_install.rb
@@ -4,24 +4,14 @@ module VagrantPlugins
       module Debian
         module DockerInstall
           def self.docker_install(machine, version)
-            package = 'lxc-docker'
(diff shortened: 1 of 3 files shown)


plugins/provisioners/docker/cap/debian/docker_install.rb
```diff
@@ -4,24 +4,14 @@ module Cap
       module Debian
         module DockerInstall
           def self.docker_install(machine, version)
-            package = 'lxc-docker'
+            package = 'docker-engine'
```
sethvargo: It doesn't look like this is used anymore?

sethvargo:
Hi @cazzerson 

This is looking really good. I have a few high-level comments:

1. What do you think about extracting the installation into a library like we do in the [Chef installer](https://github.com/mitchellh/vagrant/blob/master/plugins/provisioners/chef/installer.rb)?

1. I think we need to run `apt-get update -qq` before installing curl

1. I think we want to add `-qq` to all the apt calls

I'm not super familiar with the new Docker install script. It's on my list of things to review, but :hourglass: :smile:. Are there options that users would like to pass in? If so, we should provide a way to pass those args in. See the Chef installer for an example.

cazzerson:
@sethvargo Thanks, I'll push those fixes. Two things:

1. Of course, the more I look at this, the more I notice. This may eliminate the ability to freeze the Docker version (defaulting to :latest). This was already the case for RHEL, though. Also, when you have time, search for 'Warning: the "docker" command appears to already exist on this system.' on https://get.docker.com. The script waits for 20 seconds for the user to bail on an upgrade with Ctrl-C. Is there anything else like this in other provisioners?

2. Command line options don't seem to be an issue.

3. I don't have much experience with Chef, but I can take a look when I have :hourglass: :smile:.

cazzerson:
@sethvargo Sorry, not with Chef, **like** Chef :-P I'll take a look.

cazzerson:
@sethvargo Could you tell more about about what ideas you'd like to bring from the Chef installer into the [Docker Installer](https://github.com/mitchellh/vagrant/blob/master/plugins/provisioners/docker/installer.rb)?

sethvargo:
Hey @cazzerson yea sure, but it might be a bit until I have a chance to context switch into that :smile:

mitchellh:
This looks good. My only concern is that it appears installing specific versions has been removed. If we can support that again then I'll merge!

cazzerson:
The first issue seems to be that get.docker.com is no longer recommended--is this correct, @jfrazelle? Additionally, as of 1.9, although the docs recommend uninstalling lxc-docker, the lxc-docker packages now seem to be installing 1.9 on when provisioning a fresh machine. We could switch from lxc-docker to docker-engine (with some check for distro repo selection), but I would need to do more testing for existing installations--I ran into some weird issues when trying to replace lxc-docker by hand.

PR #6118, opened 2015-08-09
Fix Win 10 Enterprise Vagrant Error

mjaverto:
This fixes error for issue https://github.com/mitchellh/vagrant/issues/6117

### DIFF
--- a/plugins/providers/hyperv/scripts/get_vm_status.ps1
+++ b/plugins/providers/hyperv/scripts/get_vm_status.ps1
@@ -12,7 +12,7 @@ try {
     $VM = Get-VM -Id $VmId -ErrorAction "Stop"
     $State = $VM.state
     $Status = $VM.status
-} catch [Microsoft.HyperV.PowerShell.VirtualizationOperationFailedException] {
+} catch [Microsoft.HyperV.PowerShell.VirtualizationException] {
     $State = "not_created"
     $Status = $State
 }


rbcarson:
Thanks for this, @mjaverto! I was trying to use Windows 10 with Hyper-V and hit this exact issue, and your fix solves the problem.

(And I'm using Hyper-V because [VirtualBox doesn't run on Windows 10 build 10525](https://forums.virtualbox.org/viewtopic.php?f=6&t=69652).)

NickMRamirez:
Will this work with older version of PowerShell where the original exception type was defined? Or would it need to check for both types in some way?

danoz-direct:
Thanks @mjaverto, this fix worked for me also.

I was seeing the error manifest itself after a `vagrant destroy`.

rbcarson:
I hit this issue again after reinstalling Vagrant and forgetting about it. It's worth noting that it's not specific to Enterprise - it's a problem on Windows 10 in general.

mitchellh:
Does this also work in earlier versions of Windows?

mitchellh:
Just tested it locally and it looks good!

KevinB-Work:
Not certain that this works universally with previous versions. Just Installed 1.8.1 on system that had been using 1.7.2 and it fails as it doesn't know about VirtualizationException. Manually changing the script to use VirtualizationOperationFailedException again fixes. Needs some way to detect which exception is valid

Issue #6131, opened 2015-08-12
vagrant up a Windows 10 box with VMware Fusion has problems with WinRM


StefanScherer:
Host: OSX 10.10.4, Vagrant 1.7.4, VMware Fusion 7.1.2 with the following plugins:

```
$ vagrant plugin list
winrm-fs (0.2.2)
vagrant-multiprovider-snap (0.0.14)
vagrant-pristine (0.3.0)
vagrant-share (1.1.4, system)
vagrant-vmware-fusion (3.2.11)
```

I still encounter problems spinning up a Windows 10 Vagrant box created with the [boxcutter/windows](https://github.com/boxcutter/windows) repo with VMware Fusion on a Mac (`make vmware/eval-win10x64-enterprise`).

I did some testing with VirtualBox 5.0.0 in the meantime and fixed similar provisioning problem there by updating the Ruby gem winrm-fs to version 0.2.2 (see #6060).

Spinning up the box looks good, the VM starts, but at the beginning of the provisioning the following error message is shown:

```
$ vagrant up --provider vmware_fusion
Bringing machine 'default' up with 'vmware_fusion' provider...
==> default: Cloning VMware VM: 'windows_10'. This can take some time...
==> default: Verifying vmnet devices are healthy...
==> default: Preparing network adapters...
==> default: Starting the VMware VM...
==> default: Waiting for machine to boot. This may take a few minutes...
The box is not able to report an address for WinRM to connect to yet.
WinRM cannot access this Vagrant environment. Please wait for the
Vagrant environment to be running and try again.
```

After waiting some seconds and then running `vagrant provision` vagrant is able to connect via WinRM:

```
$ vagrant provision
==> default: Running provisioner: shell...
    default: Running: provision.ps1 as c:\tmp\vagrant-shell.ps1
==> default: hello
```

There is no unusual long time to boot the VM, so there shouldn't be some kind of timeout there.
I have done the same with `VAGRANT_LOG=debug`.  You can find more details in this Gist:
https://gist.github.com/StefanScherer/f9f93c5b96d4ba5718ae

StefanScherer:
Seems that the loop calling `vmrun getGuestIPAddress ...` should be more tolerant.

Interesting lines in the Gist of the `vagrant up`:

[...]

PR #6137, opened 2015-08-13
Stabilize wait_for_ready to fix Windows 10 start problem

StefanScherer:
The `wait_for_ready` helper waits until WinRM is ready. Spinning up a Windows 10 guest VM sometimes the inner function `winrm_info` raises the exection `WinRMNotReady`. By ignoring this exception the `wait_for_ready` waits correctly until an IP address could be retrieved.

This solves my problem on OSX 10.10.4 with VMware Fusion 7.1.2 spinning up a Windows 10 VM in issue #6131.

### DIFF
--- a/plugins/communicators/winrm/communicator.rb
+++ b/plugins/communicators/winrm/communicator.rb
@@ -30,7 +30,10 @@ module VagrantPlugins
           # Wait for winrm_info to be ready
           winrm_info = nil
           while true
-            winrm_info = Helper.winrm_info(@machine)
+            begin
+              winrm_info = Helper.winrm_info(@machine)
+            rescue Errors::WinRMNotReady
+            end
             break if winrm_info
             sleep 0.5
           end


blomma:
I can say that this problem exists with a windows 8 guest vm also, this is on OS X 10.10.4, VMware Fusion 7.1.2, Vagrant 1.7.4 and this fix also fixes that problem for me.

Something happened in the upgrade for either of those components because before upgrading either of these components i didn't have this problem.

StefanScherer:
I forgot to mention the vagrant-vmware-fusion plugin version. I've used 3.2.11. There is an improvement in 3.2.12 (see #6131), but still this PR stabilizes the loop until WinRM is available.

blomma:
I upgraded to the newest version of vagrant-vmware-plugin (3.2.12) and that fixed the problem for me.

mitchellh:
This was fixed already, sorry we missed this PR!

StefanScherer:
No problem ;-)

PR #6139, opened 2015-08-13
Fixes Fedora network issues when biosdevname command is not present.

tjanez:
Previously, configuring and enabling network interfaces failed with:

```

/usr/sbin/biosdevname --policy=all_ethN -i bash: /usr/sbin/biosdevname:
No such file or directory

Stdout from the command:

bash: /usr/sbin/biosdevname: No such file or directory
```

The previous attempt to fix this (ccc4162) doesn't work since it doesn't
properly parse the `bash: /usr/sbin/biosdevname: No such file or
directory` error message.

This patch works around that problem and adds a comment explaining the
meaning of the return codes.

### DIFF
--- a/plugins/guests/fedora/cap/configure_networks.rb
+++ b/plugins/guests/fedora/cap/configure_networks.rb
@@ -17,7 +17,10 @@ module VagrantPlugins
           virtual = false
           interface_names = Array.new
           interface_names_by_slot = Array.new
-          machine.communicate.sudo("/usr/sbin/biosdevname; echo $?") do |_, result|
+          machine.communicate.sudo("/usr/sbin/biosdevname &>/dev/null; echo $?") do |_, result|
+            # The above command returns:
+            #   - '4' if /usr/sbin/biosdevname detects it is running in a virtual machine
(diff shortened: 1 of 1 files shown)


tjanez:
Updated the patch and tested it with vanilla Vagrant 1.7.4 (+ this patch applied) on:
- [official Fedora 22 Vagrant image for VirtualBox](https://download.fedoraproject.org/pub/fedora/linux/releases/22/Cloud/x86_64/Images/Fedora-Cloud-Base-Vagrant-22-20150521.x86_64.vagrant-virtualbox.box),
- [official CentOS 7 Vagrant image for VirtualBox v1508.01](https://atlas.hashicorp.com/centos/boxes/7/versions/1508.01).

Things work as expected:
- On Fedora 22, the value of the `result` variable is `127` and `virtual` is set to `true`.
- On CentOS 7 the value of the `result` variable is `1` and `virtual` is set to `false`.

@dcbw, would please take a look at this PR?

dcbw:
@tjanez LGTM, not sure why I didn't see that problem with the previous patch.

tjanez:
@mitchellh, @sethvargo, it would be great if you could look at this and merge it.

PR #6145, opened 2015-08-14
Update bundler version constraint

philoserf:
Now that bundler 1.10.6 is out. Let's be pessimistic rather than forcing some users to downgrade bundler.


sethvargo:
Duplicate of https://github.com/mitchellh/vagrant/pull/6111. Please see my explanation on that PR please.

PR #6149, opened 2015-08-15
The docs typo'd the nic_type parameter

pwnall:
The parameter name in the VirtualBox provider source is nic_type, not nictype.

The typo is probably caused by the fact that the parameter maps to nictype args in the VirtualBox CLI.


mitchellh:
LGTM

PR #6150, opened 2015-08-15
Don't warn about an .1 IP for DHCP networks

pwnall:
When the network's type is :dhcp, the :ip option is used to derive the DHCP server configuration, and it doesn't actually indicate the IP that will be received by the VM(s).


mitchellh:
LGTM

PR #6157, opened 2015-08-17
website/www: Add Atlas box search to nav

justincampbell:
I think that there should be a prominent link to box discovery on the Vagrant home page.
A few times I've Googled "vagrant", ended up here, and didn't know how to get to a box search page without doing another Google search.

However, a better solution may be to add a `<section>` with "Discover Vagrant Boxes on Atlas".


sethvargo:
:+1:

PR #6160, opened 2015-08-18
Scrub Guest Paths for Windows Rsync leaving Dirty Paths for Winrm Mkdir

lonniev:
Windows offers no out-of-the-box rsync utility. By far, the most
commonly used external utilities for Windows rsync are built with the
GNU Cygwin libraries. The cost for this convenience is that rsync on
Windows has to be provided paths that begin “/cygdrive/c” rather than
“c:/“ like other Windows-API utilities. Compounding the situation,
rsync doesn’t create paths/to/sub/targets and so the vagrant plugin
code, when performing an rsync, is responsible for creating
intermediate directories in guest paths if there are any. Furthermore,
the mkdir utility in Windows is not another Cygwin utility like rsync
but the routine mkdir of Windows command.com. Therefore, while rsync
needs the /cygwin paths, mkdir uses the Windows paths. Later, the
chef_solo.rp provisioner running within the guest will expect to find
Windows-style paths in its solo.rb configuration file. Due to all this,
vagrant has to keep track of both the original, possibly dirty Windows
guest path and the cygwin-scrubbed guest path.

### DIFF
--- a/plugins/guests/windows/cap/rsync.rb
+++ b/plugins/guests/windows/cap/rsync.rb
@@ -2,8 +2,17 @@ module VagrantPlugins
   module GuestWindows
     module Cap
       class RSync
+        def self.rsync_scrub_guestpath( machine, opts )
+          # Windows guests most often use cygwin-dependent rsync utilities
+          # that expect "/cygdrive/c" instead of "c:" as the path prefix
+          # some vagrant code may pass guest paths with drive-lettered paths here
+          opts[:guestpath].gsub( /^([a-zA-Z]):/, '/cygdrive/\1' )
+        end
+
         def self.rsync_pre(machine, opts)
           machine.communicate.tap do |comm|
+            # rsync does not construct any gaps in the path to the target directory
+            # make sure that all subdirectories are created
             comm.execute("mkdir '#{opts[:guestpath]}'")
           end
         end
--- a/plugins/guests/windows/plugin.rb
+++ b/plugins/guests/windows/plugin.rb
@@ -64,6 +64,11 @@ module VagrantPlugins
         Cap::MountSharedFolder
       end
 
+      guest_capability(:windows, :rsync_scrub_guestpath) do
+        require_relative "cap/rsync"
+        Cap::RSync
+      end
+
       guest_capability(:windows, :rsync_pre) do
         require_relative "cap/rsync"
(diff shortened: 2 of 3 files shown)


Y4Rv1K:
Any idea when it will be merged to master?

lonniev:
I haven’t heard from Seth or Mitchell so, no, I can’t advise. It’s an easy patch to try out at home on your own meanwhile. If you hit a case I didn’t test that breaks, I’d add that to the commit set.

(I realize I didn’t test a path like c:\grandparent\parent\child\target. I only needed to get c:\vagrant-chef\digesthash to work. I really need one more subdir to assure that the mkdir call is doing what I expect.)

-- 
Lonnie VanZandt
303-900-3048

On 20 August 2015 at 15:15:42, Yaroslav Ravlinko (notifications@github.com) wrote:

Any idea when it will be merged to master?

Y4Rv1K:
@lonniev `digesthash` that is biggest problem right now. I saw how @ramarnat [worked around](https://github.com/mitchellh/vagrant/issues/5255) this issue in case of AWS using `USER_DATA`. 
So only thing that required it's create this folder before vagrant start sync provision folders. And I see no obvious way to do that without modifying vagrant.

lonniev:
One could use the same Digest call to get the digest value that is used in the directory path. With that, one could ask the Vagrantfile to create the missing directories with a shell provisioner block before the chef-solo provisioner block. This is ugly though.

Better approach is to work out the details inside the capability provider and its helpers. My code may not be the best way but at least it keeps the Vagrantfile clean and uses no custom configs or hardcoded path assumptions.

lonniev:
This is from chef_solo.rb which computes the future name of the directory used to receive the rsync-ed chef directories. You could use that same Digest method.
              
if File.exist?(local_path)
                # Path exists on the host, setup the remote path. We use
                # the MD5 of the local path so that it is predictable.
                key         = Digest::MD5.hexdigest(local_path)
                remote_path = "#{guest_provisioning_path}/#{key}"

Y4Rv1K:
> shell provisioner block before the chef-solo provisioner block. 

Looks like it doesn't work because creating provisioning folder happening before `shell` provisioner can execute.

mitchellh:
This is a perfect implementation. Thanks so much!

PR #6166, opened 2015-08-19
Change docs for checkpoint environment variable

cbednarski:
Now reflects the not-vagrant-specific version, `CHECKPOINT_DISABLE`


sethvargo:
Ah. I see the confusion.

We try really hard to keep all the environment variables that relate to Vagrant to start with `VAGRANT_`. Could we keep it as-is, but then add text that is like "you can also use CHECKPOINT_DISABLE"?

cbednarski:
Yes I can make that change.

cbednarski:
@sethvargo Pinging on this.

cbednarski:
:heart: Thanks!

PR #6172, opened 2015-08-21
fix network detection on pld-linux

glensc:
pld linux uses redhat as base, but lacks :flavour
this will add it


mitchellh:
LGTM. Thanks!

PR #6173, opened 2015-08-21
add better chef integration for pld

glensc:
pld packages chef as chef os package, update installation and path
however it packages knife tool as separate package
so check for chef-client, not knife

### DIFF
--- a/plugins/provisioners/chef/cap/pld/chef_install.rb
+++ b/plugins/provisioners/chef/cap/pld/chef_install.rb
@@ -1,0 +1,18 @@
+require_relative "../../omnibus"
+
+module VagrantPlugins
+  module Chef
+    module Cap
+      module PLD
+        module ChefInstall
+          def self.chef_install(machine, version, prerelease, download_path)
+            machine.communicate.sudo("poldek --up -u chef --noask")
+
+            command = Omnibus.build_command(version, prerelease, download_path)
+            machine.communicate.sudo(command)
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


sethvargo:
Hi @glensc 

Thank you for the pull request. I am very against installing Chef from anything other than the omnibus installer. The omnibus installer allows users to specify different versions of Chef to install and puts them in a well-known, expected place.

glensc:
@sethvargo chef does not support this platform. with this PR at least can run chef version that is supported by pld-linux itself.

```
21:07:44 root[load: 1.09]@pld32 ~# wget http://opscode.com/chef/install.sh
21:07:41 root[load: 1.18]@pld32 ~# bash install.sh 
Unable to determine platform version!
Version: 

Please file a Bug Report at https://github.com/opscode/opscode-omnitruck/issues/new
Alternatively, feel free to open a Support Ticket at https://www.getchef.com/support/tickets
More Chef support resources can be found at https://www.getchef.com/support

Please include as many details about the problem as possible i.e., how to reproduce
the problem (if possible), type of the Operating System and its version, etc.,
and any other relevant details that might help us with troubleshooting.
```

sethvargo:
Hi @glensc 

I do not think Vagrant should support ways of installing Chef that Chef themselves do not support. I really want to refrain from having a bespoke implementation because it means getting upstream support is much more difficult. I understand that pld-linux is not supported by the omnibus installer, but the omnibus installer is the only recommended way for installing Chef. Please raise an issue with the Chef team about adding support for your platform.

In the meantime, you can set `chef.install = false` in your Vagrantfile and add a shell provisioner _before_ the Chef provisioner that installs Chef in this custom manner :smile:.

PR #6176, opened 2015-08-22
Update to bundler 1.10.6 which fixes parallel gem installations

tknerr:
Parallel installation with (e.g. `bundle install --jobs=<n>`) got broken with earlier 1.10.x versions.

This was fixed again in 1.10.6, see bundler/bundler#3799

See also chef/omnibus-chef#464

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
   s.rubyforge_project         = "vagrant"
 
-  s.add_dependency "bundler", ">= 1.5.2", "<= 1.10.5"
+  s.add_dependency "bundler", ">= 1.5.2", "<= 1.10.6"
   s.add_dependency "childprocess", "~> 0.5.0"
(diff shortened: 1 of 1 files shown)


sethvargo:
Duplicate of https://github.com/mitchellh/vagrant/pull/6111. Please see my explanation on that PR please.

tknerr:
@sethvargo your explanation that you don't want the bundler dependency unbounded makes pretty much sense to me :+1: 

However, this specific PR is about a patch level upgrade from 1.10.5 to 1.10.6 (which fixes an imho critical dependency resolution bug), not about removing the upper bound.

Would you mind a quick check on the [diff between 1.10.5 and 1.10.6](https://github.com/bundler/bundler/compare/v1.10.5...v1.10.6) to see if the internal APIs you are using are affected?

All the earlier 1.10.x bundler versions are really unusable as soon as you use parallel installation via `--jobs=<n>` :disappointed: 

I would really be very grateful if you could spend a second :eyes: on it. Would you mind?

tknerr:
See also https://github.com/test-kitchen/kitchen-vagrant/issues/190

PR #6190, opened 2015-08-25
Fixed link in destroy command documentation

gajdaw:
The alert box within destroy command documentation contains markdown-style link.
The link should be written in raw HTML within alert box.


sethvargo:
Thanks!

PR #6193, opened 2015-08-25
Faster deploys

sethvargo:
/cc @mitchellh @phinze

### DIFF
--- a/scripts/website_push_docs.sh
+++ b/scripts/website_push_docs.sh
@@ -1,17 +1,38 @@
 #!/bin/bash
 
+# Set the tmpdir
+if [ -z "$TMPDIR" ]; then
+  TMPDIR="/tmp"
+fi
+
+# Create a temporary build dir and make sure we clean it up. For
+# debugging, comment out the trap line.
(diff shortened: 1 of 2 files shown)


scripts/website_push_docs.sh
```diff
@@ -1,17 +1,38 @@
...
+# Set the tmpdir
+if [ -z "$TMPDIR" ]; then
+  TMPDIR="/tmp"
+fi
+
+# Set the path
+DEPLOY="$TMPDIR/vagrant-docs"
+rm -rf $DEPLOY
+mkdir -p $DEPLOY
+
```
phinze: I learned this trick from @ryanuber:

```bash
# Create a temporary build dir and make sure we clean it up. For
# debugging, comment out the trap line.
BUILDDIR=`mktemp -d /tmp/vagrant-www-XXXXXX`
trap "rm -rf $BUILDDIR" INT TERM EXIT
```

phinze:
Nice, LGTM. Gonna steal this for all the www stuff.

PR #6195, opened 2015-08-25
Check if Schedule.Service com object is available

ArloL:
Schedule.Service is not available for e.g. Windows XP. This PR catches that case and simply executes the script immediately since Windows XP does not require elevation.

I have tested the script in Windows XP, Vista and 8.1.


mitchellh:
When everything ran with admin privs we lived in simpler times. </tongue-in-cheek>

Thanks!

tmoreno:
Hi!

I have the same issue using 2.1.0 version.

It seems that the bug was fixed and merged years ago but was introduced again in this commit:

725824e1dd3c2225a6274df074142621751bfb14

when winrm communicator was changed.

Could you check if I'm right?

Thanks

PR #5765, opened 2015-05-29
Store the first SSH private key in generated Ansible inventory

gildegoma:
Pending before merge:
* [x] Confirmation from @mitchellh or @sethvargo that we add this feature, even if "Ansible Parallelism" is not officially supported (but so awesome and popular ;-)
* [x] Generated-Inventory QA Feedbacks (thanks to @conorsch)
* [x] Static-Inventory QA Feedbacks
* [ ] Documentation Updates (*coming soon*)

PR #6210, opened 2015-08-29
Add ssh-user to ansible inventory

yanosz:
The configured ssh user is not added to ansible's inventory file, causing connection errors late one.
While vagrant connects as "vagrant", ansible connects as the current one.

Example:
```
GNS-3node-2 ansible_ssh_host=127.0.0.1 ansible_ssh_port=2207 ansible_ssh_private_key_file=/home/jan/vagrant/.vagrant/machines/GNS-3node-2/virtualbox/private_key 
```
Error:
```
 ansible -vvv -i inventory GNS-3node-2 -m ping
<127.0.0.1> ESTABLISH CONNECTION FOR USER: jan
<127.0.0.1> REMOTE_MODULE ping
<127.0.0.1> EXEC ['ssh', '-C', '-tt', '-q', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=60s', '-o', 'ControlPath=/home/jan/.ansible/cp/ansible-ssh-%h-%p-%r', '-o', 'Port=2207', '-o', 'IdentityFile="/home/jan/vagrant/.vagrant/machines/GNS-3node-2/virtualbox/private_key"', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'ConnectTimeout=10', '127.0.0.1', "/bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-tmp-1440842011.53-32279685471205 && chmod a+rx $HOME/.ansible/tmp/ansible-tmp-1440842011.53-32279685471205 && echo $HOME/.ansible/tmp/ansible-tmp-1440842011.53-32279685471205'"]
GNS-3node-2 | FAILED => SSH encountered an unknown error during the connection. We recommend you re-run the command using -vvvv, which will enable SSH debugging output to help diagnose the issue

```

This patches changes the way the inventory is generated by adding the configured user
Example:
```
GNS-3node-2 ansible_ssh_host=127.0.0.1 ansible_ssh_port=2207 ansible_ssh_private_key_file=/home/jan/vagrant/.vagrant/machines/GNS-3node-2/virtualbox/private_key ansible_ssh_user=vagrant
```

### DIFF
--- a/plugins/provisioners/ansible/provisioner.rb
+++ b/plugins/provisioners/ansible/provisioner.rb
@@ -127,7 +127,7 @@ module VagrantPlugins
             m = @machine.env.machine(*am)
             m_ssh_info = m.ssh_info
             if !m_ssh_info.nil?
-              inventory += "#{m.name} ansible_ssh_host=#{m_ssh_info[:host]} ansible_ssh_port=#{m_ssh_info[:port]} ansible_ssh_private_key_file=#{m_ssh_info[:private_key_path][0]}\n"
+              inventory += "#{m.name} ansible_ssh_host=#{m_ssh_info[:host]} ansible_ssh_port=#{m_ssh_info[:port]} ansible_ssh_private_key_file=#{m_ssh_info[:private_key_path][0]} ansible_ssh_user=#{m_ssh_info[:username]}\n"
               inventory_machines[m.name] = m
             else
               @logger.error("Auto-generated inventory: Impossible to get SSH information for machine '#{m.name} (#{m.provider_name})'. This machine should be recreated.")
--- a/test/unit/plugins/provisioners/ansible/provisioner_test.rb
+++ b/test/unit/plugins/provisioners/ansible/provisioner_test.rb
@@ -168,7 +168,7 @@ VF
         expect(config.inventory_path).to be_nil
         expect(File.exists?(generated_inventory_file)).to be_true
         inventory_content = File.read(generated_inventory_file)
-        expect(inventory_content).to include("#{machine.name} ansible_ssh_host=#{machine.ssh_info[:host]} ansible_ssh_port=#{machine.ssh_info[:port]} ansible_ssh_private_key_file=#{machine.ssh_info[:private_key_path][0]}\n")
+        expect(inventory_content).to include("#{machine.name} ansible_ssh_host=#{machine.ssh_info[:host]} ansible_ssh_port=#{machine.ssh_info[:port]} ansible_ssh_private_key_file=#{machine.ssh_info[:private_key_path][0]} ansible_ssh_user=#{machine.ssh_info[:username]}\n")
         expect(inventory_content).to include("# MISSING: '#{iso_env.machine_names[1]}' machine was probably removed without using Vagrant. This machine should be recreated.\n")
       }
     end


gildegoma:
Hi @yanosz. Thanks for submitting this PR. Without any playbook and `Vagrantfile` examples, I am not sure to understand what do you mean by _configured ssh user_. Are you referring to the Vagrant SSH user, or to your (custom) Ansible settings?

In your example above, you're executing an `ansible` command, not `vagrant provision` (or alike). The Vagrant Ansible provisioner passes the Vagrant SSH user via the command line parameter `--user` (you can get all the details in the vagrant log console by enabling the `ansible.verbose` option in your `Vagranfile`.

So far, I was not in favour to set `ansible_ssh_user` in the generated inventory because this [inventory parameter has the highest precedence](https://github.com/mitchellh/vagrant/pull/5044#issuecomment-69715118). Even though it would work for many users, it would be fully rigid (e.g. impossible to rely on `remote_user` parameter in a Play or a Task).
This would also not mean that the `--user` parameter could be removed, since this would be a breaking change for users that rely on a static inventory.

If you disagree with the arguments above, could you please provide a detailed scenario with attached `Vagrantfile` and Ansible playbook file, so I can better figure out your use case. I also would need to know which version of Ansible you are using (in case the parameter handling is changing, e.g. in v2). Thanks!

yanosz:
Thanks for your feedback. By configured ssh use I refer to the ssh-user configured in ansible. I'm not able to provide a playbook, yet, cause the set up used for developing a playbook refering to multiple host - thus the playbook is empty atm.

On the one hand,I don't see a problem with including `ansible_ssh_user` in the generated inventory. First, this inventory is generated on the fly, only. Impacts on existing installations are rather minimal.

On the other hand, setting a ssh-key without a corresponding username is not somewhat broken. The inventory cannot be used without further modification.

IMHO it's best to generate inventory files that can be used - by default. If settings this parameters will introduce certain problems, we can introduce configuration options.

Greetz, Jan

gildegoma:
@yanosz thanks a lot for all these inputs.

> setting a ssh-key without a corresponding username is not somewhat broken.

Yup. For the records: the `ansible_ssh_private_key_file` was introduced in the generated inventory quite recently (in Vagrant 1.7.3 via #5765) in order to resolve the issues related to the auto-generated machine key pairs introduced by Vagrant 1.7. 

> having an inventory that "just works" is what people need out there

Sure that would be nice. But it should not add (too much) extra complexity to the provisioner codebase, given that this change is motivated by "external" use cases (there is no requirement for Vagrant scope)

> If setting this parameters will introduce certain problems, we can introduce configuration options.

So Jan, what do you think if we try to go forward with the following approach:

* Add a new boolean config option (e.g. `set_ssh_user_in_generated_inventory` :thought_balloon: _hum, naming, hum..._), which will be set to `true` by default.
* The `ansible_ssh_user` setting will then be added to the generated inventory, but only if the above option is true.

gildegoma:
In the meantime, `ansible -u vagrant -vvv -i inventory GNS-3node-2 -m ping` will do the trick ;-)

yanosz:
Hello,

thanks for getting back to me.  I like your idea for the configuration option.
For the name,  I'd like to propose ansible.ssh_user for the ansible environment. This defaults to the ssh-user, can be set to false for not having a user at all and can be used to specifiy any user. This seems to short and consistens.

For the trick: I changed the inventory :-)

yanosz:
@gildegoma what do you think? Can we go one that way?

gildegoma:
@yanosz thanks for the follow up, and sorry for the delay (haven't time to sum it up before this week-end).

I don't think that your `ansible.ssh_user` proposal really address this issue (Note that the ssh user can be controlled with [`config.ssh.username`](https://docs.vagrantup.com/v2/vagrantfile/ssh_settings.html), or via an ansible extra variables if it differs from Vagrant default ssh user).

By introducing `ansible_ssh_user` in the inventory, we potentially can break projects where **several remote users** are used during an `ansible-playbook` run ([`remote_user` on Ansible Task or Play](http://docs.ansible.com/ansible/playbooks_intro.html#hosts-and-users) will be shadowed when Vagrant will set the `ansible_ssh_user` variable). That's in fact the unique tricky point that force us to introduce this option (and extend this issue thread ;-).

I stress that the "custom/static inventory" alternative should also be taken in account, otherwise it will be confusing if the remote ssh user can only be enforced when the generated inventory is in use.

Then, I actually favour the following approach:

* Introduction of a new boolean option `force_remote_user`, by default set to `true`.
* Behaviours with a *generated inventory*:
  * The `ansible_ssh_user` setting is added to the generated inventory when the above option is true.
  * The `--user=...` command line argument is passed when the above option is false.
* Behaviours with a *custom inventory*:
  * The `--extra-vars=ansible_ssh_user=...`) command line argument is passed when the above option is true
  * The `--user=...` command line argument is passed when the above option is false.

This way, we'll
 * by default, override the remote users explicitly defined in Ansible plays or tasks,
 * get a generated inventory that just works out of Vagrant usage context,
 * avoid behavioural discrepancies between custom inventory and generated inventory,
 * keep usage rather simple and still fully flexible 
 * hopefully don't taint too much the codebase
 
@yanosz do you agree with this? Are you willing to implement it?

yanosz:
Thanks for getting back to me. Refering to the comments:

`ansible.ssh_user` is very different from config.ssh_user cause it sets the user in the generated inventory. This cannot be accomplished by using `config.ssh.username`. Please note, that `config.ssh.username` has to be vagrant for using prebuilt boxes (usually), while ansible's ssh-user defaults to ansible. This is a totally different story. 

Of course, it is possble to specifiy additionally settings (`-u vagrant`) for example, but I don't think, that this "just works". By that, usability suffers. Again, the usuabiltiy argument holds for an inconsistent inventory files (private key is specified, but ssh is not).
For usability reasons I vote for an inventory that "just works". 

To be honest, I'm not happy with your idea for introducing a boolean flag. Your proposal looks complicated and it took me a while to figure out.
I don't like the idea of having a different behaviour for static and generated inventories and I don't like the fact, that it is not descriptive - meaning: You want and look for a way to set `ansible_ssh_user` in the generated inventory but all you get is a reather complicated boolean having this as a side effect. From a usability point of view this is worse than before: If you do this, people will try to understand that option (at least). If not people see, that this is just not "there".

Looking at my proposal, setting `config.ansible_ssh_user` to nil will not change anything and thus be backwards compatible. But -as said - this is not meant to be the default.

I respect your work in this project and it looks, that we cannot reach a consensus here. While I'm heading for "just works", You're concerned, that changing defaults will break things, 'cause they might be forgotten.

I think I'll maintain my own fork, here.

gildegoma:
@yanosz I'm quite sorry that we couldn't reach a consensus so far. Anyway, thanks a lot for your inputs, which really convinced me that we can certainly improve something there (although it is more a *nice to have* priority in my opinion). I'll give me some time to rethink all these ideas, and - if you want - I'll keep you informed about the next actions that might be taken.
Thanks again for your time and interest on this project!

gildegoma:
FYI @yanosz: as of Vagrant 1.8, the `ansible_ssh_user` setting will be set into the auto-generated inventory **by default**.

PR #6213, opened 2015-08-31
Winrm execution timeout

jrob:
Please see mitchellh/vagrant#5506

Adds a configurable value for WinRm and the elevated permission shell ExecutionTimeLimit.

Ex: config.winrm.execution_time_limit = "P1D"

Unit tests are passing, please consider for merge.

### DIFF
--- a/plugins/communicators/winrm/communicator.rb
+++ b/plugins/communicators/winrm/communicator.rb
@@ -210,7 +210,8 @@ module VagrantPlugins
 
         "powershell -executionpolicy bypass -file \"#{guest_script_path}\" " +
           "-username \"#{shell.username}\" -password \"#{shell.password}\" " +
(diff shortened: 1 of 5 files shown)


jrob:
I don't have a full environment setup to do vagrant dev right now. I may be able to find time to work on it over the next week.

I believe that the config value that I have added is not being initialized in the unit test. Also, it's not at the same level as the other variables for that function so it may need to be changed.

Any feedback is welcome.

jrob:
Fixed unit test.
Cleaned up commits.

Please consider for merge.

jrob:
Any feedback on this pull request?

matthewcodes:
This looks good to me

mitchellh:
Looks fantastic. Thank you!

PR #6249, opened 2015-09-08
samphippen/bugfix/5605

### LINKED ISSUE #5605: Can't create vagrant vm " inspected result must be ASCII only"
I tried vagrant up with :

- ruby version :  ruby-1.9.3-p547
- vagrant : 1.7.2
- virtualbox : 4.3.26r98988

```text
$ vagrant up
/opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/config/loader.rb:44:in `inspect': inspected result must be ASCII only or use the same encoding with default external (Encoding::CompatibilityError)
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/config/loader.rb:44:in `inspect'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/config/loader.rb:44:in `set'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/vagrantfile.rb:145:in `machine_config'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:325:in `default_provider'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/plugin/v2/command.rb:165:in `block in with_target_vms'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/plugin/v2/command.rb:192:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/plugin/v2/command.rb:192:in `block in with_target_vms'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/plugin/v2/command.rb:174:in `each'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/plugin/v2/command.rb:174:in `with_target_vms'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/plugins/commands/up/command.rb:74:in `block in execute'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:277:in `block (2 levels) in batch'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:275:in `tap'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:275:in `block in batch'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:274:in `synchronize'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:274:in `batch'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/plugins/commands/up/command.rb:58:in `execute'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/cli.rb:42:in `execute'
	from /opt/vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:301:in `cli'
	from /opt/vagrant/bin/../embedded/gems/gems/vagrant-1.7.2/bin/vagrant:174:in `<main>'
```

I change ruby version to 2.1.0 but same result.

samphippen:
Here we implement a naive solution to #5605 which catches the case that
a provided source contains an object which cannot be inspected, because
an object contained within in has an #inspect string that returns a
string that is incompatible with the encoding in
`Encoding.default_external` or a string which cannot be downcast to
7-bit ascii.

The Ruby VM implementation of "#inspect" implements this checking on
these lines of code: http://git.io/vZYNS. A Ruby level override of
this method does not cause this problem. For example:

```ruby
class Foo
  def inspect
    "😍".encode("UTF-16LE")
  end
```

will not cause the problem, because that's a Ruby implementation and the
VM's checks don't occur.

However, if we have an Object which **does** use the VM implementation
of inspect, that contains an object that has an inspect string which
returns non-ascii, we encounter the bug. For example:

```ruby
class Bar
  def inspect
    "😍".encode("UTF-16LE")
  end
end

class Foo
  def initialize
     @bar = Bar.new
  end
end

Foo.new.inspect
```

Will cause the issue.

The solution this patch provides basically catches the encoding error
and inserts a string which attempts to help the user work out which
object was provided without blowing up. Most likely, this was caused
by a user having a weird encoding coming out of one of the sources
passed in, but without a full repro case, it's not clear whether a patch
should be applied to a different object in the system.

Closes #5605.

### DIFF
--- a/test/unit/vagrant/config/loader_test.rb
+++ b/test/unit/vagrant/config/loader_test.rb
@@ -61,8 +61,11 @@ describe Vagrant::Config::Loader do
           end
         end.new(klass_with_bad_inspect_string)
       }
+
       it "does not raise the ascii encoding exception" do
-        instance.set(:arbitrary, test_source)
+        expect {
+          instance.set(:arbitrary, test_source)
+        }.to raise_error(ArgumentError, /Unknown configuration source/)
       end
     end
   end
--- a/test/unit/vagrant/config/loader_test.rb
+++ b/test/unit/vagrant/config/loader_test.rb
@@ -38,6 +38,35 @@ describe Vagrant::Config::Loader do
 
   let(:instance) { described_class.new(versions, version_order) }
 
+  describe "#set" do
+    context "with an object that cannot be inspected" do
+
+      # This represents the euro symbol in UTF-16LE. pack("c*") returns an ASCII
+      # string and so we have to force the encoding
+      UTF_16LE_STRING_THAT_CANNOT_BE_DOWNCAST_TO_ASCII = [0x20, 0xAC].pack("c*").force_encoding("UTF-16LE")
+
+
+      let(:klass_with_bad_inspect_string) do
+        Class.new do
+          def inspect
+            UTF_16LE_STRING_THAT_CANNOT_BE_DOWNCAST_TO_ASCII
+          end
+        end
+      end
+
+      let(:test_source) {
+        Class.new do
+          def initialize(collaborator)
+            @foo = collaborator.new
+          end
+        end.new(klass_with_bad_inspect_string)
+      }
+      it "does not raise the ascii encoding exception" do
+        instance.set(:arbitrary, test_source)
+      end
+    end
+  end
+
   describe "basic loading" do
     it "should ignore non-existent load order keys" do
       instance.load([:foo])
--- a/lib/vagrant/config/loader.rb
+++ b/lib/vagrant/config/loader.rb
@@ -41,10 +41,20 @@ module Vagrant
       # `set` multiple times with the same name will override any previously
       # set values. In this way, the last set data for a given name wins.
       def set(name, sources)
-        @logger.info("Set #{name.inspect} = #{sources.inspect}")
-
         # Sources should be an array
         sources = [sources] if !sources.kind_of?(Array)
+
+        reliably_inspected_sources = sources.reduce({}) { |accum, source|
+          begin
+            accum[source] = source.inspect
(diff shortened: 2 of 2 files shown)


samphippen:
@bar so sorry. `@bar` refers to the instance variable bar in ruby, which I explained in my commit message and GitHub pinged you. That's a fun expected interaction. Hope you can understand :)

test/unit/vagrant/config/loader_test.rb
```diff
@@ -38,6 +38,38 @@ def self.merge(old, new)
 
   let(:instance) { described_class.new(versions, version_order) }
 
+  describe "#set" do
+    context "with an object that cannot be inspected" do
+
+      # This represents the euro symbol in UTF-16LE. pack("c*") returns an ASCII
+      # string and so we have to force the encoding
+      UTF_16LE_STRING_THAT_CANNOT_BE_DOWNCAST_TO_ASCII = [0x20, 0xAC].pack("c*").force_encoding("UTF-16LE")
```
mitchellh: Love this!

mitchellh:
This looks great! Thanks. :)

bar:
@samphippen I know... I'll have to suggest using a different language from now on...

just kidding :p but maybe you can escape the at handle first, thanks!

PR #6254, opened 2015-09-10
Improved NFS configuration for SUSE

krig:
Use the /sbin/service script for both status and start commands.

Also adds a SUSE variant of the contrib/sudoers script.


krig:
Pushed an updated sudoers script which fixes incorrect path to `sed` on SUSE systems.

krig:
I have updated this pull request and added updated sudoers scripts for the other Linux flavors as well: #5773 broke the example sudoers scripts. This resolves that problem. One concern is that the `cp` `Cmnd_Alias` is perhaps a bit too broad.

krig:
@mitchellh Any chance of merging this?

PR #6259, opened 2015-09-11
Fix verify_binary for Puppet for Windows guests.

philipwigg:
Fixes a bug in the fix for https://github.com/mitchellh/vagrant/issues/5943 whereby 'which' is mistakenly used on Windows instead of 'where' to locate the Puppet binary.


philipwigg:
I should say that before this change I was getting..

'The puppet binary appears not to be in the PATH of the guest. This could be because the PATH is not properly setup or perhaps Puppet is not installed on this guest. Puppet provisioning can not continue without Puppet properly installed'

with a Windows 2012r2 guest on Mac when using the Puppet provisioner. Also relevant for https://github.com/mitchellh/vagrant/issues/5615

Issue #6065, opened 2015-07-30
invalid byte sequence in US-ASCII (ArgumentError)


realyst2k:
I deployed a vagrant based development environment to a team of developers, all running varying versions of OSX, however, a number of them ran into this error while attempting to bootstrap their VM via vagrant up

    ==> default: Running cleanup tasks for 'shell' provisioner...
    /opt/vagrant/embedded/gems/gems/vagrant-1.7.3/plugins/provisioners/shell/provisioner.rb:185:in `gsub!': invalid byte sequence in US-ASCII (ArgumentError)

Now all developers involved used a language other than English at some point, a few were situated in Europe, some formerly in Brazil, others in Canada.  But LANG was set to en_CA.UTF-8 universally across the board on both the VM and the native OSX.  Near as I can tell, none of the paths that have typically presented this problem such as $VAGRANT_HOME nor any of the shared directories, had any special characters in them.  Most of the OSX builds affected were installed in English and were bought in Canada.

Sadly, I was unable to isolate from these developers what characters, exactly, it was that got Ruby upset.  

I was, however, able to confirm it likely wasn't related to 
https://github.com/mitchellh/vagrant/issues/4966
or
https://github.com/mitchellh/vagrant/issues/2113

As I had found these in the process of attempting to troubleshoot the issue and had eliminated them as the cause.  Also, everyone involved had an up-to-date copy of Vagrant which ought to have contained any fixes/merges in those.

I had eventually found a solution on a Japanese blog wherein the OS character set in their Windows build was causing similar headaches.  

http://dqn.sakusakutto.jp/2014/02/vagrant_up_gsub_invalid_byte_se.html

Adding this to the Vagrantfile appears to have fixed all the issues the team had run into:

    Encoding.default_external = 'UTF-8'

So beyond chronicling this into the issue tracker should someone else run into this same issue, would it perhaps also make sense to provide UTF-8 support as default or would this introduce more issues than it would solve?  If the latter, could this be added to the official documentation as a potential issue people may run across?

Thanks!

sethvargo:
Hi @realyst2k 

We are tracking all the encoding issues here: https://github.com/mitchellh/vagrant/issues/5896. I have linked yours for reference. Thanks!

samphippen:
@realyst2k if you're still around might you be able to provide a vagrantfile

and the output of this script one one of the machines that encountered the issue for us to take a look at it?

```
[...]

PR #6260, opened 2015-09-12
Samphippen/bugfix/6065

samphippen:
The problem demonstrated in #6065 is that a string has incorrectly been
encoded with US-ASCII even though it contains invalid US-ASCII byte
sequences (any byte with the most significant bit on is invalid in the
US-ASCII encoding).

The thing about doing newline normalization is that it is not actually
sensitive to the presence of US-ASCII byte sequenzes. Additionally, it
is very unlikely that a user will ever be using an encoding where \r\n
is not encoded the same as it would be in ASCII.

This patch first tries the existing method of normalizing the newlines
in the provided script file, if that fails for any reason it force
encodes the string to ASCII-8BIT (which allows the most significant bit
to be on in any individual byte) and then performs the substitution in
that byte space.


mitchellh:
Perfect. Thanks!

Issue #6231, opened 2015-09-03
Vagrant not deleting temp files


yr0:
I'm using Vagrant with Docker as provider.

Whenever I launch vagrant with any command, the directory gets populated with four files, prefixed with 'vagrant', e.g. 'vagrant20150902-10813-1p2j0fh', and a directory, prefixed with 'd', e.g. 'd20150902-10813-14u60qc'.
Neither of those gets automatically deleted.

I traced the problem to `Vagrant::Bundler#deinit`, that is executed `at_exit`:

    def deinit
       File.unlink(ENV["BUNDLE_APP_CONFIG"]) rescue nil
       File.unlink(ENV["BUNDLE_CONFIG"]) rescue nil
       File.unlink(ENV["GEMFILE"]) rescue nil
    end

`ENV["BUNDLE_APP_CONFIG"]` is always a directory:

    # contents of Vagrant::Bundler#init!
    @appconfigpath = Dir.mktmpdir
    # ...
    ENV["BUNDLE_APP_CONFIG"] = @appconfigpath

That is why first unlink always fails, leaving temporary dir undeleted.
`ENV["GEMFILE"]` is usually nil (perhaps, `ENV["BUNDLE_GEMFILE"]` was meant to be deleted instead).

Anyway, even if `Vagrant::Bundler#deinit` worked as expected, it would still leave three temp files in directory, so I guess, there is a problem with the configuration of OS. 

I have installed Vagrant from .deb Linux x64 package, removed it with `apt-get remove --purge vagrant` and reinstalled again with problem persisting. The OS is Ubuntu 14.04.3 LTS. Linux pc 3.13.0-62-generic # 102-Ubuntu SMP Tue Aug 11 14:29:36 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
 
I've also tried asking the community at [stackoverflow](http://stackoverflow.com/questions/32352989/vagrant-creates-numerous-trash-config-files-and-directories).

Any help would be much appreciated!

remram44:
I'd very much like to see this fixed, I get thousands of these `d2015*` folders. Also a `vagrant-remram` full of `vagrant2015*` files.

This is really annoying and really not the mark of professional software.

sni:
I created a pull request, hopefully this gets fixed soon.

evaryont:
Since #3469 was re-opened, I guess this is now a dup?

Issue #6116, opened 2015-08-09
vagrant trashes /tmp with a lot of files


pitchforks:
I am noticing that after I end working with vagrant - the `/tmp` directory is in a scary state - it has lots of `vagrant*` files which I have to manually clean up. It is not a matter of disk space, I'm just directly using `/tmp` for temporary tasks very often and after vagrant it becomes harder to find my files and directories there, because the size of the output that `ls` command produces grows up considerably.

Perhaps vagrant could use a subdirectory of `/tmp` for creating the temporary files it needs, to avoid cluttering `/tmp`?

remram44:
Duplicate of #6231?

mitchellh:
Agreed with @remram44, dup.

Issue #4103, opened 2014-06-25
Generating too many tempfiles


rtkkdeng:
It seems that from running any vagrant command, my $TMPDIR gets overflowed with  vagrantYYYYMMDD-hmmss-XXXXXXX files and dYYYYMMDD-hmmss-XXXXXXX directories.

From vagrant/lib/bundler.rb, it creates these files every time I run a vagrant command and more depending on what vagrant command I run.
ie. Running 'vagrant --version' creates several of these tempfiles after each run.

Is there any reason why vagrant doesn't clean up these files after?

caljess599:
I am experiencing this as well.

varac:
same here, ubuntu 14.10, vagrant 1.7.1 debian package.

```
--- /tmp » ls /tmp/vagrant* | wc -l
1966

```

varac:
upgraded to 1.7.2, removed all leftovers from /tmp, run "vagrant", and these files were put in /tmp:
```
--- /tmp » ls -al vagrant* d2015* -d
drwx------ 2 varac varac 4096 Feb  7 18:42 d20150207-30612-401o24
-rw------- 1 varac varac    0 Feb  7 18:42 vagrant20150207-30612-1htp2k3
-rw------- 1 varac varac    0 Feb  7 18:42 vagrant20150207-30612-owies3
-rw-rw-r-- 1 varac varac  289 Feb  7 18:42 vagrant20150207-30612-owies32
-rw-rw-r-- 1 varac varac 4324 Feb  7 18:42 vagrant20150207-30612-owies32.lock

```

adamcstephens:
@mitchellh Can we reopen this?

adamcstephens:
@mitchellh Can we reopen this?

mitchellh:
@adamcstephens If I remember correctly, these final four weren't worth the effort to remove, hence the "much better" above and not "completely gone." If someone were to open a PR to fix these I'd take a look but I don't think I'd consider 4 leftover files going into the temporary directory a bug, since that directory is made for temporary files anyways that are safe enough to just delete via any other process.

But like I said, if someone pinpoints it and fixes it, I'd gladly merge it, but I distinctly remember this issue making many more files before, doing my best to do cleanups, getting to these last 4 and realizing they'd be really tricky, and just calling it pragmatically good enough.

varac:
[...]

PR #6301, opened 2015-09-22
fix typo when removing tmp files

sni:
This patch starts fixing #6116 and #6231 to remove at least the /tmp/d... folders and the temporary gemfile.

### DIFF
--- a/lib/vagrant/bundler.rb
+++ b/lib/vagrant/bundler.rb
@@ -85,6 +85,7 @@ module Vagrant
       FileUtils.rm_rf(ENV["BUNDLE_APP_CONFIG"]) rescue nil
       File.unlink(ENV["BUNDLE_CONFIG"]) rescue nil
       File.unlink(ENV["BUNDLE_GEMFILE"]) rescue nil
+      File.unlink(ENV["BUNDLE_GEMFILE"]+".lock") rescue nil
     end
 
     # Installs the list of plugins.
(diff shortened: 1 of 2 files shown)


remram44:
:heart_eyes:

sni:
Ok, this are all i found for now. vagrant now runs cleanly (i checked status, ssh, up and halt commands)

avdd:
Finally!   #4103  :+1:

avdd:
ping @mitchellh @sethvargo

remram44:
ping @mitchellh

marek-mazur:
Any plans merge this fix? This bug is really annoying... ;)

evaryont:
The unlink after creation is a good method to ensure the file is cleaned up by the OS once vagrant has exited, but I think it won't work for Windows users.

Is it necessary to call `deinit` in the ssh command? [The `at_exit` block in bin/vagrant](https://github.com/mitchellh/vagrant/blob/master/bin/vagrant#L111) is always ran.

avdd:
For persons in shell-y environments, try this:
```bash
_vagrant() {
    mkdir -p /tmp/vagrant
    TMPDIR=/tmp/vagrant /opt/vagrant/bin/vagrant "$@"
    local status=$?
    rm -rf /tmp/vagrant/*
    return $status
}
alias vagrant=_vagrant
```

sni:
The deinit in the ssh call has to be done. The ssh part calls exec so the at_exit part is not called. Just try it out :-) This patch still applies without change to the 1.8.1 release.

PR #6316, opened 2015-09-24
FIX: StatusException "no such file" in push via SFTP when file parent directory does not exist

stphivos:
Overcomes exception Net::SFTP::StatusException (2, "no such file") when using: "vagrant push" via sftp and a file parent directory does not exist. Function "upload" in plugins/pushes/ftp/adapter.rb class "SFTPAdapter" does not create the directory before uploading a file ('mkdir: true' seems to have no effect as zero directories are created while files are uploaded normally).

### DIFF
--- a/plugins/pushes/ftp/adapter.rb
+++ b/plugins/pushes/ftp/adapter.rb
@@ -106,6 +106,7 @@ module VagrantPlugins
       def initialize(*)
         require "net/sftp"
         super
+        @dirs = []
       end
 
       def default_port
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @stphivos 

Could you please provide a Vagrantfile or setup to reproduce this issue so we can verify and write tests? Thanks!

stphivos:
Hey @sethvargo, off course, here is my vagrant file:

```ruby
# -*- mode: ruby -*-

dir = File.dirname(File.expand_path(__FILE__))

require 'yaml'
require "#{dir}/puphpet/ruby/deep_merge.rb"

configValues = YAML.load_file("#{dir}/puphpet/config.yaml")

if File.file?("#{dir}/puphpet/config-custom.yaml")
  custom = YAML.load_file("#{dir}/puphpet/config-custom.yaml")
  configValues.deep_merge!(custom)
end

data = configValues['vagrantfile']

Vagrant.require_version '>= 1.6.0'

eval File.read("#{dir}/puphpet/vagrant/Vagrantfile-#{data['target']}")

  config.push.define "ftp" do |push|
    push.host = "192.168.1.200:22"
    push.username = "username"
    push.password = "password"
    push.destination = "/var/www"
    push.secure = true
    push.exclude = ".idea"
    push.exclude = ".vagrant"
    push.exclude = "docs"
  end
end
```
Let me know if you need any other info. Thanks!

PR #6318, opened 2015-09-24
Add 'Edit this page' links to `www` and `docs` websites

handlers:
Following [this pattern](https://github.com/mitchellh/packer/pull/2396), this PR adds 'edit this page' links to the footers of the Vagrant `docs` and `www` sites. Note that it also adds Makefiles for each Middleman site.


mitchellh:
LGTM!

PR #6321, opened 2015-09-25
Add custimizations for hyper-V

Chhed13:
added ability for Hyper-V manage: "auto stop actions", "time sync", "guest integration services", additional disk/controllers configuration
config.vm.provider "hyperv" do |vb|
    vb.disks_config = [[{ 'name' =>"d1", 'size' => "100"}, #outter array for controllers, inner hash for disks
                        { 'name' =>"d2", 'size' => "50"}], #name and size to create new .vhdx, path - to uses existent in place
                       [{ 'path' =>"D:\Chef\.kitchen\\test.vhdx"}]]
    vb.guest_integration_service = true
    vb.auto_stop_action = "ShutDown"
    vb.time_sync = false
end


sethvargo:
Paging @mwrock :smile:

mwrock:
I really like the added Hyper-V options here but I think some improvements to the configuration interface could make adding these settings a better experience.

Lets first look at `guest_integration_service` and `time_sync`. Both of these are different `VMServiceIntegration` components. I would consider having a single `VMServiceIntegration` setting that can take a hash of components:
```
vm.VMServiceIntegration {
  guest_service_interface: true, 
  heartbeat: true, 
  key_value_pair_exchange: true, 
  shutdown: true, 
  vss: true, 
  time_synchronization: true
}
```
This would have names that align with the Hyper-V integration components and since they are all just booleans, you now get support for all the different integrations. `false` values would naturally disable the components.

The disks are a cool feature, but it feels confusing to set up with the nested array. What would you think of being more explicit with the controller and perhaps adding a type so you could specify ide or scsi:
```
vm.controller {
  type: :ide,
  number: 1,
  disks [ 
    "some_name", size: 30 * 1024,
    "some_other_name", size: 100
  ]
}
vm.controller {
  type: :scsi,
  disks [ "c:/stuff/disk.vhdx" ]
}
```
I think this would have the following assumptions:
* if the controller has a `number` it already exists, otherwise it is added
* the `disks` first string arg is either a name or path. A path if it resolves to an existing file.
Also this allows disks to be sized in MB instead of GB which may be desirable in some cases.

There is some feature creep I am suggesting here and maybe you only implement a subset. My primary concern is that we don't introduce a config API that makes it difficult or awkward to expand upon down the road.

Chhed13:
Good ideas. I did this patch last summer as a minimum set of configurations wich we needed to use Vagrant. Now it looks to do more production variant. Bad that I don't understend when I have time to do this job.
Adout first part it's clear.
About second:
- I can't find the ability to create new IDE controllers. If you add syntax like this - it can work only in narrow case - when this controller already exist. So I didn't add any work with IDE on purpase
- About *disks* part syntax looks not clear. Is it array? It consists of string values and hashes at once and hashes are optional. If we use _name_ and _path_ at the same positions may be it's better to write it like this:
`disks { 
    "some_name": 30 * 1024,
    "some_other_name": 100,
    "c:/stuff/disk.vhdx": 1 #in case of existent disk value doesn't matter
  } `

mwrock:
Yeah you can't create new IDE controllers, but you can add disks to the existing 2. One would have to spwify the IDE controller number and if not, vagrant should raise an error stating that it cant create one.

On the disks, i has thinking you would specify an array. The first string param is required and if you specify an existing vhd(x), then size is unnecessary. I'm sure there is a more clear way. Whatever path is chosen should be clearly documented in https://github.com/mitchellh/vagrant/tree/master/website/source/docs/hyperv

StefanScherer:
I don't know if it fits into this PR, but it's the best place to discuss this.

How difficult would it be to allow nested virtualization in Hyper-V?

I found this documentation about configuring nested virtualization: https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/user_guide/nesting

As far as I understand before turning on the VM we have to run these steps:

```
Set-VMProcessor -VMName <VMName> -ExposeVirtualizationExtensions $true
Set-VMMemory -VMName <VMName> -DynamicMemoryEnabled $false
```

and some networking options.

This would allow people to run Hyper-V Vagrant boxes that have Hyper-V installed (eg. for a Win 2016 template running Nanoserver docker containers).

StefanScherer:
As a workaround right now someone can enable nested virtualization after the first `vagrant up`. This was my walkthrough that worked:

```
vagrant up --provider hyperv
vagrant halt
Set-VMProcessor -VMName WindowsServer2016TP5Docker -ExposeVirtualizationExtensions $true
Set-VMMemory -VMName WindowsServer2016TP5Docker -DynamicMemoryEnabled $false
vagrant up --provider hyperv
vagrant plugin install vagrant-reload
# install Hyper-V in VM
vagrant reload
```

But a setting in a `Vagrantfile` would be awesome to enable this before first boot.

PatrickLang:
I put in a simple PR for nested virtualization. It's something I need for my Docker labs so I coded up a minimal change for it. See #7738

chrisroberts:
If anyone wants to pick this up to continue on with @mwrock's suggestions, that would be awesome. Otherwise I will try and pick this back up soon.

Chhed13:
That would be very nice if you finish it

PatrickLang:
What do you think the rough order should be? I may be able to help with some of them. Would it be helpful to split out different issues for each proposal?

Here's what I have in mind:
1. Integration components - the proposal above seems along the right lines
2. Support setting IP on NICs - this probably requires ^^ first
3. Make VM name parameters match other providers

chrisroberts:
@PatrickLang That seems like a reasonable break down and yes, having them isolated from each other (in PRs) would make it much easier to handle.

bbrala:
I've looked through this and try to help a bit, i'll try and start on the integration service today.

Especially the Network Configuration would be helpfull on our end (making host resolution a breeze using our own DNS), but i recon starting with integration services would make sense. I'll post if I made some progress.

bbrala:
@PatrickLang @chrisroberts I made a first start on the integration services based on the discussion in this thread. It would be used as:


`vm_integration_services` (Hash) - Hash to set the state of integration services.

```ruby
config.vm.provider "hyperv" do |h|
    h.vm_integration_services = {
        guest_service_interface: true,
        heartbeat: true,
        key_value_pair_exchange: false,
        shutdown: true,
        time_synchronization: true,
        vss: true
    }
end
```

ntpttr:
Hi all, is there still a plan to merge a way to configure disks via hyperv in Vagrant? It looks like that functionality wasn't included in the integration services patch unless I'm missing something in there.

austin-laney:
Is there any recent progress made on this?

pedroengsi:
Hi all.
There is any time prevsion to include the Hyper-V disk size enhancement?
I been looking for this solution for a while.
Thanks in advance.

Chhed13:
Hi @pedroengsi,
I don't have any plans to move forward with this. I just get rid of Hyper-V and very happy with this.
If you want you can move forward with this feature

briancain:
Hey there everyone! I'm going to go ahead and close this issue, as I have just made a pull request that takes advantage of the recent generic disk config options in Vagrant. You can find that PR here: https://github.com/hashicorp/vagrant/pull/11541 It offers a lot of extra goodies such as resizing disks (similar to the VirtualBox implementation). Thanks!

PR #6342, opened 2015-10-01
IPv6 Host-Only Networks (VirtualBox)

mitchellh:
This adds support for IPv6 host-only networks. 

We need this in [Otto](github.com/hashicorp/otto).

Example: `config.vm.network "private_network", ip: "fde4:8dba:82e1::c4"`


plugins/providers/virtualbox/driver/version_4_2.rb
```diff
@@ -46,12 +46,21 @@ def create_dhcp_server(network, options)
...
+
+          # Configure
+          if ip.ipv4?
+            execute("hostonlyif", "ipconfig", name,
+                    "--ip", options[:adapter_ip],
+                    "--netmask", options[:netmask])
+          elsif ip.ipv6?
+            execute("hostonlyif", "ipconfig", name,
+                    "--ipv6", options[:adapter_ip],
+                    "--netmasklengthv6", options[:netmask].to_s)
```
phinze: Should we `else` raise here in case of shenanigans? Or can we trust this as an invariant of `IPAddr`?

plugins/providers/virtualbox/action/network.rb
```diff
@@ -258,30 +258,44 @@ def hostonly_config(options)
...
-          # will not work because the routing tables will force the
-          # traffic onto the real interface rather than the VirtualBox
-          # interface.
-          @env[:machine].provider.driver.read_bridged_interfaces.each do |interface|
-            that_netaddr = network_address(interface[:ip], interface[:netmask])
-            raise Vagrant::Errors::NetworkCollision if \
-              netaddr == that_netaddr && interface[:status] != "Down"
-          end
+          ip = IPAddr.new(options[:ip])
+          if !ip.ipv6?
```
phinze: Should probably do the `v4/v6` switch consistently in the places we do it.

I think the one below is a bit clearer:

```ruby
if ip.ipv4?
  # ...
elsif ip.ipv6?
  # ...
else
  raise "Malformed IPAddr: #{ip.inspect}"
end
```

lib/vagrant/util/network_ip.rb
```diff
@@ -1,10 +1,21 @@
...
+      # If the IP address is an IPv6 address, subnet should be a prefix
+      # length such as "64".
+      #
       # @return [String]
       def network_address(ip, subnet)
+        # If this is an IPv6 address, then just mask it
+        if subnet.to_s =~ /^\d+$/
+          ip = IPAddr.new(ip)
+          return ip.mask(subnet.to_i).to_s
+        end
```
phinze: `IPAddr#mask` supports subnet strings too, so theoretically this can do all the work for you. I can see the argument for changing less code, though, if that's what you were thinking.

plugins/providers/virtualbox/driver/version_4_2.rb
mitchellh: I'll add it to version 5, the old versions can be more wild west.

lib/vagrant/util/network_ip.rb
mitchellh: Fair, but yeah, I wanted to try to leave the IPv4 path as alone as possible since it mostly just works.

plugins/providers/virtualbox/action/network.rb
mitchellh: Done!

PR #6346, opened 2015-10-01
provisioners/ansible: Put username in the inventory

janek-warchol:
When running ansible provisioner on multiple VMs in parallel, vagrant
would by default feed just one ssh username to ansible.  Usually all
machines are provisioned using the same user, but it is possible that
they need different usernames.  By putting username in inventory we can
support this scenario.

PS this is the first time I do anything in Ruby, so please forgive me if I did something stupid!

### DIFF
--- a/plugins/provisioners/ansible/provisioner.rb
+++ b/plugins/provisioners/ansible/provisioner.rb
@@ -20,13 +20,10 @@ module VagrantPlugins
         # Ansible provisioner options
         #
 
-        # By default, connect with Vagrant SSH username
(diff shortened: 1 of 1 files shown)


gildegoma:
Thanks @janek-warchol. I've filed #6348 to track this issue, and as you can see there, the fix cannot be that trivial, especially to ensure that Vagrant stays backward compatible and open to any Ansible usages.

I'm closing this PR, but please feel free to talk further via #6348. I plan to fix it for the next Vagrant version, but of course you're welcome to work on it :)

PR #6354, opened 2015-10-03
Bump to latest bundler

iankronquist:
The changelog for Bundler 1.10.6 can be found here:
https://github.com/bundler/bundler/blob/c784ea3ade399fd487184493f7e8c0e340e7d6d7/CHANGELOG.md

I could find no changes which would affect vagrant.


sethvargo:
Hi @iankronquist 

Thank you for the Pull Request. This is a duplicate of https://github.com/mitchellh/vagrant/pull/6111 - you can read more about the reasons we can't upgrade at the moment. Thanks!

PR #6365, opened 2015-10-06
Update adapter.rb

ctammes:
https://github.com/mitchellh/vagrant/issues/5570

This change makes it possible to push from Windows to Linux and creates non-existing destination directories, which it didn't do.


ctammes:
Better merge PR 6316 instead of this one; it does the same.

PR #6367, opened 2015-10-07
Fix Slackware Host detection and nfsd checks

knixeur:
Slackware's version file is /etc/slackware-version not
/etc/slackware-release.
pidof is not on PATH by default (not running as root) so call it using
full path


mitchellh:
Thanks!

PR #6369, opened 2015-10-07
command/ssh-config: option to ignore unresponsive machines

electrofelix:
Add option '--ignore-down' to allow ignoring unresponsive machines when
outputting the ssh config settings.

Currently the command will output the ssh-config information for
machines until it reaches a machine with non available and will abort
with an expection.

With multi machine configs it is sometimes reasonable to have machines
down where you wish to retrieve the ssh config information of those
still running. This changes allows you to retrieve the information in a
single run of ssh-config instead of needing to request the information
for each machine individually, which can be expensive depending on the
number of machines and amount of configuration.

### DIFF
--- a/plugins/commands/ssh_config/command.rb
+++ b/plugins/commands/ssh_config/command.rb
@@ -23,6 +23,9 @@ module VagrantPlugins
           o.on("--host NAME", "Name the host for the config") do |h|
             options[:host] = h
           end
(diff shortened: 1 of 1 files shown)


mitchellh:
Rather than introduce this in the command, we'd rather make the providers robust to it. For now, please work around it by targeting specific machines, and we're going to work separately on making it just work for down machines. Thanks! And sorry for the delay.

PR #6374, opened 2015-10-08
Snapshot Support

mitchellh:
This adds support for `vagrant snapshot` as a top-level command. See the docs that are in the PR for more info. VirtualBox is the first provider to support this. Prior to Vagrant 1.8, we'll also be sure that the VMware plugin supports it.

Snapshots use VM snapshotting capabilities to provide checkpoint/restore functionality. 

We avoided putting this feature into Vagrant for awhile as it isn't compatible at all with Vagrant cloud providers. However, due to the rise of Terraform and us recommending Terraform as a better way to spin up cloud environments, we're now focusing Vagrant as a more local-focused tool, and we're not afraid to bring in features such as this.

GIF:

GIF


lib/vagrant/action/builtin/is_env_set.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Action
+    module Builtin
+      # This middleware is meant to be used with Call and can check if
+      # a variable in env is set.
+      class IsEnvSet
+        def initialize(app, env, key, **opts)
+          @app    = app
+          @logger = Log4r::Logger.new("vagrant::action::builtin::is_env_set")
+          @key    = key
+          @invert = !!opts[:invert]
```
sethvargo: Maybe I'm missing something, but I'm not seeing where this is used? Could you document it?

plugins/commands/snapshot/command/push.rb
```diff
@@ -0,0 +1,33 @@
...
+  module CommandSnapshot
+    module Command
+      class Push < Vagrant.plugin("2", :command)
+        include PushShared
+
+        def execute
+          opts = OptionParser.new do |o|
+            o.banner = "Usage: vagrant snapshot push [options] [vm-name]"
+            o.separator ""
+            o.separator "Take a snapshot of the current state of the machine and 'push'"
```
sethvargo: "of the given machine"

I assume this is going to behave like most other Vagrant commands - in a single-machine Vagrantfile, it will not require the name argument, but in a multi-machine Vagrantfile, you'll need to specify it?

plugins/commands/snapshot/command/push_shared.rb
```diff
@@ -0,0 +1,56 @@
...
+              vm.ui.info("Not created. Cannot push snapshot state.")
+              next
+            end
+
+            vm.env.lock("machine-snapshot-stack") do
+              m.call(vm)
+            end
+          end
+
+          0
```
sethvargo: ```# Exit 0, success```

plugins/commands/snapshot/command/push_shared.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+
+        def push(machine)
+          snapshot_name = "push_#{Time.now.to_i}_#{rand(10000)}"
```
sethvargo: Could we interpolate something meaningful from the machine here, like the machine id or machine name? I worry about a user going into the GUI or file system and seeing all these snapshots and not knowing what is what.

plugins/commands/snapshot/command/root.rb
```diff
@@ -0,0 +1,89 @@
...
+        end
+
+        def initialize(argv, env)
+          super
+
+          @main_args, @sub_command, @sub_args = split_main_and_subcommand(argv)
+
+          @subcommands = Vagrant::Registry.new
+          @subcommands.register(:save) do
+            require File.expand_path("../save", __FILE__)
```
sethvargo: I think we did benchmarks and `require_relative` was faster on Windows than expanding and requiring the absolute path

sethvargo:
Some minor comments, but otherwise LGTM

lib/vagrant/action/builtin/is_env_set.rb
mitchellh: Oh, I meant to remove it. Removing it.

plugins/commands/snapshot/command/push.rb
mitchellh: Yep

plugins/commands/snapshot/command/push_shared.rb
mitchellh: Snapshots are tied to the machine so its impossible to see them without looking exactly at the machine itself.

plugins/commands/snapshot/command/root.rb
mitchellh: SGTM, fixing now

jtopper:
I'm assuming you folks have looked at the community plugins which offer snapshot support. There are a couple of useful learnings in the history of https://github.com/scalefactory/vagrant-multiprovider-snap around GUI mode and other edges that it'd be a shame to have to revisit now that there's an official upstream interface.

mitchellh:
I took a look and I don't think our version should have problems, but if you experience otherwise please let us know. They also use different commands so worst case scenario you can still install the plugin.

PR #6377, opened 2015-10-08
"cap" command

mitchellh:
This introduces a new hidden (non-primary) command: `vagrant cap`.

This can be used to execute and check for arbitrary capabilities of Vagrant, the host, the guest, and the provider. This can be used to fingerprint the environment from the outside. This is an advanced feature that we don't intend to document publicly.

### DIFF
--- a/plugins/commands/cap/plugin.rb
+++ b/plugins/commands/cap/plugin.rb
@@ -9,7 +9,7 @@ module VagrantPlugins
       DESC
 
       command("cap", primary: false) do
-        require File.expand_path("../command", __FILE__)
+        require_relative "command"
         Command
       end
     end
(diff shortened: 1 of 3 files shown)


plugins/commands/cap/plugin.rb
```diff
@@ -0,0 +1,17 @@
...
+module VagrantPlugins
+  module CommandCap
+    class Plugin < Vagrant.plugin("2")
+      name "cap command"
+      description <<-DESC
+      The `cap` command checks and executes arbitrary capabilities.
+      DESC
+
+      command("cap", primary: false) do
+        require File.expand_path("../command", __FILE__)
```
sethvargo: relative

sethvargo:
LGTM

PR #6378, opened 2015-10-08
Clone another Vagrant environment

mitchellh:
This introduces `config.vm.clone` as a way to clone from another Vagrant environment (vs a box). 

This is useful in more complex Vagrant scenarios. I don't expect this to be a commonly used feature. For Vagrant 1.8, we're going to consider this experimental and not document/advertise it.


plugins/kernel_v2/config/vm.rb
```diff
@@ -37,6 +37,9 @@ class VMConfig < Vagrant.plugin("2", :config)
       attr_accessor :usable_port_range
       attr_reader :provisioners
 
+      # This is an experimental feature that isn't public yet.
+      attr_accessor :clone
```
sethvargo: return valud?

sethvargo: value*

sethvargo:
Code looks good - could we test this on Windows though?

plugins/kernel_v2/config/vm.rb
mitchellh: We don't document them for anything else. I think its okay. These are usually documented elsewhere.

mitchellh:
Sure. I'm very confident it'll work on Windows. We'll do a test prior to 1.8 so we can verify Otto works there too.

legal90:
@mitchellh @sethvargo 
JFYI about `config.vm.clone`: 
The target environment should be stopped prior to being cloned to the new VM. Otherwise, the following error occures:
```
==> default: Cloning VM...
There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.

Command: ["clonevm", "d1eb9efb-42dc-45ce-b3bd-9a015fbb4e5b", "--register", "--name", "temp_clone_1446224093227_60531"]

Stderr: 0%...
Progress state: VBOX_E_INVALID_OBJECT_STATE
VBoxManage: error: Clone VM failed
VBoxManage: error: Failed to lock source media '/Users/legal/VirtualBox VMs/vagrant-test-vbox-clone-2_default_1446224000635_94022/packer-virtualbox-iso-1422601639-disk1.vmdk'
VBoxManage: error: Details: code VBOX_E_INVALID_OBJECT_STATE (0x80bb0007), component MediumWrap, interface IMedium
VBoxManage: error: Context: "RTEXITCODE handleCloneVM(HandlerArg*)" at line 481 of file VBoxManageMisc.cpp
```

I've run into the same issue while implementing the similar feature for "parallels" provider. 
Do you have any thoughts about how it should be handled? 
I suggest to add some new core action with interactive prompt like:

```
==> The Vagrant environment you've specified is in '%{state}' and could not be cloned.
       Do you want to halt it in order to proceed the new VM creation? [y/N]
```

plugins/providers/virtualbox/action/prepare_clone_snapshot.rb
```diff
@@ -0,0 +1,65 @@
...
+          # Continue
+          @app.call(env)
+        end
+
+        protected
+
+        def prepare_snapshot(env)
+          name = env[:machine].provider_config.linked_clone_snapshot
+          name_set = !!name
+          name = "base" if !name
```
legal90: @mitchellh @sethvargo 
`"base"` - is it some convention? Will Otto refer to this snapshot by this explicit name?

I'm asking because in "parallels" provider we have a default snapshot name `"vagrant_linked_clone"`. So I'd like to understand - should I change it to "base" or not. 
Thanks.

plugins/providers/virtualbox/action/prepare_clone_snapshot.rb
```diff
@@ -0,0 +1,65 @@
...
+          end
+
+          # If we're not doing a linked clone, snapshots don't matter
+          if !env[:machine].provider_config.linked_clone
+            return @app.call(env)
+          end
+
+          # We lock so that we don't snapshot in parallel
+          lock_key = Digest::MD5.hexdigest("#{env[:clone_id]}-snapshot")
+          env[:machine].env.lock(lock_key, retry: true) do
```
legal90: As I've understood, `Environment#lock` provides only a process-level lock, so it will not work in parallel multi-machine environment (`vagrant up vm_1 vm_2`) due to this line
https://github.com/mitchellh/vagrant/blob/c17efbed755df621d6571742e163e8dff61dd119/lib/vagrant/environment.rb#L508

I guess there should be some global lock, like this:
https://github.com/mitchellh/vagrant/blob/b721eb62cfbfa93895d0d4cf019436ab6b1df05d/plugins/synced_folders/nfs/synced_folder.rb#L79

janvanveldhuizen:
> config.vm.clone

Oh wow.... this is exactly what I need. I have already been trying to implement something like this. I got it working but my implementation is a bit quick&dirty because it was my first experience in writing Ruby. 
I want to have a hierarchy of development machines that are all linked clones to avoid wasting disk space.
Like this:
```
base_win: from a Windows box
   base_win_vs2015: linked clone from base_win with Visual Studio 2015
   base_win_vs2017: linked clone from base_win with Visual Studio 2017
      dev_project1: linked clone from base_win_vs2017
      etc
```
This can be all be done manually in VirtualBox, of course. But I like to be able to use the VM management features of Vagrant.

Issue #6276, opened 2015-09-16
[salt] provisioner fails to copy minion config on 1.7.4


jcockhren:
Hey all! This is more than just a reproduction #5973 (which you should read for context. I won't duplicate here).

The goal of this issue is to point out problems/solutions with bootstrap script usage introduced in 1.7.3. With a fresh install of vagrant 1.7.4 (host: Trusty64), #5936 does not fix issues seemingly introduced with #5435.
### Problems
1. We need a clear and consistent way of customizing the arguments passed to the bootstrap script. This was the reason why @timoguin created #5435. The [documentation surrounding `install_args` and `bootstrap_options`](https://docs.vagrantup.com/v2/provisioning/salt.html) should properly reflect the  ~~quirks~~ behaviour of saltstack's bootstrap script.
   - `bootstrap_options` applies to the flagged options passed the the bootstrap script that are required to appear first.
   - `install_args` are for arguments such as branch name and install type (e.g. stable, git)
2. The quickest way back to a _working_ state would be to revert **both** #5936 and #5435. 
3. We need [testing around `install_args` and `bootstrap_options` anyways](https://github.com/mitchellh/vagrant/blob/master/test/unit/plugins/provisioners/salt/config_test.rb)

I'm working on all 3. PR soon.

Thoughts? @timoguin @mitchellh @cachedout @thatch45

babilen:
Is there anything  more that the community can help with to get this issue finally resolved?

chrisroberts:
Hi there,

It looks like this has been resolved within a previously shipped version of Vagrant so I am now closing this issue. If the original issue was not fully resolved, please reopen this issue or create a new one.

Cheers!

PR #6382, opened 2015-10-08
[salt] [bugfix] Restablize Salt provisioner 

jcockhren:
Hey all! This PR answers #6276. For the brave/bored, feel free to drill down through the referenced issues and PRs. 

For the everyone else, this PR reverts the show stopper bug introduced in #5435 and the work done on top of it in #5936. My goal was to simply bring the salt provisioner back to a working state. The issue seeming to involve vagrant using the Salt dummy config file instead of the file set with the `minion_config` setting.

### The Good

At provision time, the user supplied `minion_config` is copied to the inside the vagrant VM as you would expect. Therefore, the provisioner no longer hangs at "Bootstrapping Salt... (this may take a while)". This unblocks the Salt community. :confetti_ball: 

### The Bad

[There were no provision tests for the salt provisioner to begin with](https://github.com/mitchellh/vagrant/tree/master/test/unit/plugins/provisioners/salt). For the past 2 weeks, I've attempted to add some testing around the provisioner, however I've been unsuccessful. :disappointed: I would like the time/space to talk about why testing around the salt provisioner can not leverage the approach used in the [ansible](https://github.com/mitchellh/vagrant/blob/master/test/unit/plugins/provisioners/ansible/provisioner_test.rb), [shell](https://github.com/mitchellh/vagrant/blob/master/test/unit/plugins/provisioners/shell/provisioner_test.rb) and [chef](https://github.com/mitchellh/vagrant/blob/master/test/unit/plugins/provisioners/chef/provisioner/base_test.rb) provisioners, but for the sake of brevity let's make that a separate issue, if that's :cool:.

@mitchellh @timoguin @cachedout @thatch45 @wkral

### DIFF
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -33,7 +33,6 @@ module VagrantPlugins
       attr_accessor :install_args
       attr_accessor :install_master
       attr_accessor :install_syndic
-      attr_accessor :install_command
       attr_accessor :no_minion
       attr_accessor :bootstrap_options
       attr_accessor :version
@@ -62,7 +61,6 @@ module VagrantPlugins
         @install_args = UNSET_VALUE
         @install_master = UNSET_VALUE
         @install_syndic = UNSET_VALUE
-        @install_command = UNSET_VALUE
         @no_minion = UNSET_VALUE
         @bootstrap_options = UNSET_VALUE
         @config_dir = UNSET_VALUE
@@ -95,7 +93,6 @@ module VagrantPlugins
         @install_args       = nil if @install_args == UNSET_VALUE
         @install_master     = nil if @install_master == UNSET_VALUE
         @install_syndic     = nil if @install_syndic == UNSET_VALUE
-        @install_command     = nil if @install_command == UNSET_VALUE
         @no_minion          = nil if @no_minion == UNSET_VALUE
         @bootstrap_options  = nil if @bootstrap_options == UNSET_VALUE
         @config_dir         = nil if @config_dir == UNSET_VALUE
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -95,7 +95,7 @@ module VagrantPlugins
         @install_args       = nil if @install_args == UNSET_VALUE
         @install_master     = nil if @install_master == UNSET_VALUE
         @install_syndic     = nil if @install_syndic == UNSET_VALUE
-        @install_command    = nil if @install_command == UNSET_VALUE
+        @install_command     = nil if @install_command == UNSET_VALUE
         @no_minion          = nil if @no_minion == UNSET_VALUE
         @bootstrap_options  = nil if @bootstrap_options == UNSET_VALUE
         @config_dir         = nil if @config_dir == UNSET_VALUE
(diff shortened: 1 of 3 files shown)


timoguin:
@jcockhren and I have talked about this quite a bit and think this is the right direction. It seems the functionality I added was already there, but it wasn't clear in the documentation.

babilen:
How does this relate to #6073 ?

jcockhren:
@babilen 
So people get the context, #6073 addresses #5973. 

TBH, the errors thrown when this provisioner vomits, are misleading. The error shown in #5973 stopped occurring on Debian Jessie with this PR as well. In fact that error only appeared for me when using a debian/jessie base box but not ubuntu/trusty. Therefore this PR and #6073 seem related.

This PR fixes the bug introduced with the configuration. If the configuration vomits, then this provisioner uploads a dummy config file. Basically never considering the value of `minion_config`.

babilen:
Thanks for the explanation!

johntron:
Now we just need @sethvargo or @mitchellh to finally merge one of these PRs so we can finally close #6382, #6276, #5973, and #6073 and avoid all the confusion.

babilen:
What prevents this PR from being merged? Anything we can do to solve any potential issue?

plugins/provisioners/salt/config.rb
```diff
@@ -33,7 +33,6 @@ class Config < Vagrant.plugin("2", :config)
       attr_accessor :install_args
       attr_accessor :install_master
       attr_accessor :install_syndic
-      attr_accessor :install_command
```
jcockhren: dkjsdkjdkjdsl

mitchellh:
I believe this is fixed by #6073. Can someone advise on if this is different in any way? I believe this is fixed. Sorry for the delay.

timoguin:
@mitchellh, this is related pretty closely to #6073 but it fixes a different issue introduced in 1.7.4 by yours truly in PR #5435.

It should be merged as well.

jcockhren:
@mitchellh 

1. #6073 forces every single user of the salt provisioner to have the setting `salt.bootstrap_options = "-P -c /tmp"` in order to use it. Without that option, the provision step for all Vagrantfiles up until 1.7.2 will still hang when attempting to vagrant up from stratch. I know this because I've tested @johntron's [branch from his fork](https://github.com/johntron/vagrant/commits/revert-74d22069) using the following simple Vagrantfile using the salt provisioner:

```ruby
  config.vm.box_check_update = false
  config.vm.synced_folder "salt/states/", "/srv/salt/"

    vb.memory = "1024"
  end

    salt.minion_config = 'salt/minion'
    salt.run_highstate = true
    salt.install_args = 'git v2015.5.5'
    salt.verbose = true
  end
end
```
The result being, vagrant hangs during provisioning when attempting to run highstate. (note, I haven't included the state and config). Here's a [gist containing the supporting files needed to test @johntron 's PR](https://gist.github.com/jcockhren/2ed36558c37c98638230)
2. [Now use the same gist](https://gist.github.com/jcockhren/2ed36558c37c98638230) to checkout out my PR and you'll see that highstate successfully completes.

mitchellh:
Got it, I'll merge this one as well, does that fix it @jcockhren 

Sorry we don't have anyone here maintaining or testing Salt on a daily basis we rely on the community, so I apologize for my ignorance on the issue.

PR #6383, opened 2015-10-09
Document windows minion requirement for ipc_mode

PatOShea:
If this is not set to tpc, then the minion will fail to communicate with the master.

### DIFF
--- a/website/docs/source/v2/provisioning/salt.html.md
+++ b/website/docs/source/v2/provisioning/salt.html.md
@@ -90,7 +90,7 @@ a custom salt minion config file.
 * `minion_pub`  (salt/key/minion.pub) - Path to your minion
 public key
 
-* `grains_config`  (string) - Path to a custom salt grains file.
+* `grains_config`  (string) - Path to a custom salt grains file. On Windows, the minion needs `ipc_mode: tcp` set otherwise it will [fail to communicate](https://github.com/saltstack/salt/issues/22796) with the master.
 
 * `masterless`  (boolean) - Calls state.highstate in local mode. Uses `minion_id` and `pillar_data` when provided.
 


PatOShea:
@obestwalter - Hopefully others will benefit from our issues :+1:

obestwalter:
@PatOShea I am not sure if documenting this here is the right approach - since salt 2015.5.3 `ipc_mode: tcp` is set [programmatically on windows](https://github.com/saltstack/salt/commit/be43b2b394c7f994aec4c9acf5c8d5539b6bfc27). So somebody would really have to try to change this actively and therefore should know what they're doing (in this case: that it makes no sense to change the default ;)). This is [clearly documented on the salt end](https://docs.saltstack.com/en/latest/ref/configuration/minion.html#ipc-mode) already.

In general I think it's better to prevent user errors by simply making them impossible instead of documenting the problem. I see that my commit does not fully address that problem as it is still possible to change the ipc_mode on Windows although it should be simply impossible (unless one day ipc actually works, which is very unlikely). 

Maybe this should better be discussed over at Salt (in a new issue if something like this) on if how to prevent these kinds of user errors?

PatOShea:
Oliver, I want to add it to the documentation because it's completely
possible to add your own minion config and overwrite the one provided by
the installer on windows.

As for saltstack, they have a single resource to focus on windows and I
doubt they want to invest much time since vagrant is an edge case for them.

obestwalter:
@PatOShea I understand. But since 2015.5.3 you would have to explicitly try to set bogus settings on windows, as the defaults are sane now. 

Also: The PR is a bit to narrow for my taste. How about adding a warning section that explicitly warns that certain settings on certain osses don't work (e.g. in Windows it's not just the IPC setting, but also multiprocessing, and there might be other problems on other Osses, we both don't know about). So instead of just covering one specific case there could be an explicit referral to consult the salt docs for sane settings depending on the Os?

PatOShea:
@*obestwalter *- I'm just trying to help the next person who uses the
grains_config
(which overwrites the default config file) in windows.  I didn't test the
other default options, but this appeared to fixed my issue.  As for
increasing the breadth of the PR, I think project owners in vagrant and
salt would say, "PRs welcome". :)

mitchellh:
Thanks, lets use the doc change for now.

PR #6386, opened 2015-10-09
Fix network configuration in OS X (Darwin) guests

### LINKED ISSUE #6385: Network configuration failed in OS X (Darwin) guests
_I've reproduced this issue using the `parallels` provider. But I think, it is guest-specific and does not depend on the virtualization system_

If some network interfaces are configured for OS X virtual machine, they could be detected by the guest OS in the mixed order.

### Example
Vagrantfile:
```ruby
  ...
  config.vm.network "private_network"
  config.vm.network "public_network"
  ...
end
```

NIC order from the host:
```bash
# prlctl list -i a70ad143-b883-4a2f-8529-cc06605e32c8 | grep '  net'
  net0 (+) type=shared mac=001C42F6A0E6 card=e1000
  net1 (+) type=host iface='vagrant-vnet1' mac=001C4238D57C card=e1000
  net2 (+) type=bridged iface='Wi-Fi' mac=001C424EB008 card=e1000
```

NIC order from the guest:
```bash
# networksetup -listnetworkserviceorder
An asterisk (*) denotes that a network service is disabled.
(1) Ethernet
(Hardware Port: Ethernet, Device: en0)

(2) Ethernet Adaptor (en1)
(Hardware Port: Ethernet Adaptor (en1), Device: en1)

(3) Ethernet Adaptor (en2)
(Hardware Port: Ethernet Adaptor (en2), Device: en2)


# networksetup -listallhardwareports
Hardware Port: Ethernet
Device: en0
Ethernet Address: 00:1c:42:4e:b0:08

Hardware Port: Ethernet Adaptor (en1)
Device: en1
Ethernet Address: 00:1c:42:f6:a0:e6

Hardware Port: Ethernet Adaptor (en2)
Device: en2
Ethernet Address: 00:1c:42:38:d5:7c
```

We see, that "public" (bridge) interface has been detected as `en0`, "private" (host-only) - as `en2` and internal nat (shared) interface - as `en1`. 

### Actual Results

Guest capability `configure_networks` for Darwin guests expects that NIC order is strict, as in Linux. But due to the mixed order, these interfaces are configured wrongly.

### Suggested Solution 

There should be the similar logic, like in `configure_networks` for Windows guests. We should map network interfaces by their MAC addresses, not the device order.

I've fixed that: #6386

legal90:
Fixes #6385

Currently `configure_networks` guest cap for Darwin configures NICs following the device order and fails
when this order is mixed. 

While performing the network configuration on the guest side, we have to detect the appropriate NIC by its MAC address (like it is implemented for Windows guests).

/cc: @mitchellh @sethvargo

### DIFF
--- a/lib/vagrant/errors.rb
+++ b/lib/vagrant/errors.rb
@@ -248,6 +248,10 @@ module Vagrant
       error_key(:bundler_error)
     end
 
+    class CantReadMACAddresses < VagrantError
+      error_key(:cant_read_mac_addresses)
+    end
+
     class CapabilityHostExplicitNotDetected < VagrantError
       error_key(:capability_host_explicit_not_detected)
     end
--- a/plugins/guests/darwin/cap/configure_networks.rb
+++ b/plugins/guests/darwin/cap/configure_networks.rb
@@ -12,7 +12,7 @@ module VagrantPlugins
 
         def self.configure_networks(machine, networks)
           if !machine.provider.capability?(:nic_mac_addresses)
-            raise Errors::CantReadMACAddresses,
+            raise Vagrant::Errors::CantReadMACAddresses,
                   provider: machine.provider_name.to_s
           end
 
--- a/plugins/guests/darwin/cap/configure_networks.rb
+++ b/plugins/guests/darwin/cap/configure_networks.rb
@@ -6,45 +6,109 @@ module VagrantPlugins
   module GuestDarwin
     module Cap
       class ConfigureNetworks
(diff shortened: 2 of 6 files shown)


legal90:
It looks more complicated and requires `read_mac_addresses` provider capability being implemented, but this is the only correct and reliable solution, IMO.

Any other suggestions are welcome :)

mitchellh:
I'm going to trust your code and merge it, though I can't reason too much about the impl. But given your history with Vagrant we'll go with trust. Thanks!

legal90:
@mitchellh Thank you!

Issue #6343, opened 2015-10-01
More support for disabling provisioners


kruthar:
I'm using vagrant version: 1.7.4

I frequently have multiple provisioners in my Vagrantfile for organization and reuse of tasks. I commonly use 

``` sh
vagrant provision --provision-with x,y,z
```

to provision with specific provisioners. I have found that it would be really useful to do the opposite of this action, meaning specify provisioners I DO NOT want to run and vagrant will run all others. I can see this being done one (or both) of two ways:
1. adding a --[no-]provision-with x,y,z. Running `vagrant provision --no-provision-with x,y,z` should run all provisioners except x, y and z
2. adding a disabled: [true|false] parameter to provisioners in the Vagrantfile. `disabled: false` (default) would cause the provisioner to act as provisioners normally act. `disabled: true` would cause the provisioner to not run on normal provisioning steps unless explicitly told to do so with --provision-with x,y,z

Personally, I like the second one better, but both options seem useful to me. Adding this functionality would give the user more convenient control over which provisioners are being run.

chrisroberts:
Hey there,

I am going to close this due to lack of response. If this is still occurring, please open a [_new issue_](https://github.com/hashicorp/vagrant/issues/new) and follow the provided issue template that appears when you click the "New Issue" button. This will help us in getting a reproduction and fix. Thanks! 😄

PR #6387, opened 2015-10-09
kruthar/disable provisioners

kruthar:
This is addressing the feature additions I suggested in #6343. In this push are two simple features:
1. The 'run' attribute on provisioners in the Vagrantfile now accepts 'never' like so: `run: 'never'`. This will cause the provisioner with this attribute to never be run in the provisioning step of the machine unless explicitly told to do so using the --provision-with flag.
2. Added a --skip-provision-with flag that is the counter part to the --provision-with flag on actions that contain a provisioning step. Add `--skip-provision-with x,y,z` will provision with all provisioners specified for a machine except for provisioners x,y,z (type or name).

The combination of these two features give the user more control over what provisioners are getting run.

### DIFF
--- a/lib/vagrant/action/builtin/provision.rb
+++ b/lib/vagrant/action/builtin/provision.rb
@@ -102,12 +102,19 @@ module Vagrant
           type_map = provisioner_type_map(env)
           provisioner_instances(env).each do |p, options|
             type_name = type_map[p]
+
+            # Don't run if provision_types does not contain this provisioners type or name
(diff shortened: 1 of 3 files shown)


chrisroberts:
This looks great. I do think updating the option name from `--skip-provision-with` to `--provision-without` would make this a little more clear. Would you be willing to update the name and rebase the PR? Thanks!

chrisroberts:
Hey there,

Issue #6151, opened 2015-08-16
puppet_server provisioner fails with Puppet Collection 1


lesinigo:
When trying to use package `puppet-agent` from [Puppet Collection 1](https://puppetlabs.com/blog/welcome-puppet-collections) (actually `puppetlabs-release-pc1` on an Ubuntu 14.04 based box) the `puppet_server` provisioner fails to find the `puppet` binary.

Starting from *Puppet Collection 1* software will get installed in `/opt/puppetlabs/bin`, which is not in `$PATH`, and as far as I can tell the `puppet_server` provisioner of Vagrant 1.7.4 only looks for `puppet` in `$PATH` and does not apparently have any option to specify a different directory to look into.

It should try the new path (as far as I can tell the `puppet provisioner`, as in "puppet apply", does already try the new path and/or there are [pull requests](https://github.com/mitchellh/vagrant/pull/6057) about that path), or it should have a `binary_path` option to specify where to find the puppet executable. Better yet, it should look in the new directory *and* support a `binary_path` option :)

grafjo:
hi @lesinigo I've implemented the `binary_path` solution

mitchellh:
Fixed by #6389

PR #6389, opened 2015-10-10
Fix for #6151 / provisioner puppet_server with Puppet Collection 1

grafjo:
puppet_server provisioner fails with Puppet Collection 1 with the
following error:

```bash
==> default: Running provisioner: puppet_server...
The `puppet` binary appears not to be in the PATH of the guest. This
could be because the PATH is not properly setup or perhaps Puppet is not
installed on this guest. Puppet provisioning can not continue without
Puppet properly installed.
```

refs #6151


mitchellh:
LGTM. Thanks!

PR #6394, opened 2015-10-12
update

dpears:


sethvargo:
Hi there, 

Is this intentional? This looks like a bad rebase may have snuck it. Can you please rebase - the diff is too large to load in my browser :frowning:

Issue #5186, opened 2015-01-14
"Warning: Authentication failure. Retrying... " after packaging box


zunayed:
I am able to vagrant up and have ssh running fine from a base box that I've provisioned. 
I then package up the said box and keep getting errors with ssh. I've packaged numerious boxes before with this same process just yesterday. Not sure what the issue is now. 

```bash
vagrant package --base "<name from VBoxManage list vms>" 
vagrant box add mypackagedbox package.box
```
modified the Vagrant file `config.vm.box = "mypackagedbox" `

vagrant up now hangs on `Warning: Authentication failure. Retrying..`

I have tried with vagrant 1.7.1 & 1.7.2. Virtualbox version is 4.3.12r93733

vagrant file - http://pastie.org/9832360

asafvilman:
same thing here, found this http://comments.gmane.org/gmane.comp.tools.vagrant/4453
but couldnt make it happen...can anyone help?

zunayed:
Seems like an ssh issue
vagrant ssh-config on the base provisioned box(hashicorp/precise32) looks like it loads the identity file from .vagrant directory of the project

```bash
  vagrant ssh-config
  Host default
    HostName 127.0.0.1
    User vagrant
    Port 2222
    UserKnownHostsFile /dev/null
    StrictHostKeyChecking no
    PasswordAuthentication no
    IdentityFile /Users/zmorsalin/projects/vagrant_templates/project/.vagrant/machines/default/virtualbox/private_key
    IdentitiesOnly yes
    LogLevel FATAL"
```

But the packaged box that is failing loads it from`/Users/zmorsalin/.vagrant.d/insecure_private_key`

```bash
Host default
  HostName 127.0.0.1
  User vagrant
  Port 2222
  UserKnownHostsFile /dev/null
  StrictHostKeyChecking no
  PasswordAuthentication no
  IdentityFile /Users/zmorsalin/.vagrant.d/insecure_private_key
  IdentitiesOnly yes
  LogLevel FATAL
```

I've also tried looking in `/etc/udev/rules.d` but the only thing I saw was a empty folder called `70-persistent-net.rules`. I removed it prior to packaging and the bug persist.

melwinm:
Maybe there is no valid shh key loaded. Have you tried to set the key manually with config.ssh.private_key_path?

patrickheeney:
[...]

PR #6406, opened 2015-10-17
#5186: Warning: Authentication failure. Retrying... after packaging box

markus-perl:
After replacing the insecure key and writing a new key to .vagrant/machines/default/virtualbox/private_key vagrant continues to use the old insecure key ( /Users/myuser/.vagrant.d/boxes/mybox/0/virtualbox/vagrant_private_key)

This patch instructs vagrant to search first for the newly generated key and use this key if it is already generated.


sethvargo:
We had to revert this, but @mitchellh thinks he fixed this in https://github.com/mitchellh/vagrant/commit/1a7937ed50d767f328463978584f07ad546e456f.

PR #6415, opened 2015-10-19
Debian 8 machines will not shutdown gracefully

htesligte:
-H causes the machine to halt instead of poweroff and leaving a time causes the machine to wait 1 minute before shutdown.

### DIFF
--- a/plugins/guests/debian8/cap/halt.rb
+++ b/plugins/guests/debian8/cap/halt.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
       class Halt
         def self.halt(machine)
           begin
-            machine.communicate.sudo("shutdown -h -H")
+            machine.communicate.sudo("shutdown -h now")
           rescue IOError
             # Do nothing, because it probably means the machine shut down
             # and SSH connection was lost.


EmmanuelKasper:
This is a duplicate of https://github.com/mitchellh/vagrant/pull/6315

bitprocessor:
There is no need for a Debian 8 exception ; the default behaviour works just fine. While your fix will also work just fine, it's better to just get rid of the specific code for Debian 8. As in https://github.com/mitchellh/vagrant/pull/6315

gpkfr:
@bitprocessor I have made several test with jessie + vagrant + vmware-fusion. (all in latest version).
-The vagrant halt command execute the 'sudo shutdown -h now' correctly and the vm are stopped gracefully ; But vagrant seem to not detect the VM's shutdown and still wait for....
have you made a box for vmware fusion ?

bitprocessor:
@gpkfr I don't have Fusion, so I can't test. But to me, it sounds as if you'll get this behaviour with any Linux distribution? 

In fact, the code introduced in https://github.com/mitchellh/vagrant/pull/5369 actually breaks the shutdown process in VirtualBox: machines don't power off, they just halt and keep running. Vagrant eventually does a forced power off (pulling the plug on the vm in VBox).

If the shutdown -h now works correctly in Fusion, shutting the vm's down gracefully with a "powered-off" state in Fusion, it's rather the communication between Fusion & Vagrant that is having an issue in detecting the powered-off state I guess?

IMHO off course, can't test :( - Did you test with other distributions? CentOS, Ubuntu...

gpkfr:
Humm. i'm just tried the debian/jessie64 box. and it work fine with config.vm.graceful_halt_timeout set to 1.
but i do not test other distributions with vagrant ( i'm a Debian Lovers ;) )

could you add in your Vagrantfile this line : 

config.vm.graceful_halt_timeout = 1

and test with the debian/jessie box ?

(cf https://github.com/gpkfr/GenericVM/blob/config.vm.graceful_halt_timeout/scripts/vbox4dev.rb)

hope this help ;)

bitprocessor:
@gpkfr But it already works just fine in VBox ;-) Why should I add that extra line? (or was this for other Fusion users?)

gpkfr:
could you see this little Proof of concept screencast…
i used version 1.7.4 of vagrant without troubles..
https://vimeo.com/143275404

2015-10-22 17:12 GMT+02:00 Maarten De Wispelaere <notifications@github.com>:

htesligte:
So no, that doesn't work at all without troubles. As you can see, it says "Forcing shutdown". That means the device doesn't shutdown, the power is simply turned off. Your commit broke vagrant on Debian 8 boxes.

Unfortunately, it looks like mitchellh is no longer developing vagrant, so it might need to be time to look for a good maintainer that can get a fork going.

bitprocessor:
@htesligte exactly, it shouldn't say "forcing"...
However, do you have a source for the statement that @mitchellh is no longer developing vagrant?

gpkfr:
ok, perhaps we can suggest to patch the file
« plugins/guests/debian8/cap/halt.rb » with this content :

module VagrantPlugins
>   module GuestDebian8
>     module Cap
>     class Halt
>         def self.halt(machine)
>           begin
>             machine.communicate.tap do |comm|
>               if comm.test("lsmod |grep -w vboxguest")
>                 machine.communicate.sudo("shutdown -h now")
>               else
>                 machine.communicate.sudo("shutdown -h -H")
>               end
>             end
>           rescue IOError
>             # Do nothing, because it probably means the machine shut down
>             # and SSH connection was lost.
>           end
>         end
>       end
>     end
>   end
> end


2015-10-22 22:19 GMT+02:00 Maarten De Wispelaere <notifications@github.com>:

htesligte:
@gpkfr: Do you know what the command "shutdown -h -H" does? 

If your vmware box doesn't respond to a shutdown command, it makes no sense to change the command to not shutting down and then forcing the power off. The issue is somewhere else, not in this code.

bitprocessor:
@htesligte Exactly. 
@gpkfr As you are the Fusion user, could you please test the behaviour with another distribution?

gpkfr:
Hi,

Ok, perhaps i must to explain more clearly. When i emit a vagrant halt
command, the shutdown [-h -H || -h now] are both executed correctly on the
vm… in Both case the vm execute the shutdown procedure and stop running(vm
is OFF). but when option [-h now] are used, the vagrant software never go
down. it still waiting for the vm shutdown, almost the VM is Off. i don’t
know exactly why or how but for me, the issue is somewhere in the code.

hope this help.



2015-10-23 8:42 GMT+02:00 Henk te Sligte <notifications@github.com>:

bitprocessor:
@gpkfr That's why I keep on asking to test with a different distribution. We need to know if this is debian only, or some generic issue with Fusion.

Just run:

mkdir ubuntu-test
cd ubuntu-test
vagrant init ubuntu/trusty64 
vagrant up
vagrant halt

And see what it does.

Copy-paste the console output here

gpkfr:
yes, i make a try when i was less busy. perhaps today. do you have a box
like jessie(with systemd) to suggest ?

2015-10-23 8:46 GMT+02:00 Maarten De Wispelaere <notifications@github.com>:

gpkfr:
Sorry but ubuntu/trusty64 are only available for virtual box. ;) another
choice.

2015-10-23 11:06 GMT+02:00 Maarten De Wispelaere <notifications@github.com>:

bitprocessor:
@gpkfr Stop using Fusion... Problem solved ;-)

htesligte:
@gpkfr: You still don't know what the command does, you just found a magic workaround. If you understood the command, you wouldn't even suggest it. Try reading the shutdown manual. If the vagrant-vmware combination doesn't register a shutdown, you should look for a solution in the vagrant/vmware communication, not telling the machine to halt after a minute and then forcing the power off after a one-second timeout.

gpkfr:
I think have Found a strange implication with usage of systemd-sysv debian
package who break the halt command with VmWare* when installed.

I going to  investigate in this path …

you can remove or disable the debian8 diversion …

hope this help.

best regards,




2015-10-23 15:59 GMT+02:00 Henk te Sligte <notifications@github.com>:

mitchellh:
Fixed now in master! Thanks and sorry for the delay.

PR #6419, opened 2015-10-20
Allow setting of ssh LogLevel in Vagrantfile

jsyeo:
This PR allows users to set the LogLevel ssh configuration whenever they do `vagrant ssh`.


jstortz:
+1

mitchellh:
We have hopes and dreams of not depending directly on OpenSSH so we're going to start avoiding OpenSSH specific configurations. For this, we recommend using `vagrant ssh-config` and executing `ssh` directly. We'd like to restrict Vagrantfile configs to more commonly used configs. Thanks!

PR #6315, opened 2015-09-24
FIX: no shutdown exception for debian 8 needed

bitprocessor:
The debian8 exception is not needed & not functioning:
The shutdown -h -H command (on vagrant halt), doesn't work correctly, and just forces vagrant to force-shutdown the VM in the end.

The regular "shutdown -h now" works just fine and enabled vagrant to shutdown the machine gracefully.

PR #6429, opened 2015-10-22
#6315-FIX Added diversion when Virtualbox is used.

gpkfr:
The debian8 guest addition is mandatory to running GNU/Linux Debian 8 (aka jessie) with the provider plugin 'vmware_fusion'. Virtualbox users complained that shutdown -h -H has broke something and prefer to use the shutdown -h now command.
This patch try to make everybody happy.. ;)
Please before using a head Double Ax and drop the code, make some considération to the "Commercial Users" who uses the vmware provider plugins from https://www.vagrantup.com/vmware.
perhaps the vmware plugins need some patch to fix this bugs ?

Hope this help.

### DIFF
--- a/plugins/guests/debian8/cap/halt.rb
+++ b/plugins/guests/debian8/cap/halt.rb
@@ -1,10 +1,16 @@
 module VagrantPlugins
   module GuestDebian8
     module Cap
-      class Halt
+    class Halt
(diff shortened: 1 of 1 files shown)


gpkfr:
Please Do not merge. receive my Apologies.

Issue #6449, opened 2015-10-28
vagrant halt and reload unable to find shutdown command in RedHat guest


gawbul:
I have been attempting an auto-reboot using the vagrant-vbguest plugin. I have also tried using the vagrant-reload plugin, as well as manually issuing a `vagrant reload` and `vagrant halt` command from my host. In all cases I get the following error:

```
$ vagrant halt
==> default: Attempting graceful shutdown of VM...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

shutdown -h now

Stdout from the command:



Stderr from the command:

/bin/bash: line 3: shutdown: command not found
```

I'm using a rhel7-workstation image as the base box. The shutdown command is located at `/usr/sbin/shutdown` in the guest.

gawbul:
I've fixed this by adding the 'halt' 'guest_capability' to 'plugin.rb' in the 'plugins/guests/redhat' directory, and adding a 'halt.rb' file under the 'cap' subdirectory that calls '/usr/sbin/shutdown -h now'.

gawbul:
I tried modifying the default linux `halt.rb` to run a `\`which shutdown\` -h now`

gawbul:
Seems related to https://github.com/mitchellh/vagrant/pull/6415

sethvargo:
Hi @gawbul 

I just tried a few different RHEL 7 boxes and I am unable to reproduce this issue. The `shutdown` command is in the `PATH` by default. Do you have a specific box you are using? I think it may be a bug with that box and not necessarily a bug in Vagrant.

gawbul:
@sethvargo I'm using a custom rhel7 box that we've uploaded to Atlas (though not publicly available). Its running `Red Hat Enterprise Linux Server release 7.1 (Maipo)`. We created it using Packer from the rhel7 iso I believe? We use Chef for provisioning, so I should probably check that some files/environment variables aren't being modified.

The fact that this is occurring at all though, means the current method of assuming that shutdown is available in the `$PATH` is unstable in some circumstances (hence the ability to add the halt guest_capability), which required me to modify the Vagrant source. I think hard coding the PATH into the default linux halt script will mean not having to create halt scripts for individual guests in the future? Is there any reason that wouldn't be a better option?

sethvargo:
Hi @gawbul 

> assuming that shutdown is available in the $PATH is unstable in some circumstances

[...]

PR #6450, opened 2015-10-28
Fixes for vagrant halt/reload not finding shutdown command in some gu…

gawbul:
Commit in reference to issue #6449


gawbul:
The only issue I can see here is that older versions of RHEL have shutdown in a different location, so perhaps adding a halt capability there might actually break other people's workflows? I think hardcoding the `PATH` variable in the default linux guest `halt.rb` file would probably be the best option?

gawbul:
Actually, thinking about that, that might only work for specific shells (i.e. bash)? If the guest os uses a different shell it might break too?

PR #5340, opened 2015-02-15
[WIP] Add new provisioner: Ansible on the Guest VM

gildegoma:
@mitchellh @sethvargo Here is my attempt to implement #2103. My idea is to try to get it merged into 1.8 Milestone.

I created the branch directly in original repository, so that you can directly add things by your own (especially to quickly fix my English and Ruby mistakes, sorry in advance :sweat_smile:). I am looking forward to receiving your comments :grinning: !

#### The `ansible_local` Provisioner (new feature)

I'd be **very thankful to the interested Vagrant users to check the current design and report as soon as possible if their needs wouldn't be satisfied**, so we can adapt before this first release.

Note that the Ansible Installation Capability is already available for the following platforms (using the distribution package system (e.g. apt, yum, pacman, freebsd pkg, etc):

  - Debian 6+ (package is officially available as of jessie, with backports for wheezy. Older versions won't be supported).
  - Ubuntu 12.04+ (package is officially available as of Trusty, but it is preferable to use the official PPA from Ansible Inc.)
  - RedHat-6+ family via EPEL repository (validated on Fedora20, CentOS 6 & 7, OracleLinux 6 & 7)
  - ArchLinux (package is officially available)
  - OpenSUSE 13.2+ (package is officially available as of 13.2)
  - FreeBSD 10+ (package is officially available as of 10.0)

##### Open Questions

Discussed via code comments on this GitHub pull request.

##### Missing Features

The following functions are intentionally not developed in this initial version (but might be considered in future development iterations):

  - Automated installation of a specific Ansible version (e.g. via pip packaging).

    Reasons: Lack of user feedback, experience to confirm this need. User still can build their own base box.

#### The `ansible` Provisioner

By keeping the existing unit test coverage almost unchanged, I'm quite confident that the refactored `ansible` provisioner still behave the same (a.k.a free from new bugs!). I only added a few minor things that I hit on my coding path (Windows warning, and a buggy behavior when `VAGRANT_LOG` is not defined).

#### The Documentation

Many changes are also proposed here. I mainly tried to cut the things in four separate pages with the goal to help user to quickly find the relevant information:
* a common page to introduce how Ansible is integrated in Vagrant (ansible_intro)
* a common page to list the common options (ansible_common)
* a short page to quick start with `ansible` provisioner, with specific tips and tricks
* a short page to quick start with `ansible_local` provisioner, with specific tips and tricks

Like for the code, I'm open to any other way to organize the things... just let me know what do you think.

#### Finalization Check-List

[...]

PR #6348, opened 2015-11-02
Absolutely force the Ansible remote (ssh) user to correspond to Vagrant ssh user 

gildegoma:
In continuation to #6210 and #5765, I plan to propose the following change (still thinking):

* Introduction of a new boolean option `force_remote_user`:
  * If any Vagrant 1.7.{5,}: by default set to `false` (for backwards compatibility reason)
  * as of Vagrant 1.8+: by default set to `true` (to be friendly with a majority of use cases, almost nobody should be affected)

* Behaviours with a *generated inventory*:
  * The `ansible_ssh_user` setting is added to the generated inventory when the above option is true.
  * The `--user=...` command line argument is passed when the above option is false.

* Behaviours with a *custom inventory*:
  * The `--extra-vars=ansible_ssh_user=...`) command line argument is passed when the above option is true
  * The `--user=...` command line argument is passed when the above option is false.

#### Positive Side Effects:

* The vagrant auto-generated inventory could be then used "as it is" by external Ansible commands (no need to pass `--user` argument).
* As pointed in #6346, the "ansible parallel provisioning trick" could then be applied with machines that don't have the same remote username.

#### The Tricky Parts

The Ansible precedence rules for the remote ssh `user` parameter are the following:

 1. `ansible_ssh_user` in inventory file or as [extra variable](https://github.com/sovereign/sovereign/blob/b77bcdbb0c7fa7e66d4e51e17ed88ede56614fee/Vagrantfile#L32)
 1. **`remote_user` or `user` YAML attributes of an Ansible Play or an Ansible Task. Attention:** People using these features might suffer when upgrading to Vagrant 1.8.
 1. `-u / --user` ansible-playbook command line argument
 1. ANSIBLE_REMOTE_USER environment variable
 1. `remote_user` parameter in an `ansible.cfg` file
 1. username of the current user


gildegoma:
@mitchellh @sethvargo If you agree on this, this is to be merged for Vagant 1.8.0 release (see breaking change notice).

Now, my next step for 1.8 is to make #5340 mergeable again... (including this new change as well).

gildegoma:
@mitchellh @sethvargo according to https://github.com/hashicorp/otto/issues/297#issuecomment-153773472, Vagrant 1.8.0 should be released within the next week(s). I'd like to merge this PR now into master, is it okay for you?

gildegoma:
Closed in favour of #6500.

PR #6471, opened 2015-11-03
Be smarter about Ubuntu OS versions.

rogerhu:
Check the version number instead of the Ubuntu distribution name.


mitchellh:
This was fixed in another PR. Thanks!

kohenkatz:
@mitchellh Which PR are you referring to?

PR #6473, opened 2015-11-03
Fix custom minion_config upload

aneeshusa:
If a custom minion_config is specified, upload it during
the upload_configs step, then use sudo to copy into the config_dir
instead of uploading it again.

### DIFF
--- a/plugins/provisioners/salt/provisioner.rb
+++ b/plugins/provisioners/salt/provisioner.rb
@@ -344,7 +344,7 @@ module VagrantPlugins
       def call_highstate
         if @config.minion_config
           @machine.env.ui.info "Copying salt minion config to #{@config.config_dir}"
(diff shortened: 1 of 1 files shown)


aneeshusa:
Fixes #5973. Unlike #6073, uses sudo to always get around permissions errors, while retaining the ability to use a different config_dir. IMO, it is better for the default to be to respect the default configuration location - /etc/salt - and use sudo to put the configuration there, instead of to be to upload the configuration to a temporary directory and rely on salt-bootstrap to copy the minion configuration into /etc/salt. I use a custom bootstrap script to handle installation for Ubuntu and OS X for builds on Travis, and I would rather not have to have my bootstrap script copy this salt-bootstrap script behavior when it's easier to just put the file in the needed place..

However, I would not be opposed to removing the ability to use a different config_dir as in #6073.

mitchellh:
Fixed using #6073 Thanks!

PR #6475, opened 2015-11-03
Prefer xfreerdp for RDP connections on Linux hosts.

### LINKED ISSUE #6140: xfreerdp instead of rdesktop on centos 7 host
accessing windows 8 + using Centos 7 and rdp is not working due to rdeskop not being distributed with the release.
Installing or compiling rdesktop does really help, as windows wants secure connection and rdesktop seems not to be able to handle.

Xfreerdp is provided and it works when invoked.

Please consider making it an option to select which rdp client shall be used.

Thank you!

nikelmwann:
This resolves #6140.

`rdesktop` fails to connect to newer versions of Windows Server due to not supporting newer features of RDP such as Network Level Authentication.

`xfreerdp` [supports such features and is an all-around better option](https://unix.stackexchange.com/questions/140047/what-are-the-differences-between-rdesktop-and-xfreerdp/188168#188168), so I've made it the default RDP client when running `vagrant rdp` on a Linux host. If it's not installed, `rdesktop` will be used as a fallback. If neither are installed, an error message will be displayed as usual.

### DIFF
--- a/lib/vagrant/errors.rb
+++ b/lib/vagrant/errors.rb
@@ -408,8 +408,8 @@ module Vagrant
       error_key(:linux_nfs_mount_failed)
     end
 
-    class LinuxRDesktopNotFound < VagrantError
-      error_key(:linux_rdesktop_not_found)
+    class LinuxRDPClientNotFound < VagrantError
+      error_key(:linux_rdp_client_not_found)
     end
 
     class LocalDataDirectoryNotAccessible < VagrantError
--- a/plugins/hosts/linux/cap/rdp.rb
+++ b/plugins/hosts/linux/cap/rdp.rb
@@ -13,7 +13,7 @@ module VagrantPlugins
             elsif Vagrant::Util::Which.which("rdesktop")
               "rdesktop"
             else
-              raise Vagrant::Errors::LinuxRDesktopNotFound
+              raise Vagrant::Errors::LinuxRDPClientNotFound
             end
 
           args = []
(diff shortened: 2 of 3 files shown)


thor:
This works well, and I myself stumbled into the exact issue you mention, lack of NLA support with `rdesktop`.

One note though, if you connect to a box where the certificate of the machine isn't already trusted, the process will simply "hang", as `xfreerdp` doesn't get any input or output into the terminal from which 

    $ vagrant rdp

is run.

I "*solved*" this by adding the before the execution call, though I think it should probably be before the `extra_args`, so that it can be overwritten or similarly.
~~~ruby
args << "/cert-ignore"
~~~

If not--I don't know what's there in the util-class--perhaps it should be run in such a way that the user is notified.


Anyway, either way, I'd love to see this PR come through. I could even just manually add "-- /cert-ignore", but it gets a bit tedious. :)

nikelmwann:
Hm yeah you're right. `xfreerdp` already trusted the machine I was testing this with.

That could possibly be solved by examining the output for the error message and prompting for input on whether to accept unverified certificates. I'll have to take a look at [`lib/vagrant/util/subprocess.rb`](https://github.com/mitchellh/vagrant/blob/b721eb62cfbfa93895d0d4cf019436ab6b1df05d/lib/vagrant/util/subprocess.rb) and figure out how best to do that.

If I feel like I can add a clean solution, I'll implement it and update this PR. Otherwise, as you said, you can always pass in `/cert-ignore` as an extra argument for now.

mitchellh:
LGTM. Thanks!

PR #6485, opened 2015-11-05
command/provider: non-primary command to check and install providers

mitchellh:
This introduces a non-primary (hidden) command `vagrant provider` for listing the provider in use for an environment and also has the capability to install providers. I currently only implemented installing VirtualBox on OS X but other OS's will soon follow.


mitchellh:
I'm sure this can be improved, but going to merge as-is so I can begin doing the Otto integration work to test this out.

PR #6488, opened 2015-11-05
Add Errno::ENETUNREACH to SSH rescue

dvorak:
This changes the ssh ready? method to treat ENETUNREACH the same way as
EHOSTUNREACH errors.

When attempting to SSH into a box, it tries up to 5 times to connect to
the box, ignoring various errors.  Later it will catch and gracefully
handle most of those errors so that callers don't have to know the
details.

However, the Errno::ENETUNREACH error is not caught, which means that
callers that expect a clean boolean return from ready? don't get that,
and instead get an exception they probably aren't expecting.


mitchellh:
Thanks!

dvorak:
Thanks :)

PR #6490, opened 2015-11-05
Add provider install to `vagrant up`

mitchellh:
This introduces a new flag `--install-provider` (set by default) that automatically installs the requested provider if possible. This is a completely backwards compatible change: any providers that don't explicitly implement the `installed?` function will just always be seen as installed.

### DIFF
--- a/lib/vagrant/environment.rb
+++ b/lib/vagrant/environment.rb
@@ -383,6 +383,26 @@ module Vagrant
       raise Errors::NoDefaultProvider
     end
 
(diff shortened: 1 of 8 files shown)


legal90:
@mitchellh Thanks a lot!
Could you assist please - how `provider_install_<whatever>` host cap should look like? (for example, for `aws` provider)

Do we have to call `vagrant plugin install ...` explicitly, or it should be handled via `VagrantPlugins::CommandPlugin::Action::InstallGem` instance?

Issue #5933, opened 2015-07-11
Unable to mount /vagrant folder in 1.7.3


F21:
I am running vagrant 1.7.3 with VirtualBox 5.0 on a Windows 8.1 64-bit host.

The guest os is Ubuntu 15.04 64-bit.

When doing a `vagrant up`, it fails to mount the `/vagrant` folder:

```
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

chown `id -u vagrant`:`id -g vagrant` /vagrant

Stdout from the command:



Stderr from the command:

stdin: is not a tty
chown: changing ownership of ���/vagrant���: Not a directory
```

Relevant part of my vagrant file:

```ruby
config.vm.define "main" do |main_config|
    main_config.vm.network "private_network", ip: "192.168.50.4"
	
	main_config.vm.network "forwarded_port", guest: 80, host: 80
	main_config.vm.synced_folder "D:/work", "/work", type: "smb", owner: "www-data", mount_options: ['file_mode=0777', 'dir_mode=0777']

	main_config.vm.provider "virtualbox" do |v|
		v.memory = 2048
		v.cpus = 4
		v.customize ["modifyvm", :id, "--nictype1", "virtio"]
		v.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
		v.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
	end
  end
```

vany-egorov:
+1
```
==> default: Mounting shared folders...
    default: /vagrant => D:/vm/debian
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

chown `id -u vagrant`:`id -g vagrant` /vagrant

Stdout from the command:



Stderr from the command:

stdin: is not a tty
chown: changing ownership of ���/vagrant���: Not a directory
```

JeroenED:
Same on Windows 7. Tried with Cygwin and CMD. + I thought some actions were having a bold font. This does not seem to be there anymore.

    ==> jeroened.dev: Mounting shared folders...
        jeroened.dev: /vagrant => C:/Users/Jeroen De Meerleer/test-lampstack
    The following SSH command responded with a non-zero exit status.
    Vagrant assumes that this means the command failed!
    
    chown `id -u vagrant`:`id -g vagrant` /vagrant
    
    Stdout from the command:
    
    
    
    Stderr from the command:
    
[...]

PR #6493, opened 2015-11-06
Re-enable UNC paths for shared folders on Windows hosts

### LINKED ISSUE #6409: Support for long share paths on Windows hosts
As @Pomax suggested, I am starting a new issue for the post-1.7.3 revert as noone else has done it yet and I am starting to need this at work, badly. 

In short, when a file or folder is created in a folder on a guest that has been shared by a Windows host, the operation can fail if the created item exceeds host's path length limit. The most common use case is attempting to run a Linux guest on Windows host and installing packages through NPM where long dependency chains can easily exceed 260 characters. To remedy this, the folder should be shared using UNC path, which does not have a 260 character length restriction.

We had a fix in #5495 but it got reverted because something broke regarding the UNC paths in VirtualBox at the last minute. As far as I know, the latest versions still have this issue~~, but I haven't the chance to test it yet. Update: test build 5.0.7-103382 has the issue, test build 4.3.31-103278 seems to be working fine with UNC paths.~~

The course of action should be following:
* [x] Find out VirtualBox versions where the UNC paths last worked.
  * Virtualbox 4.3 has had mostly working implemenation of UNC shares for a long time
  * Virtualbox 5.0 hasn't had UNC share support at all
* [x] Find out exactly what is broken.
  * Virtualbox 4.3 - `df -h` returns `df: '/vagrant’: Not a directory`, using files in share works flawlessly
  * Virtualbox 5.0 - UNC shares aren't working at all
* [x] Work with VirtualBox team on solving this.
  * VirtualBox 4.3 - [issue](https://www.virtualbox.org/ticket/14651), test build r103933 has the issue solved.
  * VirtualBox 5.0 - [an issue had been opened already](https://www.virtualbox.org/ticket/14651), test build r103713 ~~seems to be working, but more testing is needed~~ works.
* [x] Reintroduce changes to Vagrant.

~~I've been out of the loop since #5495 got merged, so any help would be appreciated filling the blanks.~~

Old issues #1953 #4815 

CC @jfbibeau @celtric @dhampik

Ingramz:
This pull request is essentially the same as #5495 which enables use of long (>260 characters) path names for Windows hosts. This way creating longer than 260 character paths inside a shared folder doesn't fail inside the guest system (most common example: installing npm packages with long dependency chains).

Based on feedback from #5933, which caused the previous PR to be reverted, we contacted VirtualBox team and got them to fix
1. Missing support for shares using UNC paths on VirtualBox 5.0
2. Inability to get information about disk sizes for mounted shares, which used UNC paths on host, dubbed `df -h` issue on VirtualBox 4.3

The fixes are available for builds >=4.3 r103933 and >=5.0 r103713.

Are the changes to those drivers OK as-is or should we enable this only for builds higher than the ones mentioned above so that older builds won't "break"? If so, how?

Closes #6409

### DIFF
--- a/plugins/providers/virtualbox/driver/version_4_3.rb
+++ b/plugins/providers/virtualbox/driver/version_4_3.rb
@@ -610,6 +610,9 @@ module VagrantPlugins
         def share_folders(folders)
           folders.each do |folder|
             hostpath = folder[:hostpath]
+            if Vagrant::Util::Platform.windows?
+              hostpath = Vagrant::Util::Platform.windows_unc_path(hostpath)
+            end
             args = ["--name",
               folder[:name],
               "--hostpath",
--- a/plugins/providers/virtualbox/driver/version_5_0.rb
+++ b/plugins/providers/virtualbox/driver/version_5_0.rb
@@ -611,14 +611,9 @@ module VagrantPlugins
         def share_folders(folders)
           folders.each do |folder|
             hostpath = folder[:hostpath]
-
-=begin
-            # Removed for GH-5933 until further research.
             if Vagrant::Util::Platform.windows?
               hostpath = Vagrant::Util::Platform.windows_unc_path(hostpath)
             end
-=end
-
             args = ["--name",
               folder[:name],
               "--hostpath",


Ingramz:
@sethvargo you dealt with us last time, could you please review.

mitchellh:
This LGTM. Thanks for figuring this out.

BenLiyanage:
Is this in a vagrant version yet?

jonnywilliamson:
No. I downloaded the latest 1.7.4 and it's not there yet.

Looks like either a manual edit (take 2 secs) is needed until 1.7.5 is released.

BenLiyanage:
Yea, I did the manual edit last night.  Whats the eta on 1.7.5?

jonnywilliamson:
I don't know, I'd love to find out.

sethvargo:
No - this will be in Vagrant 1.8

jonnywilliamson:
I'm sorry but that confused me. Why would it miss out on 1.7.5 - 1.7.7? It's already merged into master?

Is 1.7.8 going to be the next version, where do we find this information out?

wouthoekstra:
Vagrant 1.8.0 and 1.8.1 are released, but as of yet I see no mention of this issue in the changelogs. Or do I miss something?

jonnywilliamson:
@wouthoekstra  It's in 1.8.0 already.

https://github.com/mitchellh/vagrant/blob/master/plugins/providers/virtualbox/driver/version_5_0.rb#L614-L616

wouthoekstra:
Thanks @jonnywilliamson. Apparently my problems are not over, I still get related errors (like maximum call stack size exceeded) and path related errors. I'm going to try a fresh install of everything (virtualbox, vagrant, homestead etc. when I have the time).

Ingramz:
It is probably worth mentioning that there can be still path related issues (mostly related to symbolic links) due to VirtualBox handling Windows permissions poorly (or it is a feature, I don't know). I don't know if this applies to you, @wouthoekstra, but I am going to document it anyway. But it would be useful if you wrote about the errors in detail, what do you execute, what's the output (errors), etc...

When I tried to get my two colleagues up and running on homestead, they both got stuck at `npm install`.

The most convenient workaround I have found for it is to turn off UAC completely (`EnableLUA` registry entry set to `0`). I've used it for a long time like this, which is why I didn't notice it personally, but do keep in mind that you'll sacrifice a great deal of security by doing this. One colleague found that it was a good solution because it makes him easier to follow tutorials.

Another way to solve this without doing the thing above would be to use `npm install --no-bin-links`. I don't use it personally, but the other colleague who does, says that it's fine for him. It's possible that a person could make a wrapper script for `npm` which adds `--no-bin-links` for the cases where the package is going to be installed inside a shared folder. This way it would make the use of it seamless and probably would have also worked for colleague who opted to turning UAC off.

BenLiyanage:
Hey--if you are using symlinks you need to run the shell that you boot your vagrant box as admin or they will not work.

Ingramz:
@BenLiyanage that's the gist of it, but from what I remember is that it's not as trivial as just opening a shell as administrator and booting the vagrant box from it. Let me know if that's not the case, I might have to try it again in that case.

jonnywilliamson:
I've had great success with using npm inside homestead when being run on a windows machine. But I HAD to wait this is PR was integrated...it was failing miserably until then.


What is also very interesting is this latest PR in homestead only a few days ago https://github.com/laravel/homestead/pull/320 . It might help some of you out using samba instead of nfs.

Ingramz:
@jonnywilliamson interesting. Thank you for pointing out the pull request, this should improve things a great deal. I'll try it out in the meanwhile and see how much it improves the current situation.

Ingramz:
I tried both running as administrator and using SMB (which inherently also requires administrator privileges). Starting the machine as administrator means that any interactions done with the machine have to be done also through administrator shell, this is what I meant it being not trivial.

I personally don't think that having to keep track of under which user the shell is opened is worth the effort, but it allows to use symlinks without reconfiguring any user permissions.

If UAC is turned off, I can see SMB as a way to improve filesystem performance, if an improvement exists.

BenLiyanage:
I've found it is as simple as that--could be wrong though.

BenLiyanage:
Personally, once my box is up I use putty instead of windows powershell or whatever.  I think that limits a lot of my exposure to the admin privileges issues you are running into.

wouthoekstra:
Thx guys! Running my shell as administrator did the trick. Next I might try @BenLiyanage 's workflow, by using putty instead of keeping an admin shell open. But for now, it works :)

Ingramz:
You don't even need to have putty for that, you could simply `ssh vagrant@your-vagrant-box.app` from your favorite shell and it's pretty much the same. With some tweaking you can even omit the user name and password. Kudos @BenLiyanage for keeping it simple.

wouthoekstra:
@Ingramz That's right, except that you can't use a non-admin-rights shell to vagrant ssh into a vagrant box that is booted from a shell that had admin rights.

If you try you'll get the message "VM must be running to open SSH connection. Run 'vagrant up' to start the vritual machine."

jonnywilliamson:
@BenLiyanage @Ingramz 

I have found this improvement to putty to be wonderful using windows. Much better for multiple sessions/windows etc. Hope it helps:

https://sourceforge.net/projects/putty-nd/

Ingramz:
@wouthoekstra that error comes from the fact that vagrant tries to figure out the ssh address from virtualbox information, but since vagrant cannot see administrator's machines, it will return that error.

Doing plain `ssh` from the shell will work because that's what vagrant also does behind the scenes.

wouthoekstra:
@Ingramz, well what do you know. You're right :) It works now. To others that use a Homestead box and might be struggling with the same thing, I had to use the ip out of my homestead.yaml: "ssh vagrant@<ip address> and not the name of the box, which is also defined in homestead.yaml.

Ingramz:
@wouthoekstra you'll need to map the name and ip address using hosts file. That's also mentioned in homestead documentation.

BenLiyanage:
Hey--just as a quick update.  One of my teammates has updated our node app to:

Node 4.3.1
NPM 3.7.2

I'm seeing the flatter folder structure, and it is working more or less correctly now on vagrant.

PR #6496, opened 2015-11-06
Fixing failure to use smb shared folders on Windows 7.

amirbilu:
The script being used to get all available IPs utilizes the PS command Get-NetIPAddress, which supported only on Windows 8+, causing Vagrant to explode while trying to vagrant up on lower versions (when using smb shared folders).

I've added a small Windows 7 fallback.

### DIFF
--- a/plugins/synced_folders/smb/scripts/host_info.ps1
+++ b/plugins/synced_folders/smb/scripts/host_info.ps1
@@ -1,8 +1,15 @@
 $ErrorAction = "Stop"
 
-$net = Get-NetIPAddress | Where-Object {
-    ($_.IPAddress -ne "127.0.0.1") -and ($_.IPAddress -ne "::1")
-} | Sort-Object $_.AddressFamily
+Try{
+	$net = Get-NetIPAddress | Where-Object {
+	    ($_.IPAddress -ne "127.0.0.1") -and ($_.IPAddress -ne "::1")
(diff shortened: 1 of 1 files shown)


mattgagliardi:
Just a quick thought (and I realize code can't be all things to all people)...I found this page after being effected by this issue.  My situation is such that I'm trying to set up VMs in a vSphere environment that's not located on my laptop (which has VMware Workstation installed on it).  The new code will also pick up IPs assigned to local VMnet adapters...I'm not sure how that'll play as it'll pick up the IP addresses for every adapter on the box...not sure what will happen when SMB has an assortment of 4-5 IPs it can map to.  Adding the following snippet will filter out all the VMnet adapters...

-and ($_.ServiceName -ne "VMnetAdapter")

amirbilu:
Another option is to grab the Host Only adapter..
Meaning .. -and ($_.Description -like "*Host*")

mitchellh:
This was changed to use the .NET API. Can you test this in the next version of Vagrant and report back? Thanks!

amirbilu:
Sure.

Many thanks!

pbowyer:
I've tested by manually patching my Vagrant install, replacing `vagrant/plugins/synced_folders/smb/scripts/host_info.ps1` with https://github.com/mitchellh/vagrant/commit/1f6e5aa08154700d834a2701e4fbb52956fd8ddf.

It works :+1:

PR #6502, opened 2015-11-08
SUSE-flavored systems use STARTMODE and not ONBOOT

dmacvicar:
As described in /etc/sysconfig/network/ifcfg.template

Static template was already using the right one, but the dhcp configuration seems
to be copied from a Fedora/Redhat template.
This fixes the issue that the interface does not come up after reboot.


mitchellh:
LGTM

Issue #6487, opened 2015-11-05
"which nmcli" causes error (CentOS 7)


mattcoffey:
Please see this closed issue for more details: https://github.com/mitchellh/vagrant/issues/5953

For an unknown reason, NetworkManager service becomes disabled at some point during the lifecycle of a CentOS 7 VM (I haven't worked out the pattern yet). This results in Vagrant's process of initialising the VM to be aborted (a side effect of which is that the /vagrant shared folder is left unmounted. The error output is as follows:

```
! which nmcli >/dev/null 2>&1 || nmcli c reload enp0s8

Stdout from the command:



Stderr from the command:

Error: NetworkManager is not running.
```

Despite the error you can vagrant ssh into the running VM. You can then gain root and run the following command on the box to fix the initialisation process:

```# systemctl enable NetworkManager.service```

and then reinitialise from the host operating system as follows:

```$ vagrant reload my_vm```

| don't know whether this potential solution fits within the desired scope of responsibilities for Vagrant, but maybe plugins/guests/fedora/cap/configure_networks.rb could try enabling NetworkManger after the nmcli command fails and try again? I understand that this would be working around a potential bug elsewhere though.

fedorovdima:
I'm working with RHEL7-based Linux where NetworkManager is not used and thus is disabled. However it is installed, so `! which nmcli >/dev/null 2>&1` results in 1 (true) and Vagrant's process of initializing the VM is aborted with the same error "Error: NetworkManager is not running."

I have fixed it with the following patch to vagrant/plugins/guests/fedora/cap/configure_networks.rb:

diff --git a/plugins/guests/fedora/cap/configure_networks.rb b/plugins/guests/fedora/cap/configure_networks.rb
index 364b744..3715d6c 100644
--- a/plugins/guests/fedora/cap/configure_networks.rb
+++ b/plugins/guests/fedora/cap/configure_networks.rb
@@ -113 +113 @@ module VagrantPlugins
-              machine.communicate.sudo("! which nmcli >/dev/null 2>&1 || nmcli c reload #{interface}")
+              machine.communicate.sudo("! which nmcli >/dev/null 2>&1 || ! systemctl -q is-enabled NetworkManager || nmcli c reload #{interface}")

sethvargo:
This has been fixed

leehambley:
> This has been fixed

[...]

PR #6508, opened 2015-11-10
Issue #6487: 'which nmcli' causes error (CentOS 7)

fedorovdima:


sethvargo:
Hey @fedorovdima 

I applied this approach with the similar PR in #6536. Thanks!

PR #6515, opened 2015-11-12
Slack net conf

oliviermeurice:
I added a plugin for Slackware network configuration.
Files have been added in vagrant/templates/guests/slackware and in vagrant/plugins/guests/slackware


sethvargo:
Fixes https://github.com/mitchellh/vagrant/issues/6514

PR #6519, opened 2015-11-13
Check that NetworkManager is running before issuing an nmcli command

signed8bit:
This is a followup to https://github.com/mitchellh/vagrant/pull/5931 where a
change was made to check that nmcli is installed. To further prevent errors
with vagrant up, we need to also make sure that NetworkManager is running.

The following error is seen when bringing up a VM where NetworkManager is
installed but not running.

        Configuring network adapters within the VM...

        ! which nmcli >/dev/null 2>&1 || nmcli c reload eth1

        Stdout from the command:

        Stderr from the command:

        Error: NetworkManager is not running.

This commit simply checks to see if the service is running before passing off to nmcli.

fixes mitchellh/vagrant#6518


sethvargo:
Hey @fedorovdima 

I applied this approach with the similar PR in #6536. Thanks!

signed8bit:
Great @sethvargo can't wait to see that in a Vagrant release.

PR #6522, opened 2015-11-16
Fix for detecting upstart, refs #6074

### LINKED ISSUE #6074: Regression: vagrant-mounted event is not emitted on Ubuntu < 15.04
The fix for #5377 broke the "vagrant-mounted" Upstart event on the LTS versions of Ubuntu. Upstart services that depend on this event will not run.

The Upstart binary was renamed from /sbin/init to /sbin/upstart on 15.04, so the check introduced with 92e6062f8dbb461c8b0f0a3e59c6427385851083 will cause any version of Ubuntu older than 15.04 not to emit the event.

yoshiyaka:
This should hopefully fix #6074. I don't have a working ruby environment, so this hasn't been tested, but executing `test ! -L /sbin/init && init --version | grep -q upstart` produces the expected results on both ubuntu 14.04 (upstart) and 15.10 (systemd).

### DIFF
--- a/plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
+++ b/plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
@@ -80,7 +80,7 @@ module VagrantPlugins
           end
 
           # Emit an upstart event if we can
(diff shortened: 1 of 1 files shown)


willrowe:
I can confirm that this successfully triggers the `vagrant-mounted` event on CentOS 6.6 where it was not before.

PR #6527, opened 2015-11-17
website: fix debian logo

pearkes:
Uses the real debian logo. Thanks @fenak! https://twitter.com/fenak/status/666495710049738752


sethvargo:
I think we might need to update middleman-hashicorp as well

sethvargo:
But :+1:

Issue #6395, opened 2015-10-12
access_token keeps getting appended in metadata_url 


teohhanhui:
Multiple occurrences of the same `access_token` in the URL found in `metadata_url` file.

sethvargo:
@teohhanhui this is the expected behavior. At some point, you have run `vagrant login` or set an `ATLAS_TOKEN` environment variable. Sending the access token is required because some resources are private and require authentication.

teohhanhui:
The problem is that the same access token appears more than once in the same url, instead of cleanly replacing the same query parameter:

`...?access_token=...&access_token=...&access_token=...`

sethvargo:
Hi @teohhanhui 

Can you please share your Vagrantfile and the output of the `vagrant` command in [debug mode](https://docs.vagrantup.com/v2/other/debugging.html)? Thanks!

remram44:
I'm running into this now, with vagrant 1.8.5:

```
remirampin@randy:~$ vagrant box update --box remram/debian-8-amd64
Checking for updates to 'remram/debian-8-amd64'
Latest installed version: 8.5
Version constraints: > 8.5
Provider: virtualbox
Updating 'remram/debian-8-amd64' with provider 'virtualbox' from version
'8.5' to '8.6'...
Loading metadata for box 'https://vagrantcloud.com/remram/debian-8-amd64?access_
token=MB6AQQ9Wwy44h4V9CCt5yF1pvhr-DuV9qCzFF3tM1kDovt41ZZtqrUSsh4vxDEFX9RE&access
_token=MB6AQQ9Wwy44h4V9CCt5yF1pvhr-DuV9qCzFF3tM1kDovt41ZZtqrUSsh4vxDEFX9RE&acces
s_token=MB6AQQ9Wwy44h4V9CCt5yF1pvhr-DuV9qCzFF3tM1kDovt41ZZtqrUSsh4vxDEFX9RE&acce
ss_token=MB6AQQ9Wwy44h4V9CCt5yF1pvhr-DuV9qCzFF3tM1kDovt41ZZtqrUSsh4vxDEFX9RE'
Adding box 'remram/debian-8-amd64' (v8.6) for provider: virtualbox
Downloading: https://vagrantcloud.com/remram/boxes/debian-8-amd64/versions/8.6/p
roviders/virtualbox.box
Successfully added box 'remram/debian-8-amd64' (v8.6) for 'virtualbox'!
```

PR #6534, opened 2015-11-18
Only append access_token the first time

sethvargo:
This fixes GH-6395 by only appending the access_token once. It also fixes a bug that was never reported. If a user supplied an access_token for a box URL, Vagrant would silently overwrite it.

After this commit, Vagrant only appends an access_token to the URL if no value exists at the key.


mitchellh:
LTM

PR #6535, opened 2015-11-18
core: prune machine if non-existent CWD from index [GH-4742]

### LINKED ISSUE #4742: global-status has incorrect entries and destroy doesn't clean up the incorrect entry
I realize that the information is cached but there should be a way to clean it up.  If I run ```vagrant destroy [ID of machine that no longer exists]``` I get ```The working directory for Vagrant doesn't exist!  This is the specified working directory: D:/path/to/dir/that/no/longer/exists```.  At this point Vagrant should probably clean up the cache and remove that specific entry.

mitchellh:
Fixes #4742

If `vagrant something ID` is used where ID represents a machine where the cwd was deleted, we now automatically prune the index from the global status DB. Tests included.

### DIFF
--- a/lib/vagrant/machine_index.rb
+++ b/lib/vagrant/machine_index.rb
@@ -78,7 +78,7 @@ module Vagrant
           @machines.delete(entry.id)
           unlocked_save
 
-          # Release acccess on this machine
+          # Release access on this machine
           unlocked_release(entry.id)
         end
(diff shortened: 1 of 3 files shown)


sethvargo:
:+1:

mitchellh:
Will fix syntax error on master.

PR #6537, opened 2015-11-19
Fix upstart detection and event emit

sethvargo:
- Fixes https://github.com/mitchellh/vagrant/issues/6074
- Closes https://github.com/mitchellh/vagrant/pull/6522


mitchellh:
LGTM

PR #6540, opened 2015-11-19
core: don't do HEAD request for box on non-HTTP [GH-5477]

### LINKED ISSUE #5477: Downloading from tftp:// uses HEAD query
I'm trying to get Vagrant to automatically fetch my boxfile from build server using TFTP protocol. 

```
  config.vm.box_url = "tftp://tftp.host:/devbox.box"
```

This results in Vagrant using `HEAD` query at first which downloads the entire boxfile to STDOUT, before starting actual download.

Trying it directly with embedded `curl` works as expected:

```
/opt/vagrant/embedded/bin/curl --output /tmp/box tftp://tftp.host/devbox.box
```

But when using the same URL in Vagrantfile, it does not. DEBUG output shows this:

    INFO box_add: Downloading box: tftp://tftp.host/devbox.box => /home/laas/.vagrant.d/tmp/box2f41853e07736bc3f7ba8f307ab44f0f11e9ae47
    INFO downloader: HEAD: tftp://tftp.host/devbox.box
    INFO subprocess: Starting process: ["/opt/vagrant/bin/../embedded/bin/curl", "-I", "-q", "--fail", "--location", "--max-redirs", "10", "--user-agent", "Vagrant/1.7.2", "--continue-at", "-", "-H", "Accept: application/json", "tftp://tftp.host/devbox.box"]

And after that `curl` starts to output binary data.

When not in DEBUG-mode, I can `Ctrl-C` while it's waiting for the HEAD to complete and then a regular download continues and successfully adds box (then exits due to interrupt). A next `vagrant up` will then successfully create a VM.

Other than the `HEAD` query, `tftp://` protocol seems to work just fine, so could you please see why it issues this `HEAD` query.

mitchellh:
Fixes #5477 

If the protocol isn't HTTP, we don't do a HEAD request.

### DIFF
--- a/lib/vagrant/action/builtin/box_add.rb
+++ b/lib/vagrant/action/builtin/box_add.rb
@@ -491,6 +491,9 @@ module Vagrant
             end
           end
 
+          # If this isn't HTTP, then don't do the HEAD request
+          return false if !uri.scheme.downcase.start_with?("http")
+
           output = d.head
           match  = output.scan(/^Content-Type: (.+?)$/i).last
           return false if !match
--- a/lib/vagrant/action/builtin/box_add.rb
+++ b/lib/vagrant/action/builtin/box_add.rb
@@ -492,7 +492,10 @@ module Vagrant
           end
 
           # If this isn't HTTP, then don't do the HEAD request
-          return false if !uri.scheme.downcase.start_with?("http")
+          if !uri.scheme.downcase.start_with?("http")
+            @logger.info("not checking metadata since box URI isn't HTTP")
+            return false
+          end
 
           output = d.head
           match  = output.scan(/^Content-Type: (.+?)$/i).last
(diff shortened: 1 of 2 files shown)


lib/vagrant/action/builtin/box_add.rb
```diff
@@ -491,6 +491,9 @@ def metadata_url?(url, env)
             end
           end
 
+          # If this isn't HTTP, then don't do the HEAD request
+          return false if !uri.scheme.downcase.start_with?("http")
```
sethvargo: Can we `@logger.debug` this just to be sure in the future?

sethvargo:
LGTM

lib/vagrant/action/builtin/box_add.rb
mitchellh: That's a really good idea.

mitchellh:
Done

Issue #6274, opened 2015-09-16
Salt Windows bootstrap minion download path needs update?


moop-moop:
vagrant/plugins/provisioners/salt/bootstrap-salt.ps1

line 54:
$url = "https://docs.saltstack.com/downloads/Salt-Minion-$version-$arch-Setup.exe"
change to
$url = "https://repo.saltstack.com/windows/Salt-Minion-$version-$arch-Setup.exe"

I noticed because the most recent salt minion version is not present in the original location, but is in the latter.

PatOShea:
@sethvargo, can this be closed?

iamseth:
While I'm a different Seth, I can confirm this is working for me and my coworkers :)

PR #6545, opened 2015-11-19
Fixing the Salt bootstrap script for Windows

iamseth:
The URL for the download has changed as per issue #6274


mitchellh:
Ah, Thanks!

PR #6551, opened 2015-11-19
Fix some Ruby

sethvargo:


mitchellh:
Ah yeah. This should be easy to unit test, do you mind doing that? Take a look at the other ones. Should be really simple.

PR #6554, opened 2015-11-19
push/heroku: use correct current branch [GH-6123]

### LINKED ISSUE #6123: vagrant push heroku strategy doesn't work when you have more than one local branch
The [`#git_branch`](https://github.com/mitchellh/vagrant/blob/186f3f25e37fea939d1f92a54bd51e6ff838d004/plugins/pushes/heroku/push.rb#L68) method in the Heroku push plugin doesn't return the current branch. It returns a list of all the branches.

So if you had four branches named `branch1 branch2 branch3 branch4`
the strategy then tries to run git like this (no matter what branch you are on): 

```
git ... push branch1 branch2 branch3 branch4:master
```

Lucky the command fails, rather than always deploying your last branch alphabetically.

mitchellh:
Fixes #6123 

This changes the way we get the branch to be the current branch.

This will fail (exit code != 0) on things that aren't on branches, which will be exposed to the user.

### DIFF
--- a/plugins/pushes/heroku/push.rb
+++ b/plugins/pushes/heroku/push.rb
@@ -69,7 +69,8 @@ module VagrantPlugins
         result = execute!("git",
           "--git-dir", git_dir(path),
           "--work-tree", path,
-          "branch",
+          "symbolic-ref",
+          "HEAD",
         )
 
         # Returns something like "* master"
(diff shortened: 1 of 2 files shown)


sethvargo:
:+1:

PR #6555, opened 2015-11-19
Automatically generate a node_name for Chef

sethvargo:
This is required because the Chef Server almost always needs a node name to interact. This will default to the hostname, but that's always going to be `vagrant.vm`, which will collide easily.

This generates a random hostname with `vagrant-` as the prefix and stores the result in the machine's data directory.


mitchellh:
LGTM! Awesome.

Issue #6278, opened 2015-09-17
chef provisioner is incompatible with chef <11.0


Tyrael:
since 4b1847acf30c8ef46d499c41ed54736277f232aa we always add --force-formatter to the chef_arguments but that's only available since chef 11.0.
as chef 10 is EOL I think it is ok for vagrant to drop support for it, but it would be better if we would simply check the version and bail out with a meaningful error message that you should either upgrade your chef install in the box or downgrade your vagrant version.

kisPocok:
+1

Tyrael:
:bow:

PR #6556, opened 2015-11-19
Only force the formatter if we are on Chef 11 or higher

sethvargo:
Fixes GH-6278


mitchellh:
Beautiful. Thanks

mitchellh:
LGTM* HAHA too robotic.

PR #6557, opened 2015-11-19
Allow ChefDK installation, add Windows installer support

sethvargo:
This allows users to install a different version of a Chef product (like `angrychef` or `chefdk`):

```ruby
chef.product = "chefdk"
```

This fixes https://github.com/mitchellh/vagrant/issues/6466.

The second part of this PR implements the existing Chef installation, as well as the new ChefDK installation to Windows. I could've sworn there was an open issue about that, but I guess there isn't. Well, it's fixed now :smile:


mitchellh:
:+1:

PR #6559, opened 2015-11-19
Require nodes_path for Chef Zero

sethvargo:


mitchellh:
LGTM, awesome for removing the deprecations too

Issue #6472, opened 2015-11-03
Standardize `chef_solo` path configuration with guest paths


jgonera:
I don't want to use separate shared folders for sharing Chef cookbooks. I know I can do something like this, although it's not documented:

```ruby
chef.cookbook_path = [:guest, "/vagrant/chef"]
```

if I already mounted the directory containing `Vagrantfile` in `/vagrant`.

It's confusing though, because the equivalent with host paths is:

```ruby
chef.cookbook_path = "chef/cookbooks"
```

Furthermore, if I want to have multiple cookbook directories, with host paths I can do:

```ruby
chef.cookbook_path = ["chef/cookbooks", "chef/vendor_cookbooks"]
```

There's no way to achieve this with guest paths. The following:

```ruby
chef.cookbook_path = [[:guest, "/vagrant/chef/cookbooks"], [:guest, "/vagrant/chef/vendor_cookbooks"]]
```

will result in Vagrant trying to find cookbooks in `/vagrant/chef/cookbooks/cookbooks` and `/vagrant/chef/vendor_cookbooks/cookbooks`.

The problem lies in:
* https://github.com/mitchellh/vagrant/blob/4b1847acf30c8ef46d499c41ed54736277f232aa/plugins/provisioners/chef/provisioner/chef_solo.rb#L32
* https://github.com/mitchellh/vagrant/blob/4b1847acf30c8ef46d499c41ed54736277f232aa/plugins/provisioners/chef/provisioner/chef_solo.rb#L104

Why are we doing this? Can we just get rid of this appended directory by replacing [`key`](https://github.com/mitchellh/vagrant/blob/4b1847acf30c8ef46d499c41ed54736277f232aa/plugins/provisioners/chef/provisioner/chef_solo.rb#L84) with `basename(path)`?

sethvargo:
Hi @jgonera 

Thank you for opening an issue, and I am sorry you are having issues with the Chef integration in Vagrant. I'm not sure I totally understand your issue - are you saying that you don't want Vagrant to mount additional shared folders for cookbooks?

The reason Vagrant behaves the way it does is to minimize the overhead for new users who may be unfamiliar with Vagrant/Chef.

I can experiment with your suggested changes, but unfortunately I suspect it will introduce a backwards-incompatible change.

jgonera:
Thanks for doing this! Sorry I didn't reply earlier.

PR #6560, opened 2015-11-20
Only append folder suffixes if we are sharing host folder paths

sethvargo:
Fixes GH-6472


mitchellh:
LGTM

jgonera:
Nit: those are two nested `if`s.

Issue #5988, opened 2015-07-17
Support arrays for chef-solo data_bags_path


xeron:
Chef-solo supports `Array` for `data_bags_path` since 12.0.0.

Related pull request: https://github.com/chef/chef/pull/1177

PR #6561, opened 2015-11-20
Support arrays for data_bags_path

sethvargo:
Fixes GH-5988


sethvargo:
This will not break backwards compatibility because the value has always been an array, we just called `.first` on it :smile:

mitchellh:
Ah, hah! LGTM

xeron:
Thank you guys! Small follow up: Chef 12 supports both `String` and `Array`.

sethvargo:
@xeron correct, but Chef 11 does not. So we still have to write it out as a string or else we break backwards compat.

PR #6562, opened 2015-11-20
Add support for forwarding host environment variables via SSH

sethvargo:
Rebases and fixes GH-4132


mitchellh:
LGTM. Much simpler.

chikamichi:
Hi,

In the initial issue (#4131), there was this great idea of supporting mapping to work around potential conflicts:

``` ruby
config.ssh.forward_env = { "VAGRANT_USER" => "USER" }
#                          from guest     => to host
```

It seems like it's not supported, was it dropped on purpose or would a PR be well-received?

Issue #5307, opened 2015-02-09
local-exec push not working on Windows


tsauter:
The local-exec push method is not working on Windows. 
The ruby fork statement isn't supported on this platform. It doesn't matter if an inline or external script is used.

Full error message is:

```
C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/util/safe_exec.rb:22:in `fork': fork() function is unimplemented on this machine (NotImplementedError)
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/util/safe_exec.rb:22:in `exec'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/plugins/pushes/local-exec/push.rb:41:in `execute!'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/plugins/pushes/local-exec/push.rb:36:in `execute_script!'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/plugins/pushes/local-exec/push.rb:14:in `push'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:567:in `push'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/plugins/commands/push/command.rb:28:in `execute'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/cli.rb:42:in `execute'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/lib/vagrant/environment.rb:301:in `cli'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.7.2/bin/vagrant:174:in `<main>'
```

molnara:
I get the same error in Windows 7.

sethvargo:
@mitchellh weird... how do the other shell things work?

neuron303:
Actually it is working with script files. Use files that windows is able to start (i.e. use .bat, .exe, ... files)

Since the inline script is written to a .sh temporary file this is broken.

The error message is triggered by the fallback mechanism to fork when Kernel.exec does not work.
If the fallback is disabled Kernel.exec returns following:
```
[...]

PR #6563, opened 2015-11-20
Use subprocess as a poor-man's exec for local-exec

sethvargo:
Fixes GH-5307


mitchellh:
I think we should guard this for only Windows. There are reasons `exec` is nice.

mitchellh:
LGTM

PR #6565, opened 2015-11-20
guests/linux: POSIX sed [GH-6104]

### LINKED ISSUE #6104: Failing to copy over authorized_keys on vagrant up Linux host FreeBSD guest
After 'vagrant up' process errors out with:
```bash

sed -i '/^.*ssh-rsa VERY_LONG_STRING_HERE== vagrant insecure public key.*$/d' ~/.ssh/authorized_keys

Stdout from the command:

Stderr from the command:

sed: 1: "/home/vagrant/.ssh/auth ...": invalid command code v
````

This is because BSD sed used the -i option differently than GNU sed. The -i option is not part of the [POSIX specification](http://pubs.opengroup.org/onlinepubs/009604599/utilities/sed.html).  However, there are [many possible fixes](http://unix.stackexchange.com/questions/92895/how-to-achieve-portability-with-sed-i-in-place-editing) that would render this command POSIX-compliant.

Searching the repository, there seem to be 21 lines where 'sed -i' is used.  Is that something that people would want to see fixed?  If so, I'll be glad to take a crack at it.

### PR DESCRIPTION
**Pull request by mitchellh**
Fixes #6104 

`-i` isn't a POSIX flag, just do the move manually.

### DIFF
--- a/plugins/guests/linux/cap/remove_public_key.rb
+++ b/plugins/guests/linux/cap/remove_public_key.rb
@@ -10,8 +10,10 @@ module VagrantPlugins
 
           machine.communicate.tap do |comm|
             if comm.test("test -f ~/.ssh/authorized_keys")
-              comm.execute(
-                "sed -i '/^.*#{contents}.*$/d' ~/.ssh/authorized_keys")
+              comm.execute(<<SCRIPT)
+sed -e '/^.*#{contents}.*$/d' ~/.ssh/authorized_keys > ~/.ssh/authorized_keys.new
+mv ~/.ssh/authorized_keys.new ~/.ssh/authorized_keys
+SCRIPT
             end
           end
         end

### CODE REVIEW

**Comment by sethvargo** [comment]
:+1:

**Comment by legal90** [comment]
@mitchellh @sethvargo This PR brakes permissions on `~/.ssh/authorized_keys`. Due to `mv`, they are `644`, but should be `600`. It causes vagrant can't authorize:

```code
$ vagrant up
...
==> centos: Waiting for machine to boot. This may take a few minutes...
    centos: SSH address: 10.211.55.235:22
    centos: SSH username: vagrant
    centos: SSH auth method: private key
    centos: Warning: Connection refused. Retrying...
    centos:
    centos: Vagrant insecure key detected. Vagrant will automatically replace
    centos: this with a newly generated keypair for better security.
    centos:
    centos: Inserting generated public key within guest...
    centos: Removing insecure key from the guest if it's present...
    centos: Key inserted! Disconnecting and reconnecting using new SSH key...
    centos: Warning: Authentication failure. Retrying...
    centos: Warning: Authentication failure. Retrying...
    centos: Warning: Authentication failure. Retrying...
    centos: Warning: Authentication failure. Retrying...
    centos: Warning: Authentication failure. Retrying...
...
```

PR #6567, opened 2015-11-20
core: remove saved synced folders not from Vagrantfile

### LINKED ISSUE #6198: Synced folders state can't be updated without destroying the VM
Using rsync, I changed the path that rsync is using while a machine was already built, provisioned and running. If I change the vagrant file such that it changes the path of the rsync'd folders, the commands `vagrant rsync` and `vagrant rsync-auto` do not being using the new settings. If I `vagrant reload` or `vagrant halt`/`vagrant up`, I see the changed rsync paths reflected while the machine comes up, but even after taking these steps the rsync/rsync-auto commands continue to use the old config.

I found that removing the file  `.vagrant/machines/default/virtualbox/synced_folders` and vagrant reloading fixes the issue.

mitchellh:
Fixes #6198 

This prunes items from the synced folder cache that come from the Vagrantfile and are no longer present in the Vagrantfile. Unit tests are here to help us.

### DIFF
--- a/lib/vagrant/action/builtin/mixin_synced_folders.rb
+++ b/lib/vagrant/action/builtin/mixin_synced_folders.rb
@@ -76,6 +76,16 @@ module Vagrant
           if opts[:merge]
             existing = cached_synced_folders(machine)
             if existing
+              if opts[:vagrantfile]
+                # Go through and find any cached that were from the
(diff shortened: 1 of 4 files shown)


sethvargo:
:+1:

Issue #5978, opened 2015-07-17
:public_address capability for VirtualBox provider


mkuzmin:
My [external plugin](https://github.com/mkuzmin/vagrant-guestip) makes `machine.provider.capability(:public_address)` call.
It works fine with Hyper-V and external providers, but fails on VirtualBox.

This capability is implemented within `vagrant-share` plugin, and does not return values when it's called directly.

Please move the implementation from vagrant-share to Vagrant core.

mkuzmin:
Thank you for rapid response, the change works great!

PR #6574, opened 2015-11-21
'vagrant address' command

mkuzmin:
`vagrant address` command outputs public IP address of a guest machine, the same value as `vagrant ssh` uses for connection.
```
>vagrant address
172.20.161.192
```

It supports multi-machine Vagrantfiles, but a guest name must be specified explicitly.

This feature depends on https://github.com/mitchellh/vagrant/issues/5978. At the moment Virtualbox provider does not expose `:public_address` capability.

### DIFF
--- a/plugins/commands/address/command.rb
+++ b/plugins/commands/address/command.rb
@@ -1,0 +1,24 @@
+require 'optparse'
+
+module VagrantPlugins
+  module CommandAddress
+    class Command < Vagrant.plugin("2", :command)
+      def self.synopsis
+        "outputs public IP address of a guest machine"
+      end
+
+      def execute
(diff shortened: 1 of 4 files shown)


sethvargo:
Hi @mkuzmin 

Thank you for the pull request. We would prefer to see this as an external plugin before moving this into core. Additionally, you can use the `vagrant ssh-config` output to get the machine's IP address. If the plugin gains in popularity, we can definitely move it into core. Thanks! :smile:

mkuzmin:
I'm fine with keeping this as an external plugin, but it really needs #5978 fixed. 

I'm using it with vSphere provider, but have multiple requests from other users for VirtualBox support.

PR #6582, opened 2015-11-23
Making the Fedora guest detection search string generic

LalatenduMohanty:
Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>


LalatenduMohanty:
@strzibny comments?

strzibny:
I have to say I don't know the reason for the strictness of this check. The newer versions usually works or at least its better if something work than nothing... and older versions are unsupported anyway. I agree that it can be more generic so we don't have always revisit and update.

PR #6585, opened 2015-11-23
Add presence helpers

sethvargo:


lib/vagrant/util/presence_helpers.rb
```diff
@@ -0,0 +1,41 @@
+module Vagrant
+  module Util
+    module PresenceHelpers
```
mitchellh: Just name this `Presence`

mitchellh:
One change, then LGTM

PR #6587, opened 2015-11-23
Use the new presence helpers in the Chef provisioner

sethvargo:


mitchellh:
Failing tests, are they related? Otherwise lGTM

Issue #6422, opened 2015-10-21
Vagrant "login" do not work behind an authentified proxy


psa-jforestier:
I user Vagrant 1.7.4 under Windows, but the "login" command do not work behind the corporate proxy :  

```
set HTTP_PROXY=http://login:password@proxy.com:80
set HTTPS_PROXY=http://login:password@proxy.com:80
wget www.vagrantup.com --no-check-certificate
...
index.html saved [7602]

vagrant login
... request username and password ...
C:/User/U402471/PSA_Softs/Vagrant/embedded/lib/ruby/2.0.0/net/http/response.rb:1
19:in `error!': 407 "Proxy Authentication Required" (Net::HTTPServerException)
        from C:/User/U402471/PSA_Softs/Vagrant/embedded/lib/ruby/2.0.0/net/http/
response.rb:128:in `value'

```

It seems that the http.rb object do not handle correctly the user/pass inside the HTTP_PROXY env.var.

I hard coded in http.rb the user and pass (in function HTTP.new) and it works.

I tried with "http_proxy" (lowercase), and other env var (http_proxy_username), and none works.

PR #6590, opened 2015-11-24
Respect proxy envvars for login command

sethvargo:
Fixes GH-6422


mitchellh:
Ah, there we go. LGTM!

PR #6591, opened 2015-11-24
Add `vagrant port` command

### LINKED ISSUE #1415: "vagrant port" command
Vagrant's [port collision and correction](http://docs.vagrantup.com/v2/networking/forwarded_ports.html) feature is cool!

As a user, I would like to be able to figure out what the corrected port is.

Currently, I could do this:

* read the vbox uuid from the .vagrant file
* use VBoxManage showvminfo to find the host ports chosen

Example, to find what host port maps to guest port 5601:

```
% VBoxManage showvminfo 6bfdd751-bc20-48ae-b80f-20408086b82d | grep 'guest port = 5601'
NIC 1 Rule(1):   name = 5f1-16bq, protocol = tcp, host ip = , host port = 39290, guest ip = , guest port = 5601
```

More ideally, something like `vagrant port <port>` could output the host port which forwards to `<port>` on the guest vm.

### LINKED ISSUE #2826: List port forwards of running machine
It's nice that `vagrant up` displays the forwarded ports when booting the VM, but if you miss that message I see no way to list the forwarded ports of a running VM besides opening VirtualBox or grepping the output of VBoxManage showvminfo.  Perhaps `vagrant status` could be modified to print this information?

### LINKED ISSUE #3009: vagrant up --machine-readable could return final hostname and port for all vm's
It would be handy for me if after running vagrant up successfully with the --machine-readable flag, the output could contain the following:

1. the hostname/ip for each machine that can be used to connect to it
2. the port that can be connected on for each machine (very handy in the case of port clashes)

### LINKED ISSUE #5650: Provide option for guest-to-host port-forwarding
At my company, connections to the database are tunneled over SSH. Therefore, when I want to access the database, I do something like this on my host machine:

    ssh -L 3306:localhost:3306 my.database.server.com

Great. Now I need to provide the database to my Vagrant box with an SSH remote forward:

    vagrant ssh -- -R 3306:localhost:3306

Now I've got the full chain. Local traffic on the host machine on port 3306 will be forwarded to `my.database.server.com:3306`, and I then create another remote forwarding into my Vagrant box which pushes traffic on the guest on port 3306 to the host on port 3306.

This is kind of a pain for new users. It would be really great and ease development if Vagrant could allow an option in the `Vagrantfile` to define a port forwarding from a local port on the guest to a local port on the host, basically the inverse of what it does right now for exposing guest ports to the host. 

This could be accomplished by simply having a process kick off when `vagrant up` is called which establishes a port-forwarding using SSH. Basically, in the Vagrant start phase before the provisioner is run, kick off a process like this to connect to the guest:

    ssh -R $host_port:localhost:$guest_port -P $guest_ssh_port $guest_host ... "/usr/bin/yes"

The `/usr/bin/yes` bit will keep traffic on the SSH connection to help it not get terminated. Doing the remote port forward over SSH is the easiest option, because otherwise you'll have to mess with iptables and firewall settings on both the host and guest: not good.

sethvargo:
This is still a :construction: but I have some questions around structure for @mitchellh 

- [x] Docs
- [ ] Name - should it be `vagrant port`, `vagrant ports`, `vagrant list-ports`?

- - -

- Fixes #1415
- Fixes #2826
- Fixes #3009
- Fixes #5650

### DIFF
--- a/plugins/commands/port/command.rb
+++ b/plugins/commands/port/command.rb
@@ -7,7 +7,6 @@ module VagrantPlugins
         "displays information about guest port mappings"
       end
 
-      # @todo support multiple strategies if requested by the community
       def execute
         opts = OptionParser.new do |o|
           o.banner = "Usage: vagrant port [options] [name]"
--- a/plugins/commands/port/command.rb
+++ b/plugins/commands/port/command.rb
@@ -1,0 +1,71 @@
+require "optparse"
+
+module VagrantPlugins
+  module CommandPort
+    class Command < Vagrant.plugin("2", :command)
+      def self.synopsis
+        "displays information about guest port mappings"
+      end
+
+      # @todo support multiple strategies if requested by the community
+      def execute
+        opts = OptionParser.new do |o|
+          o.banner = "Usage: vagrant port [options] [name]"
+          o.separator ""
+          o.separator "Options:"
+          o.separator ""
+
+          o.on("--machine-readable", "Display machine-readable output")
+        end
+
+        # Parse the options
+        argv = parse_options(opts)
+        return if !argv
+
+        @logger.debug("Port command: #{argv.inspect}")
+        with_target_vms(argv, single_target: true) do |vm|
+          vm.action_raw(:config_validate,
+            Vagrant::Action::Builtin::ConfigValidate)
+
+          if vm.state.id != :running
+            @env.ui.error "not running - make this a better error or use the middleware"
+            return 1
+          end
+
+          # This only works for vbox? should it be everywhere?
+          # vm.action_raw(:check_running,
+          #   Vagrant::Action::Builtin::CheckRunning)
+
+          if !vm.provider.capability?(:forwarded_ports)
+            @env.ui.error <<-EOH.strip
+The #{vm.provider_name} provider does not support listing forwarded ports. This
+is most likely a limitation of the provider and not a bug in Vagrant. If you
+believe this is a bug in Vagrant, please search existing issues before opening
+a new one.
+EOH
+            return 1
+          end
+
+          ports = vm.provider.capability(:forwarded_ports)
+
+          if ports.empty?
+            @env.ui.info("The machine has no configured forwarded ports")
+            return 0
+          end
+
+          @env.ui.info <<-EOH
+The forwarded ports for the machine are listed below. Please note that these
+values may differ from values configured in the Vagrantfile if the provider
+supports automatic port collision detection and resolution.
+EOH
+          ports.each do |guest, host|
+            @env.ui.info("#{guest.to_s.rjust(6)} (guest) => #{host} (host)")
+            @env.ui.machine("forwarded_port", guest, host, target: vm.name.to_s)
+          end
+        end
+
+        0
+      end
+    end
+  end
+end
--- a/plugins/commands/port/command.rb
+++ b/plugins/commands/port/command.rb
@@ -1,8 +1,12 @@
+require "vagrant/util/presence"
+
 require "optparse"
 
 module VagrantPlugins
   module CommandPort
     class Command < Vagrant.plugin("2", :command)
+      include Vagrant::Util::Presence
+
       def self.synopsis
         "displays information about guest port mappings"
       end
@@ -14,6 +18,9 @@ module VagrantPlugins
           o.separator "Options:"
           o.separator ""
 
+          o.on("--guest", "Output the host port that maps to the given guest port") do
+          end
+
           o.on("--machine-readable", "Display machine-readable output")
         end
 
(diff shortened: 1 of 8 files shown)


plugins/commands/port/command.rb
```diff
@@ -0,0 +1,71 @@
...
+        # Parse the options
+        argv = parse_options(opts)
+        return if !argv
+
+        @logger.debug("Port command: #{argv.inspect}")
+        with_target_vms(argv, single_target: true) do |vm|
+          vm.action_raw(:config_validate,
+            Vagrant::Action::Builtin::ConfigValidate)
+
+          if vm.state.id != :running
```
sethvargo: There is a middleware action for CheckRunning (commented out below), but it's only defined in the virtualbox stack. Should that be moved into the main middleware @mitchellh? If so, we could just use that instead of checking the state id directly.

plugins/commands/port/command.rb
```diff
@@ -0,0 +1,71 @@
...
+            @env.ui.error "not running - make this a better error or use the middleware"
+            return 1
+          end
+
+          # This only works for vbox? should it be everywhere?
+          # vm.action_raw(:check_running,
+          #   Vagrant::Action::Builtin::CheckRunning)
+
+          if !vm.provider.capability?(:forwarded_ports)
```
sethvargo: Gonna move these to i18n eventually, but just leaving them here now for early review purposes.

plugins/commands/port/command.rb
```diff
@@ -0,0 +1,71 @@
...
+          end
+
+          @env.ui.info <<-EOH
+The forwarded ports for the machine are listed below. Please note that these
+values may differ from values configured in the Vagrantfile if the provider
+supports automatic port collision detection and resolution.
+EOH
+          ports.each do |guest, host|
+            @env.ui.info("#{guest.to_s.rjust(6)} (guest) => #{host} (host)")
+            @env.ui.machine("forwarded_port", guest, host, target: vm.name.to_s)
```
sethvargo: Output looks like this:

```
The forwarded ports for the machine are listed below. Please note that these
values may differ from values configured in the Vagrantfile if the provider
supports automatic port collision detection and resolution.

  2222 (guest) => 22 (host)
  8080 (guest) => 80 (host)
  8081 (guest) => 81 (host)
  8082 (guest) => 82 (host)
  8083 (guest) => 83 (host)
  8084 (guest) => 84 (host)
  8085 (guest) => 85 (host)
  8086 (guest) => 86 (host)
  8087 (guest) => 87 (host)
  8088 (guest) => 88 (host)
  8089 (guest) => 89 (host)
```

plugins/commands/port/command.rb
```diff
@@ -0,0 +1,71 @@
...
+        argv = parse_options(opts)
+        return if !argv
+
+        @logger.debug("Port command: #{argv.inspect}")
+        with_target_vms(argv, single_target: true) do |vm|
+          vm.action_raw(:config_validate,
+            Vagrant::Action::Builtin::ConfigValidate)
+
+          if vm.state.id != :running
+            @env.ui.error "not running - make this a better error or use the middleware"
```
sethvargo: If we can leverage the middleware, then we don't even have to raise here - the middleware will do it

plugins/commands/port/command.rb
```diff
@@ -0,0 +1,71 @@
...
+            return 1
+          end
+
+          # This only works for vbox? should it be everywhere?
+          # vm.action_raw(:check_running,
+          #   Vagrant::Action::Builtin::CheckRunning)
+
+          if !vm.provider.capability?(:forwarded_ports)
+            @env.ui.error <<-EOH.strip
```
sethvargo: Is it better to print to the UI as an error and return 1 or raise an exception here?

plugins/commands/port/command.rb
```diff
@@ -0,0 +1,71 @@
...
+          o.separator ""
+
+          o.on("--machine-readable", "Display machine-readable output")
+        end
+
+        # Parse the options
+        argv = parse_options(opts)
+        return if !argv
+
+        @logger.debug("Port command: #{argv.inspect}")
```
mitchellh: All commands and their argvs are recorded higher level so you can safely omit this.

plugins/commands/port/command.rb
mitchellh: We probably don't need to even check if its running. Some providers I'm sure (VirtualBox included) can give us port mappings even if it isn't running, or can choose to error themselves.

mitchellh:
Looks good to me. The only other change I'd make is to add another flag which is `--guest`:

`vagrant port --guest 22`

which only outputs the host port that matches that guest port (with no other UI). This would make it easy to use in bash scripts like: `ssh -p $(vagrant port --guest 22)`

plugins/commands/port/command.rb
sethvargo: It actually prints a really terrible "no method encode for nil class" error

beporter:
So many :+1: 's. Thank you.

dduportal:
Thanks for that !

PR #6595, opened 2015-11-24
Make check_running a builtin action

sethvargo:


mitchellh:
This can't work because the "running" state for VMs is provider specific. :(

PR #6597, opened 2015-11-24
Fix the resolution of SafeEnv.

timuralp:
Recent change broke bundler.rb due to incorrect scope resolution (the
error is: uninitialized constant Vagrant::Bundler::SafeEnv
(NameError).


sethvargo:
Thanks!

Issue #2538, opened 2013-11-26
Cleanup Tasks for Shell Provisioner


jeremyfelt:
Support for cleanup tasks was added via #1302 in Vagrant 1.3.0 and only the Chef provisioner was given a built in method to handle this functionality.

It would be excellent if the built in shell provisioner also had the ability to run either an inline script or file on `vagrant destroy`.

pdxfixit:
+1 for this feature.

sjmarshy:
+1

komlenic:
+1

komlenic:
I think this should be a core feature due to its utility and varied use-cases, however using the vagrant-triggers plugin is an option.

To implement this functionality using vagrant-triggers plugin, in your vagrantfile you can add:
```ruby
config.trigger.before :destroy do
  info "Dumping the database before destroying the VM..."
  run_remote  "command"
end
```

If you prefer to use a shell script to contain all your teardown/cleanup tasks:
```ruby
config.trigger.before :destroy do
  info "Dumping the database before destroying the VM..."
  run_remote  "bash /vagrant/cleanup.sh"
end
```
... where /vagrant/cleanup.sh is a script that will exist on both the host and the guest, because vagrant shares the directory containing the vagrantfile on the host, with /vagrant on the guest by default.

Stono:
+1

clok:
+1

franklinkim:
+1

edux:
+1

akatz0813:
+1 Using vagrant with aws code deploy on-premises and really need to deregister the machine from within the vm on destroy.

iampuma:
+1

dennisgranath:
+1

kadaan:
@gtmtech FWIW internally we added support for vagrantfile specified parameters which can be specified in a similar way to triggers. We use it to allow users to pass in their credentials way earlier in `vagrant up` and to allow them to specify if they want to join the domain. The parameters end up in `vagrant up -?` just like the native parameters.  Would this coupled with vagrant-triggers accomplish what you want?

gtmtech:
[...]

PR #6603, opened 2015-11-26
Only run cleanup tasks when they are defined on the provisioner

sethvargo:
This helps with some confusion caused in GH-2538, since the output says:

> Running cleanup tasks for 'shell' provisioner...

But that's actually not true. It is running the cleanup tasks iff the provisioner defined a cleanup task. This commit changes the provisioner_cleanup middleware to only run cleanup tasks if the subclass defines a cleanup task.

The reason we can't just check if the provisioner `respond_to?` the `cleanup` method is because the parent provisioner base class (which all provisioners inherit from) defines a blank cleanup method. This is important because it means we never risk calling an unimplemented cleanup function, and it also helps define the public API for a provisioner.

### DIFF
--- a/lib/vagrant/action/builtin/provisioner_cleanup.rb
+++ b/lib/vagrant/action/builtin/provisioner_cleanup.rb
@@ -32,10 +32,20 @@ module Vagrant
 
           # Ask the provisioners to modify the configuration if needed
           provisioner_instances(env).each do |p, _|
-            env[:ui].info(I18n.t(
-              "vagrant.provisioner_cleanup",
-              name: type_map[p].to_s))
-            p.cleanup
+            name = type_map[p].to_s
+
+            # Check if the subclass defined a cleanup method. The parent
+            # provisioning class defines a `cleanup` method, so we cannot use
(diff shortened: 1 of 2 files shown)


gildegoma:
LGTM :shipit: in 1.8?

lib/vagrant/action/builtin/provisioner_cleanup.rb
```diff
@@ -32,10 +32,20 @@ def do_cleanup(env)
...
-              "vagrant.provisioner_cleanup",
-              name: type_map[p].to_s))
-            p.cleanup
+            name = type_map[p].to_s
+
+            # Check if the subclass defined a cleanup method. The parent
+            # provisioning class defines a `cleanup` method, so we cannot use
+            # `respond_to?` here. Instead, we have to check if _this_ instance
+            # defines a cleanup task.
+            if p.public_methods(false).include?(:cleanup)
```
mitchellh: RUBY!

mitchellh:
LGTM

Issue #6608, opened 2015-11-27
Vagrant cannot detect CoreOS network interface names


jharshman:
# Issue: 
Current regex in plugins/guests/coreos/cap/configure_networks.rb doesn't match on ethX interface naming convention.  Causes issues when trying to configure network interfaces in a cluster.
` comm.sudo("ifconfig | grep 'enp0\\|ens' | cut -f1 -d:") do |_, result|`

## To Reproduce
Vagrant Version 1.7.2
Libvirt Version 1.2.13.1
CoreOS LibVirt Base Box available from Atlas: 
vagrant init jharshman/coreos-stable; vagrant up --provider=libvirt
VagrantFile and supporting cloud init files can be found here:
[CoreOS Config Gist](https://gist.github.com/jharshman/97c7c16a40ddd562293a)

## Debug Gist from Vagrant Up
[Vagrant Debug Gist](https://gist.github.com/jharshman/869bf907f3074d1566ef)

## Suggested Fix (tested)
Modify regular expression to match on the ethX pattern and be a bit more detailed with matching on the other patterns
`comm.sudo("ifconfig | grep -E '(e[n,t][h,s,p][[:digit:]]([a-z][[:digit:]])?)' | cut -f1 -d:") do |_,result|`

PR #6610, opened 2015-11-27
Vagrant Issue #6608

jharshman:
# Summary
Re-factor and repair regular expression attempting to match present interfaces.
The re-factored regular expression will match on enp* ens* eth* variants.

#### Referenced Issue
Vagrant Cannot Detect CoreOS interface names #6608

### DIFF
--- a/plugins/guests/coreos/cap/configure_networks.rb
+++ b/plugins/guests/coreos/cap/configure_networks.rb
@@ -15,7 +15,7 @@ module VagrantPlugins
 
             # Read network interface names
             interfaces = []
-            comm.sudo("ifconfig | grep 'enp0\\|ens' | cut -f1 -d:") do |_, result|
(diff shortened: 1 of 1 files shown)


sethvargo:
LGTM, but I want to test this with some other guests first to make sure it doesn't regress anything.

jharshman:
From my understanding, there shouldn't be a need to test with other guest operating systems.  This should be exclusive to CoreOS or at least that is what the plugin file path leads me to believe.

sethvargo:
@jharshman there are multiple versions of coreos, so I want to make sure we arent breaking backwards compat

jharshman:
@sethvargo Gotcha.  Thanks for clarifying.

iquiw:
In #6608, the grep command is

`grep -E '(e[n,t][h,s,p][[:digit:]]([a-z][[:digit:]])?)'`

but in this PR, it is

` grep '(e[n,t][h,s,p][[:digit:]]([a-z][[:digit:]])?'`

No `-E` and no last close parenthesis.
I think the former is correct one.

PR #6612, opened 2015-11-28
Add checksum validation for the VirtualBox installs on Windows and OS X

### LINKED ISSUE #6611: VirtualBox downloads should be checksummed before installing
I noticed that in
https://github.com/mitchellh/vagrant/blob/7a1d17042d86ba64ca6f7d8b53ef9d5f9dac8909/plugins/hosts/darwin/cap/provider_install_virtualbox.rb#L13 and https://github.com/mitchellh/vagrant/blob/7a1d17042d86ba64ca6f7d8b53ef9d5f9dac8909/plugins/hosts/windows/cap/provider_install_virtualbox.rb#L14, VirtualBox is downloaded over http://, allowing for possible MITM during download.

Quoting @sethvargo in https://github.com/mitchellh/vagrant/issues/6531#issuecomment-157775095 (unrelated issue),

> "How will you verify it has the right contents? We could add some kind of checksum verification, but downloading arbitrary things from the internet and running them is usually discouraged"

The current code is downloading arbitrary things from the internet right now, sadly. Would be nice to just add `Vagrant::Util::FileChecksum` to validate the downloaded files. SHA-256 checksums for VirtualBox files are available at https://www.virtualbox.org/download/hashes/5.0.10/SHA256SUMS.

reedloden:
Use Vagrant::Util::FileChecksum to validate the downloaded VirtualBox
installers.

SHA-256 checksums for VirtualBox files are available at
https://www.virtualbox.org/download/hashes/5.0.10/SHA256SUMS.

Fixes #6611.

### DIFF
--- a/lib/vagrant/errors.rb
+++ b/lib/vagrant/errors.rb
@@ -524,6 +524,10 @@ module Vagrant
       error_key(:provider_cant_install)
     end
 
+    class ProviderChecksumMismatch < VagrantError
+      error_key(:provider_checksum_mismatch)
+    end
+
     class ProviderInstallFailed < VagrantError
       error_key(:provider_install_failed)
     end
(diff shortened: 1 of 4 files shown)


mitchellh:
Looks great. Thanks!

PR #6626, opened 2015-12-01
Ansible: Implement support for setting variables in the dynamic inventory

chenz:
As discussed in issue #6619

PR #6627, opened 2015-12-02
Review of GH-6626: Support of variables in the Ansible dynamic inventory

gildegoma:
@chenz I've started to test your #6626, which is very good! Please have a look at my little additions...

I'll test a bit more this week and  propose you to finalize it through this reviewing PR.


chenz:
Looks good to me. Nice catch with the symbols. Is there anything else I need to do?

plugins/provisioners/ansible/provisioner/guest.rb
```diff
@@ -128,6 +128,8 @@ def generate_inventory_machines
               else
                 machines += "#{machine_name}\n"
               end
+              host_vars = get_inventory_host_vars_string(machine_name)
+              machines.sub!(/\n$/, " #{host_vars}\n") if host_vars
```
gildegoma: How important is the `$`! I first think it was a bug, and cannot believe that my test was successful ;-)

In the future, I think that it might be good to append the `host_vars` values in a the `base.rb` shared parts. But to achieve that, `generate_inventory_machines` should then return an array of hosts instead of the "rendered text".

(This comment is here for the records. No change required in this review)

gildegoma:
@chenz No, nothing to do for you :) I'll merge it quite soon (especially because I see that @mitchellh is active again, and I suspect that the release is getting closer ;-)

gildegoma:
@chenz merged, thank you again very much :smiley:

PR #6652, opened 2015-12-08
Only consider the VM interfaces in the IPv6 fixup

sethvargo:
Closes GH-6596

### DIFF
--- a/plugins/providers/virtualbox/action/network_fix_ipv6.rb
+++ b/plugins/providers/virtualbox/action/network_fix_ipv6.rb
@@ -43,12 +43,12 @@ module VagrantPlugins
           # If we have no IPv6, forget it
           return if !has_v6
 
-          networks(env).each do |network|
-            next if !present?(network[:ipv6])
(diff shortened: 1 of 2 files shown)


plugins/providers/virtualbox/action/network_fix_ipv6.rb
```diff
@@ -41,24 +43,38 @@ def call(env)
...
+            .map { |_, i| i[:hostonly] if i[:type] == :hostonly }.compact
+        end
+
+        # The list of networks that are tied to a host-only adapter.
+        # @return [Array]
+        def networks(env)
+          iface_names = self.host_only_interface_names(env)
+          env[:machine].provider.driver.read_host_only_interfaces
+            .select { |network| iface_names.include?(network[:name]) }
+        end
```
phinze: Nit: I think networks is a bit misleading here - this still returns a list of host_only_interfaces

phinze:
One nit, code LGTM

PR #6659, opened 2015-12-10
Use <prefix>::1 as the IPv6 adapter IP.

### LINKED ISSUE #6658: [OS X + IPv6]: Unable to communicate with the VM over IPv6 over host-only interfaces
When provisioning an IPv6 host-only interface on OS X, the guest is unable to communicate with the host. This appears to be due to the handling of routes for addresses of the type of `2001:dead:beef::` and that is the address Vagrant chooses for the adapter. The error is `no route to host`, although there appears to be a routing table entry for `2001:dead:beef::/64 dev vboxnetN`. When changing the adapter IP to `2001:dead:beef::1`, the communication proceeds as expected.

I do  not understand the differences between the IPv6 stacks, but on Linux this issues does not manifest.

Given that there is no strong requirement to use `<prefix>::` as the IPv6 address for the adapter, I suggest changing it to `<prefix>::1`, akin to IPv4 configuration.

timuralp:
Set the IPv6 adapter IP to be <prefix>::1. Otherwise, guest to host
communication over IPv6 is not routed correctly. This means that
consumers should not specify <prefix>::1 IP addresses to VirtualBox,
which should be a reasonable restriction.

Fixes #6658

### DIFF
--- a/plugins/providers/virtualbox/action/network.rb
+++ b/plugins/providers/virtualbox/action/network.rb
@@ -290,8 +290,8 @@ module VagrantPlugins
             # Default subnet prefix length
             options[:netmask] ||= 64
 
-            # IPv6 we just mask the address and use that as the adapter
-            options[:adapter_ip] ||= ip.mask(options[:netmask].to_i).to_s
+            # Set adapter IP to <prefix>::1
+            options[:adapter_ip] ||= (ip.mask(options[:netmask].to_i) | 1).to_s
 
             # Append a 6 to the end of the type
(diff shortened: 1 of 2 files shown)


dbishop:
Specifically, not setting it to `::1` seemed to work with a Linux VirtualBox host, but not with OS X VirtualBox host.  On OS X host, all IPv6 networking was pretty janky until I gave the host-only interface on the host a trailing `...::1`.

mitchellh:
This LGTM! Thanks.

PR #6681, opened 2015-12-17
Fix for problem where the host_info.ps1 script returns a string instead of a list of IPs

tonyarkles:
This may have been due to a recent Windows Update? When doing a "vagrant up" on Windows 8.1, it was crashing due to the expected list of IPs returned from the host_info.ps1 script coming back as a string instead of a list.


mitchellh:
That is strange, going to merge this as-is since I can't see it hurting things.

PR #6688, opened 2015-12-19
fixes #2270, support env variables in puppet apply run

### LINKED ISSUE #2270: Passing Environment Variables through to Puppet Apply
Hi.  There seems to be no way to pass environment variables to the Puppet Apply process when using Vagrant.  Being able to do this would really help when behind a proxy server.  It's important to note that the proxy is _not_ needed to contact a Puppet master etc.; the proxy is needed for the VM to install things (and not just things from a Linux package repo).  At the minute I have to either litter my puppet modules and manifests with proxy settings or change the global config of the VM to make every process in the VM use the proxy server.  

It would be really helpful if Vagrant supported something like this (the environment property is the new bit): 

``` ruby
  config.vm.provision :puppet do |puppet|
    puppet.manifests_path = "puppet/manifests"
    puppet.manifest_file  = "init.pp"
    puppet.environment = {
      "http_proxy": "http://proxy:8080",
      "https_proxy": "http://proxy:8080"
    }
  end
```

The great thing about this approach is that once "vagrant up" is finished, it doesn't leave any proxy config in the VM.

Azulinho:
for a config of:
```
   config.vm.provision :puppet do |puppet|
     puppet.environment_variables = { 'FOO' => 'BAR' }
   end
```
it provides 'FOO' in the same execution context as the puppet apply run:
```
\>VAGRANT_LOG=info bundle exec vagrant provision 2>&1 | grep 'puppet apply'
 INFO ssh: Execute: FOO='BAR'   puppet apply --color=false
 --detailed-exitcodes --manifestdir
 /tmp/vagrant-puppet/manifests-a11d1078b1b1f2e3bdea27312f6ba513
 /tmp/vagrant-puppet/manifests-a11d1078b1b1f2e3bdea27312f6ba513/default.pp
```
this should fix #2270

### DIFF
--- a/website/source/downloads.html.erb
+++ b/website/source/downloads.html.erb
@@ -24,8 +24,8 @@ description: |-
       <a href="https://releases.hashicorp.com/vagrant/<%= latest_version %>/vagrant_<%= latest_version %>_SHA256SUMS.sig">
         verify the checksums signature file
       </a>
-      which has been signed using <a href="https://www.hashicorp.com/security.html" target="_TOP">HashiCorp's GPG key</a>.
-      You can also <a href="https://releases.hashicorp.com/vagrant/" target="_TOP">download older versions of Vagrant</a> from the releases service.
+      which has been signed using <a href="https://www.hashicorp.com/security.html" target="_blank" rel="nofollow noopener noreferrer">HashiCorp's GPG key</a>.
+      You can also <a href="https://releases.hashicorp.com/vagrant/" target="_blank" rel="nofollow noopener noreferrer">download older versions of Vagrant</a> from the releases service.
     </p>
   </div>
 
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @Azulinho 

Thank you for opening an PR. Unfortunately this will not merge cleanly and there appears to be other changes to the CFEngine provider as well. Could you please rebase and remove the unrelated changes or submit them in another PR? Thanks! :smile:

murphyj:
Any movement on this? Seems like the sort of thing you'd ideally want.

Is there a suggested workaround?

Azulinho:
not sure how the cfengine stuff had appeared there, probably branched out some random feature branch

dropped this PR in favour of: 
https://github.com/mitchellh/vagrant/pull/7252

PR #6700, opened 2015-12-22
v.use_linked_clone should be v.linked_clone

tomkins:
I'm assuming this got changed during development, but the documentation refers to an setting which doesn't work:

```

VirtualBox Provider:
* The following settings shouldn't exist: use_linked_clone
```


mitchellh:
Thanks!

PR #6706, opened 2015-12-22
linked_clone if Vagrant::VERSION =~ /^1.8/

kikitux:
To void this message on pre Vagrant 1.8

```bash

VirtualBox Provider:
* The following settings shouldn't exist: linked_clone
```

### DIFF
--- a/website/docs/source/v2/virtualbox/configuration.html.md
+++ b/website/docs/source/v2/virtualbox/configuration.html.md
@@ -54,6 +54,14 @@ config.vm.provider "virtualbox" do |v|
 end
 ```
 
+To have backward compatibility:
+
+```ruby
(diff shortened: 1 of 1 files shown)


website/docs/source/v2/virtualbox/configuration.html.md
```diff
@@ -54,6 +54,14 @@ config.vm.provider "virtualbox" do |v|
 end
 ```
 
+To have backward compatibility:
```
sethvargo: @kikitux good idea, but I think we want to encourage `Vagrant.require_version` instead.

mitchellh: I think we can mention both, I'll merge this as is for now.

PR #6714, opened 2015-12-22
Convert env values to strings before escaping

sethvargo:
Fixes https://github.com/mitchellh/vagrant/issues/6707#issuecomment-166704292


mitchellh:
LGTM

Issue #6705, opened 2015-12-22
.bundle directory is created to current directory when using Vagrant 1.8.0


tuminoid:
Vagrant 1.8 creates `.bundle` directory when using any vagrant command to current directory.

arthuroy:
Vagrant 1.8 also generates a `nul` file.

lmayorga1980:
I can see the same behavior. OSX El Capitan

PHLAK:
I'm also getting the .bundle file when issuing vagrant commands.  
  * Vagrant v1.8.0
  * Ubuntu 15.10

josephfrazier:
Using bundler version 1.10.5, I was able to `git bisect start v1.8.0 v1.7.4` the bug down to commit 2ade66443e776bf637f96a74d9d1231b5eadd44c, with the following script passed as `./bisect.sh` to `git bisect run ./bisect.sh`:

```shell
# ./bisect.sh
bundle install --path vendor/bundle
rm -rf .bundle
bundle exec vagrant
! ls -d .bundle
```

It looks like https://github.com/mitchellh/vagrant/commit/2ade66443e776bf637f96a74d9d1231b5eadd44c#diff-04eb3e04ebc04adf4ab7a99c0bc923d7R47 is the first time `::Bundler.settings` has been used, which is what's prompting it to create `.bundle/config`. I'm not very familiar with bundler, but perhaps there's a way to change where the settings are stored? For what it's worth, https://github.com/bundler/bundler/commit/83782981216ccc4f9143ec7e518687c54bc29307 introduces default settings (including `:retry => 3`) to bundler v1.10.0 and up, so maybe vagrant could rely on that instead?

geerlingguy:
+1 - I had to add `.bundle` to my project's gitignore just to prevent people from accidentally adding it in.

thachmai:
+1

Just updated to vagrant 1.8 on El Capitan. The .bundle directory is created immediately after the "vagrant up" command.

wizwjw:
I have the same issue.

mitchellh:
This is fixed, will be completely fixed in 1.8.1

PR #6717, opened 2015-12-23
[WIP] core: use environment variable for bundler retry #6705

josephfrazier:
(This attempts to address https://github.com/mitchellh/vagrant/issues/6705#issuecomment-166768147. It's marked WIP since it doesn't yet clean up the environment after itself. Also, I'm not sure if more tests are required, or if this is even the best way to approach the problem.)

This prevents a .bundle/config file from being written.

See 2ade664 for the original implementation.

### DIFF
--- a/lib/vagrant/bundler.rb
+++ b/lib/vagrant/bundler.rb
@@ -45,7 +45,9 @@ module Vagrant
       end
 
       # Configure Bundler to retry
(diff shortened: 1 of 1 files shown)


mitchellh:
This looks great! I think we can actually put this in `init!`. We don't clean up other bundler env vars so this should be fine.

PR #6722, opened 2015-12-23
Fix Docker provisioning for Debian/Jessie box

KaffeJunky123:
Docker provisioning fails on Debian Jessie without the apt-transport-https package installed.


mitchellh:
Thanks!

Issue #6653, opened 2015-12-08
Networking configuration fails on Ubuntu with systemd > v197


gMagicScott:
In Ubuntu 15.10, systemd now uses "predictable network interface names".

When Vagrant sets up the networking during `vagrant up`, it ends with an error that `/sbin/ip addr flush dev eth#{interface} 2> /dev/null` fails.

For network setup, currently Ubuntu falls back to the Debian networking capability. The template being used hard codes `eth` followed by a number from the `options[:interface]`. When I tried it, it compiled to `eth2`.

The command is failing because there is no interface `eth2`, there aren't any that start with `eth`. In my test, the interface you would want is `enp0s3`.

The only OS that's using these predictable interface names that I know of is Ubuntu 15.10, but there are probably more, or will be in the near future. You can read more on how the predictable names are determined [here](https://wiki.freedesktop.org/www/Software/systemd/PredictableNetworkInterfaceNames/).

I saw this issue was reported in boxcutter/ubuntu#34, but I don't think it's an issue for the base box to solve.

sethvargo:
Hi @gMagicScott 

This has been fixed in the master branch. I don't recall the other issue, but I know we fixed it. I was able to successfully bring up an ubuntu 15.10 box just now with virtualbox (using `ubuntu/wily64`) without error. Sorry I don't recall the exact issue, but this is fixed and will be part of Vagrant 1.8 :smile: 

```
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'ubuntu/wily64'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'ubuntu/wily64' is up to date...
==> default: A newer version of the box 'ubuntu/wily64' is available! You currently
==> default: have version '20151120.0.0'. The latest is version '20151203.0.0'. Run
==> default: `vagrant box update` to update.
==> default: Setting the name of the VM: scratch_default_1449611249397_72470
==> default: Clearing any previously set forwarded ports...
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
==> default: Forwarding ports...
    default: 22 (guest) => 2222 (host) (adapter 1)
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
    default: Warning: Remote connection disconnect. Retrying...
[...]

Issue #2648, opened 2013-12-13
use_dhcp_assigned_default_route doesn't work for Debian 7


taurus-forever:
Hi,

The bridged interface option `:use_dhcp_assigned_default_route=>true` doesn't work in Debian 7. The option was added here: https://github.com/mitchellh/vagrant/pull/868 .

We have two interfaces on VM:
  eth0 - VirtualBox NAT interface
  eth1 - VirtualBox Bridged interface
VM always use eth0 as default gateway route (use_dhcp_assigned_default_route option changes nothing).

It doesn't work even with workaround for Ubuntu 12.07: https://github.com/mitchellh/vagrant/pull/947 and with patch from https://github.com/nickrw/vagrant/commit/a0b684a4ab4a447d408b083c7d62d5f6b5e7d817

The only one way when it works well is (in exactly this order):
`post-up route del default dev eth0
post-up dhclient $IFACE
pre-down route add default dev eth0`

The problem is that Debian 7 use the default route from the first dhcp reply (eth0 replies first, so it ignores gw from dhcp reply for eth1 until you not remove the eth0 default route explicitly).

Please consider to commit the workaround above to `templates/guests/debian/network_dhcp.erb`. Thank you!

Issue #6683, opened 2015-12-17
Ubuntu Wily (15.10) vagrant ssh times out


LukeHoersten:
### Summary
Ubuntu Wily vagrant box times out when trying to `vagrant ssh` into it. Works fine with `ubuntu/vivid64` and before.

### Versions
- **Host:** OS X 10.11.2
- **Guest:** Ubuntu Wily 15.10
- **VirtualBox:** 5.0.10 r104061
- **Vagrant:** 1.7.4

### Vagrantfile
```ruby
Vagrant.configure(2) do |config|
  config.vm.box = "ubuntu/wily64"
end
```

`vagrant up --debug` output [available in this pastebin](http://lpaste.net/2317235732088881152).

### Investigation
My suspicion is that in Wily, Ubuntu started defaulting to a network bridge to linux container and this is overriding the necessary vagrant routing.

I've confirmed the VirtualBox port forwarding from Host port 2222 to Guest port 22 is configured correctly.

#### Vivid routing tables
```
# route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         10.0.2.2        0.0.0.0         UG    0      0        0 eth0
10.0.2.0        0.0.0.0         255.255.255.0   U     0      0        0 eth0
```

#### Wily routing tables
```
# route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
10.0.3.0        0.0.0.0         255.255.255.0   U     0      0        0 lxcbr0
```

This prevents the `sshd` form even seeing the connection attempt.

### Other issues with ssh timeout symptoms
This is different than another issue people were seeing with the same symptom of ssh connection timeouts. That issue was specific to 64bit virtualization being disabled at the BIOS level. This is not the issue here because `ubuntu/vivid64` box works just fine.

*Thanks to `ada` on IRC for helping debug this issue.*

adamancini:
I can duplicate this issue on OS X 10.11.2, Vagrant 1.7.4, vbox 5.0.10 r104061

Perni1984:
no time to check this right now, but could it be connected to the following issue on ubuntu/wiley64:
https://gist.github.com/brbsix/b70413dec907906ef659

LukeHoersten:
@Perni1984 you're brilliant. That was the problem. The fix worked at the bottom of the [gist you posted](https://gist.github.com/brbsix/b70413dec907906ef659):

[...]

Issue #3957, opened 2014-06-03
Request: Debian source network interfaces


ifnull:
Please consider using source interfaces for configuring network devices in Debian systems. 

```
source /etc/network/interfaces.d/*.cfg
```

https://github.com/mitchellh/vagrant/blob/master/plugins/guests/debian/cap/configure_networks.rb#L17

Reference:
https://bugs.launchpad.net/ubuntu/+bug/1262951

From Ubuntu Trusty:

```
# Source interfaces
# Please check /etc/network/interfaces.d before changing this file
# as interfaces may have been defined in /etc/network/interfaces.d
# NOTE: the primary ethernet device is defined in
# /etc/network/interfaces.d/eth0
# See LP: #1262951
```

PR #6724, opened 2015-12-23
Add networking cap for Ubuntu guest

gMagicScott:
In issue #6653, configuring custom networking failed due to changes in `systemd` in Ubuntu 15.10. Vagrant defaulted to falling back to Debian configuration which hard coded `eth` and a digit to configure the networks.

In this pull request, the Debian configuration was copied to Ubuntu, checks to see if the the `eth` networks are available. If so, defer to the Debian setup.

Otherwise we lookup the open interfaces and configure them with the new "predictable names" (that can be oddly hard to predict).

One of the templates copied from Debian mentioned issue #2648. I'm not sure if it is something that is applicable to Ubuntu or the new naming schemes. I left it in since I'm not sure how to test the issue.


gMagicScott:
Somewhat related is #6683. When I tested this, I was using the [boxcutter/ubuntu1510](https://atlas.hashicorp.com/boxcutter/boxes/ubuntu1510), whose Packer build can be [found here](https://github.com/boxcutter/ubuntu/blob/master/ubuntu1510.json).

When I tested against the **ubuntu/wily64**, I still get the ssh timeout (Connection reset by peer) error. I looked for the script that builds the Ubuntu official box, but wasn't able to find it. I'm not sure what the difference is between the two setups as to why ssh fails on the official box.

I have an idea that may work, but I'm not sure how or where to hook into the Vagrant execution to run some VBoxManage commands between boot and the initial ssh connection.

brbsix:
Do you think this PR would be a good opportunity to integrate #3957 (specifically checking `/etc/network/interfaces.d` on Ubuntu) or is that not really an issue? I'm not sure why, but https://github.com/mitchellh/vagrant/commit/e99253be6eb418ba02d66997c4643c50377c19e3 was never commited to master.

gMagicScott:
Does anyone know when the feature was introduced into Debian / Ubuntu to be able to split network interfaces into files? It's not there from the beginning. 

If vagrant would split the network config into files, it would need to be conditionally around a version test, unless support for really old OSs doesn't matter.

gabriel403:
Any news on what's happening with this? Really such in a hole atm.

gabriel403:
@gMagicScott do you know of anyway we can use this without having to wait for the pr to be merged?

gMagicScott:
@gabriel403 I tried putting this into a plugin but I couldn't get it to work. 

Currently I'm using my fork with the patch. There is documentation in the Readme and in the wiki of this repo on how to run the dev version of vagrant.

gabriel403:
@gMagicScott thanks for the help, shame you couldn't get the plugin working 😀

PR #6749, opened 2015-12-27
use SafeExec to capture history via up/down keys in powershell command

mwrock:
`Subprocess` sets up the process in such a way that pressing the up/down arrows does not cycle through command history in the `powershell` command. Using `SafeExec` instead as the `ssh` command does seems to fix this.

### DIFF
--- a/plugins/hosts/windows/cap/ps.rb
+++ b/plugins/hosts/windows/cap/ps.rb
@@ -1,7 +1,7 @@
 require "pathname"
 require "tmpdir"
 
-require "vagrant/util/subprocess"
+require "vagrant/util/safe_exec"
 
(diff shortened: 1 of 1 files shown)


mwrock:
Any chance of getting this into 1.8.2 @sethvargo ? Been using it consistently without issue on a couple machines.

mkuzmin:
Hi Matt

I applied this change on top of my 1.8.1 installation, but see no difference. Are there any other requirements?

mwrock:
Hmm. I refreshed my vagrant install from master a couple days ago and I do see the improved behavior. I'm on a windows 10 host. Its possible other hosts might behave differently but its not really easy for me to try on a different host.

PR #6756, opened 2015-12-29
Removed a tiny strange char `U+05BB`

OmarIthawi:
http://www.fileformat.info/info/unicode/char/05bb/index.htm

This prevented me from a quick copy-paste solution for the Ubuntu sudo-less vagrant up:

 - https://docs.vagrantup.com/v2/synced-folders/nfs.html


sethvargo:
Good find

OmarIthawi:
Thanks Seth

PR #6760, opened 2015-12-29
Change to use pkgng

petems:
Old pkg_* is now EOL: https://lists.freebsd.org/pipermail/freebsd-ports-announce/2014-February/000077.html

### DIFF
--- a/plugins/guests/freebsd/cap/rsync.rb
+++ b/plugins/guests/freebsd/cap/rsync.rb
@@ -8,10 +8,7 @@ module VagrantPlugins
             version = result.split('.')[0].to_i if type == :stdout
           end
 
-          pkg_cmd = "pkg_add -r"
-          if version && version >= 10
-            pkg_cmd = "pkg install -y"
-          end
+          pkg_cmd = "pkg install -y"
 
           machine.communicate.sudo("#{pkg_cmd} rsync")
         end


sethvargo:
Hi @petems 

It may be EOL, but Vagrant will still carry support for those older operating system versions, so I don't think we can introduce this change.

petems:
@sethvargo I might be wrong, but I don't think it works anymore as the old packaging style no longer exists. This is from using the latest FreeBSD images provided:

```

pkg_add -r rsync

Stdout from the command:

```
Error: Unable to get ftp://ftp.freebsd.org/pub/FreeBSD/ports/amd64/packages-9-stable/Latest/rsync.tbz: File unavailable (e.g., file not found, no access)
```

Using the new package style still exists:

```
# pkg install -y rsync
Updating FreeBSD repository catalogue...
FreeBSD repository is up-to-date.
All repositories are up-to-date.
The following 1 package(s) will be affected (of 0 checked):

New packages to be INSTALLED:
	rsync: 3.1.2_1

The process will require 569 KiB more space.
290 KiB to be downloaded.
Fetching rsync-3.1.2_1.txz: 100%  290 KiB 297.8kB/s    00:01
Checking integrity... done (0 conflicting)
[1/1] Installing rsync-3.1.2_1...
[1/1] Extracting rsync-3.1.2_1: 100%
```

Maybe changing the command to `pkg install -y rsync && pkg_add -r rsync` would meet in the middle?

sethvargo:
Hi @petems 

What is the latest version of freebsd? The code that I'm reading says it should be using `pkg install`, not `pkg_add`.

What version of Vagrant are you using?

petems:
@sethvargo This is on FreeBSD 9, that code is live on master here: https://github.com/mitchellh/vagrant/blob/master/plugins/guests/freebsd/cap/rsync.rb#L11

sethvargo:
@petems correct, the code is saying that pkg install is only for FreeBSD 10+ (which seems to be latest). I don't know enough about the FreeBSD package ecosystem to know if this is a safe change (i.e. how will this affect fbsd 7, 6, etc?)

jungle-boogie:
hi @sethvargo,

Your concerns are legitimate but invalid.

https://www.freebsd.org/security/security.html#sup

As you see from the above link, 9-stable, 9.3-release, 10.1-release, 10.2-release will EOL end of this year.

This link is also relevant:
https://www.freebsd.org/security/unsupported.html

7.x and 7-stable have been EOL since at least February 2013.

6.x and 6-stable have been EOL since November 2010.

Hope this helps!

jungle-boogie:
Additionally, this post from October 2013 is relevant, too:
https://lists.freebsd.org/pipermail/freebsd-pkg/2013-October/000107.html

The pkg_install suite of tools pkg_create(1), pkg_add(1), and
pkg_info(1) (which ports also use), are deprecated and will be
discontinued in roughly 6 months. A communication regarding the
deprecation of the pkg_install suite of tools will be sent separately in
the future.

jungle-boogie:
The command pkg_add exists in 9.3 but as already pointed out by @petems, pkg_ tools have been deprecated: 
https://www.freebsd.org/cgi/man.cgi?query=pkg_add&apropos=0&sektion=0&manpath=FreeBSD+9.3-RELEASE&arch=default&format=html

The same command doesn't exist in 10.0+:
https://www.freebsd.org/cgi/man.cgi?query=pkg_add&apropos=0&sektion=0&manpath=FreeBSD+10.0-RELEASE&arch=default&format=html

petems:
@sethvargo Can we get this merged? :+1:

jungle-boogie:
Yes, please merge and support pkg!

PR #6820, opened 2016-01-07
Add quoting for ansible's `extra_vars`

### LINKED ISSUE #6726: Vagrant 1.8, ansible_local `extra_vars` being ignored
Porting applications to use the lovely new `ansible_local`, and running into a few issues.

In my Vagrantfile, I have:

```
config.vm.provision "ansible_local" do |ansible|
    ansible.provisioning_path = "/my-app/playbooks"
    ansible.playbook = "site.yml"
    ansible.inventory_path = "development"
    ansible.extra_vars = {
      is_vagrant: true,
      should_seed: true,
      source: "/my-app"
    }
    ansible.verbose = "vvvv"
  end
```

Values have been scrubbed and simplified. Many of them actually inspect environment variables to determine there value, preventing me from moving them to my inventory group_var file.

default values defined in `group_vars/all` is `is_vagrant=false` and `should_seed=false`

which generates (what looks correctly): 

```bash
cd /my-app/playbooks && PYTHONUNBUFFERED=1 ANSIBLE_FORCE_COLOR=true ansible-playbook --limit='default' --inventory-file=development --extra-vars={"is_vagrant":true,"should_seed":true,"source":"/cirro-legacy"} -vvvv site.yml
```

But then I get (in a dummy task created to display variable output):
```
TASK: [common | display vars] *************************************************
<default> REMOTE_MODULE command echo "is_vagrant=False should_seed=False source=/cirro-legacy" #USE_SHELL
```

I can work around this temporarily, but *something* seems amiss.

kamazee:
`--start-at-task` and `--limit` are quoted, so is `extra_vars` as of the commit referenced below.

This seems to fix #6726 
I know absolutely no ruby, but hope that things I ended up with after some googling are not too scary.
I also modified a test that checked quoting to expect `extra-vars` quoted as well (and made `extra-vars` contain a hashtable as checking just a filename seemed pretty useless to me). I also supported nested single quotes for all 3 options that are quoted now.

This fix is checked with `ansible` provisioner (as it might have appeared broken because of the suggested fix). I'm happy to report that `ansible` did very same job as `ansible-local`  provisioning my guest machine.

### DIFF
--- a/plugins/provisioners/ansible/helpers.rb
+++ b/plugins/provisioners/ansible/helpers.rb
@@ -15,8 +15,8 @@ module VagrantPlugins
 
         shell_arg = []
         command.each do |arg|
-          if arg =~ /(--start-at-task|--limit)=(.+)/
-            shell_arg << "#{$1}='#{$2}'"
+          if arg =~ /(--start-at-task|--limit|--extra-vars)=(.+)/
+            shell_arg << "%s='%s'" % [$1, $2.gsub("'", %q(\\\'))]
           else
             shell_arg << arg
           end
--- a/test/unit/plugins/provisioners/ansible/provisioner_test.rb
+++ b/test/unit/plugins/provisioners/ansible/provisioner_test.rb
@@ -741,7 +741,7 @@ VF
         ssh_info[:private_key_path] = ['/my/key1', '/my/key2']
 
         # command line arguments
-        config.extra_vars = "@#{existing_file}"
+        config.extra_vars = {test: "string with 'apostrophes' and ="}
         config.sudo = true
         config.sudo_user = 'deployer'
         config.verbose = "vvv"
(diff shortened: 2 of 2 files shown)


gildegoma:
@kamazee sorry for the delay to review this. Your solution is actually not working on the following aspects:

* In your example the generated `--extra-vars='{test: "string with 'apostrophes' and ="}'` is accepted by `ansible-playbook` but the single quotes (') around the _apostrophes_ word are trimmed, and the final result is `string with apostrophes and =`
* with single quotes enclosure, there is no way to pass escaped values (e.g. an escaped double quote character (`\"`)).

While reviewing and testing this, I've created an alternative pull request (#7103), which is based on double quotes. I'd really love to get your feedback about this one. For the moment this feature branch is unfortunately not based on your initial work, but this can be arranged if you want to :smile:

kamazee:
@gildegoma you're right. Not sure why I didn't pay attention to it (especially given that I kind of knew it).
Will check your fix against my setup ASAP (but not later than tomorrow evening) and will post the findings into #7103 (although I doubt there will be any).

Thanks!

gildegoma:
@kamazee :+1: :heart: :heart: :heart:

Issue #936, opened 2012-05-17
Add feature to mount folders post-provisioning?


tmotley:
I don't think this feature or config setting exists having perused the docs and support.  My Vagrant file mounts shared folders, and assigns them a group and user ownership. However, if the case of a new box, the user and group I assign the mounted folder ownership do not exist until the chef provisioning is done on the box.  I tried to control the order of provisioning and mounting, but the mounting is always attempted first, which obviously breaks if the provisioning has not occurred yet.

For example:

config.vm.define :dpils_slave_1, :primary => false do |cfg|
    cfg.vm.forward_port 8080, 8081
    #cfg.vm.network :hostonly, "1.1.1.11"
    cfg.vm.provision :chef_solo do |chef|
      chef.cookbooks_path = ["chef/cookbooks", "chef/site-cookbooks", "chef/site-cookbooks-tk"]
      chef.roles_path = ["chef/roles-tk"]
      chef.add_role "jboss"
      # chef.log_level = :debug
    end

```
cfg.vm.share_folder 'crx_node_2', '/foo/bar/dpils_repository/dpils_whq_02/crx', 'foo/bar/crx_node_2', :owner => "jboss", :group => "jboss", :create => true
```

  end

Which gives this error on a clean vagrant load:

[dpils_slave_1] -- v-csr-4: /tmp/vagrant-chef-1/chef-solo-4/roles
[dpils_slave_1] -- v-csc-1: /tmp/vagrant-chef-1/chef-solo-1/cookbooks
[dpils_slave_1] -- v-csc-2: /tmp/vagrant-chef-1/chef-solo-2/cookbooks
[dpils_slave_1] -- v-csc-3: /tmp/vagrant-chef-1/chef-solo-3/cookbooks
[dpils_slave_1] -- crx_node_2: /nike/vault/dpils_repository/dpils_whq_02/crx
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

mount -t vboxsf -o uid=`id -u jboss`,gid=`id -g jboss` crx_node_2 /nike/vault/dpils_repository/dpils_whq_02/crx

Ghoughpteighbteau:
I hit this issue as well. Deving a server applet for tomcat, would like to mount the webapps folder so I can do partial compiles from my dev environment, but I can't mount a webapps folder that doesn't exist. Doc's had no answers for me. This is a needed feature. 

+1

danomatic:
[...]

PR #6836, opened 2016-01-09
Allow overriding the inferred name from `config.vm.synced_folder`

### LINKED ISSUE #6835: Ability to prepare, but not enable, a synced folder
I want to make a Vagrantfile where the provisioner runs before the synced folder is mounted. This is similar to #936 and #2701. In those issues, some workarounds were mentioned, but I haven't managed to get them to work for myself.

I need to have some setup run before the folder is mounted because I need the synced folder to be inside another folder that doesn't exist in the Virtual Machine until after my provisioning. I either need the option for synced folders to be done after provisioning or I could also use preparing the synced folder and disable the auto mount so I can use my provisioner to call `mount` manually.

I tried to use `disabled: true` but this seems to cause the folder to never be prepared. I tried some of the advice mentioned [here](https://github.com/mitchellh/vagrant/issues/936#issuecomment-23467018), but the following don't work:

```
# requires at least two arguments
config.vm.synced_folder "."
# the guestpath is used as the name and the guestpath, causes an error when booting
config.vm.synced_folder ".", nil
# sets the name correctly, but errors out because the guestpath gets used at some point
options = config.vm.synced_folder ".", "name"
options[:guestpath] = nil
```

Is there anyway to do what I want?

I am using Vagrant 1.8.1.

jsternberg:
If a `name` is specified as an option, it will be used as the id instead
of inferring it from the `guestpath`. If `guestpath` is nil, the `name`
needs to be specified so the folder can be mounted with a name.

This also fixes the synced folder code to allow `guestpath` to be nil.
It was allowed in a previous version for the purpose of preventing a
directory from being auto mounted (#936), but seems to have become an
error at some point after that.

An example of modifying the default /vagrant folder so it doesn't
auto-mount anymore:

```
config.vm.synced_folder ".", nil, name: "/vagrant"
```

An example of sharing another folder, but not auto-mounting it:

```
config.vm.synced_folder ".", nil, name: "foobar"
```

Fixes #6835.

### DIFF
--- a/plugins/kernel_v2/config/vm.rb
+++ b/plugins/kernel_v2/config/vm.rb
@@ -198,6 +198,7 @@ module VagrantPlugins
       #   folder.
       # @param [Hash] options Additional options.
       def synced_folder(hostpath, guestpath, options=nil)
+        name = (options && options.delete(:name)) || guestpath
         if Vagrant::Util::Platform.windows?
           # On Windows, Ruby just uses normal '/' for path seps, so
           # just replace normal Windows style seps with Unix ones.
@@ -214,7 +215,7 @@ module VagrantPlugins
         # Make sure the type is a symbol
         options[:type] = options[:type].to_sym if options[:type]
 
-        @__synced_folders[options[:guestpath]] = options
+        @__synced_folders[name] = options
       end
 
       # Define a way to access the machine via a network. This exposes a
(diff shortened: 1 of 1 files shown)


jsternberg:
@sethvargo is it possible to review this PR in relation to #6835? It's very important in terms of being able to mount a shared directory to an alternate location.

PR #6847, opened 2016-01-11
Bump net-scp dependency version

anatol:


sethvargo:
Hi @anatol 

Thank you for opening a pull request. What is the motivation for bumping this dependency? Is there a security vulnerability or a feature that we need? In general, we are very pessimistic about version constraints in Vagrant because we have been bitten in the past by tiny "patch" upgrades that introduce breaking changes.

sethvargo:
Ping @anatol

sethvargo:
Hi there,

I am going to close this due to lack of response. If you would like to see this merged, please submit a new pull request and include the answers to the questions above. Thanks! 😄

anatol:
> the motivation for bumping this dependency?
The motivation is to keep the project up-to-date and avoid potential diamond dependency problems.

> tiny "patch" upgrades that introduce breaking changes
If such change caused a problem in a released version then it is definitely a "smell" that the project does not have a good test coverage.

PR #6848, opened 2016-01-11
Bump nokogiri dependency to the latest stable version

anatol:


sethvargo:
Hi @anatol 

Thank you for opening a pull request. What is the motivation for bumping this dependency? Is there a security vulnerability or a feature that we need? In general, we are very pessimistic about version constraints in Vagrant because we have been bitten in the past by tiny "patch" upgrades that introduce breaking changes.

anatol:
https://github.com/mitchellh/vagrant/issues/6766

NewyorkDev:
I'm getting the same error - it should be updated.

noko

jroweboy:
@sethvargo With the latest version of vagrant, nokogiri is now broken on windows. Nokogiri <1.6.7 does not support ruby 2.2 on windows and fails with the error that @NewyorkDev posted. I'm using a vagrant plugin that depends on nokogiri and vagrant, but I can't update the nokogiri version in the plugin since it creates incompatible dependencies between vagrant's nokogiri version and the changed version in the plugin.

Is there a reason why nokogiri is listed as a dependency in vagrant at all? I can't seem to find any references to it in the codebase. I just really need this problem resolved either by updating to >1.6.7, or by removing it as a dependency, or by doing something else.

StefanScherer:
The vagrant-vcloud plugin also suffers from this issue. I first thought it was an issue inside the plugin frapposelli/vagrant-vcloud#122

taliesins:
Any chance of getting vagrant 1.8.2 released with this patch?

dbarrett84:
+1 
Have to use version 1.7.4 currently but need some of the bug fixes from 1.8.1 for other issues.

PrahlM93:
+1

shaned24:
+1

jeverton:
I think this is affecting vagrant-azure, too.  Looking forward to getting this fix.

bheuvel:
Also vagrant-cloudstack is affected. Any idea when a new msi build is expected?

As using Vagrant as a gem is not the way to go but is something I could build myself, is there some way (script/doc) I could make my own (unsupported/development) msi/distribution of current master?

sethvargo:
https://github.com/mitchellh/vagrant/wiki/Installing-Vagrant-from-Source

bheuvel:
@sethvargo Thanks!
(And sorry you had to past a link to the wiki :pensive:, I overlooked the section on binstubs )

bheuvel:
@sethvargo As I need Vagrant from Kitchen, the binstubs (shebang (`#!/`)) don't really work in that combo (Windows, Kitchen, Vagrant). Which is why I made a PR at kitchen-vagrant. But would love your view on that https://github.com/test-kitchen/kitchen-vagrant/issues/218 if possible.

PR #6850, opened 2016-01-11
[WIP] Stop the Docker host VM on vagrant halt

cweagans:
Pretty rough, but this is kind of what I'm thinking as a way to solve https://github.com/mitchellh/vagrant/issues/6832

Feedback and/or assistance welcome!

### DIFF
--- a/plugins/providers/docker/action.rb
+++ b/plugins/providers/docker/action.rb
@@ -97,6 +97,7 @@ module VagrantPlugins
             end
 
             b2.use Stop
+            b2.use HostMachineStop
           end
         end
       end
@@ -305,6 +306,7 @@ module VagrantPlugins
       autoload :HostMachinePortChecker, action_root.join("host_machine_port_checker")
       autoload :HostMachinePortWarning, action_root.join("host_machine_port_warning")
       autoload :HostMachineRequired, action_root.join("host_machine_required")
+      autoload :HostMachineStop, action_root.join("host_machine_stop")
       autoload :HostMachineSyncFolders, action_root.join("host_machine_sync_folders")
       autoload :HostMachineSyncFoldersDisable, action_root.join("host_machine_sync_folders_disable")
       autoload :InitState, action_root.join("init_state")
--- a/plugins/providers/docker/action/host_machine_stop.rb
+++ b/plugins/providers/docker/action/host_machine_stop.rb
@@ -1,0 +1,62 @@
+require "log4r"
+
+module VagrantPlugins
+  module DockerProvider
+    module Action
+      # This action is responsible for creating the host machine if
+      # we need to. The host machine is where Docker containers will
+      # live.
+      class HostMachineStop
+        def initialize(app, env)
+          @app    = app
+          @logger = Log4r::Logger.new("vagrant::docker::hostmachine")
+        end
+
+        def call(env)
+          host_machine = env[:machine].provider.host_vm
+
+          begin
+            env[:machine].provider.host_vm_lock do
+              stop_host_machine(host_machine, env)
+            end
+          rescue Vagrant::Errors::EnvironmentLockedError
+            sleep 1
+            retry
+          end
+
+          @app.call(env)
+        end
+
+        protected
+
+        def stop_host_machine(host_machine, env)
+          # Create a UI for this machine that stays at the detail level
+          proxy_ui = host_machine.ui.dup
+          proxy_ui.opts[:bold] = false
+          proxy_ui.opts[:prefix_spaces] = true
+          proxy_ui.opts[:target] = env[:machine].name.to_s
+
+          # Reload the machine so that if it was created while we didn't
+          # hold the lock, we'll see the updated state.
+          host_machine.reload
+
+          # If the machine is running, then we can stop it.
+          if host_machine.communicate.ready?
+            env[:machine].ui.detail(
+              I18n.t("docker_provider.host_machine_stopping"))
+            env[:machine].ui.detail(" ")
+            host_machine.with_ui(proxy_ui) do
+              host_machine.action(:halt)
+            end
+
+            # If we can still communicate with the machine, we have a problem.
+            if host_machine.communicate.ready?
+              raise Errors::HostVMCommunicatorNotHalted,
+                id: host_machine.id
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/plugins/providers/docker/errors.rb
+++ b/plugins/providers/docker/errors.rb
@@ -17,6 +17,10 @@ module VagrantPlugins
         error_key(:host_vm_communicator_not_ready)
       end
 
+      class HostVMCommunicatorNotHalted < DockerError
+        error_key(:host_vm_communicator_not_halted)
+      end
+
       class ImageNotConfiguredError < DockerError
         error_key(:docker_provider_image_not_configured)
       end
--- a/templates/locales/providers_docker.yml
+++ b/templates/locales/providers_docker.yml
@@ -33,6 +33,9 @@ en:
     host_machine_starting: |-
       Vagrant will now create or start a local VM to act as the Docker
       host. You'll see the output of the `vagrant up` for this VM below.
+    host_machine_stopping: |-
+      Vagrant will now stop the local Docker host VM. You'll see the output
+      of `vagrant halt` for this VM below.
     host_machine_syncing_folders: |-
       Syncing folders to the host VM...
     logging_in: |-
@@ -162,6 +165,13 @@ en:
 
         If this does not fix it, please verify that the host VM provider
         is functional and properly configured.
+
+        Host VM ID: %{id}
+      host_vm_communicator_not_halted: |-
+        The Docker provider was not able to stop the host VM successfully, and
+        the host VM is still reporting that SSH is available. The ID of the host
+        VM is shown below for convenience. You can attempt to manually stop the
+        host VM by running `vagrant halt %{id}`.
 
         Host VM ID: %{id}
       package_not_supported: |-
--- a/website/docs/source/v2/docker/basics.html.md
+++ b/website/docs/source/v2/docker/basics.html.md
@@ -109,7 +109,7 @@ Vagrant.configure("2") do |config|
 end
 </pre>
 
-The host VM will be spun up at the first `vagrant up` where the provider
-is Docker. To control this host VM, use the
-[global-status command](/v2/cli/global-status.html)
-along with global control.
+The host VM will be created and started at the first `vagrant up` where the
+provider is Docker. This machine will also be stopped when `vagrant halt` is
+run. For any other control of this machine, use the
+[global-status command](/v2/cli/global-status.html) along with global control.


sethvargo:
Hi @cweagans 

As I said in 6832, the challenging part here is that the boot2docker VM is _global_. If you have multiple Vagrantfiles that use Docker, we only spin up one instance of boot2docker. This PR will destroy that virtual machine, even if other docker-based Vagrantfiles on the system are still using it. 

This is why we clearly call this out in the documentation (https://docs.vagrantup.com/v2/docker/basics.html - see "Host VM"). 

> The host VM will be spun up at the first vagrant up where the provider is Docker. To control this host VM, use the global-status command along with global control.

As I said in my comments on 6832, Vagrant intentionally masks a lot of magic to make for an amazing development experience. For power users, the global control commands exist to provide a more controlled experience.

cweagans:
1. Why is it global in the first place? That doesn't seem like a good decision to me when everything else about a vagrant environment is more or less isolated from the rest of the environments on the host. As an example, I use the same computer for both personal and day job work stuff. My day job has a Core OS vagrant machine that is pre-loaded with some credentials that should only be accessible when I'm working on day job stuff. Disk space is cheap, so I wouldn't mind having multiple vms sitting around to run my docker containers.

2. There are other circumstances to consider here too, though. For instance, in the same example as above, I have the `vagrant_vagrantfile` option added to my containers. When the first container tries to start, it first creates the Docker host VM that's *specifically for this set of containers*. When I run `vagrant halt`, then, I'd expect that machine to shut down as well, but it doesn't.

I see where you're coming from though - if someone just wants to run containers, having b2d running globally probably makes sense. Would it make sense, though, to shut the machine down if there are no more containers running on it after the last `vagrant halt`? Something like this:

* `cd project1 && vagrant up`
  * Host VM is created and started, containers are started
* `cd ../project2 && vagrant up`
  * Containers are started on the existing host vm
* `cd ../project3 && vagrant up`
  * Same as ^
* `cd ../project1 && vagrant halt`
  * Containers are stopped. Host VM remains running
* `cd ../project3 && vagrant halt`
  * Same ^
* `cd ../project2 && vagrant halt`
  * Containers are stopped, and since vagrant sees no more containers running in that VM, the host VM is stopped too.

sethvargo:
Hi @cweagans 

As I said before, it is global to make for a really amazing development experience. We want developers to feel as though Docker is running natively on their local machine. In fact, if you run Vagrant on an OS that supports Docker, it will just just run docker directly and _not_ spin up a VM. I understand this is outside of what one may typically think of Vagrant's behavior, which is why we clearly document the behavior.

I am failing to understand your hesitation to just manage the docker-proxy machine lifecycle on your own. If you desire the complete Vagrant workflow for your docker-proxy, you can have two Vagrantfiles (or one multi-machine Vagrantfile) and just export the proper environment variables. Then you can control everything including memory, CPU, IP, etc.

> Would it make sense, though, to shut the machine down if there are no more containers running on it after the last vagrant halt?

This is reasonable, but as you will see, it's actually impossible to determine in Vagrant's current state. The global status is a cache, and Vagrant only knows a limited set of information. In short, Vagrant _doesn't_ know if there exists a machine that still has that dependency. It also adds unnecessary startup times as opposed to just leaving the machine running.

I think, if you want this much control over the docker-proxy, you should just manage it yourself. I don't think the tradeoff of complicating the workflow (which most users actually love) for a rather small use case that is already solvable :smile:

cweagans:
You're coming across as very condescending. I *know* how it works because I spent all weekend digging through the code to try to understand wtf is going on.

I can't reiterate this enough: leaving things running in the background is *not an amazing developer experience*. It uses more energy, it uses more CPU/RAM, and frankly, it's kind of annoying.

I *am* managing the VM separately, but having to jump through hoops to make it work is pretty dumb, if you ask me. Right now, this is a stripped down version of my Vagrantfile:

```
# -*- mode: ruby -*-
# vi: set ft=ruby :

# This is needed for OS detection later.
require 'rbconfig'

# If the docker-host machine isn't created yet, manually create that with the
# Virtualbox provider before moving on to the container provisioning. This isn't
# needed on Linux hosts (and the docker-host machine isn't defined on Linux hosts
# anyway).
unless RbConfig::CONFIG['host_os'] =~ /linux/
  if ARGV[0] == "up" && ARGV[1] != "docker-host" && !File.exist?(".vagrant/machines/docker-host/virtualbox/id")
    system("vagrant up docker-host --provider=virtualbox")
  end
end

# Set the default provider to Docker so that our containers come up properly.
ENV['VAGRANT_DEFAULT_PROVIDER'] = 'docker'

# Parallel provisioning doesn't work because of cross-container dependencies.
# See https://github.com/mitchellh/vagrant/issues/6831
ENV['VAGRANT_NO_PARALLEL'] = 'yes'

  # The Docker host machine isn't needed on Linux hosts, so skip defining it if
  # we can.
  unless RbConfig::CONFIG['host_os'] =~ /linux/
    config.vm.define "docker-host" do |dh|
      dh.vm.box = "coreos-beta"
      dh.vm.box_url = "http://beta.release.core-os.net/amd64-usr/current/coreos_production_vagrant.json"
      dh.vm.network "private_network", ip: "172.25.128.20"
      dh.vm.synced_folder "..", "/home/core/share",
          id: "core",
          :nfs => true,
          :mount_options => ['rw', 'vers=3', 'tcp', 'fsc', 'actimeo=1', 'nolock', 'udp']
      # Disable the default mount.
      dh.vm.synced_folder ".", "/vagrant", disabled: true


      dh.vm.provider "virtualbox" do |v|
        v.memory = 2048

        # If Virtualbox is version 5.0.0 or higher, enable the kvm paravirtualization
        # provider for better performance.
        begin
          if VagrantPlugins::ProviderVirtualBox::Driver::Meta.new.version >= "5.0.0"
            v.customize ['modifyvm', :id, '--paravirtprovider', 'kvm']
          end
          rescue Vagrant::Errors::VirtualBoxNotDetected
        end
      end
    end
  end

  def shared_container_config(d)
    d.vagrant_machine = "docker-host"
    d.auth_server = "quay.io"
    # d.email = "dpim.pubcentral.bot@nbcuni.com"
    d.email = " "
    d.username = " "
    # d.username = "nbcuniversal+coreos_vm"
    d.password = "bmJjdW5pdmVyc2FsK2NvcmVvc192bTo0NUdDUVNQRTNQTFg0S1Q1QUJFN05RSjlDNURaREVEUFhZSU9FNVVTV1NKTFM1WUUyWElPSTNPTkdPQVBPN1k1"
  end

  # Set up the database container.
  config.vm.define "db" do |db|
    db.vm.synced_folder ".", "/vagrant", disabled: true
    db.vm.provider "docker" do |d|
      d.name = "db"
      d.image = "mysql:5.6"
      d.create_args = ["-m", "512M"]
      d.ports = [ "3306:3306" ]
      d.env = {
        "MYSQL_DATABASE" => "pubcentral",
        "MYSQL_ALLOW_EMPTY_PASSWORD" => "yes"
      }
      shared_container_config(d)
    end
  end

  # Set up the memcached container.
  config.vm.define "memcached" do |memcached|
    memcached.vm.synced_folder ".", "/vagrant", disabled: true
    memcached.vm.provider "docker" do |d|
      d.name = "memcached"
      d.image = "memcached:latest"
      d.create_args = ["-m", "128M"]
      d.ports = [ "11211:11211" ]
      d.env = {
        "MEMCACHED_SERVER" => "memcached:11211"
      }
      shared_container_config(d)
    end
  end

  # Set up the web container. Note that this block needs to live after any
  # other containers that it links to.
  config.vm.define "web" do |web|
    web.vm.synced_folder ".", "/vagrant", disabled: true
    web.vm.provider "docker" do |d|
      d.name = "web"
      d.image = "quay.io/nbcuniversal/pubcentral-cms-dev"
      d.create_args = ["-m", "512M"]
      d.ports = [ "80:8080" ]
      d.link("db:db")
      d.link("memcached:memcached")
      if (RbConfig::CONFIG['host_os'] =~ /linux/)
        # If we're on Linux, bind mount the current directory into the container...
        d.volumes = [ ".:/var/www/pubcentral" ]
      else
        # ...otherwise, we're using the Docker Host VM, so we need to mount the files
        # from the NFS share.
        d.volumes = [ "/home/core/share:/var/www/pubcentral" ]
      end
      d.env = {
        "PC_DEV" => "true",
        "APACHE_RUN_USER" => "pubcentral",
        "APACHE_RUN_GROUP" => "staff"
      }
      shared_container_config(d)
    end
  end

end

```

cweagans:
You're coming across as very condescending, @sethvargo. Immediately dismissing my concerns as irrelevant and closing the issue and the PR is kind of a dick move. I *know* how this stuff works because I've spent a significant amount of time over the last couple weeks digging through the code trying to figure it out. I'm saying that the way that it works right now is *wrong*. I can't reiterate this enough: leaving *an entire vm* running in the background without even *telling* the user is *not an amazing developer experience*. It's wasteful (in terms of energy and system resources) and frankly kind of annoying.

I *am* managing the machine's lifecycle manually, but my point is that I shouldn't have to. If b2d would just shut down *like it should* when I tell my environment to shut down, I wouldn't have to maintain yet another set of provisioning scripts (stripped out of the below Vagrantfile for brevity).

> it's actually impossible to determine in Vagrant's current state.

That's not true at all. On vagrant halt, ssh into the host vm and run `docker ps`. Done. You don't need to track cross machine dependencies to do this.

Below is a stripped down version of my `Vagrantfile`. This is the only way I could get it to work properly without doing stupid things with environment variables or writing a plugin specifically for our company (which is *exactly* what I had to do in order to use the vagrant_vagrantfile option - at the end of `vagrant halt`, the plugin would run `docker ps` on the host vm and see if there's anything left, and if not, stop the machine).

Reiterating yet again: *having to do all of this stuff myself is NOT an amazing experience*.

```

# This is needed for OS detection later.
require 'rbconfig'

# If the docker-host machine isn't created yet, manually create that with the
# Virtualbox provider before moving on to the container provisioning. This isn't
# needed on Linux hosts (and the docker-host machine isn't defined on Linux hosts
# anyway).
unless RbConfig::CONFIG['host_os'] =~ /linux/
  if ARGV[0] == "up" && ARGV[1] != "docker-host" && !File.exist?(".vagrant/machines/docker-host/virtualbox/id")
    system("vagrant up docker-host --provider=virtualbox")
  end
end

# Set the default provider to Docker so that our containers come up properly.
ENV['VAGRANT_DEFAULT_PROVIDER'] = 'docker'

# Parallel provisioning doesn't work because of cross-container dependencies.
# See https://github.com/mitchellh/vagrant/issues/6831
ENV['VAGRANT_NO_PARALLEL'] = 'yes'

  # The Docker host machine isn't needed on Linux hosts, so skip defining it if
  # we can.
  unless RbConfig::CONFIG['host_os'] =~ /linux/
    config.vm.define "docker-host" do |dh|
      dh.vm.box = "coreos-beta"
      dh.vm.box_url = "http://beta.release.core-os.net/amd64-usr/current/coreos_production_vagrant.json"
      dh.vm.network "private_network", ip: "172.25.128.20"
    end
  end

  # Set up the database container.
  config.vm.define "db" do |db|
    db.vm.synced_folder ".", "/vagrant", disabled: true
    db.vm.provider "docker" do |d|
      d.name = "db"
      d.image = "mysql:5.6"
    end
  end

  # Set up the memcached container.
  config.vm.define "memcached" do |memcached|
    memcached.vm.synced_folder ".", "/vagrant", disabled: true
    memcached.vm.provider "docker" do |d|
      d.name = "memcached"
      d.image = "memcached:latest"
    end
  end

  # Set up the web container. Note that this block needs to live after any
  # other containers that it links to.
  config.vm.define "web" do |web|
    web.vm.synced_folder ".", "/vagrant", disabled: true
    web.vm.provider "docker" do |d|
      d.name = "web"
      d.image = "quay.io/mycompany/myimage"
      if (RbConfig::CONFIG['host_os'] =~ /linux/)
        # If we're on Linux, bind mount the current directory into the container...
        d.volumes = [ ".:/var/www/pubcentral" ]
      else
        # ...otherwise, we're using the Docker Host VM, so we need to mount the files
        # from the NFS share.
        d.volumes = [ "/home/core/share:/var/www/pubcentral" ]
      end
    end
  end
end

```

sethvargo:
Hi @cweagans 

Thank you for your reply. It is certainly not my intent to come off as condensing, so I apologize for that. We routinely close PRs or issues, but, as you can see here, we continue the conversation long after they are closed. It is merely a means of triage.

> I'm saying that the way that it works right now is _wrong_.

I mean this in the nicest way possible :smile:, but this is an opinion that is based on your own personal experience, and it differs greatly from most of the Vagrant-Docker users I've interacted with. The current behavior that exists in Vagrant is based off of many conversations with users and their desired experiences. We _do_ document the fact that the Virtual Machine is left running. Furthermore, you can search back through many of the previous issues since the Vagrant Docker provider was added - never once has this issue come up, yet there are hundreds of daily Vagrant+Docker users. For 99% of people, it is a non-issue, and they _like_ the fact that it just works out of the box.

The thing that I'm trying to re-iterate is that the boot2docker VM is global, not per-environment. Even though your environment may no longer be using the Docker provider, there could exist a different environment with a different Vagrantfile that still depends on the boot2docker VM.

> That's not true at all. On vagrant halt, ssh into the host vm and run `docker ps`.

Unfortunately that doesn't guarantee that no other environments need the boot2docker instance - that just guarantees that no other environments have a long-running docker process _right now_. Other environments could exist that issue one-off Docker commands that return immediately. In fact, this is a very commonly-requested use case, and it's the reason users have requested `vagrant docker-exec`. Just because a docker process isn't currently running doesn't mean there are no dependencies on the system that rely on Docker. One could argue that you should just re-spin up that VM in that instance, but that's not an ideal experience for most users.

The boot2docker VM is incredibly small in terms of disk space and resource consumption. I don't know why you need such a complex Vagrantfile when you could just run `vagrant global status` to determine the b2d machine ID and then `vagrant destroy` with that machine ID, if your desired behavior is to truly destroy the boot2docker instance when running destroy. We do document this process as I linked above.

cweagans:
> we continue the conversation long after they are closed. It is merely a means of triage.

Unfortunately, that also removes visibility from the issue and discourages other people from participating in the issue, leading toward two people talking without outside input. You should rethink that approach.

> The current behavior that exists in Vagrant is based off of many conversations with users and their desired experiences. We do document the fact that the Virtual Machine is left running. Furthermore, you can search back through many of the previous issues since the Vagrant Docker provider was added - never once has this issue come up, yet there are hundreds of daily Vagrant+Docker users. For 99% of people, it is a non-issue, and they like the fact that it just works out of the box.

If you're going to rely on documentation for this, it should be a big, bold, stupidly noticeable thing that gets displayed in one's terminal right after running `vagrant halt`, rather than a sentence buried in the middle of a paragraph on some documentation page somewhere. Furthermore, the documentation page *does not make it clear* that the machine will continue running. For instance:

> Vagrant will spin up a single instance of a host VM and run multiple containers on this one VM. This means that with the Docker provider, you only have the overhead of one virtual machine, and only if it is absolutely necessary.

This makes it sound like if I've declared multiple containers in my Vagrantfile, I'll get one VM that runs them all, which is cool. This *does not* tell me that I'll have one VM for my entire host to work with and that it will *continue running* even after I've told vagrant to stop, and that's the core of the issue here. The fact that it doesn't *tell me* what is happening makes the fact that it's doing something dumb even worse.

Again, I'm not arguing against having Docker containers work out of the box. It's a great feature. I like it. Other people like it. Pretty much everyone likes it.

I'm arguing against this being a global thing that gets left running. Literally nothing else in Vagrant is global from the perspective of a user unless they specifically ask for it (e.g. `vagrant global-status`). Why should this be any different? Because nobody has noticed it and complained about it? C'mon - I thought we were shooting for better here, and not just "good enough that nobody complains".

If the b2d machine weren't global, the Docker provider could conceivably manage the b2d machine on a per-Vagrantfile basis, no? That is, it's turned on as needed when I run `vagrant up` and it's turned off when I run `vagrant halt`, and we don't have to worry about cross-project dependencies of any kind. If that were the case, we wouldn't even be having this conversation.

> I don't know why you need such a complex Vagrantfile when you could just run vagrant global status to determine the b2d machine ID and then vagrant destroy with that machine ID, if your desired behavior is to truly destroy the boot2docker instance when running destroy.

Because having the Vagrantfile this complex allows our environment to work *as people expect it to*. If it's not set up like this, people end up with the b2d machine running in the background for all of eternity and don't even realize it because it's not specifically called out in the CLI, nor in the documentation. The primary use case here is halt, not destroy. We have several applications, all of which take up large amounts of RAM, so our system resources are at a premium.

mitchellh:
Hey @cweagans, I'll step in here to give another point of view. Let's just start from a fresh slate, I won't respond to anything in particular above and instead talk about the feature itself.

I can see where you're coming from. I myself have run into the issue where my Docker host remains running for much longer than I expected since I forget it is running. This is absolutely a real thing. 

We've talked about this before and @sethvargo is correct in that is our current viewpoint. I've implemented this before and reverted it since it actually made things a bit worse since when stopping/starting containers the behavior I _really_ wanted (but isn't possible for current Vagrant core) is to bring it down if we don't start a container in X minutes. This isn't possible now. When I implemented it [similarly to how you did], I ran into the issue where every time I stopped/started a dev container my host machine was also restarting. 

I don't think there is a good way to solve this. I'm open to bringing in this functionality if we can start by making it optional (an opt-in with a Vagrantfile Docker-specific flag). What do you think of this?

I only briefly looked over the code but it looks like its still missing checking if any containers are running on the machine. Could you add this? I'll reopen the PR in the mean time. 

How does all this sound?

darylteo:
Hi @cweagans, 

 had a similar use case, and agree with your point about isolation of environments (was definitely caught out by the unexpected behaviour of the docker host) but also understand the reasonings behind such. 

BUT I did find a workaround that may or may not have existed at the time of your post (perhaps this should be the officially sanctioned method of achieving docker host isolation). This is still in trial personally but from the look of it I can't see anything "wrong" so there shouldn't be any weird side-effects.

```yaml
# -*- mode: ruby -*-
# vi: set ft=ruby :

    config.vm.define "docker_host", primary: true do |host|
        host.vm.provider "virtualbox" do |v|
          v.memory = 1024
          v.cpus = 2
        end

        host.vm.box = "ubuntu/xenial64"
        host.vm.network "private_network", ip: "192.168.33.10"
        host.vm.provision "shell", path: "./bootstrap.sh"

        host.vm.network "forwarded_port", guest: 80, host: 80
        host.vm.network "forwarded_port", guest: 3306, host: 3306
    end

    config.vm.synced_folder "public", "/var/www/html", :mount_options => ["dmode=777", "fmode=666"]
    config.vm.define "default" do |app|
        app.vm.provider "docker" do |d|
            d.build_dir = "."
            d.ports = ["80:80", "443:443"]
            d.create_args = ["--net=host"]

            d.vagrant_vagrantfile = "./Vagrantfile"
            d.vagrant_machine = "docker_host"
        end
    end

end

````

Basically:

 - define the docker host yourself in the Vagrantfile. Technically, this could have just been hashicorp/boot2docker but I wanted other stuff in there as well (like mysql and other utilities).
 - obviously, docker must be installed if you're rolling your own image (this is done in bootstrap.sh)
 - use vagrant_vagrantfile and vagrant_machine properties to specify where to find information about the dockerhost.
 - vagrant up provisions both docker host and container (i've spuriously had issues with this, but hopefully they should iron themselves out once I've provisioned the docker host properly).
 - vagrant halt halts both docker host and container.

Here's the log when I run `vagrant up`.
```
> vagrant up 
Bringing machine 'docker_host' up with 'virtualbox' provider...
Bringing machine 'default' up with 'docker' provider...
==> docker_host: Checking if box 'ubuntu/xenial64' is up to date...
==> docker_host: A newer version of the box 'ubuntu/xenial64' is available! You currently
==> docker_host: have version '20170106.1.0'. The latest is version '20170113.0.0'. Run
==> docker_host: `vagrant box update` to update.
==> docker_host: Clearing any previously set forwarded ports...
==> docker_host: Clearing any previously set network interfaces...
==> docker_host: Preparing network interfaces based on configuration...
    docker_host: Adapter 1: nat
    docker_host: Adapter 2: hostonly
==> docker_host: You are trying to forward to privileged ports (ports <= 1024). Most
==> docker_host: operating systems restrict this to only privileged process (typically
==> docker_host: processes running as an administrative user). This is a warning in case
==> docker_host: the port forwarding doesn't work. If any problems occur, please try a
==> docker_host: port higher than 1024.
==> docker_host: Forwarding ports...
    docker_host: 80 (guest) => 80 (host) (adapter 1)
    docker_host: 3306 (guest) => 3306 (host) (adapter 1)
    docker_host: 2375 (guest) => 2375 (host) (adapter 1)
    docker_host: 22 (guest) => 2222 (host) (adapter 1)
==> docker_host: Running 'pre-boot' VM customizations...
==> docker_host: Booting VM...
==> docker_host: Waiting for machine to boot. This may take a few minutes...
    docker_host: SSH address: 127.0.0.1:2222
    docker_host: SSH username: ubuntu
    docker_host: SSH auth method: password
    docker_host: Warning: Remote connection disconnect. Retrying...
==> docker_host: Machine booted and ready!
[docker_host] GuestAdditions 5.0.30 running --- OK.
==> docker_host: Checking for guest additions in VM...
==> docker_host: Configuring and enabling network interfaces...
==> docker_host: Mounting shared folders...
    docker_host: /vagrant => /home/user/project/path
    docker_host: /var/www/html => /home/user/project/path/public
==> docker_host: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> docker_host: flag to force provisioning. Provisioners marked to run always will still run.
==> default: Docker host is required. One will be created if necessary...
    default: Docker host VM is already ready.
==> default: Syncing folders to the host VM...
    docker_host: Mounting shared folders...
    docker_host: /var/lib/docker/docker_1484398236_5869 => /home/user/project/path/public
    docker_host: /var/lib/docker/docker_1484398236_60542 => /home/user/project/path
    docker_host: /var/lib/docker/docker_build_6bd3159f2ccb0ad76ae7b39f3519a2a3 => /home/user/project/path
==> default: Image is already built from the Dockerfile. `vagrant reload` to rebuild.
==> default: Vagrant has noticed that the synced folder definitions have changed.
==> default: With Docker, these synced folder changes won't take effect until you
==> default: destroy the container and recreate it.
==> default: Starting container...
==> default: Provisioners will not be run since container doesn't support SSH.
````

briancain:
Hey there! I'm going to close this due to a lack of response from the changes requested for this in quite some time. We will be happy to take a look at a new PR with the changes requested, thanks!

Issue #5633, opened 2015-04-25
remove outdated


hryamzik:
It would be nice to have a command to remove outdated boxes. Cleaning them manually is a bit frustrating.

```
ubuntu/precise64            (virtualbox, 12.04.4)
ubuntu/precise64            (virtualbox, 20150420.1.0)
ubuntu/trusty64             (virtualbox, 14.04)
ubuntu/trusty64             (virtualbox, 20150420.1.1)
```

Here I've had to run `vagrant box remove ubuntu/precise64 --box-version 12.04.4` and the same for 14.04 box...

dragon788:
@hryamzik The tough part is these version numbers don't necessarily compare for a computer. A human might know that one is newer, but since they are both valid versions but not directly comparable  it would be hard to code that in. The other challenge is whether the box is still in use by any running VMs, in order to remove any box that's in use you would need to pass a force/override command, doing this for large numbers of boxes could be dangerous.

sethvargo:
https://github.com/mitchellh/vagrant/pull/6863

Issue #4412, opened 2014-08-29
Removing outdated boxes


chrisvire:
I can run "vagrant box outdated" to see the list of boxes that need to be updated.
I can run "vagrant box update" that will update these boxes.
I really want to have an easy way to remove the now outdated boxes.

dbajal:
Plusing this request. I think "remove --outdated" could be enough.

everflux:
Can you just create a pull request for your proposal?

thom8:
:+1: 

@chrisvire This would be an awesome addition!

In the meantime you can use this command -->

```
curl -LSs http://bit.ly/box-update-all | bash
```

which will update all your base boxes and remove all old versions.

kylegibson:
Has anyone started on this implementation?

bbrala:
This feature would help me a lot so i decided to give it a go. I pretty much have it working, and will make a pull request soon hopefully.

I ended up doing a command since putting i tinto the other commands felt weird. It seemed they differ quite a lot in scope or purpose.

``vagrant box remove-old-versions --provider docker``

I probably use the internal action to remove the boxes so it is less new code. So i might also add some more options like confirming.

bbrala:
Well i made a pull request #6863. I might also extract this to a seperate plugin so the feature can be used while we wait what is said about the pull request.

bbrala:
Ok, i've also created a plugin with this feature. Ready to install to your local vagrant :)

https://github.com/swisnl/vagrant-remove-old-box-versions

kylegibson:
@bbrala - That is amazing! Thank you for your efforts, you've saved me several days of work!

dbajal:
@bbrala Well done. Waiting for merge.

amontalban:
+1

@bbrala great work! I have installed your plugin and works like a charm. Thanks!

vinceskahan:
excellent - your PR should definitely be merged - works great (vagrant 1.7.4)

bbrala:
Added tests for all options except --force. And i added the options --dry-run and --name which were contributed to the plugin. Think everything should be ready for a merge.

bbrala:
Ping @sethvargo

bbrala:
@sethvargo is there any timeline on which this might be merged or at least reviewed?

zho:
is there any question to this documentation?

PRUNING OLD VERSIONS

[...]

PR #6863, opened 2016-01-12
New command: vagrant box remove-old-versions with tests

bbrala:
Added a command `vagrant box remove-old-versions` to remove all but the newest version of a box in vagrant. It builds a list of  boxes that are installed and finds the newest versions. It then removes every box that had a newer version installed.

```
Usage: vagrant box remove-old-versions [options]

Options:

    -p, --provider PROVIDER          The specific provider type for the boxes to destroy.
    -n, --dry-run                    Only print the boxes that would be removed.
        --name NAME                  The specific box name to destroy.
    -f, --force                      Destroy without confirmation even when box is in use.

```

Fixes GH-5633


bbrala:
I extracted the code to a plugin also so the feature can be used before it is merged. See: https://github.com/swisnl/vagrant-remove-old-box-versions

kylegibson:
:+1: I suspect they will want some test cases with this code

bbrala:
Yeah probably, might do that soonish, but was getting a bit late yesterday evening. :)

bbrala:
Added tests for all options except --force. And i added the options --dry-run and --name which were contributed to the plugin. Think everything should be ready for a merge (issue #4412)

sethvargo:
Hi @bbrala 

I haven't forgotten about this issue, I've just been thinking. I'm not a huge fan on introducing a new command, and I think it might make more sense to have a `--prune` option on the existing `vagrant box remove` command. What do you think?

bbrala:
Hi Seth, glad to hear it had not been forgotten.

I understand the reluctance of introducing a new command, i agree you should always be weary of putting in more and commands. I started coding this searching for the command to extend to add this functionality. But I really felt the code for removing old boxes globally was out of place in all the other commands.

I thought about putting it in the `vagrant box remove`, but that piece of code was very much aimed at handling single boxes. It felt weird to fit a option in there that would go through all your boxes, when the remove command is really made to remove one single, specific box. 

If you look at the use of language, it might fit somewhere in `vagrant box outdated`, but this is also really aimed at a single box and has a completely different goal.

I really think mixing single box command with something more global like this is a bad idea and will confuse the end user. 

If I read your proposal with perhaps a `vagrant box --prune`, i feel one would read that as pruning old versions for that given box, which could be useful. But I do feel the worth of this piece of code is really just cleaning up everything in one command, no one likes a full SSD :smile: So maybe a `—prune-all` or `--prune-global` would work. I’m not entirely sure.

I do like the feel of the word prune though. Perhaps "vagrant box prune" might be better. remove-old-versions feels clunky.

If you don't agree on the weirdness of putting it inside the remove command, that is of course fine, and I wouldn’t mind. I think my opinion is mostly a feeling of inconsistency in the back of my head.

bbrala:
@sethvargo Think you can find the time to check my reply? I can refactor if you really want, but i do need a decision before i can do that.

sethvargo:
Hi @bbrala 

I'm going through issues over the next few days. I'm not sure I'll get to it though.

bbrala:
Cheers @sethvargo, i do understand the volume of issues is quite large :)

chrisroberts:
@bbrala Hi! Reading through the history in this issue my initial reaction was a `--prune` option on the `box remove` command would be a good location for this. However, I am in agreement with the argument that `box remove` is specifically aimed at a single box, and this option would likely make the usage a bit confusing. So, I'm :+1: on the new command, but would like to see it named as `prune` (and it seems like you agree with that based on previous comment). Could you update the command name on this PR? Thanks!

bbrala:
Cool, just to be sure, would the full command be ``vagrant prune``? Will make the time to refactor as soon as possible.

chrisroberts:
@bbrala it would be: `vagrant box prune`

bbrala:
Ok cool, will rebase and update.

bbrala:
@chrisroberts i've refactored the name to prune and rebased onto master. I also added the command to the documentation. :)

bbrala:
Hmm, looks like github is a bit confused because of the rebase. When icheck in my IDE i only see commits on my branch... weird.

chrisroberts:
Looks like the rebase got hosed. If you can get a clean one to push, that would be great, but it's not a huge deal. I can cherrypick items out into a new PR to stage easy enough :smiley:

bbrala:
Hmm, i'll make a  clean one and reference this pr and the issue.

bbrala:
@chrisroberts i've openend a new pull request #7978 which is just the changes applied to master.

chrisroberts:
Awesome, thanks! I'll close up this in favor of the new PR. Cheers!

PR #6867, opened 2016-01-13
FreeBSD: Use quotes around hostname in rc.conf

jlduran:
Use double quotes around the `hostname` value in `/etc/rc.conf`

### DIFF
--- a/plugins/guests/freebsd/cap/change_host_name.rb
+++ b/plugins/guests/freebsd/cap/change_host_name.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
       class ChangeHostName
         def self.change_host_name(machine, name)
           if !machine.communicate.test("hostname -f | grep '^#{name}$' || hostname -s | grep '^#{name}$'", {shell: "sh"})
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @jlduran 

Is there an associated issue or bug report this is fixing?

jlduran:
No. Should I file one first?

From the [man page], `hostname` takes a _str_ as a value, so quotes are needed.

[man page]: https://www.freebsd.org/cgi/man.cgi?query=rc.conf&apropos=0&sektion=0&manpath=FreeBSD+10.2-RELEASE&arch=default&format=html

sethvargo:
Hi @jlduran 

No, you don't have to. I just wanted to make sure we cross-linked appropriately if this was referencing another issue.

jlduran:
:+1:  No problem.

The hostname is being set by the `hostname` command in the [line below](https://github.com/mitchellh/vagrant/pull/6867/files#diff-afbfe36b60fd5cfca052d76bdec06cffR8) effectively.

In `/etc/rc.conf`, however is being set without quote marks. If you have, for instance, an Ansible provisioner that verifies if the hostname is set, double entries start to appear, for example, in `/etc/rc.conf`:

    hostname=foo.example.com
    hostname="foo.example.com"

It's really no big deal, but annoying.

PR #6874, opened 2016-01-14
Update nfs docs

### LINKED ISSUE #6872: Please document that `sudo visudo` should be used to edit `/etc/sudoers`.
The documentation on [NFS Synced Folders](https://docs.vagrantup.com/v2/synced-folders/nfs.html) is awesome. However, there's no mention of how to edit that `/etc/sudoers` file.

I didn't know about the `visudo` command and went ahead and edited the file. What I entered had a syntax error and fixing it took a really long time because I lost the ability to `sudo` at all.

I'm sure using `visudo` is obvious to those who know. But I'm also certain that no one who knows about `visudo` will think twice before editing that file by another means.

If you can point me to where documentation can be added, I'd be happy to make a PR.

theengineear:
Closes #6872

Adds a warning to the [synced folders docs](https://docs.vagrantup.com/v2/synced-folders/nfs.html) about using `visudo` vs any other editor as syntax errors can leave you stuck without privileges.

### DIFF
--- a/website/docs/source/v2/synced-folders/nfs.html.md
+++ b/website/docs/source/v2/synced-folders/nfs.html.md
@@ -138,6 +138,8 @@ Below, we have a couple example sudoers entries. Note that you may
 have to modify them _slightly_ on certain hosts because the way Vagrant
 modifies `/etc/exports` changes a bit from OS to OS.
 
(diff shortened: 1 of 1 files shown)


theengineear:
@sethvargo Thanks for pointing me in the right direction.

PR #6879, opened 2016-01-15
Add missing features to snapshot restore/pop

jtopper:
In our test environments, it's good to be able to roll back to the same,
anonymous, snapshot repeatedly.  This patch adds a `--no-delete` option
to the `snapshot pop` command allowing this.

This makes the new core snapshot behaviour more consistent with what we
were doing with [vagrant-multiprovider-snap](https://github.com/scalefactory/vagrant-multiprovider-snap) which I intend to deprecate eventually.

### DIFF
--- a/plugins/commands/snapshot/command/push_shared.rb
+++ b/plugins/commands/snapshot/command/push_shared.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
   module CommandSnapshot
     module Command
       module PushShared
-        def shared_exec(argv, m)
+        def shared_exec(argv, m, opts={})
           with_target_vms(argv) do |vm|
             if !vm.id
               vm.ui.info("Not created. Cannot push snapshot state.")
@@ -12,7 +12,7 @@ module VagrantPlugins
             end
 
             vm.env.lock("machine-snapshot-stack") do
-              m.call(vm)
+              m.call(vm,opts)
             end
           end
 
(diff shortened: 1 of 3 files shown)


jtopper:
I've also added a patch to allow the provisioning flags (available with the `up` and `reload` commands) to be passed to the `snapshot restore` and `snapshot pop` commands, so that it is not necessary to run the provisioner on rollback.

ferventcoder:
The default snapshot rollback behavior should be not to run any provisioning - it should be an explicit opt in with `--provision`. Thoughts?

plugins/commands/snapshot/command/push_shared.rb
```diff
@@ -20,14 +20,14 @@ def shared_exec(argv, m)
...
-        def push(machine)
+        def push(machine,opts={})
           snapshot_name = "push_#{Time.now.to_i}_#{rand(10000)}"
 
           # Save the snapshot. This will raise an exception if it fails.
           machine.action(:snapshot_save, snapshot_name: snapshot_name)
         end
 
-        def pop(machine)
+        def pop(machine,opts={})
```
phinze: Style nit - comma space on all these args.

phinze:
LGTM - thanks @jtopper!

rfletcher:
In case anyone else ends up here the same way I did: It looks like updates to [the docs](https://www.vagrantup.com/docs/cli/snapshot.html#restore) have been released prematurely.

As I write this they include `--[no-]provision` for these snapshot commands, but 1.8.1 is the current release, and it does not support those options:

````shell
$ vagrant --version
Vagrant 1.8.1
$ vagrant snapshot restore --help
Usage: vagrant snapshot restore [options] [vm-name] <name>

Restore a snapshot taken previously with snapshot save.
````

Happy to find that --no-provision is coming though!

PR #6893, opened 2016-01-20
Prerelease version fix

mmickan:
This should fix https://github.com/mitchellh/vagrant/issues/6892


sethvargo:
Nice! Adds tests too! :smile:

PR #6894, opened 2016-01-20
redhat/nfs_client: Ignore systemctl nfs return code

fabiand:
By default there are no exports defined inside the Fedora guest,
because of this the nfs service fails to restart.
Let's ignore the returncode, because it's not to critical.

### DIFF
--- a/plugins/guests/redhat/cap/nfs_client.rb
+++ b/plugins/guests/redhat/cap/nfs_client.rb
@@ -25,7 +25,7 @@ module VagrantPlugins
 
         def self.restart_nfs(machine)
           if systemd?(machine)
-            machine.communicate.sudo("/bin/systemctl restart rpcbind nfs")
+            machine.communicate.sudo("/bin/systemctl restart rpcbind nfs || :")
           else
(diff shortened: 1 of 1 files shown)


rmohr:
@fabiand a good solution might be to check if /etc/exports is empty or contains only comments instead of ignoring the exit code. What do you think?

fabiand:
@rmohr Good idea, but then we should also check exports.d and check that they contain non-comments.

sethvargo:
Hi @fabiand 

I just tried a few RHEL boxes from Atlas and I have been unable to reproduce this. Can you supply more information please? I'm very hesitant to `|| true` on a command because it could hide a real error.

fabiand:
Wow - I think that it was also fixed in the distribution.

sethvargo:
Hey @fabiand 

Okay - are you okay if we close this out then? Sorry for the delayed reply on our prat.

fabiand:
No worries. Closing according to the chapter 42 of the book of the flying spaghetti monster - and because of the previous two comments.

PR #6912, opened 2016-01-21
Fixed an issue locating puppet binary. Issue-6876

gavioto:
As explained here http://stackoverflow.com/questions/11009598/whats-the-cmd-powershell-equivalent-of-which-on-bash in PS may be necessary to use "where.exe" instead of "where" to avoid call "Where-Object"

The change is compatible with CMD also.


liv3d:
+1

ferventcoder:
Huge :+1:

Issue #6834, opened 2016-01-09
Commands using machine ID fail


noplanman:
Since I updated to Vagrant 1.8.1, using the machine ID with Vagrant commands has stopped working properly.
It seems as if the `vagrant` command gets executed twice.

If I start up the machine from the project directory with `vagrant up`, it starts up perfectly and `vagrant halt` shuts it down properly.

Yet if I use the same commands using the machine ID from `vagrant global-status`, I get an error.

`vagrant up 1a2b3c4d` now gives me:
```
Vagrant can't use the requested machine because it is locked! This
means that another Vagrant process is currently reading or modifying
the machine. Please wait for that Vagrant process to end and try
again. Details about the machine are shown below:

Name: default
Provider: virtualbox
```

and `vagrant halt 1a2b3c4d` gives me:
```
/project-dir/Vagrantfile:50: warning: already initialized constant CONF
/project-dir/Vagrantfile:50: warning: previous definition of CONF was here
```

The last version I was using was 1.7.4, where it worked perfectly.

kamigerami:
Initially it looks like it will hold a lock to the index so it can be read/updated
 then it reloads the data (does not lock the index since it is already locked from outside of the call)
 then locking the machine exclusively and never releasing the entry (unlocking it).

so if the entry is released by running the unlocked_release method everything works fine.

Will investigate further during the weekend. 

@sethvargo  initial thoughts?

noplanman:
@kamigerami Any news on this? Is this a general problem or just some configuration issue?
Thanks for checking it out :blush:

kamigerami:
looks general to me! I'll update when I have something useful.

tdebatty:
+1
`vagrant halt, suspend and resume <id>` are working properly
`cd <directory> && vagrant up` works too
`but vagrant up <id>` gives
`Vagrant can't use the requested machine because it is locked!`

kamigerami:
@tdebatty you are right! Did not notice that the first time around! I have a new fix for it! Will update soon

kamigerami:
I am running Version: 1.8.2.dev and all these commands (halt, suspend, up and resume) are working fine for me with the machine id.

Please confirm this on your end as well @noplanman @tdebatty so we can close this ticket.

noplanman:
Where can I get 1.8.2.dev from? Do I need to build from source or is there an installer around?
Thanks for your efforts on fixing this :+1:

kamigerami:
pull the master branch

[...]

PR #6920, opened 2016-01-24
fixes-#6834-commands using machine id fail

kamigerami:
#6834-commands using machine id fail: 
fixes issue with lock never being released while using vagrant up / halt with machine id, releasing it for it to continue without failures.
also fixing a typo in the comments.


kamigerami:
@sethvargo let me know what you think

kamigerami:
I believe this can be closed! No longer needed in 1.8.2b - have a look at the ticket #6834

PR #6923, opened 2016-01-24
remove call to deprecated toggle_nori_type_casting in winrm

mwrock:
Removing call to method deprecated in v1.3. This will also prevent v1.5+ from leaking deprecation warning.

cc @sneal


sneal:
:shipit:

PR #7014, opened 2016-02-11
set longer timeout for net::ssh

dictcp:
timeout is set from 15 sec to 30 sec. it will fix some timeout error
like "timeout during server version negotiating" for some guest setting

### DIFF
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -339,7 +339,7 @@ module VagrantPlugins
           paranoid:              false,
           password:              ssh_info[:password],
           port:                  ssh_info[:port],
-          timeout:               15,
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @dictcp 

Thank you for the PR. My concern here is that, for an invalid configuration, a user now has to wait a longer period of time. I think a better approach is to leave the 15s and make this configurable in the ssh_info hash. What do you think?

sethvargo:
Hi there,

I am going to close this due to lack of response. If you would still like to see this patch merged, please open a _new_ pull request and answer the questions above. Thanks! 😄

scott-cornwell:
Having this same issue... I think this may be a virtualbox problem as I don't have this problem at all with vmware. Setting timeout to 30 worked for me as well.

PR #7025, opened 2016-02-12
New Alpine Linux support

Eric-Guo:
I'm new to [Alpine](http://linux.softpedia.com/blog/docker-founders-hire-alpine-linux-developer-to-move-the-official-image-to-ubuntu-500007.shtml) and Vagrant, found Vagrant can not properly halt the Alpine Linux, by looking at Vagrant, seems the code is well structured and only add 3 files will fix halt problem.

https://atlas.hashicorp.com/Eric-Guo/alpine_dev

So here is a PR.

### DIFF
--- a/plugins/guests/alpine/cap/halt.rb
+++ b/plugins/guests/alpine/cap/halt.rb
@@ -1,0 +1,16 @@
+module VagrantPlugins
+  module GuestAlpine
+    module Cap
+      class Halt
+        def self.halt(machine)
+          begin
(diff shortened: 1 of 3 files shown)


Eric-Guo:
@maier I just found there is a [plugin existing for alpine](https://github.com/maier/vagrant-alpine), so maybe official will perfer it's supported as plugin?

sethvargo:
Hi @Eric-Guo 

We generally prefer to see new OSes like this added as plugins before we bring them into core. Thanks!

PR #7027, opened 2016-02-12
Added two sections to README oriented towards beginners:

evisboy:
* What to Use Vagrant For
* When to Use Vagrant

To make the Vagrant project more approachable to a larger number of users—from informed developers to curious passersby and new and interested visitors—I have added two new sections to the README. The new sections explain in plain and straightforward language what Vagrant is used for and when one should use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,14 +7,23 @@
 
 Vagrant is a tool for building and distributing development environments.
 
+
+
+
+## What to Use Vagrant For
+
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @evisboy 

Thank you for opening a Pull Request. This is very similar to https://github.com/mitchellh/vagrant/pull/7008. Please see my comments on that Pull Request.

Issue #6356, opened 2015-10-04
The WinRM ready? method raises an error instead of returning false


Sharpie:
The following Vagrantfile includes a plugin which defines a command that attempts to print whether a VM can be communicated with:

```ruby
class TestCommand <Vagrant.plugin(2, :command)
  def execute
    with_target_vms(@argv) do |vm|
      puts "#{vm.name} #{vm.communicate.ready?}"
    end
  end
end

class TestPlugin < Vagrant.plugin('2')
  command(:running) do
    TestCommand
  end
end

Vagrant.plugin("2").manager.register(TestPlugin)

Vagrant.configure("2") do |config|
  config.vm.define 'posix-vm' do |node|
    # `vagrant running posix-vm` executes fine.
    node.vm.communicator = :ssh
  end

  config.vm.define 'windows-vm' do |node|
    # `vagrant running windows-vm` throws an error.
    node.vm.communicator = :winrm
  end
end
```

Executing `vagrant running windows-vm` throws an error:

```
vagrant running windows-vm
The box is not able to report an address for WinRM to connect to yet.
WinRM cannot access this Vagrant environment. Please wait for the
Vagrant environment to be running and try again.
```

I expect the return value to be `false`, as is the case with the SSH communicator used by `vagrant running posix-vm`:

```
vagrant running posix-vm
posix-vm false
```

Sharpie:
This example fails with Vagrant v1.7.4, but succeeds with v1.6.5, which suggests a regression of some sort in the WinRM communicator.

RyanJarv:
I'm experiencing this issue as well

RyanJarv:
The ssh communicator is communicator is catching Vagrant::Errors::VagrantError for ready? so this is what I'm doing to work around this issue atm.

https://github.com/RyanJarv/vagrant/commit/d46c61f613ecc3f73f95c6ef9a4050e6b329996b

sethvargo:
Hi all,

Thank you for reporting this. This has been fixed in 1e84cc4. Thanks!

Sharpie:
@sethvargo This is still an issue in Vagrant 1.8.1. The WinRM `ready?` method is throwing an error instead of returning `false` as the SSH `ready?` method does.

Sharpie:
@sethvargo Can we re-open this issue now that #7028 has been closed?

Sharpie:
@sethvargo Ping again. Could this be re-opened? This issue is still present in the latest Vagrant release.

PR #7028, opened 2016-02-12
communicators/winrm return false on ready? failure

Sharpie:
Prior to this patch, the WinRM `.ready?` method raised an error if there was
a communication failure. This behavior differs from the SSH communicator which
returns `false` from `ready?` if a communication failure occurs. The difference
in behavior makes it difficult to write code that works with both types of
communicators.

This patch changes the WinRM `.ready?` method to return false if an error of
type `Vagrant::Errors::VagrantError` occurs, which matches the behavior of the
SSH communicator.

Fixes mitchellh/vagrant#6356.

### DIFF
--- a/plugins/communicators/winrm/communicator.rb
+++ b/plugins/communicators/winrm/communicator.rb
@@ -107,9 +107,9 @@ module VagrantPlugins
 
         @logger.info("WinRM is ready!")
         return true
-      rescue Errors::TransientError => e
-        # We catch a `TransientError` which would signal that something went
-        # that might work if we wait and retry.
+      rescue Vagrant::Errors::VagrantError => e
+        # We catch a `VagrantError` which would signal that something went
+        # wrong expectedly in the `connect`, which means we didn't connect.
         @logger.info("WinRM not up: #{e.inspect}")
 
         # We reset the shell to trigger calling of winrm_finder again.
--- a/test/unit/plugins/communicators/winrm/communicator_test.rb
+++ b/test/unit/plugins/communicators/winrm/communicator_test.rb
@@ -66,9 +66,9 @@ describe VagrantPlugins::CommunicatorWinRM::Communicator do
       expect(subject.ready?).to be_false
     end
 
-    it "raises an error if hostname command fails with an unknown error" do
+    it "returns false if hostname command fails" do
       expect(shell).to receive(:powershell).with("hostname").and_raise(Vagrant::Errors::VagrantError)
-      expect { subject.ready? }.to raise_error(Vagrant::Errors::VagrantError)
+      expect(subject.ready?).to be_false
     end
 
     it "raises timeout error when hostname command takes longer then winrm timeout" do


Sharpie:
@sethvargo Any thoughts on this change? I'd love to eventually remove all the shim code that I've added to Vagrant plugins wrapping `communicator.ready?` calls.

sethvargo:
@mwrock

mwrock:
The problem with this approach is that `ready?` is consumed by `wait_for_ready` and it specifically rescues a `VagrantError` in its call to `ready?` as a signal that not only is the box not ready but it will likely *never* be ready given its current configuration. While a `TransientError` signifies the error is "recoverable" and should be retried.

Looking at the ssh communicator, while you are correct that it rescues `VagrantError`, it looks like even a communication failure would not actually raise. It just calls `connect` which rescues on `Exception` so its very unlikely even a `VagrantError` will be raised. That said, you are correct that the basic behavior of

mwrock:
The problem with this approach is that `ready?` is consumed by `wait_for_ready` and it specifically rescues a `VagrantError` in its call to `ready?` as a signal that not only is the box not ready but it will likely *never* be ready given its current configuration. While a `TransientError` signifies the error is "recoverable" and should be retried.

Looking at the ssh communicator, while you are correct that it rescues `VagrantError`, it looks like even a communication failure would not actually raise. It just calls `connect` which rescues on `Exception` so its very unlikely even a `VagrantError` will be raised. That said, you are correct that the basic behavior of `ready?` differs between the two and the winrm implementation tries to determine if the connection is "repairable."

Sharpie:
The main usecase for this is easy detection of running machines in a provider-independent way. The `machine.state` method isn't very helpful as different providers define different arbitrary sets of states. So, we can't assume there is a universal `:running` state.

A better approach is to ask "can I execute commands on this machine?" which naturally maps to the the `machine.communicator.ready?` method. However, in the case of WinRM, this method does not return `false` in many common cases where communication isn't possible --- such as the machine being powered off.

mwrock:
The retry logic could definitely be refactored making WinRM's implementation of `retry?` align better with the intent you point out. Unfortunately this PR as it stands would interfere with retries. For example, if the machine was powered off, we want to "fail fast" and not retry. WinRM is a more "sensitive" protocol than SSH and the retry logic is very desirable here.

mwrock:
cc @sneal

sneal:
:+1: Definitely agree that winrm comm `ready?` method should behave like the SSH provider `ready?` method. To make that happen we'll need to refactor the winrm communicator to look more like the SSH communicator. At a minimum that looks like we'll need to:

1. `ready?` shouldn't throw, just return true/false.
2. `wait_for_ready?` should call `connect` first which can throw any number of retryable/non-retryable exceptions, instead of _only_ calling `ready?`.
3. The `wait_for_ready?` catch block should re-raise any non-retryable exceptions.

On an semi-related note, `Helper.winrm_info` should behave more like ssh_info does.

sethvargo:
Hey @Sharpie 

Did you see Matt and Sean's comments?

sethvargo:
Hey there,

Sharpie:
Hey sorry for the lack of response. At this point, I don't think this PR is waiting on me. It sounds like the patch I've proposed is not satisfactory, but the proposed fix requires a deeper re-factoring of the WinRM communicator than I can do.

@sethvargo I'm fine with closing this PR, but could we re-open the original issue, #6356? That contains a complete, minimal, re-production case.

mwrock:
If its any consolation, we'll be doing some winrm refactoring here relatively soon after winrmv2 releases and will keep this issue in mind when that work is done.

PR #7035, opened 2016-02-14
Fix for #4608: Support for port forwarding in an IP aliased environment

pravinchandar:
Added support for Port forwarding in an IP aliased environment where the **same host port but different host IP addresses** are used. The change makes the following forwarding rule(s) possible.

Ex: `eth0` on the host is ip aliased to have a range of IP addresses `10.20.30.0/24.`

In the Vagrant file, we can now have an entry like the following and
it will just work! Note the host port `8081` is same for both .1 and .2.

```
    config.vm.network "forwarded_port", guest: 80, host: 8080
    config.vm.network "forwarded_port", guest: 81, host: 8081, host_ip: 10.20.30.1
    config.vm.network "forwarded_port", guest: 82, host: 8081, host_ip: 10.20.30.2
  end
```

### DIFF
--- a/lib/vagrant/action/builtin/handle_forwarded_port_collisions.rb
+++ b/lib/vagrant/action/builtin/handle_forwarded_port_collisions.rb
@@ -62,7 +62,7 @@ module Vagrant
           @logger.info("Detecting any forwarded port collisions...")
 
           # Get the extra ports we consider in use
-          extra_in_use = env[:port_collision_extra_in_use] || []
+          extra_in_use = env[:port_collision_extra_in_use] || {}
 
           # Get the remap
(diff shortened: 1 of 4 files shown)


pravinchandar:
Any updates on this one @sethvargo?

ArjanSchouten:
Does this PR have any progress? I'm still having the same problems.

MarcinOrlowski:
Bump in 2017. Not sure why this issue is considered low priority as it's really PITA with some more complex setups to deal with it with currently

MarcinOrlowski:
Good news. Will that be available in next release? What's the ETA?

pravinchandar:
Thanks @chrisroberts, look forward to the next release!

Issue #7040, opened 2016-02-16
HyperV Provider ignores the name of disk in BoxFile when cloning the disk


interdev:
When performing vagrant up on a new HyperV machine the clone operation for the disk image detects the correct extension but ignores what the file is called and creates a clone called disk.vhd or disk.vhdx

This means that disks that may be parented to a common base disk image cannot be exported from HyperV as both images will be named disk.

interdev:
Fix submitted https://github.com/mitchellh/vagrant/pull/7041

interdev:
1.8.3 resolves

PR #7041, opened 2016-02-16
Fix for #7040 to Preserve disk name

interdev:
Keep the disk name that is in the box file rather than copying to disk.*


sethvargo:
LGTM /cc @mwrock

mwrock:
:+1:

PR #7047, opened 2016-02-17
Fix berks

BackSlasher:
Firstly, I want to squash before merging  

This change restores the original environment for processes run by `Vagrant::Util::Subprocess`. This is something that a lot of methods in the Vagrant codebase are supposed to be doing, but eventually didn't.  
The basic issue was that Bundler was double-executed (in addition to the env variable modification in the Vagrant Golang launcher), so it didn't restore the original Ruby variables properly.

After applying this, Vagrant successfully runs non-ChefDK Berkshelf properly (the ChefDK Berkshelf always works because it's ignoring Ruby environment variables like `GEM_PATH`.

I might have not understood the way `jailbreak` is supposed to work. It might need to run under the original env contex as well. I'll be happy to get guidance for testing it

### DIFF
--- a/lib/vagrant/util/env.rb
+++ b/lib/vagrant/util/env.rb
@@ -9,7 +9,7 @@ module Vagrant
         if Vagrant.original_env.any?
           ENV.replace(proxy_env)
         elsif defined?(::Bundler::ORIGINAL_ENV)
-          #ENV.replace(::Bundler::ORIGINAL_ENV)
+          ENV.replace(::Bundler::ORIGINAL_ENV)
         end
         yield
       ensure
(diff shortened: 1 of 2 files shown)


ramonsnir:
+1

BackSlasher:
For completeness:  
http://blog.backslasher.net/running-external-ruby-code-from-vagrant.html

BackSlasher:
@mitchellh did anyone review this by any chance?

chrisroberts:
@BackSlasher Hi there. Is this still an issue you are having with berks? If so, can you rebase and update this PR? Thanks!

BackSlasher:
@chrisroberts I don't think I have any context on this anymore as I haven't worked on anything using Vagrant / Berkshelf for ~1y now :disappointed: 
I should have written a test or a repro for this I guess.
Best way to test it is to run  vagrant with a Bundler-installed berkshelf, but I don't have the dev env for doing this atm.

chrisroberts:
@BackSlasher Okay, great. I just wanted to check in before I close this out. If I do find issues against the current releases, I'll open up a new PR. Thanks for letting me know. Cheers!

PR #7090, opened 2016-03-01
hyperv differencing disk

### LINKED ISSUE #7089: Hyper-V Differencing Disks
In Hyper-V there is [differencing disk](https://technet.microsoft.com/en-us/library/cc720381(v=ws.10).aspx) concept (e.g. disk not being copied, we just create new empty disk and map it as a layer on top of original disk - similar to what docker doing with their layers).

### Advantages:

1. Improved `vagrant up` time - creating differencing disk take less than second no matter how big is original disk
2. Reduced space usage - differencing disk contains only difference data

### Vagrant version
Vagrant 1.8.1

### Host operating system
Windows Server 2012R2 Datacenter

### Guest operating system
any

### Vagrantfile

Here is suggested usage example

```ruby
  config.vm.provider "hyperv"
  config.vm.box = "hashicorp/precise64"
  config.vm.provider :hyperv do |hv|
    hv.differencing_disk = true # use differencing disk instead copying whole image
  end
end
```

mac2000:
### Overview
Allow Hyper-V users use differencing disks instead of copying whole image each time to save space and time.

Creating differencing disk takes less than second and it does not matter how big is original disk

Added `differencing_disk` (Boolean) (optional, default is false) switch for hyperv.

Usage example:

```ruby
  config.vm.provider "hyperv"
  config.vm.box = "hashicorp/precise64"

  config.vm.provider :hyperv do |hv|
    hv.differencing_disk = true
  end
end
```

### References
- Fixes #7089
- References #7089

### DIFF
--- a/plugins/providers/hyperv/action/import.rb
+++ b/plugins/providers/hyperv/action/import.rb
@@ -18,11 +18,13 @@ module VagrantPlugins
           maxmemory = env[:machine].provider_config.maxmemory
           cpus = env[:machine].provider_config.cpus
           vmname = env[:machine].provider_config.vmname
+          differencing_disk = env[:machine].provider_config.differencing_disk
 
           env[:ui].output("Configured Dynamic memory allocation, maxmemory is #{maxmemory}") if maxmemory
           env[:ui].output("Configured startup memory is #{memory}") if memory
           env[:ui].output("Configured cpus number is #{cpus}") if cpus
           env[:ui].output("Configured vmname is #{vmname}") if vmname
+          env[:ui].output("Configured differencing disk instead of cloning") if differencing_disk
 
           if !vm_dir.directory? || !hd_dir.directory?
             raise Errors::BoxInvalid
@@ -95,7 +97,11 @@ module VagrantPlugins
           env[:ui].detail("Cloning virtual hard drive...")
           source_path = image_path.to_s
           dest_path   = env[:machine].data_dir.join("#{image_filename}#{image_ext}").to_s
-          FileUtils.cp(source_path, dest_path)
+          if differencing_disk
+            env[:machine].provider.driver.execute("clone_vhd.ps1", {Source: source_path, Destination: dest_path})
+          else
+            FileUtils.cp(source_path, dest_path)
+          end
           image_path = dest_path
 
           # We have to normalize the paths to be Windows paths since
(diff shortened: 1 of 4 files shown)


sethvargo:
Ping @mwrock

mwrock:
:100: 

This is awesome. The only thing that would make this better is an option to have vagrant use this as the default. Nice work @mac2000!

mac2000:
Thank you @mwrock I also think that it will be nice to have that feature on by default actually

Bad things that *may* happen:

1. In some edge cases with weird/outdated setups without differencing disk support people may get into troubles, but turning off that switch will make them happy again - should be clarified in release notes
2. It is not recommended to use differencing disks in production, but the same is true for dynamic expanding disk which is used by most of boxes - so I do not see any issue here

Good things that *will* happen:

1. People will not believe their eyes when they will be able to star 100 virtual machines in few minutes :)

So, if that is ok I will make one more commit which will make this thing turned on by default

PS: Also I tried to search a way to make this thing trying to check is differencing disk supported and if so - use it, if not - copy it, but stopped after realizing that there may be cases when differencing disk are not desired solution

mwrock:
Yep. Totally agree this should be an option.

The other value here beyond speed is space. Now 10 vagrant instances != 10 * (1 gajillion gigabytes) but 10* (<1 GB).

mac2000:
Yep, you are totally right about sizes and it is critical for people who use SSD,

As about "Totally agree this should be an option" - I did not get your point - we leave all as is, or should I make one more commit which will make this thing turned on by default? (Its up to you)

BTW: differencing disk will take exactly 4mb no matter how big is original disk was

mwrock:
sorry for the confusion. Yes I'd say leave as is. I just meant I agree that it should not be enabled all of the time and I also think that for now, it should NOT be the default. I'll probably hack my personal vagrant install to make it the default but thats my own choice.

mac2000:
Heh, I actually did same trick and set it to enabled by default in my installation, ok, thank you for your time, I am leaving all as is, hope to see it somewhere in the future

Thank you for Vagrant!

codekaizen:
Wow, great work. Is there anything I can do to help getting this PR accepted?

mwrock:
@sethvargo would love to see this included in the next version.

Issue #7073, opened 2016-02-24
specification.rb:945:in `all=': undefined method `group_by' for nil:NilClass (NoMethodError)


tpenguin:
### Vagrant version

1.8.1

### Expected behavior

`lib/vagrant/bundler.rb` line 268 should not have performed what it has. The call to method `#all=` causes `nil` to be passed into *rubygems* which clearly shows in their source code that this is **not** actually the way to use the method.

### Actual behavior

```
/usr/local/lib/ruby/site_ruby/2.1/rubygems/specification.rb:945:in `all=': undefined method `group_by' for nil:NilClass (NoMethodError)
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/bundler.rb:268:in `with_isolated_gem'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/bundler.rb:225:in `internal_install'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/bundler.rb:97:in `install'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/plugin/manager.rb:62:in `block in install_plugin'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/plugin/manager.rb:72:in `call'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/plugin/manager.rb:72:in `install_plugin'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/plugins/commands/plugin/action/install_gem.rb:37:in `call'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/action/warden.rb:34:in `call'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/action/builder.rb:116:in `call'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/action/runner.rb:66:in `block in run'
	from /usr/local/lib/ruby/gems/2.1/gems/vagrant-1.8.1/lib/vagrant/util/busy.rb:19:in `busy'
[...]

PR #7092, opened 2016-03-01
Fixes bug report #7073 with RubyGems later than 2.4.6.

tpenguin:
### Overview
RubyGems later than version 2.4.6 introduced a change which prevented Vagrant from functioning, due to `NilClass:#group_by` method missing.

### References
- Fixes GH-7073
- References rubygems/rubygems@044b0e2685e4b219b013f1067d670918a48c1f62


sethvargo:
Hi @tpenguin 

Thank you for opening a PR. I don't think silently rescuing the error is a good practice, and it also is going to open a new bug in Vagrant, since the specs won't actually be reset.

Issue #5852, opened 2015-06-20
FreeBSD 10 Private network DHCP issue


emve-dev:
When using DHCP with private network the first `vagrant up` is ok. But after trying `vagrant reload` or `vagrant halt` and `vagrant up` i get this error:
```
==> default: Configuring and enabling network interfaces...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

dhclient em1

Stdout from the command:



Stderr from the command:

dhclient already running, pid: 705.
exiting.
```

Heres my Vagrantfile:
```
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure(2) do |config|
  config.vm.box = "chef/freebsd-10.0"
  config.ssh.forward_agent = true
  config.vm.network "private_network", type: "dhcp"
  config.vm.network "forwarded_port", guest: 5000, host: 5000
  config.vm.synced_folder ".", "/vagrant", type: "nfs", id: "vagrant-root"
end
```
I am still able to ssh login but my /vagrant folder is empty (shared folder wasn't mounted).
To fix this issue I edit the `/etc/rc.conf` remove these lines:
```
#VAGRANT-BEGIN
ifconfig_em1="DHCP"
#VAGRANT-END
```
And after running `vagrant reload` everything works until the next `vagrant reload` (so I need to repeat the whole `rc.conf` editing.
There is still a line in `rc.conf` which makes DHCP work:
```
ifconfig_em0="DHCP"
``` 

Full `/etc/rc.conf`
```
zfs_enable="YES"
hostname="packer-freebsd-10.0-amd64"
ifconfig_em0="dhcp"  # This makes DHCP work
sshd_enable="YES"
rpcbind_enable="YES"
nfs_server_enable="YES"
mountd_flags="-r"
vboxnet_enable="YES"
vboxguest_enable="YES"
vboxservice_enable="YES"
ifconfig_vtnet0_name="em0"
ifconfig_vtnet1_name="em1"
ifconfig_vtnet2_name="em2"
ifconfig_vtnet3_name="em3"
#VAGRANT-BEGIN
ifconfig_em1="DHCP" # This makes `vagrant up` fail
#VAGRANT-END
```

ifconfig:
```
em0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
[...]

PR #7093, opened 2016-03-01
[MOD] Change the way FreeBSD network interfaces are configured. Fixes…

amontalban:
### Overview
Change the way FreeBSD network interfaces are configured by using netif service instead of running manual commands via shell.

### References
- Fixes GH-5852


sethvargo:
Hi @amontalban 

I'm not familiar with BSD - what versions is this packages available on? Will it break on older/newer releases?

amontalban:
Hi @sethvargo,

Netif is present in freebsd since 5.x so it should work on older versions
and also it's the proper way to restart interfaces in freebsd.

For more info check [1] and [2].

Thanks!

[1] https://forums.freebsd.org/threads/22494/

[2]
http://www.linuxscrew.com/2007/09/05/how-to-restartstopstart-networking-in-freebsd/

PR #7103, opened 2016-03-05
ansible_local: use double quoting for 'extra-vars', 'limit' and 'start-at-task' options

gildegoma:
### Overview

When the `extra_vars` option is a hash, the `ansible_local` provisioner must quote the generated JSON parameter. 
I also propose here that the two other quoted parameters (`limit` and `start_at_task`) also rely on double quotes (for output consistency and content flexibility).

### References

- Fixes GH-6726
- Replaces GH-6820

### DIFF
--- a/plugins/provisioners/ansible/config/base.rb
+++ b/plugins/provisioners/ansible/config/base.rb
@@ -113,8 +113,16 @@ module VagrantPlugins
           end
 
           if raw_arguments
-            @raw_arguments = Helpers::as_array(raw_arguments)
+            if raw_arguments.kind_of?(String)
+              @raw_arguments = [raw_arguments]
+            elsif !raw_arguments.kind_of?(Array)
+              @errors << I18n.t(
+                "vagrant.provisioners.ansible.errors.raw_arguments_invalid",
+                type:  raw_arguments.class.to_s,
+                value: raw_arguments.to_s)
+            end
           end
+
         end
       end
     end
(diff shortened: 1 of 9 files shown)


gildegoma:
Call for reviewers: @kamazee @elsom25 @fdemmer @bertramn @warrenseine

(sorry for the massive ping broadcast, but I'm not confident with all these backslashes, and I'm afraid to have missed one of your possible use cases ;-)

kamazee:
:+1: 
worked for me

fdemmer:
sorry for delay, works for me! :+1:

gildegoma:
@kamazee @fdemmer thank you very much for taking time to test this :smiley: 

Before merging, I'd like to fix the issue with the `raw_arguments` option, which values (e.g. an `--extra-vars` or `limit`) shouldn't be modified by this "quoting stuff". I'll try to do this within the next days...

gildegoma:
Okay, with ab036dd, any `--extra-vars` parameter set via the `raw_arguments` provisioner option (a possible [typical workaround](https://github.com/mitchellh/vagrant/issues/6726#issuecomment-208836260) to this bug #6726) won't be broken with the next Vagrant release.

I'm not 100% happy with this implementation (because of the code duplications around the handling of the `raw_arguments`), but these are well covered and provides a better behaviour. I'm merging now, but please let me know if you see how we could improve it :)

PR #7104, opened 2016-03-05
Fixed centos 7 networks if using more than 3 ethernet adapters.

altexdim:
### Overview
Fixed sorting of interface names in centos7


sethvargo:
Hi @altexdim 

Thank you for opening a PR. I think the `-S` flag sorts the output by file size, largest first. Is that the intended behavior here?

Refs: http://man7.org/linux/man-pages/man1/ls.1.html

altexdim:
Hi
It was my mistake, the correct line is

`ls -v`

PR #7108, opened 2016-03-06
HyperV: Reinstate compatibility with PowerShell 4

jtopper:
### Overview
On `vagrant status` for HyperV users, we catch an exception to determine the status of the machine.  This changed in order to support Windows 10, but this broke compatibility with powershell 4. This PR makes the exception we're looking for conditional on PS version.

### References
A comment on https://github.com/mitchellh/vagrant/commit/587c88e65ae294c4f2939937c2079c545efd17cb shows that this commit broke back-compat with PowerShell 4.

### DIFF
--- a/plugins/providers/hyperv/scripts/get_vm_status.ps1
+++ b/plugins/providers/hyperv/scripts/get_vm_status.ps1
@@ -7,12 +7,18 @@ Param(
 $Dir = Split-Path $script:MyInvocation.MyCommand.Path
 . ([System.IO.Path]::Combine($Dir, "utils\write_messages.ps1"))
 
-# Get the VM with the given name
+
+if($PSVersionTable.PSVersion.Major -le 4) {
+  $ExceptionType = [Microsoft.HyperV.PowerShell.VirtualizationOperationFailedException]
+} else {
+  $ExceptionType = [Microsoft.HyperV.PowerShell.VirtualizationException]
+}
+
 try {
     $VM = Get-VM -Id $VmId -ErrorAction "Stop"
     $State = $VM.state
     $Status = $VM.status
-} catch [Microsoft.HyperV.PowerShell.VirtualizationException] {
+} catch $ExceptionType {
     $State = "not_created"
     $Status = $State
 }


sethvargo:
Hi @jtopper 

Thank you for the PR. Unfortunately I'm not a PS expert, so I'm going to defer to @mwrock when he has time. Thanks! :smile:

mwrock:
Great to see this fix!

I think this is fine but I would be inclined to `SilentlyContinue` on the `ErrorAction` and set `$state` and `$status` based on whether `$vm` is `$null` or not. While generally I think a try/catch on a specific exception type is a very good idea, I personally find branching off version here to be counter productive. But its not horrible.

I would ask that you correct your git line ending handling so that the commit does not come across changing the entire file and instead only shows the addition of the new code.

jtopper:
@mwrock Thanks for the feedback, I've taken care of the line endings problem.

I'm not hugely familiar with PowerShell myself, so I don't exactly follow your other suggestions, but it sounds like you're commenting on the code that was there previously, not my modification to it?

mwrock:
Thanks for fixing the line endings and you are right about the code I'm commenting on being unrelated directly to your pr so please disregard.

After eyeballing again today, it looks like your conditions are the opposite of what we want. You want to either swap the types or change `le` to `gt`.

jtopper:
@mwrock No, I think the conditions are correct. In the current HEAD, we're catching `Microsoft.HyperV.PowerShell.VirtualizationException` which means the plugin doesn't work with PowerShell 4. This change was introduced in https://github.com/mitchellh/vagrant/commit/587c88e65ae294c4f2939937c2079c545efd17cb#diff-15ddfe653598f0ea25832265a1e69cc0

My PR here ensures we check for the older `Microsoft.HyperV.PowerShell.VirtualizationOperationFailedException` in v4 and earlier (reinstating the old behaviour), and for `Microsoft.HyperV.PowerShell.VirtualizationException` otherwise (which leaves it working in 5).

mwrock:
gotcha @jtopper . I was checking for `[Microsoft.HyperV.PowerShell.VirtualizationException]` on win 10 and it was not loaded even after importing the `Hyper-v` module. However I now see after running `get-vm` on a bogus vm, The type is just loaded more lazily than I anticipated.

Yup this looks good to me @sethvargo :+1:

Issue #7109, opened 2016-03-06
No snapshot support exists for the Hyper-V provider


jtopper:
Vagrant 1.8 added snapshot support, but this didn't cover all the core providers.  I've maintained a plugin (https://github.com/scalefactory/vagrant-multiprovider-snap) supporting snapshotting across the Virtualbox, VMWare and Hyper-V providers using a consistent CLI interface and will shortly provide a PR containing a port of the Hyper-V support into Vagrant core.

PR #7110, opened 2016-03-06
Add snapshot support to Hyper-V provider

jtopper:
### Overview
This PR is a port of the Hyper-V snapshot support from https://github.com/scalefactory/vagrant-multiprovider-snap into Vagrant core.  It supports all of the snapshot behaviour supplied by the Virtualbox and VMWare providers.

### Known Bugs
Restored VMs using SMB shared folders may error as they boot, during the mounting of those shared folders, though in the cases I've observed, the pre-snapshot mounts are still functional inside the guest.  

Presumably this is down to the filesystem and network state in the guest taking a trip back in time, but the host not going along for the ride.

I believe this to be bug-compatible with the other providers, and with my original implementation, but this may be a concern here since the default folder sharing mechanism on Hyper-V hosts is SMB.

### References
- Fixes GH-7109

### DIFF
--- a/plugins/providers/hyperv/action.rb
+++ b/plugins/providers/hyperv/action.rb
@@ -207,6 +207,58 @@ module VagrantPlugins
         end
       end
 
+      def self.action_snapshot_delete
+        Vagrant::Action::Builder.new.tap do |b|
+          b.use ConfigValidate
+          b.use Call, IsState, :not_created do |env, b2|
+            if env[:result]
+              b2.use Message, I18n.t("vagrant_hyperv.message_not_created")
+              next
+            end
+
+            b2.use SnapshotDelete
+
+          end
+        end
+      end
+
+      def self.action_snapshot_restore
+        Vagrant::Action::Builder.new.tap do |b|
+          b.use ConfigValidate
+          b.use Call, IsState, :not_created do |env, b2|
+            if env[:result]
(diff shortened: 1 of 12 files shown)


sethvargo:
Hi @jtopper 

Thank you for the PR. This is looking really good! I have two main questions:

1. Since this is already as OSS project under scalefactory, I think we need written permission to bring this code into core.
2. Are there any documentation pages that need to be updated? You can find all the docs in the `website/` folder in the Vagrant repo.

jtopper:
1. How would you like to receive written permission?  I co-own The Scale Factory, and authored the plugin, and it's fine by me, but I understand the need to cross all the 't's. 
2. I'll take a look over the documentation, no doubt there will be some changes to make.

jtopper:
@sethvargo I've updated the docs as appropriate.  Please also find attached a letter of permission to use code from the scalefactory project.

[PR 7110 permission.pdf](https://github.com/mitchellh/vagrant/files/263699/PR.7110.permission.pdf)

sethvargo:
LGTM /cc @mwrock for the powershells tuff

mwrock:
:+1: looks good and I tested it too on my laptop with great success. created, listed, restored and deleted a hyper-v snapshot. Very nice!!

sethvargo:
- Thanks @mwrock! 
- Thanks @jtopper!

PR #7122, opened 2016-03-08
Fixes #7121: Corrected binary logic description.

### LINKED ISSUE #7121: Confusion about binary option description in Shell provisioner.
Currently [the documentation page](https://www.vagrantup.com/docs/provisioning/shell.html) states:

> binary (boolean) - Vagrant automatically replaces Windows line endings with Unix line endings. If this is true, then Vagrant will not do this. By default this is "false". If the shell provisioner is communicating over WinRM, this defaults to "true".

This should be opposite, so it should state:

> If this is false, then Vagrant will not do this.

See this [post](http://stackoverflow.com/q/28651275/55075) for more details.

kenorb:
### Overview
The description of binary option for shell is not correct.

### References
- Fixes GH-7121
- As per http://stackoverflow.com/q/28651275/55075 thread, this seems to be confusing.

> CRLFs are changed to LFs if this is set to true.

### DIFF
--- a/website/source/docs/provisioning/shell.html.md
+++ b/website/source/docs/provisioning/shell.html.md
@@ -48,7 +48,7 @@ The remainder of the available options are optional:
   the keys remain untouched.
 
 * `binary` (boolean) - Vagrant automatically replaces Windows line endings with
-  Unix line endings. If this is true, then Vagrant will not do this. By default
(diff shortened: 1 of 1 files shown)


PetrGlad:
Now the setting name becomes incorrect :) Normally "binary" implies that test is passed as is....

PR #7123, opened 2016-03-08
website: add a page to send folks to get to a signup link for atlas

pearkes:
This adds a page to describe Atlas as Atlas no longer has an easy to understand homepage.


sethvargo:
:+1:

PR #7126, opened 2016-03-10
Fixed linked clone failure when master VM is missing

### LINKED ISSUE #6742: VirtualBox: Vagrant does not recover if master VM for linked clones is deleted
Step to repeat:

- Create a vagrantfile with 
```ruby
  ...  
 vb.linked_clone = true
end
````
- run `vagrant up`: 
  - master VM for linked clones is created
  - new VM linking to master VM is created
- run `vagrant destroy`:
  - only VM linking to master VM is deleted
- manually delete master VM for linked clones in virtual box  
- run again `vagrant up` 

you get an error because vagrant still search for the previously created master VM for linked clone (I suppose the uuid of the VM is stored somewhere...).

How can I recover from this situation?
It is possible to improve Vagrant support to Virtual box linked clone in order to recover automatically from such situation?
It is possible to improve Vagrant support to Virtual box in order to allow giving a human readable name to the master VM for linked clones 

Environment:
Vagrant 1.8.1
VirtualBox  5.0.10 r104061
Osx 10.11.2

### LINKED ISSUE #7127: Fix Rspec Tests
`Rake.application.last_comment` was removed in Rake version 11 which came out today.
https://github.com/ruby/rake/commit/e76242ce7ef94568399a50b69bda4b723dab7c75#diff-53e46efb0e8e7ddcaa1826b1c508c756

The current version of `rspec` still uses `last_comment` but a PR has been merged to update to the newer `Rake.application.last_description`.
https://github.com/rspec/rspec-core/pull/2197

To get builds working we should pin to a rake version `< 11.0` until `rspec` can catch up.

### PR DESCRIPTION
**Pull request by jjulien**
### Overview
When using linked_clones, If the master VM is manually removed from VirtualBox a user will get a Box with UUID xxx-xxx-xxx Not Found error when trying to run `vagrant up` on a new machine.

This is because the master_id file exists `action/import_master.rb` sets ` env[:clone_id] = master_id_file.read.chomp if master_id_file.file?`

The `action/import.rb` code assumes if `env[:clone_id]` is set then it does not need to import the box, it can just clone it.  This is not true when the master VM has manually been removed.

The `action/import_master.rb` code  seemed to be setup to handle this situation with a nice `if` statement.  All the if statement did though was log a message.
```
          if env[:clone_id] && env[:machine].provider.driver.vm_exists?(env[:clone_id])
            @logger.info(
              "Master VM for '#{env[:machine].box.name}' already exists " +
              " (id=#{env[:clone_id]}) - skipping import step.")
            return
          end
```

This pull request just adds an `else` to this `if` statement to remove the `env[:clone_id]` setting so that when `action/import.rb` is called it knows it needs to first import the master, then clone.

Fixes #6742

### DIFF
--- a/plugins/providers/virtualbox/action/import_master.rb
+++ b/plugins/providers/virtualbox/action/import_master.rb
@@ -51,6 +51,8 @@ module VagrantPlugins
               "Master VM for '#{env[:machine].box.name}' already exists " +
               " (id=#{env[:clone_id]}) - skipping import step.")
             return
+          else
+            env.delete(:clone_id)
           end
 
           env[:ui].info(I18n.t("vagrant.actions.vm.clone.setup_master"))


jjulien:
@sethvargo Hey, can you take a look at this?

jjulien:
@sethvargo Is there a process I should be following for contributions?  Trying to figure out why this hasn't been closed or commented on.

sethvargo:
Thanks!

Issue #4304, opened 2014-08-06
[salt] There is no way to know if provisionning has failed 


trivoallan:
It would be convenient to be able to activate Salt's --retcode-passthrough flag, in order to have Vagrant fail when Salt provisionning fails.

see https://github.com/saltstack/salt/issues/6973

PR #7137, opened 2016-03-14
Pass through retcode of masterless Salt provisioner

aneeshusa:
### Overview
Pass through the return status of the Salt provisioner when run in masterless mode.

### References
- References GH-4304

The other calls to state.highstate all had this option added by commit
1cc78dc224677b9f2bf3a273199b53258722852e, but the masterless option
did not yet exist.


sethvargo:
No longer required as https://github.com/mitchellh/vagrant/pull/7207. Thank you though! 😄

Issue #7155, opened 2016-03-18
ubuntu 15.10 & 16.04 guests wont work with private networks


kiorky:
### Vagrant version
Vagrant 1.8.1

### Host operating system
ubuntu 15.04 (any ubuntu)

### Guest operating system
ubuntu 16.04 (15.10+ exhibit the problem)

### Vagrantfile
```rb
Vagrant.configure(2) do |config|
  config.vm.box = "xenial64"
  config.vm.box_url = "http://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-amd64-vagrant.box"
  config.vm.network "private_network", ip: "192.168.33.10"
end
```
### Debug output
Vagrant fails first with
```
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

hostname -f

Stdout from the command:
Stderr from the command:
sudo: unable to resolve host ubuntu-xenial
mesg: ttyname failed: Inappropriate ioctl for device
hostname: No address associated with hostname
```

If we create /etc/network/interfaces.d/eth1.cfg like this
```
:!cat /etc/network/interfaces.d/eth1.cfg
# The primary network interface
auto eth1
iface eth1 inet manual
```

And powerdown/vagrant up the vm, the provisionning will work

sethvargo:
Hi @kiorky 

Thank you for opening an issue. I am unable to reproduce this issue with the following vanilla Vagrantfile. The Vagranfile you supplied is doing a lot of things including monkey-patching the ruby, defining helpers, and performing conditional execution that makes it very difficult to reason about the actual Vagrantfile that is being used and the parameters. Are you able to reproduce this with a vanilla Vagrantfile?

```ruby
Vagrant.configure(2) do |config|
  config.vm.box = "bento/ubuntu-15.04"
  config.vm.network "private_network", ip: "33.33.33.10"
end
```

There are so many variables in your Vagrantfile that it's not possible to narrow down where the problem might be:

1. You are binding to specific adapters
2. You are pulling in data from the environment which could be changing
3. You are using the non-stable versions of the ubuntu OS
4. You are making system calls to determine resources
5. You are monkey-patching Ruby's core library, which can have unintended side-effects on the way Vagrant evaluates code
6. You are reading data from an external yaml file
[...]

Issue #6871, opened 2016-01-14
Ubuntu 15.10 error when enabling network interfaces


gabriel403:
Previously using a custom 15.04 all worked fine

```
==> dev: Setting hostname...
==> dev: Configuring and enabling network interfaces...
==> dev: Mounting shared folders...
```

new custom 15.10 box, virtually no changes to box just an updated ubuntu version

```
==> dev: Setting hostname...
==> dev: Configuring and enabling network interfaces...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

/sbin/ip addr flush dev eth1 2> /dev/null

Stdout from the command:



Stderr from the command:

mesg: ttyname failed: Inappropriate ioctl for device


```

Similar to #6577 but I'm already on vagrant 1.8

```
$ vagrant -v
Vagrant 1.8.1
```

schmunk42:
Same here with `1.8.1` when running

    config.vm.provision "shell", run: "always", inline: "route add default gw 1.2.3.4"

gabriel403:
Anyone had a chance to look into this?

RudyJessop:
Take a look at this http://askubuntu.com/questions/671228/lost-network-device-on-upgrade-to-15-10-on-test-vm

`eth0` was changed to `enp0s3` 

In moving forward in fixing this issue. Would need more looking into.

gabriel403:
@RudyJessop this is a fresh vm so makes perfect sense

```
enp0s3 is the name of an Ethernet device as given by the new version of udev, which enables stateless persistant network interface names. 
```

gMagicScott:
I think this is a duplicate of #6653 and would be fixed in pull #6724 if it gets merged.

gMagicScott:
@schmunk42, I think you are experiencing a related but different issue. The `ubuntu/wily64` box seems to be failing at the initial ssh connection during boot and is outlined in #6683.

schmunk42:
@gMagicScott Yes, I think you're right. I just get the error `mesg: ttyname failed: Inappropriate ioctl for device` and I must not use `run: "always",` - it's working fine for me so far.

jknightihiji:
That's what it seems that I'm getting now.

gabriel403:
@gMagicScott that makes sense thanks, let's hope the pr get's merged sometime

dch:
[...]

PR #7159, opened 2016-03-20
Do not return an error if ifdown fails

sethvargo:
Ubuntu versions prior to 16.04 always returned a successful exit status, even if one tried to down an interface that does not exist. This behavior changed in Ubuntu 16.04 to return an error. This commit preserves the old behavior.

- Fixes GH-7155
- Fixes GH-6871

### DIFF
--- a/plugins/guests/debian/cap/configure_networks.rb
+++ b/plugins/guests/debian/cap/configure_networks.rb
@@ -42,7 +42,10 @@ module VagrantPlugins
             # each specifically, we avoid reconfiguring eth0 (the NAT interface) so
             # SSH never dies.
             interfaces.each do |interface|
-              comm.sudo("/sbin/ifdown eth#{interface} 2> /dev/null")
+              # Ubuntu 16.04+ returns an error when downing an interface that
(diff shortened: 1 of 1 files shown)


valkum:
Works for me.

kiorky:
Maybe a better bypass whould be to only return true if we dont grep the interface into /etc/network/interfaces /etc/network/interfaces.d/*; because ifdown can return 1 in case of a real problem, if the interface is well configured, but cant go down for some reason

kiorky:
but yes, this fix as--is for this use case works also for me.

s12v:
works for me as well

unattended-user:
Does anybody know when new version with this fix is being released?

rowanembarks:
Does anybody know when new version with this fix is being released?

dongweiming:
@sethvargo  +1

thiagotalma:
@sethvargo @mitchellh please release! :pray:

foxx:
@sethvargo @mitchellh Any chance of a release? Seems we're having to wait months between patches being made available :/

sethvargo:
Tomorrow or monday at the latest

PR #7179, opened 2016-03-27
Add section to NFS sync folder docs for non-standard system utility locations

c0state:
Add section in sync folder documentation on how to modify sudoers file to not require a password when user's system utility location(s) are non-standard.


sethvargo:
Thanks!

PR #7180, opened 2016-03-28
Specifically ask for help page

jamescarr:
When I tried this out the piece that detects ansible fails because running just `ansible` returns a non-zero exit code.


sethvargo:
Hi @jamescarr 

`command` shouldn't actually execute the command; it's just checking for its existence. Can you please share the debug output?

jamescarr:
@sethvargo I actually made a mistake. I thought this is where it was failing but debugging led me to the same defect fixed in https://github.com/mitchellh/vagrant/commit/9dbdb9397a92d4fc489e9afcb022621df7f60d11

jamescarr:
BTW when will 1.8.2 be released? https://github.com/mitchellh/vagrant/commit/9dbdb9397a92d4fc489e9afcb022621df7f60d11 is pretty much blocking me here. Is it possible to just monkey patch the module to provide the same behavior?

PR #7181, opened 2016-03-29
Remove sysvinit else clauses in NFS for Arch Linux

Barthalion:
Currently the code tries to detect if Arch uses systemd via checking
comm= of PID 1. As access to proc filesystem might be restricted due to
hideproc set and systemd is default for all Arch Linux installations
since October 2012, let's just ditch that check.


nebril:
!wspaniale

sethvargo:
919f3ee

PR #7190, opened 2016-03-31
Ansible parallel race

### LINKED ISSUE #6526: Ansible provisioner has a race when used with multi machine configuration
This is a follow on from https://github.com/mitchellh/vagrant/pull/5551, which was an initial attempt to fix a race condition in the ansible provisioner.

The race occurs when bringing multiple VM's up in parallel and using the ansible provisioner as the first provisioner. It will only appear with providers where the SSH information relies on interrogating the system after it is booted (vagrant-libvirt behaves this way), although it is likely that a different race condition of not being able to connect to other machines would occur if the SSH information was available from a provider as soon as it started booting, but SSH was not accessible until it was finished booting and the Vagrant environment wished to override the '--limit' option to reference more than one host.

Using the vagrant-libvirt provider, the race condition will exhibit the follow errors from the ansible subprocess:
`provided hosts list is empty`
or
`Specified --limit does not match any hosts`

The author of the recent rewrite of the ansible provisioner to support in guest execution of ansible appears to be familiar with some parts of the issue given the comment at https://github.com/mitchellh/vagrant/blob/a3c077cbe0b27339bb14c7bcd404ea64fefd16d4/plugins/provisioners/ansible/provisioner/base.rb#L84

The race condition remains due to the following:
* the same inventory file is used for all ansible provisioner subprocesses
* each machine thread generates the inventory file contents and attempts to update the file if contents differ
* the code incorrectly assumes that all valid running machines will be ready to provide ssh_info upon the first machine starting to execute the ansible provision

This is due to the file being truncated by another machine thread while an ansible subprocess in the current thread is attempting to read the file.

Likely the solution will come down to the following items to look at:
* unique inventory file for each machine
  * retain the inventory directory behaviour and generate a unique inventory file for each machine containing just that machine
  * revert to passing the path to the inventory file and generate a unique inventory file for each machine containing information for all machines
* wait till all machines that could be part of the inventory to be running before generating
  * retry multiple times to get the ssh_info for each active machine
    * this is somewhat inefficient without a way to determine if some machines will never be accessible because they failed to boot
  * register an internal status, or have a code pattern to determine when each machine has reached certain builtin actions.
    * internal status: currently the machine state is provided by provider, which means it can be unique to that provider forcing a provisioner to know what states providers will return in order to support them
     * code pattern action mechanism: being able to determinately wait until machines will be ready to respond to ssh connections or spot that they have been destroyed so as to know not to wait for them to be ready

There may be other items worth looking at, but think it would be useful for some guidance here as to best way to proceed in fixing.

electrofelix:
Combined, these two changes should reduce the number of race issues when bringing up multiple machines in parallel and using the ansible provisioner to configure each guest.

The state lock is required any time active_machines is used and anything that can trigger a state check on the other machines occurs. Currently the machine index assumes that it will only be updated by itself, however if you skip syncing of the current directory to the machine, it's possible to get to the ansible provisioner checking all the other machines for ssh_info while they are in the process of changing state, which will trigger a Locked Machine exception.

The use of a temp file to write out the inventory and then replace should ensure that if the contents have changed that ansible will only ever see the old or new contents, and never the truncated file contents that is caused by opening with the 'w' mode.

Ansible ignores files starting with '.' since version 1.2, so using the same directory to ensure we are on filesystem to use rename instead of a move should be safe. Anyone using ansible < 1.2 should really upgrade.

These should fix #6526 for most use cases where the provisioner is only running against the current machine. Handling the provisioner configured to run against all other machines using parallel up without random failures due to some machines still coming up, will require some form of waiting till all running machines are ready.

This has mostly been developed against the vagrant-libvirt provider, which triggers this due to needing to wait to get the SSH information from the machine after it has booted.

### DIFF
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -109,6 +109,7 @@ module Vagrant
       @provider_options = provider_options
       @ui              = Vagrant::UI::Prefixed.new(@env.ui, @name)
       @ui_mutex        = Mutex.new
+      @state_mutex     = Mutex.new
 
       # Read the ID, which is usually in local storage
       @id = nil
@@ -505,11 +506,17 @@ module Vagrant
       # master index.
       uuid = index_uuid
       if uuid
-        entry = @env.machine_index.get(uuid)
-        if entry
-          entry.state = result.short_description
-          @env.machine_index.set(entry)
-          @env.machine_index.release(entry)
+        # active_machines provides access to query this info on each machine
+        # from a different thread, ensure multiple machines do not access
+        # the locked entry simultaneously as this triggers a locked machine
+        # exception.
+        @state_mutex.synchronize do
+          entry = @env.machine_index.get(uuid)
+          if entry
+            entry.state = result.short_description
+            @env.machine_index.set(entry)
+            @env.machine_index.release(entry)
+          end
         end
       end
 
--- a/plugins/provisioners/ansible/provisioner/host.rb
+++ b/plugins/provisioners/ansible/provisioner/host.rb
@@ -132,8 +132,15 @@ module VagrantPlugins
           inventory_file = Pathname.new(File.join(inventory_path, 'vagrant_ansible_inventory'))
           @@lock.synchronize do
             if !File.exists?(inventory_file) or inventory_content != File.read(inventory_file)
-              inventory_file.open('w') do |file|
-                file.write(inventory_content)
+              begin
+                # ansible dir inventory will ignore files starting with '.'
+                inventory_tmpfile = Tempfile.new('.vagrant_ansible_inventory', inventory_path)
+                inventory_tmpfile.write(inventory_content)
+                inventory_tmpfile.close
+                File.rename(inventory_tmpfile.path, inventory_file)
+              ensure
+                inventory_tmpfile.close
+                inventory_tmpfile.unlink
               end
             end
           end


sethvargo:
@gildegoma

gildegoma:
The idea (of an atomic replacement of the generated inventory file) looks very good to me. 

But before I start testing it, I'd really appreciate to get more confident about the following concurrency questions I have:

 1. @electrofelix have you already verified that there is no deadlock danger by using now two mutexes (the inventory generation lock in `ansible/provisioner.rb` introduced by eb6aa2ac8c6238dc4aff71bd5b4463b992bc7189 and this new `state_mutex` in core `lib/vagrant/machine.rb` ?
 2. @sethvargo (or @mitchellh) can you please confirm that this Mutex addition in `lib/vagrant/machine.rb` is safe and enough performant for all "well known" usages of the `Vagrant::Machine::state` method?

Thanks a lot in advance for your help!

electrofelix:
@gildegoma we monkey patched this into our environment using the following snippet loaded at the start of the Vagrantfile:
```
require 'vagrant/machine'

# following is required to load the classes in order to monkey patch them
require Vagrant.source_root.join('plugins/provisioners/ansible/config').to_s

module Vagrant
  class Machine

    alias_method :old_initialize, :initialize
    def initialize(*args)
      @state_mutex = Mutex.new
      old_initialize(*args)
    end

    # state uses 'machine_index.get(uuid)' which returns a locked entry
    # that must be explicitly released by the calling entity
    # use a lock to ensure that if state is accessed by multiple threads
    # they do not attempt to access the locked entity simultaneously
    alias_method :old_state, :state
    def state
      result = nil
      @state_mutex.synchronize do
        result = old_state
      end
      result
    end
  end
end
```
This has been in place since around October with averaging ~400 CI runs a day bringing up 4-7 machines in parallel, without us hitting a Machine Locked error or other related error.

The lock introduced by eb6aa2a is only applied when updating the inventory file, all the state information has already been retrieved at that stage, so there is no code path that could involve needing both locks, they are fully independent.

gildegoma:
@electrofelix thanks a lot for this detailed feedback. I'll make some tests on my own and plan to bring this into Vagrant 1.8.2.

sethvargo:
Adding the mutex is fine

gildegoma:
LGTM 🌅, but on principle I'll run some tests before merging. Looking forward :-)

PR #7191, opened 2016-03-31
Exit the SSH session before vagrant destroy

ondine:
Suggestion: How to exit the session: http://superuser.com/questions/704032/how-to-exit-ssh-of-vagrant-in-terminal


sethvargo:
Thanks!

PR #7192, opened 2016-03-31
Fix Subprocess problem with closure of stdin by caller

driskell:
Hello,

I've been writing a plugin or two for Vagrant and encountered an issue with the Subprocess utility. If you call it with a block and request notification of stdin, that's great so you can feed in data as and when it's ready for it. However, if you then close the stdin because you've finished sending data (which you have to or it will busy loop your block) you get the following exception and it crashes:

    E, [2016-03-31T19:52:09.690798 #24234] ERROR -- : exception while processing events: closed stream

This PR fixes the issue by only scheduling stdin for IO::Select if it's requested AND if it's open.

### DIFF
--- a/lib/vagrant/util/subprocess.rb
+++ b/lib/vagrant/util/subprocess.rb
@@ -146,7 +146,7 @@ module Vagrant
 
         @logger.debug("Selecting on IO")
         while true
-          writers = notify_stdin ? [process.io.stdin] : []
+          writers = notify_stdin && !process.io.stdin.closed? ? [process.io.stdin] : []
           results = ::IO.select([stdout, stderr], writers, nil, 0.1)
           results ||= []
           readers = results[0]


sethvargo:
Hi @driskell 

Thank you for the PR. This seems reasonable to me, but it feels weird that you'd pass in an IO that is closed. I feel like that should raise some kind of exception because otherwise the behavior is swallowed.  What do you think?

driskell:
I didn't pass in anything?

When you call execute with a block and options :notify_stdin => true then it calls my block with the stdin IO object when write is ready. If I then close stdin to signal EOF to the subprocess Vagrant crashes. This is all internal to Vagrant and I'm not passing anything. Vagrant is responsible the IO object not the caller.

As to why I need to close stdin, if I don't then Vagrant keeps calling my block and if i ignore it it hits 100% CPU, also the remote process never finishes as its waiting for EOF.

driskell:
Please let me know if you have any further questions or need anything adjusting. Thanks!

driskell:
@sethvargo Might you be able to take another look at this? Thanks!

sethvargo:
Hi @driskell 

> As to why I need to close stdin, if I don't then Vagrant keeps calling my block and if i ignore it it hits 100% CPU, also the remote process never finishes as its waiting for EOF.

That's not what I'm seeing in the code. Is your subprocess exiting? Vagrant breaks if the subprocess exits.

driskell:
@sethvargo No. My subprocess will only exit if I close stdin. In my plugin I am writing to the stdin of `rsync` and then I need to wait for it to exit gracefully. However, unless I close stdin, `rsync` will never exit, as it keeps waiting until I finish giving it data.

So either:
1. I close stdin and vagrant crashes because it attempts to poll a closed FD
2. I do not close stdin and it is always writable and vagrant uses 100% CPU because it keeps calling my block, which is unable to do anything (there's nothing left to write).

Hope this explains it!

sethvargo:
@driskell why does your application only stop if you close stdout? I'm still not convinced that it's Vagrant's responsibility to check if the io is writable/not closed.

driskell:
> why does your application only stop if you close stdout?

@sethvargo The program is `rsync` and is not mine. `rsync` is reading, processing, and outputting because I'm using the `--files-from=-` parameter to read the files to transfer from the stdin. However, this issue would apply to every single command that processes stdin. `cat` does this. `gzip`, `grep`, `wc`, almost every single command that takes input will continue to read input and never exit until input is closed.

Looking into Vagrant source it seems that nothing uses the stdin capabilities, thus why this issue is not seen thus far in Vagrant. The only place that references it is Powershell for Hyper-V but it seems a block is never given so I can only presume it busy loops until the command completes.

> I'm still not convinced that it's Vagrant's responsibility to check if the io is writable/not closed.

That's fine. The main question however is who is responsible for closing stdin? Logically, only the caller knows *when* stdin should be closed, so there needs to be some way at least for the caller to tell Vagrant to close stdin if it indeed owns the descriptor. If the caller is given the right to closer the descriptor (like I am doing) then Vagrant needs to handle that.

Ultimately, there's an issue here. Maybe I shouldn't be closing the descriptor the way I am - but the question then remains, how do I close it? I do feel though my solution is the smallest and the best fit solution - that Vagrant API allows the passed IO descriptor for stdin to be closed by the caller when it has completed. At the moment, it doesn't, it crashes.

driskell:
> why does your application only stop if you close *stdout*?

Sorry! I read this as "stdin" and responded so. Was it a typo? I have to close *STDIN*. Not *STDOUT*.

driskell:
@mitchellh @sethvargo Hey guys. Did you get any chance to look at this again?
Thanks!

justjake:
I have just written my own implementation of this pull request, only to find this branch! Consider this my upvote on this issue -- this is a real difficiency in `Subprocess`'s API.

Here's my pull request: https://github.com/mitchellh/vagrant/pull/7778

chrisroberts:
:+1: Same thing as #7778, can you add some tests to cover the behavior? Thanks!

chrisroberts:
This is fixed via #7778. Thanks!

PR #7207, opened 2016-04-05
Salt: move masterless config to apply to all platforms

### LINKED ISSUE #6915: Salt provisioner ignores run_highstate setting in masterless mode
On vagrant 1.8.1 i am experiencing issues when trying to bring up a masterless minion without running highstate. The relevant snippet from my config looks like this:

    salt.masterless = true
    salt.run_highstate = false

My expectation is that i end up with an installed salt-minion and that no highstate call is made during provisioning. After taking a look at the relevant [code](https://github.com/mitchellh/vagrant/blob/v1.8.1/plugins/provisioners/salt/provisioner.rb#L336) its quite obvious that the [call_masterless method] (https://github.com/mitchellh/vagrant/blob/v1.8.1/plugins/provisioners/salt/provisioner.rb#L323) silently ignores the run_highstate setting. From my point of view this is wrong or should (if there are reasons for this behaviour) be at least documented and an information passed back to the user like it happens for other "senseless" stuff like [calling orchestrate on a minion] (https://github.com/mitchellh/vagrant/blob/v1.8.1/plugins/provisioners/salt/provisioner.rb#L377).

If there are no reasons and this is simply a bug i would like to fix this behaviour.

msarahan:
This makes the masterless config setting more integrated with the other highstate logic by making it control a string (--local) instead of a separate code block.  In doing so, it also avoids the masterless section overriding the highstate setting.

This was essential for fixing masterless configurations on Windows, because the separate code block for masterless config did not include Windows logic.

Closes #6924
Fixes #6915

If you can point me to tests, I'm happy to try to work some in with this PR.  I have to warn you, though, I'm a Python-head, and not very good with Ruby.

### DIFF
--- a/plugins/provisioners/salt/provisioner.rb
+++ b/plugins/provisioners/salt/provisioner.rb
@@ -320,43 +320,38 @@ module VagrantPlugins
         end
       end
 
-      def call_masterless
-        @machine.env.ui.info "Calling state.highstate in local mode... (this may take a while)"
-        cmd = "salt-call state.highstate --local#{get_loglevel}#{get_colorize}#{get_pillar}"
-        if @config.minion_id
-          cmd += " --id #{@config.minion_id}"
-        end
-        @machine.communicate.sudo(cmd) do |type, data|
-          if @config.verbose
-            @machine.env.ui.info(data)
-          end
-        end
-      end
-
       def call_highstate
-        if @config.masterless
-          call_masterless
-        elsif @config.run_highstate
+        if @config.run_highstate
+          local=""
+          if @config.masterless
+            local=" --local"
+          end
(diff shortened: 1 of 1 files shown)


jakirkham:
Sorry to be noisy, but is there any chance we could get a review of this? Seems to have sat quietly here for two weeks.

msarahan:
Great!  Thanks for merging!

jakirkham:
Thanks @sethvargo. Do you know if there is a release schedule and whether this can be included?

sethvargo:
@jakirkham this week or next

jakirkham:
Awesome! Thanks. Looking forward to it. 😄

jakirkham:
Just for future ref (in case other people come looking for this), it is in 1.8.3+.

PR #7214, opened 2016-04-06
Plugin install do not highlight the fact we can refer to a gem file stored locally

renoirb:
In some case, we may want to use Vagrant in networks that aren’t connected to the Internet.

This snippet shows how we can achieve this.

If it wasn’t for this [Stack Overflow response][superuser], I would still be working with this.

   [superuser]: https://superuser.com/questions/801745/cant-vagrant-install-a-plugin-gem-dependency-fails-despite-being-installed/899414#answer-899414

### DIFF
--- a/website/source/docs/cli/plugin.html.md
+++ b/website/source/docs/cli/plugin.html.md
@@ -62,6 +62,19 @@ This command accepts optional command-line flags:
   to a more complex constraint by comma-separating multiple constraints:
   "> 1.0.2, < 1.1.0" (do not forget to quote these on the command-line).
 
+## Behind corporate firewalls
+
+### Offline package install
+
+Installing is done the same way, but can refer to a Gem file downloaded
+from [RubyGems.org](https://rubygems.org/) in a folder (e.g. `C:\somewhere`)
+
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @renoirb 

If you look up on line 29:

> This installs a plugin with the given name or file path. If the name
 is not a path to a file, then the plugin is installed from remote
 repositories, usually [RubyGems](https://rubygems.org). This command will
 also update a plugin if it is already installed, but you can also use
 `vagrant plugin update` for that.

It specifically calls out that the name can be a local file path. I don't want to add a complete section to the docs for this, but I would welcome a suggestion to make that sentence clearer.

renoirb:
Oh. I haven’t realized this.

Maybe because I read diagonally, like many people does, and ignored that block of text.

How about we add some more command line examples, such as in this use-case, so that it TL;DR readers will at least see the command and realize it’s possible.

sethvargo:
https://github.com/mitchellh/vagrant/commit/4157082bf3a91292c649f2694bbcfa01881cadc4

sethvargo:
Thanks!

PR #7223, opened 2016-04-10
update bash completion(snapshot)

tKammy:
Add command: snapshot
Add Sub command for snapshot: delete list pop push restore save


sethvargo:
Thanks!

PR #7228, opened 2016-04-12
update cmd format

tKammy:
Add parameters to below commands' usage.

destroy, halt, package, port, provision, reload, resume, ssh, ssh-config,
status, suspend, up

### DIFF
--- a/website/source/docs/cli/destroy.html.md
+++ b/website/source/docs/cli/destroy.html.md
@@ -9,7 +9,7 @@ description: |-
 
 # Destroy
 
-**Command: `vagrant destroy`**
+**Command: `vagrant destroy [name|id]`**
 
 This command stops the running machine Vagrant is managing and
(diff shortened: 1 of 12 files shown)


sethvargo:
Hi @tKammy 

Thank you for opening an issue. We generally do not document the full CLI signature in the docs, since it's available when you run `vagrant command -h`. Is there an issue you are experiencing?

tKammy:
I understood.

>  Is there an issue you are experiencing?
I didn't know that we can use 'id' for these CLI via both `vagrant command -h` and the docs.
So, is it OK to fix 'vagrant command -h' like following?

`Usage: vagrant destroy [options] [name|id]`

sethvargo:
Hi @tKammy 

When you say "id", do you mean the machine-id from the output of global-status?

tKammy:
Yes, I mean the machine-id.

PR #7241, opened 2016-04-18
Handle new 'predictable' network interface naming style

dougneal:
Explicitly discover the names of the ethernet interfaces, and don't assume that they are eth0, eth1 etc as this does not hold true under newer systemd-based distributions.

Addresses: https://github.com/mitchellh/vagrant/issues/7155 https://github.com/mitchellh/vagrant/issues/6871


cnk:
I would love to be able to try this out; it looks like it fixes a number of the issues I am having building an Ubuntu 16.04 LTS dev VM. 

Where do the scripts live to build a binary release locally? I tried building the vagrant gem from source and installing it along with its dependencies. But when I run it, I just get the 'vagrant doesn't run as a gem anymore' warnings.

dougneal:
Hey @cnk - not sure on the process for doing a local build but you should be able to run it from the checked out source tree: https://github.com/mitchellh/vagrant/wiki/Installing-Vagrant-from-Source

cnk:
Yeh I figured that out. I still need to figure out how to run with a salt provisioner. 

@dougneal I am still getting the error from `sudo hostname -f` not resolving. So I will have to do some more poking around.

dougneal:
I ran into that too, it's because the image doesn't have `libnss-myhostname`. Try the `boxcutter/ubuntu1604` image for now, it seems to work better!

plugins/guests/debian/cap/configure_networks.rb
```diff
@@ -16,6 +16,17 @@ def self.configure_networks(machine, networks)
             comm.sudo("sed -e '/^#VAGRANT-BEGIN/,$ d' /etc/network/interfaces > /tmp/vagrant-network-interfaces.pre")
             comm.sudo("sed -ne '/^#VAGRANT-END/,$ p' /etc/network/interfaces | tac | sed -e '/^#VAGRANT-END/,$ d' | tac > /tmp/vagrant-network-interfaces.post")
 
+            kernel_ifnames = []
+            comm.sudo("/usr/bin/find /sys/class/net -type l -print | sort") do |type, data|
+              if type == :stdout then
+                data.split("\n").each do |line|
+                  if (line =~ /^\/sys\/class\/net\/(enp[0-9]+s[0-9]+|eth[0-9]+)$/) then
+                    kernel_ifnames << $1
```
caldwell: I'm not sure this is technically correct. My systemd based machine (though not Ubuntu) has an ethernet adapter named `enp0s31f6`. This wouldn't be caught by this RE.

ahrenstein:
+1 for this fix

plugins/guests/debian/cap/configure_networks.rb
geerlingguy: This would only be for debian-based guests... is your machine running a Debian derivative, or something else (e.g. RHEL)?

caldwell: Debian GNU/Linux stretch/sid

caldwell: The name format appears to be described in [systemd's udev-builtin-net_id.c](https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c#L20)

geerlingguy:
Here's a workaround for those interested in getting a Xenial base box working prior to this PR getting pulled in: https://github.com/geerlingguy/packer-ubuntu-1604/issues/1#issuecomment-213130111

hnakamur:
I also created a patch for vagrant 1.8.1.
https://gist.github.com/hnakamur/f23e35083bf60d5c573392bfdad94466

krzysbaranski:
for now I use the following script in ubuntu/xenial64, after that I can reload box with provisioning

```
  # fix hosts ubuntu/xenial
  echo 127.0.0.1 `hostname` |sudo tee -a /etc/hosts

  # fix network ubuntu/xenial
  # disable Predictable Network Interface Names
  ip a show dev eth1 > /dev/null
  if (($? > 0)); then
    echo "#### eth1 not exists (eth1 and eth2 is expected by vagrant)"
    echo "#### turning off Predictable Network Interface Names"
    grep -q "net.ifnames=0" /etc/default/grub
    if (($? > 0)); then
      echo "GRUB_CMDLINE_LINUX=\"net.ifnames=0"\" | sudo tee -a /etc/default/grub
      sudo update-grub
      echo "#### restart box to disable Predictable Network Interface Names"
    fi
  fi
```

hnakamur:
I created a simpler pull request https://github.com/mitchellh/vagrant/pull/7253 to fix this issue and also a patch to vagrant 1.8.1 https://github.com/mitchellh/vagrant/pull/7253#issuecomment-216125836

domnulnopcea:
when is this going to be merged?

sebi-hgdata:
Just tested it out with
```
  
    v.memory = 1024
    v.cpus = 1
  end
 
  config.vm.define "b1" do |b1|
    b1.vm.network "private_network", ip: "192.168.33.2"
  end
  
end
```
 and it works.

PR #7249, opened 2016-04-21
Fix issues with virtual interfaces

dzamorano:
This change fix this issue https://github.com/openshift/openshift-ansible/issues/1341

After provisioning vagrant configure private_network on interface eth0 instead eth1 because there are br0 in result of ls /sys/class/net | egrep -v lo\|docker

### DIFF
--- a/plugins/guests/fedora/cap/configure_networks.rb
+++ b/plugins/guests/fedora/cap/configure_networks.rb
@@ -25,7 +25,7 @@ module VagrantPlugins
           end
 
           if virtual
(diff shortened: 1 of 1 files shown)


dzamorano:
The command 

> find -L /sys/class/net -xtype l -name device -maxdepth 2 2> /dev/null | awk -F/ '{print \$5}'

only returns the physical interfaces, this way it doesn't display the virtuals ones (eg docker, bridge...) because vagrant shouldn't manage it.

PR #7253, opened 2016-04-23
Modify configure_networks to support udev's predicatable network interface names on Ubuntu/Debian

hnakamur:
This pull request is another try to fix to #7241 and #6871.
This is based on [Add networking cap for Ubuntu guest by gMagicScott · Pull Request #6724 · mitchellh/vagrant](https://github.com/mitchellh/vagrant/pull/6724/files) and modified so that:

- Use `ip route | awk '$1=="default"{print $NF;exit}'` to get the main_interface.
- Use `ls /sys/class/net | grep -v -E '^(lo$|docker|lxc)'` to get available_interfaces then remove the main_interface from them.
- Use `if [ `/bin/cat /sys/class/net/#{interface}/operstate` = up ]; then /sbin/ifdown #{interface} 2> /dev/null; fi` to bring down a network interface.
- Modify network interface templates to support both naming style like eth0 and enp0s8.
- Use `hostnamectl --static status` to get the current hostname on distributions with systemd.

Tested with following Vagrantfiles.

Ubuntu Xenial with predicatable network interface names

```
  config.vm.network "private_network", ip: "192.168.33.13"
end
```

Ubuntu Trusty with old network interface names

```
  config.vm.network "private_network", ip: "192.168.33.9"
end
```

Debian Wheezy  with old network interface names

```
  config.vm.box = "debian/wheezy64"
  config.vm.network "private_network", ip: "192.168.33.16"
end
```

Debian Jessie with old network interface names

```
  config.vm.box = "debian/jessie64"
  config.vm.network "private_network", ip: "192.168.33.17"
end
```

Tested with
* Vagrant 1.8.1
* VirtualBox 5.0.16
* box versions
  - debian/jessie64            (virtualbox, 8.4.0)
  - debian/wheezy64            (virtualbox, 7.10.0)
  - ubuntu/trusty64            (virtualbox, 20160406.0.0)
  - ubuntu/xenial64            (virtualbox, 20160420.3.0)

Note you need to use VirtualBox 5.0.16, not the latest 5.0.18.
[Bug #1573058 “Ubuntu 16.04 current not booting in Vagrant (gurum...” : Bugs : cloud-images](https://bugs.launchpad.net/cloud-images/+bug/1573058)


hnakamur:
Confirmed ubuntu/xenial64 (virtualbox, 20160420.3.0) on Vagrant 1.8.1 with this patch works well with VirtualBox 5.0.20

hnakamur:
I also created a patch for vagrant 1.8.1
https://gist.github.com/hnakamur/f23e35083bf60d5c573392bfdad94466

plugins/guests/debian/cap/configure_networks.rb
```diff
@@ -9,8 +10,22 @@ module Cap
...
+            comm.execute(%q!ip route | awk '$1=="default"{print $NF;exit}'!) do |type, data|
+              main_interface = data.chomp if type == :stdout
+            end
+            @logger.debug("debian configure_networks. main_interface=#{main_interface.inspect}")
+            available_interfaces = []
+            comm.execute("ls /sys/class/net | grep -v -E '^(lo$|docker|lxc)'") do |type, data|
+              available_interfaces = data.chomp.split("\n") if type == :stdout
+            end
+            available_interfaces.delete(main_interface)
+            @logger.debug("debian configure_networks. available_interfaces=#{available_interfaces.inspect}")
```
chrisvire: Do you still need the debug logging?  Should "require log4r" and logger references be removed?

chrisvire:
Do you still need the debug logging?

plugins/guests/debian/cap/configure_networks.rb
hnakamur: @chrisvire I removed the debug logging. Thanks!

plugins/guests/ubuntu/cap/change_host_name.rb
```diff
@@ -6,6 +6,23 @@ def self.change_host_name(machine, name)
           super
         end
 
+        def initialize(machine, new_hostname)
+          super
+          @logger = Log4r::Logger.new("vagrant::plugins::guest")
```
chrisvire: Still some logging

plugins/guests/ubuntu/cap/change_host_name.rb
```diff
@@ -6,6 +6,23 @@ def self.change_host_name(machine, name)
...
+        end
+
+        def get_current_hostname
+          return super unless systemd?
+
+          hostname = ""
+          sudo "hostnamectl --static status" do |type, data|
+            hostname = data.chomp if type == :stdout && hostname.empty?
+          end
+          @logger.debug("ubuntu get_current_hostname got hostname=#{hostname} by hostnamectl.")
```
chrisvire: And here

hnakamur: @chrisvire I removed them too. Thanks for a catch!
This time I checked no logging is in the result of `git diff master`.

plugins/guests/debian/cap/configure_networks.rb
```diff
@@ -11,6 +11,16 @@ class ConfigureNetworks
 
         def self.configure_networks(machine, networks)
           machine.communicate.tap do |comm|
+            main_interface = ''
+            comm.execute(%q!ip route | awk '$1=="default"{print $NF;exit}'!) do |type, data|
+              main_interface = data.chomp if type == :stdout
+            end
+            available_interfaces = []
+            comm.execute("ls /sys/class/net | grep -v -E '^(lo$|docker|lxc)'") do |type, data|
```
ctrlrsf: I think `^(lo$|docker|lxc)` should be `^(lo$|docker|lxc|lxd)`. The lxd project recently started calling their bridge interfaces lxdbrX instead of lxcbrX.

hnakamur: Thanks for a good catch!
I add a commit to support lxd too.

hnakamur:
I also updated my patch for vagrant 1.8.1 to ignore network interfaces whose name matches the regex ^lxd as well as ^lo$, ^docker and ^lxc
https://gist.github.com/hnakamur/f23e35083bf60d5c573392bfdad94466

sni:
Using this patch makes my ubuntu12-04 vm hang on "Configuring and enabling network interfaces..."

plugins/guests/debian/cap/configure_networks.rb
```diff
@@ -11,6 +11,16 @@ class ConfigureNetworks
 
         def self.configure_networks(machine, networks)
           machine.communicate.tap do |comm|
+            main_interface = ''
+            comm.execute(%q!ip route | awk '$1=="default"{print $NF;exit}'!) do |type, data|
+              main_interface = data.chomp if type == :stdout
+            end
+            available_interfaces = []
+            comm.execute("ls /sys/class/net | grep -v -E '^(lo$|docker|lx[cd]|veth)'") do |type, data|
```
rmoriz: somewhat related, CoreOS also uses systemd naming, maybe we want to unify it? https://github.com/mitchellh/vagrant/pull/6610/files

hnakamur:
@sni Could you share your Vagrantfile? Did you use the patch for Vagrant 1.8.1 or for master?

plugins/guests/debian/cap/configure_networks.rb
hnakamur: I think commas in the regex `(e[n,t][h,s,p][[:digit:]]([a-z][[:digit:]])?` in https://github.com/mitchellh/vagrant/pull/6610/files should be deleted to say the least.

* [PredictableNetworkInterfaceNames](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/)
* [systemd/udev-builtin-net_id.c at 25b0e6cb99c2a13c8f1ae9c5057f6941b722e764 · systemd/systemd](https://github.com/systemd/systemd/blob/25b0e6cb99c2a13c8f1ae9c5057f6941b722e764/src/udev/udev-builtin-net_id.c#L20)

I think `^(en|eth)` or `^e(n|th)` would do for the whitelist approach.
However [PredictableNetworkInterfaceNames](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/) says
> You create your own manual naming scheme, for example by naming your interfaces "internet0", "dmz0" or "lan0"
So I think the blacklist approach may be better.

On the other hand, you can change the network interface name for lxd from the default `lxdbr0` to some other value. So the ideal way is detecting the type of network interfaces, not checking the names. But I had not found the way to do that, so I took the blacklist approach this time.

sni:
My Vagrantfile contains nothing special, except i create a bridged interface like:
```
config.vm.network "public_network", type: "dhcp", :bridge => "eth0"
```

iv597:
Any word on this? There's been at least a few bug reports about this behavior and the PR has been outstanding for now literally months. I find it a little hard to believe this has sat for so long - private IP addresses on Xenial guests (my broken use case) can't be *that* rare, can it?

If there's anything I can do to test or build this (though I'm more a JS/Python guy than Ruby), please let me know. #7155 is an increasingly problematic bug for me and my team as we look to upgrade our infrastructure from Trusty to Xenial with local testing.

awesomebing1:
Thank you so much for this patch, it took forever to find but finally ended my public network suffering!

domnulnopcea:
is this going to be merged?

lobsterdore:
This is a showstopper for a lot of people I think, the bug addressed by this fix is stopping me from adopting Xenial at all.

chasebolt:
I would like to see this merged as well

endeavour:
Me too

PR #7254, opened 2016-04-23
Adds Guest detection and features for Amazon Linux

petems:
* Amazon Linux is like RHEL 6 mostly, with a few small differences
* So lets import the RedHat code and hardcode the flavor to 6

I couldn't find any examples of spec or unit tests for the guest detection and plugin code, if there's an example I wouldn't mind writing tests too 👍

### DIFF
--- a/plugins/guests/amazon/cap/flavor.rb
+++ b/plugins/guests/amazon/cap/flavor.rb
@@ -1,0 +1,14 @@
+module VagrantPlugins
+  module GuestRedHat
+    module Cap
+      class Flavor
+        def self.flavor(machine)
+          # Amazon AMI is a frankenstien RHEL, mainly based on 6
+          # Maybe in the future if they incoporate RHEL 7 elements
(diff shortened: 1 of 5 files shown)


petems:
This was probably missed before because the old Amazon Linux had `/etc/redhat-release` so it assumed it was CentOS. The new one does not:

```
Last login: Sat Apr 23 21:08:11 2016 from 10.0.2.2

       __|  __|_  )
       _|  (     /   Amazon Linux AMI
      ___|\___|___|

https://aws.amazon.com/amazon-linux-ami/2014.03-release-notes/

[root@localhost vagrant]# cat /etc/redhat-release
CentOS
```

sethvargo:
Hi @petems 

I think this is more likely to belong in the vagrant-aws plugin. What do you think?

petems:
Hmm, I mean all of the guest detection stuff is in core?

Technically this affects local vagrant if you make a Amazon Linux Vagrant box too...

PR #7283, opened 2016-05-03
Adds instructions concerning Bundler versions

hedinfaok:
Because Vagrant requires certain versions of Bundler in order to work properly, the correct version of Bundler must be installed manually.  I gathered this information from a couple of sources.

It's handy to have this information in this specific guide :)


sethvargo:
Hi @hedinfaok 

This is already described in the WIKI: https://github.com/mitchellh/vagrant/wiki/Installing-Vagrant-from-Source. I think we should just link there instead of having two sets of instructions.

hedinfaok:
@sethvargo Oh, cool. I think you're right. I just edited the wiki to include the bundler version and updated this PR to point to the Wiki page.

sethvargo:
Great thanks!

Issue #7202, opened 2016-04-05
ssh.forward_agent = true not evaluated when vm.guest = windows


james-powis:
### Vagrant version
Observed on 1.8.1 and 1.7.4, however ssh.forward_agent works properly on 1.6.5

### Host operating system
OSX El Capitan

### Guest operating system
Windows 2012R2

### Vagrantfile
```ruby
Vagrant.configure("2") do |c|
  c.ssh.forward_agent = true
  c.ssh.insert_key = false
  c.vm.define 'win-2012R2-std' do |v|
    v.vm.hostname = 'win-2012R2-std'
    v.vm.box = 'opentable/win-2012r2-standard-amd64-nocm'
    v.vm.box_url = 'opentable/win-2012r2-standard-amd64-nocm'
    v.vm.box_check_update = 'true'
    v.vm.network :private_network, ip: "10.255.33.129", :netmask => "255.255.0.0", :mac => "080027B45925"
    v.vm.network :forwarded_port, guest: 3389, host: 3389, id: 'rdp', auto_correct: true
    v.vm.network :forwarded_port, guest: 5985, host: 5985, id: 'winrm', auto_correct: true
    v.vm.guest = :windows
    v.vm.provider :virtualbox do |vb|
      vb.customize ['modifyvm', :id, '--memory', '1024', '--cpus', '1']
    end
  end
end
```

### Debug output
```
vagrant up on 1.8.1 and 1.7.4 are the same:
18:04:48:~/git/personal/vw/devops/spec/acceptance/.vagrant/beaker_vagrant_files/win2012r2.yml
james@MacBook-Pro (add_acceptance *%)> vagrant up 
Bringing machine 'win-2012R2-std' up with 'virtualbox' provider...
==> win-2012R2-std: Importing base box 'opentable/win-2012r2-standard-amd64-nocm'...
==> win-2012R2-std: Matching MAC address for NAT networking...
==> win-2012R2-std: Checking if box 'opentable/win-2012r2-standard-amd64-nocm' is up to date...
==> win-2012R2-std: Setting the name of the VM: win2012r2yml_win-2012R2-std_1459814762312_83182
==> win-2012R2-std: Clearing any previously set network interfaces...
==> win-2012R2-std: Preparing network interfaces based on configuration...
    win-2012R2-std: Adapter 1: nat
    win-2012R2-std: Adapter 2: hostonly
==> win-2012R2-std: Forwarding ports...
[...]

PR #7287, opened 2016-05-04
[#7202] Always forward SSH on Windows

KierranM:
As per the details in #7202 this has been fixed before by @mitchellh in 8655d21 but was regressed in 3e8fdc0.

This re-implements the changes, and adds a test that explicitly defines this behaviour.

All the unit tests still pass, and I have tested creating a Windows vagrant box with this change


james-powis:
@KierranM Thanks for hopping on this, I was going to tackle it but found little time to do so... 👍

Issue #7289, opened 2016-05-05
Incorrect `/etc/exports` generation.


phyber:
Vagrant configured as follows:
* Adapter 1: Standard Vagrant NAT with an `ssh` port forward.
* Adapter 2: VirtualBox IntNet for VMs to communicate with each other over, static IP.
* Adapter 3: Host-only adapter set to `type: 'dhcp'`.

Vagrant will only put the address of Adapter 2 in the `/etc/exports` file, but requests to the host obviously go out via Adapter 3, so the mount request is denied.

This issue disappears if Adapter 3 is configured with a static IP address. In this case Vagrant populates `/etc/exports` with the addresses of both Adapter 2 and Adapter 3.
So it would appear that Vagrant is doing something strange when asked to use DHCP for the host-only adapter.

Vagrant version: 1.8.1
VirtualBox version: 5.0.20r106931

### References
- GH-5424

phyber:
PR #7290 has been prepared alongside this issue.

Note that the PR currently breaks the `allows statically configured guest IPs to work for NFS, even when guest property would fail`.

I am unsure how that test should be addressed.

phyber:
The spec test in PR #7290 now correctly passes and an additional spec test was added to check for the new behaviour.

PR #7290, opened 2016-05-05
prepare_nfs_settings: Try harder to get all machine IPs.

phyber:
Vagrant was not behaving correctly in configurations where there was
a static IP on a VirtualBox `intnet` interface and a DHCP `:hostonly`
interface configured.

Since `machine_ip` attempted to get static addresses `||` dynamic
addresses, it would simply use the static machine address and
continue.

This commit corrects this behaviour by collecting all static and
dynamic addresses into the `machine_ip` array instead of just one or
the other.

The result of this is a correctly generated `/etc/exports` on the
host machine, allowing NFS mounts to work correctly in this type of
environment.

### References
- GH-7289

### DIFF
--- a/plugins/providers/virtualbox/action/prepare_nfs_settings.rb
+++ b/plugins/providers/virtualbox/action/prepare_nfs_settings.rb
@@ -36,8 +36,19 @@ module VagrantPlugins
         #
         # The ! indicates that this method modifies its argument.
         def add_ips_to_env!(env)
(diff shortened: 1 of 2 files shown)


phyber:
Note that this PR currently breaks the `allows statically configured guest IPs to work for NFS, even when guest property would fail`.

I am unsure how that test should be addressed.

Issue #6904, opened 2016-01-21
SMB synced folder support for linux


varac:
https://github.com/mitchellh/vagrant/issues/1749 was about SMB support in general, and the windows part has been integrated. It should not be too hard to implement it on linux as well.

chrisroberts:
Hi there,

We would really like this, but this issue has been open for over a year with no one working on it, Leaving it open is unfortunately making our issue count look higher than it is. I’m going to close this and if someone wants to work on it I’d still be open to a PR!

Cheers!

PR #7291, opened 2016-05-05
Expanding SMB sharing to be compatible with more than just Windows hosts (#6904)

mdt-software:
This only supports the generic Linux host and the pre-existing Windows host, but puts the infrastructure in place so that others with relevant hosts can easily expand on what's been done.

### DIFF
--- a/plugins/hosts/linux/cap/smb.rb
+++ b/plugins/hosts/linux/cap/smb.rb
@@ -1,0 +1,31 @@
+require "pathname"
+require "log4r"
+
+module VagrantPlugins
+  module HostLinux
+    module Cap
+      class SMB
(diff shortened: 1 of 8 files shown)


plugins/hosts/linux/cap/smb.rb
```diff
@@ -0,0 +1,31 @@
+require "pathname"
+require "log4r"
+
+module VagrantPlugins
+  module HostLinux
+    module Cap
+      class SMB
+        @@logger = Log4r::Logger.new("vagrant::linux_host::capabilities::smb")
```
sethvargo: Please remove the logger from the capability.

chrisroberts:
SMB implementation has been redone in Vagrant 2.0 so I'm going to close this up as it is no longer applicable to the current code base. Thank you for this PR!

PR #7293, opened 2016-05-08
Set max_network_adapters to 36 for Virtualbox

ericpulvino:
Updated ./plugins/providers/virtualbox/driver/base.rb file, set max_network_adapters to 36 for Virtualbox so all 36 adapters are cleared prior to programming interfaces. This fix is per issue 7286 --> https://github.com/mitchellh/vagrant/issues/7286

## Limited Testing
Performed limited testing on the provided vagrantfile in issue 7286. To compare performance timing before and after the change. No Difference.
Performed Vagrant up on leaf1 5 times in a row and averaged the completion times using the following commands.

## Test Commands
time vagrant up leaf1
vagrant destroy -f leaf1
time vagrant up leaf1
vagrant destroy -f leaf1
time vagrant up leaf1
vagrant destroy -f leaf1
time vagrant up leaf1
vagrant destroy -f leaf1
time vagrant up leaf1
vagrant destroy -f leaf

## Set to 8 Max Adapters
Vagrant up 1: 39.073 sec
Vagrant up 2: 37.678 sec
Vagrant up 3: 37.069 sec
Vagrant up 4: 37.629 sec
Vagrant up 5: 37.737 sec
Average: 37.837 sec

## Set to 36 Max Adapters
Vagrant up 1: 37.613 sec
Vagrant up 2: 37.041 sec
Vagrant up 3: 37.095 sec
Vagrant up 4: 37.056 sec
Vagrant up 5: 37.483 sec
Average: 37.257 sec

### DIFF
--- a/plugins/providers/virtualbox/driver/base.rb
+++ b/plugins/providers/virtualbox/driver/base.rb
@@ -178,7 +178,7 @@ module VagrantPlugins
 
         # Returns the maximum number of network adapters.
         def max_network_adapters
-          8
+          36
         end
 
         # Returns a list of forwarded ports for a VM.
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -1364,7 +1364,7 @@ en:
         There is no available slots on the VirtualBox VM for the configured
         high-level network interfaces. "private_network" and "public_network"
         network configurations consume a single network adapter slot on the
-        VirtualBox VM. VirtualBox limits the number of slots to 8, and it
+        VirtualBox VM. VirtualBox limits the number of slots to 36, and it
         appears that every slot is in use. Please lower the number of used
         network adapters.
       virtualbox_not_detected: |-


sethvargo:
Thanks!

sethvargo:
FYI - we had to revert this. Vbox on windows only has 8 network interfaces.

ericpulvino:
Thanks for the heads-up, I think this might depend on the chipset in use
specified by the virtualbox VM...

Per Virtualbox Documentation:
<https://www.virtualbox.org/manual/ch03.html#settings-motherboard>
Chipset

Here you can select which chipset will be presented to the virtual machine.
Before VirtualBox 4.0, PIIX3 was the only available option here. For modern
guest operating systems such as Mac OS X, that old chipset is no longer
well supported. As a result, VirtualBox 4.0 introduced an emulation of the
more modern ICH9 chipset, which supports PCI express, three PCI buses,
PCI-to-PCI bridges and Message Signaled Interrupts (MSI). This allows
modern operating systems to address more PCI devices and no longer requires
IRQ sharing. Using the ICH9 chipset it is also possible to configure up to
36 network cards (up to 8 network adapters with PIIX3). Note that the ICH9
support is experimental and not recommended for guest operating systems
which do not require it.

Perhaps chipset detection should be added? What was the error seen?
-Eric

sethvargo:
@ericpulvino #7417

ericpulvino:
Looks like back to the drawing board on this fix... I see two methods to fix this:

1. Host OS Detection -- take the right action for the host on which you are running.
Downside can get muddy over time as things change. This would have to be mirrored in the error output.

2. Soft Error -- Don't hard-stop/fail on an error here try to clear the adapter... if it's there great, if not, no harm. Upside here is that it doesn't introduce messy Host OS detection and should work everywhere.

Which would you prefer I investigate?

PR #7298, opened 2016-05-08
Change: disable the linux switch (Docker for Mac/Windows, etc.)

till:
The assumption that Docker is only native on Linux is no longer true with Docker for Mac/Windows and other efforts. :)


sethvargo:
Hi @till 

I'm going to merge this, but then I'm going to change the logic a bit. Since the Windows/Mac stuff is still in beta, people may want the "it just works" experience. Additionally, some users just want to install Vagrant and be done (they don't want to install _other_ tools), so I'm going to make this opt-in behavior.

till:
@sethvargo sounds good, is there another commit/PR to look for as in how to use this going forward?

sethvargo:
8c11b53

PR #7299, opened 2016-05-10
Added 'vagrant reload' info on collision_error

tiemevanveen:
Although the current message does tell you that the forwarded port can't be auto corrected due to the current state of the machine, it does not provide a clear solution. This happens to me all the time and after changing my vagrant file I've found out that I could just do a `vagrant reload`! 

Could it be nice to add this do the docs?


sethvargo:
Thanks!

PR #7308, opened 2016-05-11
Adding example of using the preserve_order flag

davidsiefert:
I couldn't really understand how to use this feature.... I might be dumb, didn't rtfm, I dunno.  Here is a PR anyway to help clarify by example.


sethvargo:
Thanks!

PR #7318, opened 2016-05-16
fix setting hostname for Debian guests

evgeni:
Without that setting `config.vm.hostname` for `debian/jessie64` and similar boxes will fail:

```

hostname -f

Stdout from the command:



Stderr from the command:

sudo: unable to resolve host jessie.raw
stdin: is not a tty
hostname: Name or service not known
```

While this should also be fixed in the box itself, the Debian guests plugin should be more tolerant to badly set-up machines.

While at it:
* don't use `sudo` for just calling `hostname`
* don't consider `localhost` a valid hostname

### DIFF
--- a/plugins/guests/debian/cap/change_host_name.rb
+++ b/plugins/guests/debian/cap/change_host_name.rb
@@ -33,7 +33,10 @@ module VagrantPlugins
 
         def get_current_hostname
           hostname = ""
-          sudo "hostname -f" do |type, data|
+          execute "hostname -f", error_check: false do |type, data|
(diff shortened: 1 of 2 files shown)


sethvargo:
Hi @evgeni 

Thank you for the PR - is there an associated issue as well? There doesn't have to be, but we like to cross-link so they close when we merge.

evgeni:
@sethvargo no, I did not open an issue for that, went straight for the patch :)

evgeni:
@sethvargo ping? :)

sethvargo:
Hi @evgeni 

I am working to get through the Vagrant issues as quickly as possible. Thanks

PR #7332, opened 2016-05-20
Much faster rsync auto

wk8:
This patch aims to make `vagrant rsync-auto` much faster than it used to.

Profiling `rsync-auto`, it was clear that most of the time was spent calling
guest capabilities.

This patch tries to alleviate that issue by refactoring
`VagrantPlugins::SyncedFolderRSync::RsyncHelper` into an instantiable class
(as opposed to only static methods before), so that it can cache a lot of the
results of the calls to the guest capabilities.

We're using that new API in `rsync-auto`, resulting in a dramatic improvement:
on my dev VMs (Mac OS X host, VirtualBox, Ubuntu 14.04 guests), `rsync-auto` now
takes about 300ms per iteration (as opposed to 5-6 secs before); and those 300ms
correspond to almost no overhead compared to running the `rsync` command manually.

The old API with static methods (i.e.
`VagrantPlugins::SyncedFolderRSync::RsyncHelper.single_rsync`) is still there for
backward compatibility reasons, for plugins that might use it (gatling-rsync-auto
is one of these); but a deprecation warning was added to it.

I'm aware this patch will need tests to be merged, but before I spend time writing
those, I'd like to know if this approach is acceptable to the Vagrant core team.

### DIFF
--- a/plugins/synced_folders/rsync/command/rsync.rb
+++ b/plugins/synced_folders/rsync/command/rsync.rb
@@ -59,7 +59,7 @@ module VagrantPlugins
 
             # Sync them!
             folders.each do |id, folder_opts|
-              RsyncHelper.rsync_single(machine, ssh_info, folder_opts)
+              RsyncHelper.new(machine, ssh_info, folder_opts).rsync_single
             end
           end
 
--- a/plugins/synced_folders/rsync/command/rsync_auto.rb
+++ b/plugins/synced_folders/rsync/command/rsync_auto.rb
@@ -27,6 +27,7 @@ module VagrantPlugins
 
         def execute
           @logger = Log4r::Logger.new("vagrant::commands::rsync-auto")
+          @rsync_helpers = {}
 
           options = {}
           opts = OptionParser.new do |o|
@@ -76,7 +77,7 @@ module VagrantPlugins
             if ssh_info
               machine.ui.info(I18n.t("vagrant.rsync_auto_initial"))
               folders.each do |id, folder_opts|
-                RsyncHelper.rsync_single(machine, ssh_info, folder_opts)
+                rsync_helper(machine, id, folder_opts).rsync_single
               end
             end
 
@@ -186,7 +187,7 @@ module VagrantPlugins
               ssh_info = opts[:machine].ssh_info
               begin
                 start = Time.now
-                RsyncHelper.rsync_single(opts[:machine], ssh_info, opts[:opts])
+                rsync_helper(opts[:machine], opts[:id], opts[:opts]).rsync_single
                 finish = Time.now
                 @logger.info("Time spent in rsync: #{finish-start} (in seconds)")
               rescue Vagrant::Errors::MachineGuestNotReady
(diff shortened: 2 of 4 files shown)


wk8:
@sethvargo : as mentioned in the commit message, I'm aware tests are missing, but I'd like to know if that approach is acceptable before spending time writing tests. Tested manually though, and works great

Addresses https://github.com/mitchellh/vagrant/pull/7332

sethvargo:
Hi @wk8 

Thank you for opening a pull request and issue. I'd like to better understand your benchmarks - could you please share? Is the issue that we are not caching guest capabilities?

wk8:
@sethvargo : can't share that as is, private repo, you know how it goes :) gonna try to set up a public repo that reproduces the issue.

And I do believe indeed that the issue comes from not caching guest capabilities; seems each of these take about 2 secs on my machine.

wk8:
@sethvargo : upon further investigation, seems that it comes from the fact that the Vagrantfile does all kind of stuff, including reading files from disk and the like, and that each call on guest capabilities re-parses said Vagrantfile?

sethvargo:
@wk8 that is correct. The Vagrantfile is parsed and loaded multiple times: https://www.vagrantup.com/docs/vagrantfile/

wk8:
@sethvargo : thanks, that's indeed the issue here. Any reason to not cache it within a single Vagrant run? This is kinda what this patch does, but I suspect it could be done deeper within Vagrant.

In any case, I'd say that it's not totally unreasonable to have a decent amount of logic in a Vagrantfile (what's your take on that?) and that in that case, rsync-auto being slow is an issue that this patch addresses. Is the general idea acceptable to you or not at all?

sethvargo:
Hi @wk8 

I think the reason we cannot cache this is because some of the capabilities actually perform commands on the system. I think we might need a higher-level function that marks a capability as cacheable.

wk8:
Hi @sethvargo ,

What takes a long time is not actually running the capabilities, it's parsing the Vagrantfile to know what action a given capability actually maps to. That part sounds cacheable to me?

Thanks

sethvargo:
Hey @wk8 

Sorry - I'm not sure I know what you mean. Could you explain a bit more?

wk8:
@sethvargo : well, in the concrete example of rsync and say the `:rsync_post` capability, which by default on linux maps to running a `chmod` on the rsync files, what takes most time is parsing the Vagrantfile after each rsync to know what action the `:rsync_post` capability maps to (the `chmod`), much more so than running the actual capability (running `chmod` over SSH); seems to me we could cache that running the `:rsync_post` capability is that `chmod` instead of having to "resolve" what that capability means every time.

Don't know if it makes more sense that way?

danez:
Thanks for looking into that @wk8, I encountered the same slowness.

Hope that this or any other solution is going to be integrated which makes it faster.

wk8:
@danez : I made this patch into a Vagrant plugin at https://github.com/wk8/vagrant-instant-rsync-auto (note that you need rsync 3.1+ for it to work, see https://github.com/wk8/vagrant-instant-rsync-auto#troubleshooting); please let me know if that works for you, or else please feel free to open issues against it.

@sethvargo : does my last comment above make more sense?

sethvargo:
Hi @wk8 

Sorry, it doesn't make sense to me. The Vagrantfile is only parsed at the invocation of the Vagrant command to the best of my knowledge. It should be parsed once you run `rsync-auto` and all of that is loaded in memory. Do you have any evidence that supports it's being loaded multiple times?

My biggest fear is that you are caching a required operation here. It's very difficult to reason about this pull request due to the number of unrelated changes to quoting syntax and restructuring that are unrelated to the root of this.

Vincent--:
Thank you for that, @wk8. I'll give a try to the plugin but just by curiosity, why do you next rsync 3.1+? Is it a different requirement than from the existing "vagrant rsync-auto"?

wk8:
@Vincent-- it uses both `--usermap` and `groupmap` rsync options, which were added in rsync 3.1

Vincent--:
If this is of any help, please find the debug output with the existing rsync-auto (24.68398 sec) and with the instant-rsync-auto plugin (1.871291 sec)

Test case: 
* around 50 000 files in the folder
* only one file was updated to test the sync

[instant-rsync-auto.txt](https://github.com/mitchellh/vagrant/files/647409/instant-rsync-auto.txt)
[rsync-auto.txt](https://github.com/mitchellh/vagrant/files/647410/rsync-auto.txt)

Part of the gain might be related to the chown command (run by rsync-auto afterwards but included directly in the rsync command for the instant-rsync-auto plugin).

apolopena:
What will it take to move this forward?

chrisroberts:
Resolved with #10529 

Cheers!

PR #7334, opened 2016-05-21
Update up.html.md

annag42:
s/created and interacting with/created and interacted with/
That is, I changed the word whose root is "interact" from "interacting" to "interacted", 
to better match tense with "created", for grammatical correctness (parallel construction).


ajlanghorn:
👍

PR #7347, opened 2016-05-26
Fixing second paragraph indentation and link.

dmerrick:
Without the two spaces on line 62, the docs look like this:

screenshot

This is confusing, because it looks like the second paragraph is wrapping up the "ALL CHEF PROVISIONERS" section. Instead, we should indent the second paragraph so it is clear it is talking about the `version` attribute.

I also fixed the broken `install` link.


dmerrick:
Cheers!

PR #7354, opened 2016-05-27
Fix file leaking in bundler

### LINKED ISSUE #3469: cleanup /tmp
After some time using Vagrant on one of my boxes (I don't reboot this box) my /tmp directory looks really terrible. There are a lot of files and directories prefixed with vagrant. I think Vagrant should remove the files and directories by itself.

```
berendt@zitrone /tmp % ls -l | grep vagrant | wc -l
5494
```

### LINKED ISSUE #6231: Vagrant not deleting temp files

sethvargo:
This fixes a leak that occurs when interacting with Vagrant's bundler instance. Each call to Tempfile creates and generates a tempfile, but they are not cleaned up unless GC runs. Additionally, we were not deleting a few of the files.

This adds a new helper that creates a tmp path using `Dir::Tmpname`, and then returns a new open file handle for that path.

- Fixes #6301 
- Fixes #3469
- Fixes #6231

### DIFF
--- a/lib/vagrant/bundler.rb
+++ b/lib/vagrant/bundler.rb
@@ -2,6 +2,7 @@ require "monitor"
 require "pathname"
 require "set"
 require "tempfile"
+require "fileutils"
 
 require "bundler"
 
@@ -61,7 +62,7 @@ module Vagrant
       end
 
       # Setup the Bundler configuration
-      @configfile = File.open(Tempfile.new("vagrant").path + "1", "w+")
+      @configfile = tempfile("vagrant-configfile")
       @configfile.close
 
       # Build up the Gemfile for our Bundler context. We make sure to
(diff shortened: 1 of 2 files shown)


remram44:
Everything happens 😄 Thanks!

PR #7359, opened 2016-05-29
Cache the results of common operations

sethvargo:
This commit changes Vagrant::Util::Platform to cache the result of some common operations. These values are highly unlikely to change over the course of a single Vagrant run and they are only cached for that run.


sethvargo:
https://github.com/mitchellh/vagrant/pull/7359/files?w=1

phinze:
LGTM

Issue #4275, opened 2014-07-31
allow overriding keys_only and IdentitiesOnly


WyseNynja:
I started trying to make a PR for this, but got stuck.  I don't know how to add new options to the Vagrantfile.

Here is my situation:

I have an SSH key on my local machine that allows me to connect to a vagrant

I can `vagrant up` on my local machine and everything works as expected.

I want to ssh to a remote machine (that does not have direct access to my private keys, but DOES have access to my ssh-agent) and run vagrant up there.  This does not work because 'IdentitiesOnly' and 'keys_only' are both forced to yes/true; it hangs on "Waiting for SSH"

I want to have an option in my Vagrantfile that lets me use my ssh-agent instead of a password/custom ssh key/insecure ssh key.  This seems like a simple enough request.

I hard coded the changes into my copy of vagrant and this worked.  I just don't know how to parameterize it.

```
diff --git a/plugins/communicators/ssh/communicator.rb b/plugins/communicators/ssh/communicator.rb
index 2438d74..d850909 100644
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -293,7 +293,7 @@ module VagrantPlugins
           config:                false,
           forward_agent:         ssh_info[:forward_agent],
           keys:                  ssh_info[:private_key_path],
-          keys_only:             true,
+          keys_only:             false,
           paranoid:              false,
           password:              ssh_info[:password],
           port:                  ssh_info[:port],
diff --git a/templates/commands/ssh_config/config.erb b/templates/commands/ssh_config/config.erb
index 2cdb2db..d31cee9 100644
--- a/templates/commands/ssh_config/config.erb
+++ b/templates/commands/ssh_config/config.erb
@@ -12,7 +12,7 @@ Host <%= host_key %>
   IdentityFile <%= path %>
 <% end -%>
 <% end -%>
-  IdentitiesOnly yes
+  IdentitiesOnly no
   LogLevel FATAL
 <% if forward_agent -%>
   ForwardAgent yes
diff --git a/test/unit/plugins/commands/ssh_config/command_test.rb b/test/unit/plugins/commands/ssh_config/command_test.
index e458776..6d96aab 100644
--- a/test/unit/plugins/commands/ssh_config/command_test.rb
+++ b/test/unit/plugins/commands/ssh_config/command_test.rb
@@ -51,7 +51,7 @@ Host #{machine.name}
   UserKnownHostsFile /dev/null
[...]

PR #7360, opened 2016-05-29
Allow customization of keys_only & paranoid SSH

sethvargo:
This adds two new SSH configuration options:

- `keys_only`
- `paranoid`

These values were previously hard-coded, but can now be user-specified.

Fixes GH-4275


zmt:
So, I have been trying to make this work, but have found that the keys_only override from a Vagrantfile is getting dropped soemwhere between the ssh_config and the machine objects and is still always true in the ssh communicator call to Net::SSH.start. So, this doesn't quite close GH-4275. I'm prying my way into the problem and will submit a PR if/when I solve it.

PR #7374, opened 2016-05-31
Improving resilience of some VirtualBox commands.

nicolasvan:
These changes are an attempt at reducing the number of race conditions we seem to be getting when running multiple vagrant environments using virtualbox on the same host machine.

The `read_used_ports` and `delete_unused_host_only_networks` functions use VBoxManage to retrieve a list of VMs and call `vboxmanage showvminfo` on each of them.

A call to `vboxmanage showvminfo` will fail if the VM has just been deleted by a different vagrant process.

Shall we just swallow any VBoxManage error that contain `VBOX_E_OBJECT_NOT_FOUND` (as this is likely to mean that the VM is gone and we shouldn't have to worry about it any more)?

### DIFF
--- a/plugins/providers/virtualbox/driver/version_4_0.rb
+++ b/plugins/providers/virtualbox/driver/version_4_0.rb
@@ -422,8 +422,15 @@ module VagrantPlugins
               # Ignore our own used ports
               next if uuid == @uuid
 
-              read_forwarded_ports(uuid, true).each do |_, _, hostport, _|
-                ports << hostport
+              begin
+                read_forwarded_ports(uuid, true).each do |_, _, hostport, _|
+                  ports << hostport
+                end
+              rescue Vagrant::Errors::VBoxManageError => e
+                raise if !e.extra_data[:stderr].include?("VBOX_E_OBJECT_NOT_FOUND")
+
+                # VirtualBox could not find the vm. It may have been deleted
+                # by another process after we called 'vboxmanage list vms'? Ignore this error.
               end
             end
           end
--- a/plugins/providers/virtualbox/driver/version_4_0.rb
+++ b/plugins/providers/virtualbox/driver/version_4_0.rb
@@ -85,11 +85,18 @@ module VagrantPlugins
 
           execute("list", "vms").split("\n").each do |line|
             if vm_name = line[/^".+?"\s+\{(.+?)\}$/, 1]
-              info = execute("showvminfo", vm_name, "--machinereadable", retryable: true)
-              info.split("\n").each do |line|
-                if network_name = line[/^hostonlyadapter\d+="(.+?)"$/, 1]
-                  networks.delete(network_name)
+              begin
+                info = execute("showvminfo", vm_name, "--machinereadable", retryable: true)
+                info.split("\n").each do |line|
+                  if network_name = line[/^hostonlyadapter\d+="(.+?)"$/, 1]
+                    networks.delete(network_name)
+                  end
                 end
+              rescue Vagrant::Errors::VBoxManageError => e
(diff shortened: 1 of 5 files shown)


sethvargo:
Hi @nicolasvan 

Thank you for opening a Pull Request. I would like to learn more about your situation and the way you are reproducing this error. I just tried to bring up multiple VMs in parallel with VirtualBox and was unable to reproduce this issue.

If there truly is a race condition here:

1. It's either fixed in master (which is what I run)
2. It's fixed in upstream vbox

In any event, I don't think retrying is the proper approach here. Vagrant has a global mutex locking system which is probably a better approach here.

nicolasvan:
Hi @sethvargo

The way to reproduce is a bit tricky. I'm on windows and I can sometime get that error by running multiple powershell sessions in parallel (up to 5) running this kind of thing:
```
# session 1
$LASTEXITCODE = 0; while($LASTEXITCODE -eq 0) {vagrant destroy vm1 -f; vagrant up vm1 }
```
```
# session 2
$LASTEXITCODE = 0; while($LASTEXITCODE -eq 0) {vagrant destroy vm2 -f; vagrant up vm2 }
```
and so on...
And using this type of Vagrantfile:
```
	[1,2,3,4,5].each do |i|
		config.vm.define "vm#{i}" do |node|
			node.vm.box = 'red-gate/windows-2012r2-ssms2014' # That vm is not publicly available sorry!
		end
	end
end
```

Having 5 vagrant processes destroying/creating virtualbox vms continuously at the same time might be asking for trouble? :sweat_smile: 

If we're unlucky, `vagrant destroy vm1` will delete vm1 while `vagrant up vm2` is getting virtualbox to query for all the ports used by all the vms including vm1 which will cause `vboxmanage showvminfo <guid-of-vm1>` to be called after vm1 is destroyed.

Note that my change **doesn't cause anything to be retried**. We just ignore the `VBOX_E_OBJECT_NOT_FOUND` error returned by virtualbox and keep on querying for the other vms.

===
To be honest, it does look like vboxmanage is not great at dealing with parallel calls. I can fairly easily get errors by calling this twice at the same time:
```
vboxmanage showvminfo <guid-of-vm>
```
One of these call will almost always give me some kind of error looking like this: 
```
VBoxManage.exe: error: The object is not ready
VBoxManage.exe: error: Details: code E_ACCESSDENIED (0x80070005), component SessionMachine, interface IMachine, callee IUnknown
VBoxManage.exe: error: Context: "COMGETTER(BandwidthControl)(bwCtrl.asOutParam())" at line 2243 of file VBoxManageInfo.cpp
```

So maybe locking all the calls to vboxmanage could be the way to go?
Although it looks like the current approach with vboxmanage is to retry? (As per https://github.com/mitchellh/vagrant/blob/master/plugins/providers/virtualbox/driver/version_5_0.rb#L29)
Personally I find debugging locking issues pretty scary so I would go for retrying things when possible. Maybe all calls to vboxmanage could use `retryable: true`? :fearful:

===
Some background info as to why we're doing this:
We are currently trying to use vagrant to quickly start testing environments for our developers and CI builds (we use Teamcity). We are seeing some issues
on our CI side of things, mostly because we are currently having up to 3 Teamcity agents running on a single (powerful) machine.
Locking our CI builds that use vagrant to only 1 Teamcity agent per machine does improve reliability a lot. But it would be nice if we didn't have to :smile:

sethvargo:
I'm going to defer to @mitchellh on this one, but it's unlikely we are going to merge this for 1.8.2, just due to the significance of the change. Sorry!

nicolasvan:
no worries :smile:

chrisroberts:
Rebased in #8464

PR #7376, opened 2016-05-31
Don't restore Gem::Specification.all if it's not being replaced

karlkfi:
Revert 3f9fb2ef03a78bc9177a52d4629a5e80e6e42b2e from GH-2769

- For compatibility with ba77d4b533e449fa610531a7832b8fd1837eb9db

Fixes GH-7073


sethvargo:
Thanks!

PR #7378, opened 2016-06-01
VirtualBox provider: Extension to allow disk resizing on VM import

cawka:
The extension consists of two parts:
- allow VM import using `--options importtovdi` option
- enable substitution of :diskN (e.g., :disk0) during VM customization
  with UUID of the Nth disk associated with the VM

Example:

```
  v.use_vdi = true
  v.customize 'pre-boot', ['modifyhd', :disk0, '--resize', '30000']
end
```

### DIFF
--- a/plugins/providers/virtualbox/action/customize.rb
+++ b/plugins/providers/virtualbox/action/customize.rb
@@ -22,6 +22,21 @@ module VagrantPlugins
             customizations.each do |command|
               processed_command = command.collect do |arg|
                 arg = env[:machine].id if arg == :id
+
+                if arg =~ /^disk(\d+)$/
+                  disk_id = $1.to_i
+                  disks = env[:machine].provider.driver.read_disks
+
+                  if not disks or disk_id >= disks.size
+                    raise Vagrant::Errors::VMCustomizationFailed, {
+                      command: "Discovering UUID for the specified disk",
+                      error:   "Disk with index #{disk_id} does not exist"
+                    }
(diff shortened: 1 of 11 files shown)


sethvargo:
Hi @cawka 

Thank you for opening a Pull Request. Could you please explain the motivation for this change here? We are changing some pretty core functionality and method airty here, so I don't think we will be able to merge this until Vagrant 2.0.

cawka:
The main motivation is that some boxes, specifically `ubuntu/xenail32` (besides other problems) comes with 10gb disk, which may not be useful for people, including me.  With the update, it is relatively easy to fix the size of the VMs disk without the need to do repackaging.

I'm ok if you want to postpone to next release.  If you haven't noticed, I've added implementation only for VirtualBox 5, as I don't have the older version to play with.

chrisroberts:
Defaulting the `use_vdi` value might be an easy way to introduce this without the introduction of breaking changes:

```ruby
def import(ovf, use_vdi=false)
```

Will need to have a closer look at this on older versions of virtualbox, raising an exception if it can't be supported.

chrisroberts:
After merging this and testing out the functionality, I'm going to close this due to it being vbox specific and only implemented on 5.x versions. I would like to see a generic api that the providers can then implement to provide consistency within the providers as well as the user interface. Thank you for the contribution and I will circle back to the implementation provided here when looking at how to make this a generic feature. Cheers!

PR #7380, opened 2016-06-01
Added box_download_proxy

davidgaleano:
You have to set your environmental variables `http_proxy` or `https_proxy` correctly when using vagrant behind a proxy, which makes your `Vagrantfile` not self-sufficient. Commands like `vagrant up` will check if the box is up-to-date, failing potentially in unfriendly ways if the proxy environment variables are not set correctly (in my case it would take 60s to time-out before continuing...). And managing your environment variables gets messy when you have to use different proxies for different domains.
This pull request only attempts to solve the proxy problems when downloading boxes, a new configuration option `box_download_proxy` would allow you to specify the proxy to use when downloading a box or when checking for updates.

### DIFF
--- a/lib/vagrant/action/builtin/box_add.rb
+++ b/lib/vagrant/action/builtin/box_add.rb
@@ -414,6 +414,7 @@ module Vagrant
           downloader_options[:headers] = ["Accept: application/json"] if opts[:json]
           downloader_options[:ui] = env[:ui] if opts[:ui]
           downloader_options[:location_trusted] = env[:box_download_location_trusted]
+          downloader_options[:proxy] = env[:box_download_proxy]
 
           Util::Downloader.new(url, temp_path, downloader_options)
         end
--- a/lib/vagrant/action/builtin/box_check_outdated.rb
+++ b/lib/vagrant/action/builtin/box_check_outdated.rb
@@ -45,7 +45,8 @@ module Vagrant
             client_cert: env[:client_cert] ||
                            machine.config.vm.box_download_client_cert,
             insecure: !env[:insecure].nil? ?
-                        env[:insecure] : machine.config.vm.box_download_insecure
+                        env[:insecure] : machine.config.vm.box_download_insecure,
+            proxy: machine.config.vm.box_download_proxy
           }
 
           env[:ui].output(I18n.t(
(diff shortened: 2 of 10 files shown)


davidgaleano:
First time with Ruby and Vagrant so it could be totally wrong code or wrong solution.

sethvargo:
HI @davidgaleano 

Thank you for the Pull Request. The `http(s)_proxy` environment variables are not unique to Vagrant; they are respected as the "standard" place to retrieve proxy information from. While I definitely understand the desire to have a top-level configuration option for this, I think you can do something like this in your Vagrantfile:

```
ENV["HTTP_PROXY"] = "my-proxy.corp"
```

to achieve the same result. Can you give that a test and report back please?

davidgaleano:
I know that those environment variables are not unique to Vagrant, but it is an insufficient solution when there are different proxies for different domains, which is why browsers and other clients support the WPAD protocol (which is a solution with its own problems, like for example requiring a Javascript interpreter...).

Would `ENV["HTTP_PROXY"]` apply to all the HTTP requests by Vagrant? for example, would it use that setting when checking the licence for the VMware plugin? if so, it would not solve my problem. In the best possible case I would need at least two different proxies: one for the external site (VMware plugin license check or similar), and another for the internal site (the one hosting the box). My change at least solves one of the proxies, leaving the other one for the environment variable.

In any case I did not know that I could do `ENV["HTTP_PROXY"]`, thanks :-)

sethvargo:
Hi @davidgaleano 

Proxies are generally a global, top-level configuration, which is why environment variables are almost always the go-to solution. 

The ENV will apply to all requests by Vagrant which use that Vagrantfile. Since the `plugin` commands do not evaluate the Vagrantfile, they will not be used in that case (same with `box list`, etc).

While we definitely appreciate the contribution, I worry that the added complexity is not worth the gain (trade-offs). This is, from my perspective, a very rare edge case that has never historically occurred in the history of Vagrant. Most users have a single proxy through which they must make requests.

davidgaleano:
The WPAD protocol is universally used in big companies exactly because a single global environment variable is hopelessly insufficient. I had to deal with complex proxy setups many times in the past.

The fact that ENV does not apply to all operations makes it actually a viable solution for the dual proxy case so I will accept that as a solution, thanks.

PR #7408, opened 2016-06-09
Fixed documentation for channel chef option value from stable to latest.

repejota:
While I was affected by https://github.com/chef/chef/issues/4948, I've noticed that documentation was saying the default value for `chef.channel` option was `stable` but this is not true.

In fact, by default installs `latest` so I changed a little the documentation as it was confusing.


sethvargo:
The docs will be correct with the next release. This was a mistake and we deployed the docs too quickly. Thanks!

PR #7424, opened 2016-06-11
Remove punctuation from URL

### LINKED ISSUE #7423: Punctuation on URL's confusing text editor
This is _really minor_ but it annoyed me enough to make a change...

Documentation URL's present in Vagrantfile generated with `vagrant init` have a period/fullstop at the end.

This confuses some text editors in GUI envionments that wrongly include this when the URL is activated sent to the browser leading to a 404

mcrmonkey:
- Removed period from end of URL so as not to confuse URL handlers in some text editors

Ultimately closes #7423

### DIFF
--- a/templates/commands/init/Vagrantfile.erb
+++ b/templates/commands/init/Vagrantfile.erb
@@ -8,10 +8,10 @@
 Vagrant.configure("2") do |config|
   # The most common configuration options are documented and commented below.
   # For a complete reference, please see the online documentation at
-  # https://docs.vagrantup.com.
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @mcrmonkey 

This is a complete sentence, so it should end in period. I tried in VIM, atom, and sublime, and this is not an issue. We are going to favor on the side of grammar here, but thank you for the PR!

mcrmonkey:
Yeah but...

... yeah, you're actually right here.
 It is a complete sentence and I should stop using half baked tools.

PR #7448, opened 2016-06-14
For for issue #7447

### LINKED ISSUE #7447: Cannot get status of VM in HyperV provider on Windows Server 2012R2 & Windows 8.1
### Vagrant version
1.8.3
Visual check of code has same issue in 1.8.4

### Host operating system
Windows 8.1 & Windows Server 2012R2

### Guest operating system
Windows Server 2012R2

### Vagrantfile
N/A provider issue affecting all Vagrantfiles

### Expected behaviour

**Only For HyperV provider**
Should have detected HyperV VM status

### Actual behavior

Failed due to use of dynamic testing of exception type which is not supported on the platform.

### Steps to reproduce
1. vagrant up for anything with hyperV provider on named platforms results in the error

```
At C:\HashiCorp\Vagrant\embedded\gems\gems\vagrant-1.8.4\plugins\providers\hyperv\scripts\get_vm_status.ps1:21 char:8
+ } catch $ExceptionType {
+        ~
The Catch block is missing its statement block.
At C:\HashiCorp\Vagrant\embedded\gems\gems\vagrant-1.8.4\plugins\providers\hyperv\scripts\get_vm_status.ps1:21 char:24
+ } catch $ExceptionType {
+                        ~
Unexpected token '{' in expression or statement.
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : MissingCatchHandlerBlock
```

interdev:
Change the exeception testing to look for the String name of the exception type. Dynamic use of the exception type does not work on Windows 8.1 and Windows Server 2012R2

Fixes #7447

### DIFF
--- a/plugins/providers/hyperv/scripts/get_vm_status.ps1
+++ b/plugins/providers/hyperv/scripts/get_vm_status.ps1
@@ -9,18 +9,24 @@ $Dir = Split-Path $script:MyInvocation.MyCommand.Path
 
 
 if($PSVersionTable.PSVersion.Major -le 4) {
-  $ExceptionType = [Microsoft.HyperV.PowerShell.VirtualizationOperationFailedException]
+  $ExceptionType = "Microsoft.HyperV.PowerShell.VirtualizationOperationFailedException"
 } else {
-  $ExceptionType = [Microsoft.HyperV.PowerShell.VirtualizationException]
+  $ExceptionType = "Microsoft.HyperV.PowerShell.VirtualizationException"
 }
 
 try {
     $VM = Get-VM -Id $VmId -ErrorAction "Stop"
     $State = $VM.state
     $Status = $VM.status
-} catch $ExceptionType {
-    $State = "not_created"
-    $Status = $State
+} catch [System.Exception] {
+    $type = [String]$_.Exception.GetType()
+    if ($type -eq $ExceptionType ) {
+      $State = "not_created"
+      $Status = $State
+    } else {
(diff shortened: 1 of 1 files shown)


sethvargo:
I defer to @mwrock since this is out of my wheelhouse 😄

codekaizen:
There are two problems with get_vm_status.ps1. One is the invalid Powershell which the PR here fixes. The other is that the check on PS version is incorrect, and instead should depend on the Hyper-V version to determine which exception type to throw.

mwrock:
I can confirm that this is definitely breaking and this pr fixes the issue in my windows 10 build 14366 prevew build. I also think @codekaizen is correct to be concerned about Hyper-V version vs Powershell version. I have tested with 2 2012 R2 VMs: 1 has WMF 5 installed and the other does not. Both do not have access to type `Microsoft.HyperV.PowerShell.VirtualizationException`. 

I think looking at the version of the Hyper-V module is likely the best test and `Microsoft.HyperV.PowerShell.VirtualizationException` should be used on versions 2.0.0 and greater.

There is also a syntax error in this PR that would occur if a more general exception occurs where it uses `Write-Hosts` and not `Write-Host`.

I think this is important to fix since it is totally breaking Hyper-V environments. I've been swamped but will try to submit a PR later unless someone beats me to it.

sethvargo:
Closing in favor of https://github.com/mitchellh/vagrant/pull/7487. Thanks!

PR #7456, opened 2016-06-15
Append newline to each key added

jonnywilliamson:
The change to this file [10 days ago](https://github.com/mitchellh/vagrant/commit/e2b7e28082a4b5e546655526773477e8df833bdd#diff-715ea9eefd185460afff2f7cf7eb0c3b) removed a newline character at the end of each ssh public key added to the `authorized_keys` file. This meant that when another key was added, it continued on the same line as the one before and thus wasn't being detected/able to be used when a ssh connection came in with a key file.

With regards to https://github.com/mitchellh/vagrant/issues/7455 this is an (ugly) fix.

I'm sure someone knows a better command to concat a file and a string and append it to the `authorized_keys` file. But this does fix the problem.

### DIFF
--- a/plugins/guests/linux/cap/insert_public_key.rb
+++ b/plugins/guests/linux/cap/insert_public_key.rb
@@ -19,6 +19,7 @@ module VagrantPlugins
             mkdir -p ~/.ssh
             chmod 0700 ~/.ssh
             cat '#{remote_path}' >> ~/.ssh/authorized_keys
+            echo "\n" >> ~/.ssh/authorized_keys
             chmod 0600 ~/.ssh/authorized_keys
 
             # Remove the temporary file
--- a/plugins/guests/linux/cap/insert_public_key.rb
+++ b/plugins/guests/linux/cap/insert_public_key.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
       class InsertPublicKey
         def self.insert_public_key(machine, contents)
           comm = machine.communicate
-          contents = contents.chomp
+          contents = contents.chomp+"\n"
 
           remote_path = "/tmp/vagrant-authorized-keys-#{Time.now.to_i}"
           Tempfile.open("vagrant-linux-insert-public-key") do |f|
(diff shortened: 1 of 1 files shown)


karlkfi:
does it need to be `\\n`? That's what it was using when it was a print statement.

jonnywilliamson:
I think the fact that it's surrounded by  double quotes takes care of the need to escape the \

However, please test it out on your machine, takes about 2 secs to add the line to that file.

Currently working for me...

This PR is just to show what needs to be done, not the best way to implement it!

karlkfi:
`echo` will supply the line break, you don't need to `echo "\n"`.

Also, I like this impl a bit better, using the old string content way, without a file. 

It's possible they deliberately wanted a file so it's not in the bash history, but since it's a public key and not a private key, it doesn't really affect security.

```
      def insert_public_key(machine, contents)
		comm = machine.communicate
		contents = contents.chomp
		contents = Vagrant::Util::ShellQuote.escape(contents, "'")

		comm.execute <<-EOH.gsub(/^ {12}/, '')
		  mkdir -p ~/.ssh
		  chmod 0700 ~/.ssh
		  printf '#{contents}\\n' >> ~/.ssh/authorized_keys
		  chmod 0600 ~/.ssh/authorized_keys
		EOH
	  end
```

sethvargo:
@karlkfi the old way doesn't work on non-linux hosts with linux guests - we have to upload the file and `cat` it into authorized keys. 

@jonnywilliamson can you update this PR to ensure the newline exists in the Ruby of `contents` before we shell into the system? It looks like we currently `chomp`, and that might not be the right behavior. Instead we should `strip` and then ensure we have a newline at the end.

jonnywilliamson:
@sethvargo 

Hi, so I took another look at the code (a little more sober too :) ) and I think this is a more streamlined approach.

The chomp vs strip query? I think chomp is fine here. It tidies up the end of the line rather than try and clean up both ends...I can't see any need to change this...does anyone else?

I've run this on my own system and it was much better than the previous ugly `echo` patch I made last night.

Hope it helps.

jonnywilliamson:
Just after reading that `<<` is much better than `+` in ruby concatenation, so I updated the PR again.

http://stackoverflow.com/questions/377768/string-concatenation-in-ruby

karlkfi:
@sethvargo: Curious. Why doesn't the old method work on non-linux hosts? I didn't see that mentioned in the commit that made the change.

PR #7482, opened 2016-06-18
guests/redhat: fix network configuration in CentOS 5

kaorimatz:
ip command is not on PATH by default in CentOS 5 so call it with the full path.


sethvargo:
Merged in de9abc8. Thanks!

PR #7484, opened 2016-06-18
guests/bsd: Centralize logic for halting guests

sethvargo:


wca:
It seems like this change wasn't tested on a FreeBSD guest, which yields this error:

/sbin/shutdown -p -h now

Stdout from the command:



Stderr from the command:

shutdown: incompatible switches -h, -k, -p and -r
usage: shutdown [-] [-h | -p | -r | -k] [-o [-n]] time [warning-message ...]

PR #7487, opened 2016-06-19
Fixing version check and catch statement in get_vm_status

codekaizen:
Use the version of Hyper-V instead of Powershell to determine which
exception is thrown by Get-VM. Also fixing an invalid catch statement,
since Powershell can not use variables to match thrown exceptions in
catch.


codekaizen:
This is an additional fix to the issue #7447, which duplicates the catch statement fix, but also works on Windows 2012 server which has a different version of Hyper-V.

mwrock:
:+1: this looks good to me and works well on my win 10 machine. I can also confirm the version check is correct for win 2012r2 both with and without wmf5.

sethvargo:
Thanks!

M-OC:
+1

Thanks a lot!

PR #7501, opened 2016-06-20
Ubuntu/xenial64 base box requires us to modprobe vboxsf ourselves

jcshort:
Attempts to use virtualbox shared folders with the Ubuntu 16.04 Cloud Image base box from https://vagrantcloud.com/ubuntu/boxes/xenial64 result in mount errors about invalid filesystem type.  The base box does ship the vboxsf module in /lib/modules, but only inserts the `vboxguest` module.  This plugin runs before any user-defined provisioners, so adding the modprobe to a shell provisioner is insufficient.

### DIFF
--- a/plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
+++ b/plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
@@ -6,7 +6,7 @@ module VagrantPlugins
           expanded_guest_path = machine.guest.capability(
             :shell_expand_guest_path, guestpath)
 
-          mount_commands = []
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @jcshort 

Thank you for opening a PR. I don't think this linux NFS option is accepted across all the linux guests, so I'm 👎 for adding it. I'm also generally not in favor of adding these types of workarounds due to bugs in boxes themselves. You can specify this option manually in the Vagrantfile using `mount_options` or use a well-behaved box like `bento/ubuntu-16.04`. There are countless other issues with the box published by canonical that need to be fixed upstream. Thanks!

Issue #7493, opened 2016-06-20
vagrant plugin install failure using local gem


lawrenceong:
### Vagrant version
1.8.4

### Host operating system
Ubuntu 16.04

### Debug output
https://gist.github.com/lawrenceong/a587ccf77fdaa51edc573979391977a1

### Expected behavior
The plugin installs without any issue.

### Actual behavior
We get a bundler error when attempting to install the local plugin.  Install works in Vagrant 1.8.1.  Installing a remote plugin works without any issue.

### Steps to reproduce
1. Install Vagrant 1.8.4
2. curl -O -L https://rubygems.org/downloads/vagrant-google-0.2.4.gem
3. vagrant plugin install ./vagrant-google-0.2.4.gem

vdloo:
I am experiencing the same issue. After upgrading to 1.8.4 'vagrant plugin install' can no longer install from local .gem files. Tested on OS X Yosemite and Ubuntu 16.04.

```
$ vagrant plugin install vagrant-hypconfigmgmt-0.0.1.gem 
Installing the 'vagrant-hypconfigmgmt-0.0.1.gem' plugin. This can take a few minutes...
Bundler, the underlying system Vagrant uses to install plugins,
reported an error. The error is shown below. These errors are usually
caused by misconfigured plugin installations or transient network
issues. The error from Bundler is:

Could not find gem 'vagrant-hypconfigmgmt (= 0.0.1)' in any of the gem sources listed in your Gemfile or available on this machine.
```

Tested with a random other vagrant plugin as well
```
$ vagrant plugin install pkg/vagrant-sshfs-1.1.0.gem 
Installing the 'pkg/vagrant-sshfs-1.1.0.gem' plugin. This can take a few minutes...
Bundler, the underlying system Vagrant uses to install plugins,
reported an error. The error is shown below. These errors are usually
caused by misconfigured plugin installations or transient network
issues. The error from Bundler is:

Could not find gem 'vagrant-sshfs (= 1.1.0)' in any of the gem sources listed in your Gemfile or available on this machine.
```

budhrg:
I had the similar issue with installing local gem in 1.8.4. Thanks for fixing again in 1.8.5.

PR #7505, opened 2016-06-20
core/bundler: Reset specs in a way that works with new Rubygems

sethvargo:
Thanks to @mitchellh's excellent bisecting skilz

Fixes GH-7493


mitchellh:
LGTM

Issue #7046, opened 2016-02-17
rsync functionality on Windows fails; requires cygwin and/or cygwin pathing?


jkymarsh:
On Vagrant 1.8.1, Windows 10.

I've attempted to get rsync shared folders working using both MinGW and cwRsync, with the same result on both:

```powershell
==> default: Rsyncing folder: /c/Users/Josh/Projects/vagrant-boilerplate/ => /vagrant
==> default:   - Exclude: [".vagrant/", ".git/"]
There was an error when attempting to rsync a synced folder.
Please inspect the error message below for more info.

Host path: /c/Users/Josh/Projects/vagrant-boilerplate/
Guest path: /vagrant
Command: rsync --verbose --archive --delete -z --copy-links --chmod=ugo=rwX --no-perms --no-owner --no-group --rsync-path sudo rsync -e ssh -p 2222 -o ControlMaster=auto -o ControlPath=C:/Users/Josh/AppData/Local/Temp/ssh.49 -o ControlPersist=10m -o StrictHostKeyChecking=no -o IdentitiesOnly=true -o UserKnownHostsFile=/dev/null -i 'C:/Users/Josh/Projects/vagrant-boilerplate/.vagrant/machines/default/virtualbox/private_key' --exclude .vagrant/ --exclude .git/ /c/Users/Josh/Projects/vagrant-boilerplate/ vagrant@127.0.0.1:/vagrant
Error: Could not create directory '/home/Josh/.ssh'.
Warning: Permanently added '[127.0.0.1]:2222' (ECDSA) to the list of known hosts.
mm_receive_fd: no message header
process_mux_new_session: failed to receive fd 0 from slave
mux_client_request_session: read from master failed: Connection reset by peer
Failed to connect to new control master
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: error in rsync protocol data stream (code 12) at io.c(226) [sender=3.1.2]
```

The root of the problem SEEMS to be some pathing issues: in the error, the `/c/Users/Josh/Projects/vagrant-boilerplate/` and `/home/Josh/.ssh` are both pretty alarming, given that this particular instance of `vagrant reload` was run within Powershell, so obviously neither of those directories exists in a Windows filesystem context.

`rsync` and `ssh` are both properly referenced in the PATH before any other entries.

[...]

PR #7506, opened 2016-06-21
Fix for #7046: forgo using ControlMaster for rsync on Windows

rafd123:
ControlMaster support is broken in cygwin.

See http://www.cygwin.com/ml/cygwin/2015-01/msg00123.html


rafd123:
There's already a PR for this it turns out: https://github.com/mitchellh/vagrant/pull/7012

PR #7510, opened 2016-06-23
Clarify build from source needs the bundler version.

kitsunde:
It wasn't clear what version the documentation was asking for, I thought it was asking for the Vagrant version at first. I've also made the bundler version consistent across the document.


sethvargo:
Hi @kitsunde 

Thank you for the Pull Request, but I'm not going to merge this. I intentionally used 1.2.3 instead of hard coding the real version. It's not possible to keep that documentation in sync with the gemspec, which is why it specifically says:

>  You will need to inspect the `vagrant.gemspec` to determine the version when you are compiling from source.

As also described at the top of that file, installing from source is only recommended for users with experience with Ruby and the Ruby ecosystem. We do not recommend compiling from source; please use the official installers instead.

PR #7519, opened 2016-06-24
Support installing synced folder clients in Arch

evaryont:
Add support for installing the NFS client, SMB/CIFS client, and rsync in
an Arch Linux guest.

### DIFF
--- a/plugins/guests/arch/cap/nfs_client.rb
+++ b/plugins/guests/arch/cap/nfs_client.rb
@@ -6,7 +6,7 @@ module VagrantPlugins
           comm = machine.communicate
           comm.sudo <<-EOH.gsub(/^ {12}/, '')
             pacman -Sy --noconfirm
-            pacman -S --noconfirm nfs-client
+            pacman -S --noconfirm nfs-utils
           EOH
         end
       end
--- a/plugins/guests/arch/cap/nfs_client.rb
+++ b/plugins/guests/arch/cap/nfs_client.rb
@@ -1,0 +1,16 @@
+module VagrantPlugins
+  module GuestArch
+    module Cap
+      class NFSClient
+        def self.nfs_client_install(machine)
+          comm = machine.communicate
+          comm.sudo <<-EOH.gsub(/^ {12}/, '')
+            pacman -Sy --noconfirm
+            pacman -S --noconfirm nfs-client
+          EOH
+        end
+      end
+    end
+  end
+end
+
--- a/plugins/guests/arch/cap/rsync.rb
+++ b/plugins/guests/arch/cap/rsync.rb
@@ -4,10 +4,12 @@ module VagrantPlugins
       class RSync
         def self.rsync_install(machine)
           comm = machine.communicate
-          comm.sudo <<-EOH.gsub(/^ {12}/, '')
-            pacman -Sy --noconfirm
-            pacman -S --noconfirm rsync
-          EOH
+          if !comm.test("test -f /usr/bin/rsync")
+            comm.sudo <<-EOH.gsub(/^ {14}/, '')
+              pacman -Sy --noconfirm
+              pacman -S --noconfirm rsync
+            EOH
+          end
         end
       end
     end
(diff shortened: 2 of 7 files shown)


plugins/guests/arch/cap/nfs_client.rb
```diff
@@ -0,0 +1,16 @@
+module VagrantPlugins
+  module GuestArch
+    module Cap
+      class NFSClient
+        def self.nfs_client_install(machine)
+          comm = machine.communicate
+          comm.sudo <<-EOH.gsub(/^ {12}/, '')
+            pacman -Sy --noconfirm
+            pacman -S --noconfirm nfs-utils
```
ogarcia: You can make this in one line: `pacman -Sy --noconfirm nfs-utils`

evaryont: That is not recommended, per this quote in the [Arch wiki](https://wiki.archlinux.org/index.php/Pacman#Installing_packages) and reiterated multiple times by developers:

> Warning: When installing packages in Arch, avoid refreshing the package list without upgrading the system (for example, when a package is no longer found in the official repositories). In practice, do not run pacman -Sy package_name instead of pacman -Syu package_name, as this could lead to dependency issues. See [System maintenance#Partial upgrades](https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported) are unsupported and [BBS#89328](https://bbs.archlinux.org/viewtopic.php?id=89328).

ogarcia: Thats OK, but the help says that you can do the work in one line. :wink:

ogarcia: Anyway, in one or two lines, the PR is pefect for me, please @mitchellh aprove it. :smile:

plugins/guests/arch/cap/rsync.rb
```diff
@@ -0,0 +1,18 @@
+module VagrantPlugins
+  module GuestArch
+    module Cap
+      class RSync
+        def self.rsync_install(machine)
+          comm = machine.communicate
+          if !comm.test("test -f /usr/bin/rsync")
```
sethvargo: Hi there,

This isn't the recommended way to do this anymore. We now have a capability for `rsync_installed` and a cap for `rsync_install`. The "installed" cap should perform the check if rsync is installed, and the install should actually install it.

sethvargo:
Hey there,

I tested this out and looked at some alternate implements. The problem is that pacman can be _huge_ to download and sync, resulting in a really slow first up. I think this is an OS where we really need to have the tool installed at build-time. Sorry!

evaryont:
Apologies, but I don't follow. Pacman is preinstalled on Arch. And it will
need to synchronize with the repositories, yes, but so would yum or apt.
Why would Arch be a special case? Also, a configuration issue could be the
repositories pointed at Arch directly (which is severely rate limited)
instead of a mirror. Could you have encountered that?

sethvargo:
Hi @evaryont 

Using `terrywang/archlinux` from Atlas, sync took 33 minutes on a 150/150mbps connection. It downloaded many gigabytes of data.

evaryont:
Oh dear! That is over two years old. Since then, Arch had gone through a number of major renovations. And at least one glibc rebuild. Since Arch is a rolling release distribution you were upgraded all the way to the modern 'version', so you ended up downloading the 2 years worth of upgrades.

I would recommend using @ogarcia's boxes instead: https://atlas.hashicorp.com/ogarcia/boxes/archlinux-x64 They are updated monthly, at the same cadence as Arch puts out snapshot installers.

(Though Arch doesn't really have versions, you could call terrywang's base box 24 versions old)

sethvargo:
Hey @evaryont 

Okay - at least I'm not crazy 😄 . I'm not super familiar with arch, but I thought "having the user wait 2 hours to boot a box is not a good UX". I thought that was just the norm.

sethvargo:
@evaryont can you rebase this quickly and I'll get it into the next release please?

evaryont:
Do you still want the rsync install/installed cap split as well?

sethvargo:
b31b240 - I just grabbed it myself 😄

PR #6869, opened 2016-01-13
fixing incorrect missing ansible error

daniel-sc:
At least for ansible 2.0.0.1 the command `ansible-galaxy --help` is inappropriate for testing if ansible is installed, as it yields an error:

```
vagrant@vagrant-ubuntu-trusty-64:~$ ansible-galaxy --help && echo "OK"          
Usage: ansible-galaxy [delete|import|info|init|install|list|login|remove|search|setup] [--help] [options] ...

Options:
  -h, --help     show this help message and exit
  -v, --verbose  verbose mode (-vvv for more, -vvvv to enable connection
                 debugging)
  --version      show program's version number and exit
ERROR! Missing required action
```

PR #7536, opened 2016-06-30
More robust Ansible detection for local Ansible provisioner

hph86:
For the local Ansible provisioner, I would like to suggest a more robust way of Ansible guest detection. The current way of [checking return values](https://github.com/mitchellh/vagrant/blob/a4f45a18a653bd5b157a00c5ffaa3ba0c74ea5f5/plugins/provisioners/ansible/provisioner/guest.rb#L55-L60) has already caused trouble for the Ansible 2.0 detection (#6869). Future changes to the Ansible CLI and the return codes may break the detection suggested in #6869 again. Therefore, this PR changes the detection behavior to check that `ansible-galaxy` and `ansible-playbook` are both included in `$PATH` and executable. What do you think?

### DIFF
--- a/plugins/provisioners/ansible/provisioner/guest.rb
+++ b/plugins/provisioners/ansible/provisioner/guest.rb
@@ -54,7 +54,7 @@ module VagrantPlugins
 
           # Check that ansible binaries are well installed on the guest,
           @machine.communicate.execute(
-            "ansible-galaxy info --help && ansible-playbook --help",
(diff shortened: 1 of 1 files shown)


gildegoma:
@hph86 thank you very much for this pull request. Please let me some time to think a bit more about it, but at first glance, I really love the idea!

hph86:
@gildegoma: Thank you for your feedback. I have just updated the PR to be more POSIX-compatible.

gildegoma:
@hph86 This got finally merged, with some modifications... See a55590b for more details.

Thanks again :)

PR #7550, opened 2016-07-06
plugin: Write out state file via create-new-then-rename

cgwalters:
This makes updates atomic, so it's useful generally.  However
I'm making this change specifically to support rpm-ostree.

For more information, see:
https://bugzilla.redhat.com/show_bug.cgi?id=1352152

### DIFF
--- a/lib/vagrant/plugin/state_file.rb
+++ b/lib/vagrant/plugin/state_file.rb
@@ -1,4 +1,6 @@
 require "json"
+require "fileutils"
+require "tempfile"
 
 module Vagrant
   module Plugin
@@ -91,8 +93,11 @@ module Vagrant
 
       # This saves the state back into the state file.
       def save!
-        @path.open("w+") do |f|
+        Tempfile.open(@path.basename.to_s, @path.dirname.to_s) do |f|
           f.write(JSON.dump(@data))
+          f.chmod(0644)
+          f.close
+          FileUtils.mv(f.path, @path)
         end
       end
 


lib/vagrant/plugin/state_file.rb
```diff
@@ -91,9 +92,10 @@ def sources
 
       # This saves the state back into the state file.
       def save!
-        @path.open("w+") do |f|
-          f.write(JSON.dump(@data))
-        end
+        tmpfile = Tempfile.new(@path.basename.to_s, @path.dirname.to_s)
+        tmpfile.write(JSON.dump(@data))
+        tmpfile.close
+        File.rename(tmpfile.path, @path)
```
sethvargo: Yea, so there's a reason we don't do this...

1. `File.rename` has historically been a problem on Windows
1. `File.rename` does not work across multiple volumes, particularly if the volumes are different formats
1. The tmpfile could be removed by GC before the mv takes place (aka there's a slight race between lines 97 and 98 here)

I think it's probably a better idea to do something like:

```ruby
Tempfile.open("vagrant-plugins") do |f|
  f.binmode
  f.write(JSON.dump(@data))
  f.close
  FileUtils.mv(f.path, @path)
end
```

What do you think?

cgwalters: Re 2: The tempfile is created in the same directory as the target
Re 3: I'll be honest and say this is my first Ruby patch, but I have a strong suspicion it can't be removed by GC because the local variable holds a strong reference across the function scope

But yes, this is just an implementation of a general pattern for "replace file atomically" which is harder on Windows indeed.  I'm an upstream develope for GLib where for `g_file_set_contents()`  we do https://git.gnome.org/browse/glib/tree/glib/gfileutils.c#n1217

It's the sort of thing I'd expect Ruby to have in the standard library, but if it exists I couldn't find it.

sethvargo: I think `FileUtils.mv` handles this better than `File.rename`

cgwalters:
Okay, updated :arrow_up: to use `FileUtils.mv`

sethvargo:
@cgwalters This is still going to leave the tempfile lying around. The Ruby docs are wrong... you have to call #unlink on the file to remove it, but the block form that I posted in my earlier comment automatically does that.

cgwalters:
`FileUtils.mv` will use `rename()` on Unix which does unlink the
tempfile, and on Windows it will to after deleting the original.
 
The tempfile *does* still call `unlink` in its cleanup on the old name,
but that's relatively harmless...not sure offhand if there's a way to
avoid that.
 
(There has to be some Ruby codebase that does things like this right and
has a "write file atomically" utility function but I don't know Ruby
well enough to know which codebases are good to look at)
 
 
On Thu, Jul 7, 2016, at 03:19 PM, Seth Vargo wrote:
> @cgwalters[1] This is still going to leave the tempfile lying around.
> The Ruby docs are wrong... you have to call #unlink on the file to
> remove it, but the block form that I posted in my earlier comment
> automatically does that.
 

Links:

  1. https://github.com/cgwalters

cgwalters:
Sorry, when reworking this I introduced a regression - we need to chmod back to 0644.  Will fix.

PR #7574, opened 2016-07-13
Add VirtualBox 5.1 driver

### LINKED ISSUE #7411: Support Virtualbox 5.1
I'm not on a machine I can send a PR from, but [these exact steps](https://github.com/mitchellh/vagrant/issues/5572) with the obvious modifications work.

I looked briefly at the release notes and I don't think there are any new configuration options worth tweaking the driver over.

Ingramz:
Fixes #7411

Not sure why nobody has submitted a PR yet.

### DIFF
--- a/plugins/providers/virtualbox/driver/meta.rb
+++ b/plugins/providers/virtualbox/driver/meta.rb
@@ -61,6 +61,7 @@ module VagrantPlugins
             "4.2" => Version_4_2,
             "4.3" => Version_4_3,
             "5.0" => Version_5_0,
+            "5.1" => Version_5_1,
           }
 
           if @@version.start_with?("4.2.14")
--- a/plugins/providers/virtualbox/driver/version_5_1.rb
+++ b/plugins/providers/virtualbox/driver/version_5_1.rb
@@ -1,0 +1,734 @@
+require 'log4r'
+
+require "vagrant/util/platform"
+
+require File.expand_path("../base", __FILE__)
+
+module VagrantPlugins
+  module ProviderVirtualBox
+    module Driver
+      # Driver for VirtualBox 5.1.x
+      class Version_5_1 < Base
+        def initialize(uuid)
+          super()
+
+          @logger = Log4r::Logger.new("vagrant::provider::virtualbox_5_1")
+          @uuid = uuid
+        end
+
+        def clear_forwarded_ports
+          args = []
+          read_forwarded_ports(@uuid).each do |nic, name, _, _|
+            args.concat(["--natpf#{nic}", "delete", name])
+          end
+
+          execute("modifyvm", @uuid, *args) if !args.empty?
+        end
+
+        def clear_shared_folders
+          info = execute("showvminfo", @uuid, "--machinereadable", retryable: true)
+          info.split("\n").each do |line|
+            if line =~ /^SharedFolderNameMachineMapping\d+="(.+?)"$/
+              execute("sharedfolder", "remove", @uuid, "--name", $1.to_s)
+            end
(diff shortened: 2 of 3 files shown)


apetresc:
I've never worked on the Vagrant code before, but this seems suboptimal. Does Vagrant really expect every point release of VirtualBox to have its own bit-for-bit copy (except for the version number) of a 734-line driver?

```diff
$ wc -l version_5_*
  734 version_5_0.rb
  734 version_5_1.rb
 1468 total
$ diff version_5_0.rb version_5_1.rb
10,11c10,11
<       # Driver for VirtualBox 5.0.x
<       class Version_5_0 < Base
---
>       # Driver for VirtualBox 5.1.x
>       class Version_5_1 < Base
15c15
<           @logger = Log4r::Logger.new("vagrant::provider::virtualbox_5_0")
---
>           @logger = Log4r::Logger.new("vagrant::provider::virtualbox_5_1")
```

I notice that the other drivers do the same thing (although they at least have a few minor implementation differences), so I don't doubt you're doing the right thing - but it seems like a better design should exist in Vagrant. For example, what's stopping us right now from either:

- rename the `5.1` driver to `Version_5_x` and then changing `meta.rb` to simply map `5.0` and `5.1` to it?
- having the `5.1` driver subclass the `5.0` driver and override only the version number/logger?

Just trying to brainstorm a better alternative :)

Ingramz:
> Does Vagrant really expect every point release of VirtualBox to have its own bit-for-bit copy (except for the version number) of a 734-line driver?

Yes. In fact 4.3 and 5.0 are functionally identical as well.

thom-nic:
@apetresc or one could use straightforward inheritance? 

```ruby
class Version_5_x < Base
  # implementation from old Version_5_0...
end

class Version_5_0 < Version_5_x
  @logger = Log4r::Logger.new("vagrant::provider::virtualbox_5_0")
end
class Version_5_1 < Version_5_x
  @logger = Log4r::Logger.new("vagrant::provider::virtualbox_5_1")
end
```

apetresc:
@Ingramz Well, not quite *identical*, but still extremely similar in a way that could easily be subclassed away:

```diff
$ diff version_4_3.rb version_5_0.rb
1d0
< require 'ipaddr'
11,12c10,11
<       # Driver for VirtualBox 4.3.x
<       class Version_4_3 < Base
---
>       # Driver for VirtualBox 5.0.x
>       class Version_5_0 < Base
16c15
<           @logger = Log4r::Logger.new("vagrant::provider::virtualbox_4_3")
---
>           @logger = Log4r::Logger.new("vagrant::provider::virtualbox_5_0")
74a74,75
>           else
>             raise "BUG: Unknown IP type: #{ip.inspect}"
86,90d86
<         def reconfig_host_only(interface)
<           execute("hostonlyif", "ipconfig", interface[:name],
<                   "--ipv6", interface[:ipv6])
<         end
< 
589a586,590
>         end
> 
>         def reconfig_host_only(interface)
>           execute("hostonlyif", "ipconfig", interface[:name],
>                   "--ipv6", interface[:ipv6])
```

@thom-nic Yup, exactly my second suggestion. So I guess the question is – is there any reason things couldn't be done this way, and why previous drivers don't do this?

joelhandwell:
@sethvargo today, I knew VirtualBox 5.1 is released as stable version, and installed and faced issue https://github.com/mitchellh/vagrant/issues/7411. Can releasing 1.8.4.1 with VirtualBox 5.1 support be considered as option?

gggeek:
+1 for 1.8.4.1

PR #7579, opened 2016-07-14
Fix issuee 6690: rsync-path is not quoted

cnadeau:
Added quotes to rsync-path argument parameters to ensure it runs properly on Windows Command Line and Git Bash

Resolves issue https://github.com/mitchellh/vagrant/issues/6690


sethvargo:
I'm actually working on a more top-level solution for this. Thanks for the PR, but we actually need to fix this across all guests: https://github.com/mitchellh/vagrant/issues/7580

cnadeau:
Good! thank you very much!

PR #7584, opened 2016-07-14
winrm communicator: the task is only really finished when $out_file is closed

rgl:
I was getting the following error:

```
==> test: del : Cannot remove item C:\Windows\Temp\WinRM_Elevated_Shell.log: The process cannot access the file
==> test: 'C:\Windows\Temp\WinRM_Elevated_Shell.log' because it is being used by another process.
==> test: At C:\tmp\vagrant-elevated-shell.ps1:19 char:3
==> test: +   del $out_file
==> test: +   ~~~~~~~~~~~~~
==> test:     + CategoryInfo          : WriteError: (C:\Windows\Temp\WinRM_Elevated_Shell.log:FileInfo) [Remove-Item], IOExcepti
==> test:    on
==> test:     + FullyQualifiedErrorId : RemoveFileSystemItemIOError,Microsoft.PowerShell.Commands.RemoveItemCommand
```
My Vagrantfile has something like:

    config.vm.provision "shell", inline: "c:/tmp/provision/install-a.ps1", name: "Install a"
    config.vm.provision "shell", inline: "c:/tmp/provision/install-b.ps1", name: "Install b"

The error happens when its trying to run the second provision script.

It turns out that the install-a.ps1 scheduled task was not really finished until the $out_file is closed.

PS my Host and Guest is Windows 10 64-bit. Vagrant is 1.8.4. VirtualBox is 5.0.24r108355.

### DIFF
--- a/plugins/communicators/winrm/scripts/elevated_shell.ps1.erb
+++ b/plugins/communicators/winrm/scripts/elevated_shell.ps1.erb
@@ -95,6 +95,29 @@ do {
   $cur_line = SlurpOutput $out_file $cur_line
 } while (!($registered_task.state -eq 3))
 
+function Test-FileLock {
+  param (
+    [Parameter(Mandatory=$true)]
+    [String]$Path
+  )
+  if (!(Test-Path $Path)) {
+    return $false
+  }
+  try {
+    (New-Object IO.FileInfo $Path).
+      Open([IO.FileMode]::Open, [IO.FileAccess]::ReadWrite, [IO.FileShare]::None).
+      Close()
+    $false
+  } catch {
+    return $true
+  }
+}
+
+# the task is only really finished when $out_file is closed.
+while (Test-FileLock $out_file) {
+  Start-Sleep -m 100
+}
+
 $exit_code = $registered_task.LastTaskResult
 [System.Runtime.Interopservices.Marshal]::ReleaseComObject($schedule) | Out-Null
 


mwrock:
@sneal broke out the elecvated task a while back to a [separate gem](https://github.com/WinRb/winrm-elevated) and it has addressed this scenario as well. Once we release [winrm-v2](https://github.com/WinRb/winrm-elevated/pull/11) we will roll that all into vagrant. In the meantime I recomend taking [the approach](https://github.com/WinRb/winrm-elevated/blob/master/lib/winrm-elevated/scripts/elevated_shell.ps1) that winrm-elevated takes.

That approach can be best seen in [this commit](https://github.com/WinRb/winrm-elevated/pull/6/commits/fad4e008e574d3b92266298f99ee8b1c68ff19f1). It creates uniquely named temp files for the stderr/out and deletes them at the end. I think its a bit simpler than the one here and less prone to potential hangs.

@rgl if you don't mind altering this PR to do that, I think you will find that it solves your scenario just as well.

rgl:
@mwrock I think the problem you mention is unrelated to this one. I'm not trying to run multiple provision scripts at the same time, hence, there are no multiple/concurrent WinRM connections.

Deleting the temporary files at the end will not work in this case, because the file(s) is still being used at that point.

The file being used might be a manifestation that something is not right. In my case it seems its a InstallShield setup (the /sms flag does not really help...) that is forking another process into the background, and returns immediately, and because of that, the scheduled task seems to be finished, but the logical provision step is not really done, and as such, I think the possible solution/workaround/compromise is to really wait until the file is unlocked at the vagrant level as this PR does.

mwrock:
You are likely correct about the file uniqueness issue. I have certainly seen issues where locks on a file are not released before another process seeks to access it regardless of the fact that it is no longer used.

However given this is an installation scenario there are 2 other options:

1. Probably not the issue here - One can rub against the default 5 minute timeout at which point a scheduled task ends but keeps the child process running. We ran into this with winrm-elevated when trying to download and then install the chef client on slow internet. The fix for that was to add: `<ExecutionTimeLimit>PT24H</ExecutionTimeLimit>` to the `<Settings>` node.
2. Based on your observation, its the forking of the installer process causing control to return to the task script immediately.  I do think a better route in this case would be to use `start-process` with the `-wait` argument to force powershell to wait for the installer process to terminate.

We don't want to introduce special behavior here that would cause the script to perform differently than outside of the task. In other words, it it returns immediately outside of a task, then it should do the same here. So we want to change the script given to the task to properly behave.

Hope that makes sense.

rgl:
Well, it does and does not make sense! I can now see that I was trapped by yet another POSH pitfall... `&` vs `Start-Process` to call external processes.

But yes, with `Start-Process -Wait` it does work! Why I need to do that is beyond my astonishment... if you could provide some pointers, they would be greatly appreciated!

Oh and thanks for the Start-Process pointer :-)

mwrock:
I live my life in a constant state of astonishment.

The `&` (or call) operator in powershell simply invokes the script or executable given to it. 

Powershell (I'm honestly not sure if this is just PS or cmd,exe too) behaves differently depending on the type of executable. It will block on console executables but not windows subsystem executables like MSIEXEC and other installers. So `Start-Process` just gives you more control and the ability to wait and even capture the launched process.

chrisroberts:
Looks like using the `Start-Process` with the `-Wait` flag is the correct solution here so I'm going to close this up. Thanks!

PR #7589, opened 2016-07-15
Support rest-client 2.x as a dependency

junaruga:
Hi,
I want to use rest-client latest version 2.0 as vagrant's dependency.
https://rubygems.org/gems/rest-client
Is it possible?
I checked to pass the test suite by Ruby 2.2.5 and 2.3.1 by myself after I updated.


junaruga:
Thanks for your merging!

PR #7621, opened 2016-07-22
redux of pr 7398 for ssh-agent key fix

zmt:
This is probably easier...

### DIFF
--- a/plugins/commands/ssh_config/command.rb
+++ b/plugins/commands/ssh_config/command.rb
@@ -37,13 +37,20 @@ module VagrantPlugins
             ssh_host: ssh_info[:host],
             ssh_port: ssh_info[:port],
             ssh_user: ssh_info[:username],
-            private_key_path: ssh_info[:private_key_path],
+            keys_only: ssh_info[:keys_only],
+            paranoid: ssh_info[:paranoid],
             forward_agent: ssh_info[:forward_agent],
             forward_x11:   ssh_info[:forward_x11],
             proxy_command: ssh_info[:proxy_command],
             ssh_command:   ssh_info[:ssh_command],
             forward_env:   ssh_info[:forward_env],
           }
+          if ssh_info[:private_key_path]
+            variables['private_key_path'] = ssh_info[:private_key_path]
+          end
+          if ssh_info[:log_level]
+            variables['log_level'] = ssh_info[:log_level]
+          end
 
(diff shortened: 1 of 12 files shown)


plugins/commands/ssh_config/command.rb
```diff
@@ -37,13 +37,20 @@ def execute
...
             proxy_command: ssh_info[:proxy_command],
             ssh_command:   ssh_info[:ssh_command],
             forward_env:   ssh_info[:forward_env],
           }
+          if ssh_info[:private_key_path]
+            variables['private_key_path'] = ssh_info[:private_key_path]
+          end
+          if ssh_info[:log_level]
+            variables['log_level'] = ssh_info[:log_level]
+          end
```
chrisroberts: Any reason for conditionally setting these variable entries? If there's not, can they be placed back into the hash construction?

plugins/synced_folders/rsync/helper.rb
```diff
@@ -63,7 +63,11 @@ def self.rsync_single(machine, ssh_info, opts)
         opts[:owner] ||= ssh_info[:username]
         opts[:group] ||= ssh_info[:username]
 
+        # set log level
+        log_level = ssh_info[:log_level] || "FATAL"
+
```
chrisroberts: This `log_level` looks to be extraneous here, or was logging level meant to be set within the options?

chrisroberts:
@zmt Looks great! Left notes on a couple small things. If you could address those, that would be great. Thanks!

chrisroberts:
Applied some updates to this PR via #7703. Thanks for getting this PR together!

raatti:
This is still not fixed in Released version, when is this getting out?

plugins/commands/ssh_config/command.rb
zmt: yes - we wanted to do ssh-agent only, which would mean that ssh_info[:private_key_path] would not be set

plugins/synced_folders/rsync/helper.rb
zmt: I don't recall at the moment, but this was either copied from older code or prevented errors from log_level not being set in my testing way back when.

zmt:
@raatti This went out with 1.8.6 on 9/27/16 apparently.

PR #7623, opened 2016-07-22
Support Vagrantfile-relative VAGRANT_DOTFILE_PATHs

carlosefr:
These changes allow the VAGRANT_DOTFILE_PATH variable to contain paths relative to the Vagrantfile's location. This allows setting this variable globally in a `.bashrc` to avoid conflicts in Dropbox-shared directories while still allowing Vagrant commands to be issued from any project subdirectory.

This seems to be a common request and sidesteps the usual hack of setting the variable at the top of the Vagrantfile and reinvoke itself.

A placeholder `{}` at the start of the path is replaced by the path to the Vagrantfile. I'm personally using this to have non-conflicting `.vagrant` directories like so:
```
export VAGRANT_DOTFILE_PATH="{}/.vagrant-${USER}@${HOSTNAME}"
```

### DIFF
--- a/lib/vagrant/environment.rb
+++ b/lib/vagrant/environment.rb
@@ -157,9 +157,13 @@ module Vagrant
       end
 
       # Setup the local data directory. If a configuration path is given,
-      # then it is expanded relative to the working directory. Otherwise,
+      # it is expanded relative to the root path or to the working directory
+      # depending on whether it starts with the "{}" placeholder. Otherwise,
       # we use the default which is expanded relative to the root path.
       opts[:local_data_path] ||= ENV["VAGRANT_DOTFILE_PATH"] if !opts[:child]
+      if opts[:local_data_path] && opts[:local_data_path].start_with?("{}") && !root_path.nil?
+	opts[:local_data_path] = File.join(root_path, opts[:local_data_path].sub(/^\{\}/, ''))
+      end
       opts[:local_data_path] ||= root_path.join(DEFAULT_LOCAL_DATA) if !root_path.nil?
       if opts[:local_data_path]
         @local_data_path = Pathname.new(File.expand_path(opts[:local_data_path], @cwd))
--- a/lib/vagrant/environment.rb
+++ b/lib/vagrant/environment.rb
@@ -160,9 +160,13 @@ module Vagrant
       # it is expanded relative to the root path or to the working directory
       # depending on whether it starts with the "{}" placeholder. Otherwise,
       # we use the default which is expanded relative to the root path.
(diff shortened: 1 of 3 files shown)


sethvargo:
Hi @carlosefr 

Thank you for the pull request. This seems like a really hacky solution to a problem that we should ultimately fix upstream. While I appreciate the work, I'm 👎 because I think `{}` and random variable substitution is not a great UX. I'll leave this open to think on it though. Thanks!

carlosefr:
Well, it is kind of hacky. Another possible solution would be to control the used root path based on another environment variable, but that's hacky too (even more so, I think).

Actually, I don't understand why `.vagrant` is always placed in the same location as the Vagrantfile **but** `VAGRANT_DOTFILE_PATH` is interpreted as relative to the current directory. This doesn't make sense to me and I believe the proper solution would be to also consider it relative to the Vagrantfile's location. It is actually surprising to find out that the interpretation isn't consistent.

Besides breaking compatibility, I don't see any other bad consequences from this. But I don't known enough about Vagrant to be sure.

As for breaking compatibility, that would imply people are creating distinct `.vagrant` directories on different root directories on purpose. It doesn't seem too realistic and maybe it's worth breaking.

carlosefr:
What about making `VAGRANT_DOTFILE_PATH` relative to the Vagrantfile's location and introducing a new environment variable (eg. `VAGRANT_DOTFILE_PATH_USE_CWD`) that, when set, keeps the old "relative-to-cwd" behavior for compatibility reasons?

`VAGRANT_DOTFILE_PATH_USE_CWD` could be maintained indefinitely or documented as transitional and marked for removal in a future release.

sethvargo:
Hi @carlosefr 

From what I understand, the dotfile path is relative to the CWD, which is almost always where the Vagrantfile lives. If that's not the case, we should fix that. 

The other option is to make it relative to `VAGRANT_VAGRANTFILE`. What do you think?

carlosefr:
That's not the case indeed. When `VAGRANT_DOTFILE_PATH` is unset, the Vagrantfile is searched up the directory tree starting at `VAGRANT_CWD` and then `.vagrant` is created next to it. When `VAGRANT_DOTFILE_PATH=.vagrant` the semantics stays the same for the Vagrantfile but changes for the `.vagrant` directory.

I've updated the pull request to make the semantics stay the same in both cases. Does it look like something that could be acceptable?

sethvargo:
Hi @carlosefr 

Would you be willing to add a few tests for this, just so we don't regress in the future?

sethvargo:
LGTM

carlosefr:
I've topped it off with a negative test and this should be it.

chrisroberts:
:+1:

Issue #7625, opened 2016-07-22
change_host_name fails on bootup


kevwil:
Vagrant version: 1.8.5
Host OS: Mac OS X 10.10.5
Guest OS: Ubuntu 12.04.1

Vagrantfile:
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  config.vm.box = "precise64"
  config.vm.box_url = 'http://files.vagrantup.com/precise64.box'
  config.vm.host_name = 'aws-nv-c-qstest01'

  # This is the port for a web server
  config.vm.network "forwarded_port", guest: 8080, host: 8080

  # This is the port for the Jacoco server to retrieve code coverage data
  config.vm.network "forwarded_port", guest: 6300, host: 6300

  # #  Puppet config. Super fast Puppet dev!
  config.vm.provision :puppet do |puppet|
    puppet.manifests_path = ENV['PUPPET_ROOT'] + '/manifests'
    puppet.module_path    = ENV['PUPPET_ROOT'] + '/modules'
    puppet.manifest_file  = 'site.pp'
  end
end
```

Debug Output: https://gist.github.com/kevwil/5f2a347764a689e338556ddde4f656c0

### Expected behavior
What should have happened?

Clean bootup of virtual machine, possibly with puppet errors as that's what I'm working on.

### Actual behavior
What actually happened?

Bootup is aborted because 'change_host_name' on this linux guest is 'not supported'.

### Steps to reproduce
1. Install latest Virtualbox and Vagrant
2. Run 'vagrant up' with the above Vagrantfile
3. Boom!

### References
Are there any other GitHub issues (open or closed) that should be linked here?
For example:
- GH-6332

kevwil:
I tried going back to Vagrant 1.8.4, but it didn't work at all with Virtualbox 5.1.x, so I went back to the latest 1.8.5. I then tried rolling back Virtualbox to 5.0.26, and I still get this error.

sethvargo:
I'm pretty sure this is a result of https://github.com/mitchellh/vagrant/pull/7524, which is what I was afraid of. 

@kevwil can you run 

```
. /etc/os-release && test xubuntu = x$ID
```

and report the output? I'm willing to bet that older ubuntu doesnt have that file and detection is failing.

/cc @nishidayuya

kevwil:
[...]

PR #7524, opened 2016-06-27
A more reliable way to detect Ubuntu

nishidayuya:
`lsb_release` command is included in lsb-release package. Sometimes this is not installed; ex: Docker official Ubuntu image.

```sh
$ docker run --rm -it ubuntu:precise bash -l -c 'lsb_release -i'
bash: lsb_release: command not found
$ docker run --rm -it ubuntu:trusty bash -l -c 'lsb_release -i' 
Distributor ID: Ubuntu
$ docker run --rm -it ubuntu:xenial bash -l -c 'lsb_release -i'
bash: lsb_release: command not found
```

`/etc/os-release` is included in base-files package, so this is more reliable than lsb_release.

```sh
$ docker run --rm -it ubuntu:precise bash -l -c '. /etc/os-release && echo $ID'  
ubuntu
$ docker run --rm -it ubuntu:trusty bash -l -c '. /etc/os-release && echo $ID' 
ubuntu
$ docker run --rm -it ubuntu:xenial bash -l -c '. /etc/os-release && echo $ID'
ubuntu
```

This pull-request uses `/etc/os-release` for Ubuntu detection.

PR #7632, opened 2016-07-23
guests/ubuntu: Revert detection

sethvargo:
This fixes Ubuntu guest detection to fallback to the old detection method.

- Reverts GH-7524 (mostly)
- Fixes GH-7625


kikitux:
This fix on ubuntu

```
Vagrant attempted to execute the capability configure_networks on the detect guest OS linux, but the guest doesn't support that capability
```

Adding command to help internet search

PR #7643, opened 2016-07-26
Update development vagrant VM

carlosefr:
While looking into adding some tests for PR #7623 I noticed that the development vagrant VM doesn't install the proper version of Ruby neither does it install the necessary dependencies to immediately run the test suite.

These changes fix this.


sethvargo:
Thanks!

Issue #5070, opened 2014-12-27
vagrant package fails on undefined method 'join'


almacro:
Hi There:

I hit breakage when running vagrant package using 1.7.1.
The output log is here: https://gist.github.com/almacro/485fd31916e64921b9a5

Here are some system fingerprints:

$ vagrant --version
Vagrant 1.7.1
$ ruby -v
ruby 2.0.0p353 (2013-11-22) [x86_64-linux]
$ uname -a
Linux localhost.localdomain 3.10.0-123.9.3.el7.x86_64 #1 SMP Thu Nov 6 15:06:03 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
$ cat /etc/centos-release 
CentOS Linux release 7.0.1406 (Core)

almacro:
The problem reproduced on MacOS X [10.9.5] with Vagrant 1.7.1 and patch from #4962.
Any workaround for this?

almacro:
Backed out #4962, applied #4972.
Failure Error changes to 
/opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:83:in `core#hash_merge_kwd': wrong argument type String (expected Symbol) (TypeError)
        from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:83:in `package_vm'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:66:in `package_base'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/plugins/commands/package/command.rb:42:in `execute'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/lib/vagrant/cli.rb:42:in `execute'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.7.1/lib/vagrant/environment.rb:301:in `cli'
        from /opt/vagrant/bin/../embedded/gems/gems/vagrant-1.7.1/bin/vagrant:174:in `<main>'

Fantus:
iam using ubuntu 14 and vagrant 1.7.1 which results in the same problem.
i also applied #4962.

i also tried installing from the github with vagrant 1.7.2 dev version which results in the last error almarco had: https://gist.github.com/almacro/aa0b74cd2207c1770dd2

mitchellh:
Fixed in master, 1.7.2 coming out soon.

[...]

PR #7658, opened 2016-07-28
Fix for #5070

hatxor:
Fix for the issue in https://github.com/mitchellh/vagrant/issues/5070

### DIFF
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -168,7 +168,7 @@ module VagrantPlugins
 
           # Write out the private key in the data dir so that the
           # machine automatically picks it up.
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi there,

I don't think this is the correct fix. We want to use the machine's data_dir, not the environment, if I understand correctly.

chrisroberts:
@sethvargo :+1: the code is expecting to use the machine's data directory, not the environment. The core issue looks to be the `Machine` instance is not provided a data directory on instantiation which results in the nil join errors.

chrisroberts:
I have pushed a new PR to address this issue: #7725 which creates a temporary machine data directory. Closing this out in favor of the new PR. Thank you very much for submitting this PR!

PR #7659, opened 2016-07-28
darwin rdp: File was missing newlines

docwhat:
The RDP file being created for Darwin was missing newlines.

It works with both \n(unix) and \r\n(dos) but I decided to go the
MS route.

---
This change is [Reviewable](https://reviewable.io/reviews/mitchellh/vagrant/7659)

### DIFF
--- a/plugins/hosts/darwin/cap/rdp.rb
+++ b/plugins/hosts/darwin/cap/rdp.rb
@@ -40,12 +40,12 @@ module VagrantPlugins
             f.binmode
 
             opts.each do |k, v|
-              f.write("#{k}:#{v}")
+              f.write("#{k}:#{v}\r\n")
             end
 
             if rdp_info[:extra_args]
               rdp_info[:extra_args].each do |arg|
-                f.write("#{arg}")
+                f.write("#{arg}\r\n")
               end
             end
 
--- a/test/unit/plugins/hosts/darwin/cap/rdp_test.rb
+++ b/test/unit/plugins/hosts/darwin/cap/rdp_test.rb
@@ -18,6 +18,9 @@ describe VagrantPlugins::HostDarwin::Cap::RDP do
     expect(result).to match("full address:s:host:port")
     expect(result).to match("prompt for credentials:i:1")
     expect(result).to match("username:s:username")
+    result.lines.each do |line|
+      expect(line).to match(/\r\n$/)
+    end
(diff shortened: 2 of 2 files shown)


docwhat:
Of course, it would be even more portable if we used the [RDP Url](https://technet.microsoft.com/en-us/library/dn690096(v=ws.11).aspx) instead.  Since that should work on Windows, Linux and Darwin.

plugins/hosts/darwin/cap/rdp.rb
```diff
@@ -40,12 +40,12 @@ def self.generate_config_file(rdp_info)
             f.binmode
 
             opts.each do |k, v|
-              f.write("#{k}:#{v}")
+              f.write("#{k}:#{v}\r\n")
```
sethvargo: hmm... Is `\r\n` the right linebreak for Darwin?

docwhat: It's an RDP file. The version of Microsoft Remote Desktop I have (latest from App Store) doesn't care...

The code above it does `f.binmode` just above, so I thought someone was being paranoid and went with it.

Sharpie:
Noticed this today as well and opened #7698 which contains a variation of the same fix. I think either will probably be fine, left some notes about the unit test.

test/unit/plugins/hosts/darwin/cap/rdp_test.rb
```diff
@@ -18,13 +18,19 @@
     expect(result).to match("full address:s:host:port")
     expect(result).to match("prompt for credentials:i:1")
     expect(result).to match("username:s:username")
+    result.lines.each do |line|
+      expect(line).to match(/\r\n$/)
+    end
```
Sharpie: This test asserts that one or more lines in the file end in `\r\n`, which isn't the same guarantee that each setting appears on it's own line. The matches should be re-written as array inclusion tests against the content of the file:

```ruby
result = File.readlines(path).map(&:chomp)
expect(result).to include("drivestoredirect:s:*")
expect(result).to include("full address:s:host:port")
expect(result).to include("prompt for credentials:i:1")
expect(result).to include("username:s:username")
```

chrisroberts:
Closing this as #7698 was merged and resolves this issue. Thanks for the report and PR!

PR #7675, opened 2016-08-03
guests: Prevent ssh disconnect from causing error on halt

chrisroberts:
When halting the VM the sshd may be stopped before the ssh connection in use has been closed. This causes an SSHDisconnected error to bubble up and the UI to display an error message. Change rescues out the SSHDisconnected exception to prevent error message from surfacing.


mitchellh:
LGTM, makes sense

sethvargo:
👍

chrisroberts:
Updated to make consistent against all guests providing halt capability

Issue #7613, opened 2016-07-20
1.8.5 failling to bring private interface up using vbox


rafamanzo:
After upgrading to 1.8.5 and VirtualBox 5.1 the following stared to happen. Thanks for any help!

### Vagrant version
Vagrant 1.8.5

### Host operating system
Arch Linux (Virtualbox 5.1)

### Guest operating system
Arch Linux

### Vagrantfile
https://github.com/rafamanzo/pcsetup/blob/e2ab89d0bb3af524ca805b11aa2a22e8cde78895/Vagrantfile

### Debug output
https://gist.github.com/rafamanzo/aa6178bfffa1094a15f88eb569aa8002

### Expected behavior
It should generate the script for enabling private network with the correct name.

### Actual behavior
The generated script set the interface name to `Wednesday, 20 July 2016, 03:03:16 PM`

### Steps to reproduce
1. `vagrant up`

sethvargo:
Hi there,

It looks like the command `/bin/ip` returned:

```
Wednesday, 20 July 2016, 03:11:59 PM
Linux 4.3.3-3-ARCH x86_64 GNU/Linux
```

That seems odd. What is the output of `ip` on your system?

rafamanzo:
Thanks for the response. Really odd.

`ip` output looks no different than usual on the host (I've removed my interface's MAC from the output):

```shell
.-(~)---------------------------------------------------------(manzo@WALL-A)-
`--> ip
Usage: ip [ OPTIONS ] OBJECT { COMMAND | help }
       ip [ -force ] -batch filename
where  OBJECT := { link | address | addrlabel | route | rule | neigh | ntable |
                   tunnel | tuntap | maddress | mroute | mrule | monitor | xfrm |
                   netns | l2tp | fou | tcp_metrics | token | netconf }
       OPTIONS := { -V[ersion] | -s[tatistics] | -d[etails] | -r[esolve] |
                    -h[uman-readable] | -iec |
                    -f[amily] { inet | inet6 | ipx | dnet | mpls | bridge | link } |
                    -4 | -6 | -I | -D | -B | -0 |
                    -l[oops] { maximum-addr-flush-attempts } | -br[ief] |
                    -o[neline] | -t[imestamp] | -ts[hort] | -b[atch] [filename] |
                    -rc[vbuf] [size] | -n[etns] name | -a[ll] | -c[olor]}
.-(~)---------------------------------------------------------(manzo@WALL-A)-
`--> ip addr
[...]

PR #7676, opened 2016-08-03
Filter command output in SSH communicator

chrisroberts:
Fixes issues reported in #7613. Filters command stdout output to remove any output that may be been previously generated while setting up the shell. This also adds test coverage to the SSH communicator.

### DIFF
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -19,8 +19,12 @@ module VagrantPlugins
   module CommunicatorSSH
     # This class provides communication with the VM via SSH.
     class Communicator < Vagrant.plugin("2", :communicator)
+      # Marker for start of PTY enabled command output
       PTY_DELIM_START = "bccbb768c119429488cfd109aacea6b5-pty"
+      # Marker for end of PTY enabled command output
       PTY_DELIM_END = "bccbb768c119429488cfd109aacea6b5-pty"
+      # Marker for start of regular command output
+      CMD_GARBAGE_MARKER = "41e57d38-b4f7-4e46-9c38-13873d338b86-vagrant-ssh"
 
       include Vagrant::Util::ANSIEscapeCodeRemover
       include Vagrant::Util::Retryable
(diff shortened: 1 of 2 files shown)


plugins/communicators/ssh/communicator.rb
```diff
@@ -21,6 +21,7 @@ module CommunicatorSSH
     class Communicator < Vagrant.plugin("2", :communicator)
       PTY_DELIM_START = "bccbb768c119429488cfd109aacea6b5-pty"
       PTY_DELIM_END = "bccbb768c119429488cfd109aacea6b5-pty"
+      CMD_GARBAGE_MARKER = SecureRandom.uuid.freeze
```
sethvargo: I think we might want to actually hard code this just like we do the PTY delimiters. As written, each vagrant invocation will get a different UUID, which might not be desirable, especially for debugging.

mitchellh: I agree. I recommended the UUID, but instead of generating it on every invocation I'd grab one from an irb shell or something and paste it in here. For additional help with recognizability I usually prefix it with "vagrant ssh marker: " or something.

mitchellh: Also I'd add a comment above this about what this is for

mitchellh:
Tests look good, one note but otherwise looks great.

chrisroberts:
Updated to static constant marker and tossed in comments for the new marker as well as the existing pty constants.

test/unit/plugins/communicators/ssh/communicator_test.rb
```diff
@@ -0,0 +1,287 @@
...
+      and_yield(nil, nil, command_stderr_data)
+    allow(machine).to receive(:ssh_info).and_return(host: '10.1.2.3', port: 22)
+    expect(channel).to receive(:exec).with(core_shell_cmd).
+      and_yield(command_channel, '').and_return channel
+    expect(command_channel).to receive(:on_request).with('exit-status').
+      and_yield(nil, exit_data)
+  end
+
+  describe ".wait_for_ready" do
+    before(&connection_setup)
```
sethvargo: I've never seen procs used like this in rspec, but I kinda like it

sethvargo:
@chrisroberts looks like the test failures are due to a gem which updated the required Ruby version 😦

chrisroberts:
@sethvargo Yeah, all the new PRs are failing now :disappointed:

sethvargo:
@chrisroberts it might be a good idea to open a new PR off master that locks that gem to an older version that doesn't require a newer ruby. We can explore upgrading Ruby to latest closer to the release

chrisroberts:
:+1:

mitchellh:
LGTM!

PR #7677, opened 2016-08-03
Handle DragonFly BSD as FreeBSD.

rickard-von-essen:
Add support for DragonFly BSD.
https://www.dragonflybsd.org/


chrisroberts:
Detection provided in #7701. Thanks for the PR!

PR #7679, opened 2016-08-03
Use comma to set off nonrestrictive clause.

jeveleth:


eyoung100:
Was just writing a vagrant file, and ran across this.  To fix it, see http://stackoverflow.com/questions/38799829/rack-requires-ruby-version-2-2-2

In short, change `"rvm": "2.2.3",` to `"rvm": "2.2.5",`

PR #7684, opened 2016-08-04
Constrain ruby_dep gem versions to prevent constraints on ruby versions

chrisroberts:
Prevents recent version of ruby_dep from being used to satisfy the listen gem dependency as it requires a minimum ruby version of 2.2.5


mitchellh:
LGTM

dcosson:
Any plans for a new vagrant release that incorporates this change? I'm running into issues installing a vagrant plugin, it complains about ruby_dep 1.4.0 not being compatible with ruby version below 2.2.5. It looks like this PR addresses that.

PR #7688, opened 2016-08-07
Remove --interactive flag from Docker command

hunterboerner:
See: https://github.com/mitchellh/vagrant/issues/7597.

Due to the childprocess gem setting STDIN to a pipe, Docker will fail as it requires STDIN to be a TTY if run with --interactive.


chrisroberts:
@hunterboerner can you rebase this please? Thanks!

hunterboerner:
There you go.

chrisroberts:
Awesome. Thanks @hunterboerner !

PR #7693, opened 2016-08-08
util/network_ip: Simplify #network_address helper

legal90:
`Vagrant::Util::NetworkIP#network_address` could be simplified significantly.

In fact, `IPAddr#mask` method supports string and integer arguments as well, so this PR also allows to set IPv6 masks in string notation, like `"ffff:ffff:ffff:ffff::"`


legal90:
@chrisroberts Could you please review it?

legal90:
Rebased

legal90:
I believed it will be finally merged! Thank you! 😉

PR #7699, opened 2016-08-10
provider/virtualbox: Refactor host-only network settings

legal90:
I've refactored `VagrantPlugins:: ProviderVirtualBox::Action:: Network#hostonly_config` in order to make it more simpler, readable and safer to use:

* Simplified calculations of IP and DHCP settings. Handle IPs as `IPAddr` instances, not strings.
Use `IPAddr` methods instead of `Vagrant::Util::NetforkIP#network_address` and string splitting.

* Added basic validation of IP settings. If the specified IP is invalid or netmask could not be used with that IP family, then the human-readable exception will be raised. Currently, it is done only for `private_network` (aka host-only), but the validation could be moved to the global scope if you think it's reasonable. Examples:

```
# Vagranttile
config.vm.network "private_network", ip: "invalid-ip", netmask: '255.255.0.0'


# `vagrant up` output:
Network settings specified in your Vagrantfile are invalid:

Network settings: {:auto_config=>true, :mac=>nil, :nic_type=>nil, :type=>:static, :ip=>"invalid-ip", :netmask=>"255.255.0.0", :protocol=>"tcp", :id=>"d05cca65-6cb5-46f1-ba07-c891923e3219"}
Error: invalid address
```

```
# Vagranttile
config.vm.network "private_network", ip: "dead:beef::", netmask: '255.255.0.0'


# `vagrant up` output:
Network settings specified in your Vagrantfile are invalid:

Network settings: {:auto_config=>true, :mac=>nil, :nic_type=>nil, :type=>:static6, :ip=>"dead:beef::", :netmask=>"255.255.0.0", :protocol=>"tcp", :id=>"dddf6dea-0ee7-4afc-9d93-7dbd297661f6"}
Error: address family is not same
```

Existing unit tests for this class are passed. I've also verified it with manual tests on real VMs.


legal90:
@chrisroberts Could you please review this?

legal90:
I've rebased the branch and fixed a merge conflict.

legal90:
Rebased

PR #7701, opened 2016-08-10
guests/bsds: shutdown fixes and dragonfly detection

chrisroberts:
Adds guest detection for DragonFly BSD. Removes Darwin from matching base BSD guest. Provide custom halt behavior for OpenBSD which requires `-h` option to be used on versions prior to 5.7.

http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sbin/shutdown/shutdown.8?rev=1.40&content-type=text/x-cvsweb-markup
http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sbin/shutdown/shutdown.c?rev=1.40&content-type=text/x-cvsweb-markup

### DIFF
--- a/plugins/guests/bsd/guest.rb
+++ b/plugins/guests/bsd/guest.rb
@@ -2,7 +2,7 @@ module VagrantPlugins
   module GuestBSD
     class Guest < Vagrant.plugin("2", :guest)
       def detect?(machine)
-        machine.communicate.test("uname -s | grep -i 'Darwin|BSD'")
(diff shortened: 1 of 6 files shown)


mitchellh:
Looks great to me thanks for solving the mysteries of the BSDs.

As a verification: does Darwin implement all the same guest capabilities? Just making sure we aren't losing any guest caps by not matching the BSD anymore.

chrisroberts:
Inherits capabilities from BSD so all good :+1:

rickard-von-essen:
Nice! :+1:

PR #7703, opened 2016-08-11
ssh-agent interaction fix

chrisroberts:
This is a continuation of #7621 to include a couple cleanups and rsync fix before merge.


zmt:
What's the eta on the next release milestone that will include this?

Issue #7662, opened 2016-07-29
virtualbox shared folder options wrongly used


vpereira:
I'm running my guest on opensuse, where we have user vagrant, group users. So as synced_folder parameters I must set the parameters as described here: https://github.com/mitchellh/vagrant/blob/f8439869e7003255593093d3c7a2e87695368118/website/source/docs/synced-folders/basic_usage.html.md

I set the owner and group in my Vagrantfile as

````
  config.vm.synced_folder ".", "/vagrant", group:"users", owner: "vagrant"
````

and then the output from vagrant up looks like:

````
  set -e
  # looks good
  mount -t vboxsf -o uid=`id -u vagrant`,gid=`getent group users | cut -d: -f3` vagrant /vagrant
  # its wrong, here gid should be gid=`id -g vagrant`
  # since its asking the gid from the user. gid from the group doesnt make sense
  mount -t vboxsf -o uid=`id -u vagrant`,gid=`id -g users` vagrant /vagrant
````

looking the source code in:

https://github.com/mitchellh/vagrant/blob/1277f192514910f01ecd71d9acd5383867ce3372/plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb#L27

it should be:

````
mount_gid_old = "`id -g #{options[:owner]}
````
right?

vpereira:
funny thing is that with the ````set -e```` vagrant shouldn't exit with one of both commands fail. However for me,  if one of both commands fail, vagrant aborts. I'm able to ````vagrant ssh```` it and see my sync folder mounted. However because of the abortion the subsequent events (i.e provisioning) doesn't happens.. any idea?

chrisroberts:
Fixed via #7720. Thanks for the report!

PR #7720, opened 2016-08-15
Only use effective group when appropriate

chrisroberts:
Current usage of `id -g` incorrectly uses group name instead of owner name. Pulling the effective group of the user should only be used in cases where a special group has not been requested. This issue was referenced in https://github.com/mitchellh/vagrant/issues/7616#issuecomment-239135781 and an issue was created #7662

This PR adjusts the use of the effective group to only happen when the group matches the user and the group is not found.

Additional updates: 

* allow owner and group IDs to be provided as integers or strings
* tests 

_NOTE: Replaces #7683_

### DIFF
--- a/plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
+++ b/plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
@@ -6,49 +6,76 @@ module VagrantPlugins
   module GuestLinux
     module Cap
       class MountVirtualBoxSharedFolder
+        @@logger = Log4r::Logger.new("vagrant::guest::linux::mount_virtualbox_shared_folder")
+
         extend Vagrant::Util::Retryable
 
         def self.mount_virtualbox_shared_folder(machine, name, guestpath, options)
           guest_path = Shellwords.escape(guestpath)
 
-          mount_commands = ["set -e"]
+          @@logger.debug("Mounting #{name} (#{options[:hostpath]} to #{guestpath})")
 
-          if options[:owner].is_a? Integer
+          if options[:owner].to_i.to_s == options[:owner].to_s
             mount_uid = options[:owner]
+            @@logger.debug("Owner user ID (provided): #{mount_uid}")
           else
-            mount_uid = "`id -u #{options[:owner]}`"
+            output = {stdout: '', stderr: ''}
+            uid_command = "id -u #{options[:owner]}"
+            machine.communicate.execute(uid_command,
+              error_class: Vagrant::Errors::VirtualBoxMountFailed,
+              error_key: :virtualbox_mount_failed,
+              command: uid_command,
+              output: output[:stderr]
+            ) { |type, data| output[type] << data if output[type] }
+            mount_uid = output[:stdout].chomp
+            @@logger.debug("Owner user ID (lookup): #{options[:owner]} -> #{mount_uid}")
           end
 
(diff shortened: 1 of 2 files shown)


plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
```diff
@@ -6,62 +6,85 @@ module VagrantPlugins
...
       class MountVirtualBoxSharedFolder
+        @@logger = Log4r::Logger.new("vagrant::guest::linux::mount_virtualbox_shared_folder")
+
         extend Vagrant::Util::Retryable
 
         def self.mount_virtualbox_shared_folder(machine, name, guestpath, options)
           guest_path = Shellwords.escape(guestpath)
 
-          mount_commands = ["set -e"]
+          @@logger.debug "Mounting #{name} (#{options[:hostpath]} to #{guestpath})"
```
mitchellh: Sorry for this nitpick since it may not jive with your style but I'm strongly against parenthesis-less function calls. I'm sure there are some littered throughout Vagrant mostly through contributions but stylistically I just find it easier to scan (physically) lines that have parens to know when they end. 

Are you comfortable with that?

mitchellh:
Looks great, great tests.

plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
chrisroberts: :+1:

cincodenada:
Awesome, thanks for this! Saved my ass today, upgraded Vagrant and ran into this.

For those who want a quick fix: just copying `mount_virtualbox_shared_folder.rb` from this PR ([direct link to raw file](https://raw.githubusercontent.com/chrisroberts/vagrant/dd6ad2fbf3bc935dbb127bf5552dcc5990b8e3e6/plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb)) to the corresponding place in my Vagrant plugins directory (`/opt/vagrant/embedded//gems/gems/vagrant-1.8.5/plugins/guests/linux/cap/` in my case on Arch) fixed it right up. 
@chrisroberts'

plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
```diff
@@ -6,62 +6,85 @@ module VagrantPlugins
...
-          # First mount command uses getent to get the group
-          mount_options = "-o uid=#{mount_uid},gid=#{mount_gid}"
-          mount_options += ",#{options[:mount_options].join(",")}" if options[:mount_options]
-          mount_commands << "mount -t vboxsf #{mount_options} #{name} #{guest_path}"
-
-          # Second mount command uses the old style `id -g`
-          mount_options = "-o uid=#{mount_uid},gid=#{mount_gid_old}"
-          mount_options += ",#{options[:mount_options].join(",")}" if options[:mount_options]
-          mount_commands << "mount -t vboxsf #{mount_options} #{name} #{guest_path}"
+          mount_options = options.fetch(:mount_options, [])
```
dol: The change of the folder option ordering introduces the issue related to #7859.

E.g:
Old ordering
`mount -t vboxsf -o uid=1001,gid=1001,uid=201,gid=201 foo /foo`

New ordering
`mount -t vboxsf -o uid=201,gid=201,uid=1001,gid=1001 foo /foo`

PR #7721, opened 2016-08-16
commands/up: don't instantiate machines while installing providers

tchebb:
In `install_providers()`, we previously used `with_target_vms()` to iterate through and regex expand the command line arguments, determine which environment applied to each specified VM, and figure out which provider the core would use for that VM so that we could install it.

`with_target_vms()` has unintentional side-effects, however: it calls `@env.machine()` on each machine it parses from the command line, which causes any machines that don't exist to be created. However, since no provider has been determined yet, these machines are instantiated with the default provider, paying no respect to providers passed to `up` using `--provider`. This means that, if the first provider in the defaults list happens to fail to instantiate, Vagrant reports an error even when the user has explicitly specified a different provider to use.

A possible fix would have been to pass through `--provider` to `with_target_vms()`. However, a cleaner solution is to simply not instantiate any machines while installing providers. A `Machine` object is not required to get the current environment's configuration values, so creating one is unnecessary and inefficient. Instead, this patch splits the argument parsing loop from `with_target_vms()` into a separate function which yields machine names instead of `Machine` objects. Using the new function, `install_providers()` can determine each machine's default provider without creating a `Machine` object and invoking potentially-unrelated providers.

It's not hard to hit this bug during normal usage. The way I discovered it was as follows:
1. Start with a Linux machine with Vagrant installed but not VirtualBox or Docker.
2. Use `vagrant init` to create a Vagrantfile, then realize you can't `up` it because no providers are compatible.
3. Install the DigitalOcean third-party provider because you plan to eventually deploy to the cloud.
4. Install the `libvirt` provider for local testing (make sure you do this after installing DigitalOcean, so that DigitalOcean is earlier in the defaults list).
5. Attempt to start your VM with `vagrant up`.
6. See a DigitalOcean "unauthorized" error and realize that it's choosing the wrong provider.<sup>1</sup>
7. Attempt to start your VM with `vagrant up --provider libvirt`.
8. Observe that the same DigitalOcean error appears, even though you've explicitly specified a provider.

<sup>1</sup>I realize that this is partially the DigitalOcean provider's fault for not being non-defaultable, and I've already submitted a patch to them to fix that. However, one misbehaving provider shouldn't be able to prevent `up` from respecting its command line parameters.

### DIFF
--- a/lib/vagrant/plugin/v2/command.rb
+++ b/lib/vagrant/plugin/v2/command.rb
@@ -62,13 +62,62 @@ module Vagrant
           raise Errors::CLIInvalidOptions, help: opts.help.chomp
         end
 
+        # Yields the name (as a symbol) of each target VM for the command.
+        #
+        # This method should be used when the targets of a command need to
+        # be resolved to names but the side-effects of actually getting
+        # (and possibly creating) Machine objects from those names is
+        # undesired. For example, when looking up the provider that should
+        # be used to create each machine, we can't yet instantiate each
+        # machine because that requires us to already know the provider.
+        #
+        # @param [String, Array<String>] names Names or regexes of VMs.
+        #   Nil if every VM.
+        def with_matching_names(names=nil)
+          # Require that names be an array
+          names ||= []
+          names = [names] if !names.is_a?(Array)
+
+          if names.length > 0
+            names.each do |name|
+              if pattern = name[/^\/(.+?)\/$/, 1]
+                @logger.debug("Finding machines that match regex: #{pattern}")
+
+                # This is a regular expression name, so we convert to a regular
+                # expression and allow that sort of matching.
+                regex = Regexp.new(pattern)
+
+                @env.machine_names.each do |machine_name|
+                  if machine_name =~ regex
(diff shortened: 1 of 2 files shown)


chrisroberts:
Hi there,

Cheers!

tchebb:
Hey Chris,

I don't believe this issue has been resolved. I could be wrong, since it's been over two years since I wrote this code and about as long since I've used Vagrant. It looks to me, though, like the code this PR affects has barely changed in the intervening time. As such, it seems very likely to me that Vagrant still tries to instantiate machines with a default provider even when `--provider` is explicitly passed.

Please let me know if I'm missing an architectural change that makes this not the case.

chrisroberts:
Hi @tchebb. I believe this commit prevents the behavior you were experiencing: https://github.com/hashicorp/vagrant/commit/7c4252e24a8c475efafef9fb7a24acd887b7d75c#diff-ef0906b42c651d55b265858febfd97be

tchebb:
@chrisroberts Ah yep, that looks like it'd do it. Thanks!

PR #7725, opened 2016-08-16
Machine data directory for base box package

chrisroberts:
Fixes an issue referenced in #5070 by providing a temporary machine data directory. Includes some initial basic tests on the `package` command that includes testing for a valid machine data directory when using `--base` option.


sethvargo:
👍

mitchellh:
You should create the temporary directory in the env's temp dir instance var that it has. People can control this temporary directory env vars and they usually do in the case that they have a larger tmp directory they use for Vagrant and things like that.

mitchellh:
Besides that LGTM

Issue #5256, opened 2015-01-28
rsync-auto sets owner/group in excluded folder


etienneq:
Hi,

when using synced folders with rsync, I've got the following problem. Owner and group are set in excluded folders too. One has to note that I did not exclude the folder itself but only its content.

I would expect that the chown does not traverse into excluded folders.

Synced folders are configured like this:
```
type: "rsync",
owner: "xyz", group: "xyz",
rsync__auto: true,
rsync__exclude: [ ".git/", "myfolder/*" ],
rsync__args: ["--verbose", "--archive", "--delete", "-z"]
```

Cheers
Etienne

ezpuzz:
This is causing issues on our systems, because the `find` command issued to change the permissions of rsync'd files is giving `No such file or directory` errors for `"tmp/"` files in our project directory.

Command:
```
find '/vagrant' '!' -type l -a '(' ! -user vagrant -or ! -group vagrant ')' -print0 | xargs -0 -r chown vagrant:vagrant
```

Error:
```
find: `/vagrant/tmp/custom_static_compiler-tmp_dest_dir-464MbVTn.tmp': No such file or directory
```

Vagrantfile:
```
config.vm.synced_folder ".","/vagrant", type: "rsync", rsync__exclude: ["node_modules", "bower_components", "dist", "tmp/"]
```

Expected behavior:
The find Command above ignores the directories/files passed to `rsync__exclude` in the Vagrantfile such that the Command becomes:
```
find '/vagrant' '!' -type l -a '(' ! -user vagrant -or ! -group vagrant ')' -not \( -path /vagrant/node_modules -prune \) -not \( -path /vagrant/bower_components -prune \) -not \( -path /vagrant/dist -prune \) -not \( -path /vagrant/tmp -prune \) -print0
```

vlopato:
Hi, All!

Any news with this?

niamleeson:
Has this issue been fixed yet? Running into the same issue

`ERROR -- : exception while processing events: The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

find '/vagrant' '!' -type l -a '(' ! -user vagrant -or ! -group vagrant ')' -print0 | xargs -0 -r chown vagrant:vagrant

Stdout from the command:



Stderr from the command:

find: ‘/vagrant/frontend/tmp/build.lock’: No such file or directory
chown: cannot access ‘/vagrant/frontend/tmp/build.lock’: No such file or directory
[...]

PR #7726, opened 2016-08-16
Ignore files excluded from sync on chown

chrisroberts:
This address #5256 by ignoring any files that are excluded from the rsync when changing the owner. This PR includes the initial refactoring of rsync capability implementations and also includes the `:chown` option where it was missing before. Also updates `find | xargs` to POSIX compat `find -exec`.

### DIFF
--- a/plugins/guests/darwin/cap/rsync.rb
+++ b/plugins/guests/darwin/cap/rsync.rb
@@ -1,33 +1,10 @@
-require "shellwords"
+require_relative "../../../synced_folders/rsync/default_unix_cap"
 
 module VagrantPlugins
   module GuestDarwin
     module Cap
       class RSync
-        def self.rsync_installed(machine)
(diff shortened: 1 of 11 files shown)


sethvargo:
LGTM, but curious about why you chose a module instead of a subclass

mitchellh:
Ah, a constant source of pain. Hopefully this gets it this time. Look sgood!

chrisroberts:
@sethvargo Using a module as it was just common method extraction to a shared source. I'm pretty indifferent implementation wise between using a module or a common parent class. Any preference?

sethvargo:
Nope - was just curious 👍

mradcliffe:
I think the change to `find -exec` (a very good change) means there's a hard dependency on the user/group existing on the box on initial provision. I think this bug obscured this and let sync gracefully fail (?) when doing the chown before provision.

PR #7734, opened 2016-08-18
Documenting the need for ruby 2.2 to install listen gem

cole-christensen:
When running `bundle install` before `rake install` you get `Gem::InstallError: listen requires Ruby version >= 2.2.3, ~> 2.2.`


chrisroberts:
This was addressed in #7684. Thanks for the report and PR!

PR #7740, opened 2016-08-19
Add NFSv4 warning and information

chrisroberts:
Include warning when mounting synced folder using NFS with possibly invalid options. Adds note to documentation about NFSv4 and UDP.


mitchellh:
LGTM

PR #7751, opened 2016-08-23
Fix CentOS7 network config - do not prompt before overwriting

wakeful:
`vagrant version - 1.8.5`
`VirtualBox version - 5.1.4r110228`

assuming a scenario - vagrant up -> power off -> vagrant up
vagrant will stuck on the network configuration

`DEBUG ssh: stdout: mv: overwrite '/etc/sysconfig/network-scripts/ifcfg-eth0'?`
`DEBUG ssh: Sending SSH keep-alive...`

### DIFF
--- a/plugins/guests/redhat/cap/configure_networks.rb
+++ b/plugins/guests/redhat/cap/configure_networks.rb
@@ -50,6 +50,11 @@ module VagrantPlugins
             EOH
           end
 
+          commands << <<-EOH.gsub(/^ {14}/, '')
+            # Restart network
+            service network restart
+          EOH
+
           comm.sudo(commands.join("\n"))
         end
       end
--- a/plugins/guests/redhat/cap/configure_networks.rb
+++ b/plugins/guests/redhat/cap/configure_networks.rb
@@ -43,7 +43,7 @@ module VagrantPlugins
               /sbin/ifdown '#{network[:device]}' || true
 
               # Move new config into place
-              mv '#{remote_path}' '#{final_path}'
+              mv -f '#{remote_path}' '#{final_path}'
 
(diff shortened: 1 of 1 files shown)


plugins/guests/redhat/cap/configure_networks.rb
```diff
@@ -43,13 +43,18 @@ def self.configure_networks(machine, networks)
...
               # Bring the interface up
               ARPCHECK=no /sbin/ifup '#{network[:device]}'
             EOH
           end
 
+          commands << <<-EOH.gsub(/^ {14}/, '')
+            # Restart network
+            service network restart
+          EOH
+
```
chrisroberts: Can you explain what the network restart is providing here? Just want to confirm that I understand the reasoning for the addition. Thanks!

lucabelluccini:
Thank you :+1:

plugins/guests/redhat/cap/configure_networks.rb
```diff
@@ -43,13 +43,18 @@ def self.configure_networks(machine, networks)
...
               # Move new config into place
-              mv '#{remote_path}' '#{final_path}'
+              mv -f '#{remote_path}' '#{final_path}'
 
               # Bring the interface up
               ARPCHECK=no /sbin/ifup '#{network[:device]}'
             EOH
           end
 
+          commands << <<-EOH.gsub(/^ {14}/, '')
```
karlkfi: This indent removal doesn't work. There are only 12 spaces, not 14. Looks like it was lazily copied without knowing what it does. It doesn't really make sense to have to specify the exact number of space... should just replace all leading spaces with `/^ */`.

PR #7752, opened 2016-08-24
support ansible for remote servers and vm

pronix:
require for working with ansible and libvirt on remote hosts

### DIFF
--- a/plugins/provisioners/ansible/provisioner/host.rb
+++ b/plugins/provisioners/ansible/provisioner/host.rb
@@ -234,7 +234,7 @@ module VagrantPlugins
             proxy_cmd += " exec nc %h %p 2>/dev/null"
 
             ssh_options << "-o ProxyCommand='#{ proxy_cmd }'"
-          elsif @machine.provider_name == :libvirt && @machine.provider_config.connect_via_ssh
+          elsif @machine.ssh_info[:proxy_command] && @machine.provider_config.connect_via_ssh
             proxy_cmd = @machine.ssh_info[:proxy_command]
             ssh_options << "-o ProxyCommand='#{ proxy_cmd }'"
           end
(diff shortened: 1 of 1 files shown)


gildegoma:
Thank you @pronix for reporting this bug with a fix proposal.

More generally, the Ansible Provisioner should honour the existence of the Vagrant `ssh.proxy_command` [setting](https://www.vagrantup.com/docs/vagrantfile/ssh_settings.html), and this is not tied to any provider. It would be very welcome if you could test the changes I proposed on your side. If it works for you, I'll appreciate if you update this PR. Special Kudos if you can add a unit test to prevent any accidental code removal in the future (and for sure speed up the merge process). Thanks again!

gildegoma:
Hmm... there is apparently no way to edit the new github code review comments, and my previous code snippet should rather be something like:

```ruby
if @machine.ssh_info[:proxy_command] && @machine.provider_config.connect_via_ssh
  proxy_cmd = @machine.ssh_info[:proxy_command]
  ssh_options << "-o ProxyCommand='#{ proxy_cmd }'"
end
```

although I am not very sure that it is necessary to add the `@machine.provider_config.connect_via_ssh` condition. Maybe you can better figure out with the different `libvirt` use cases, for instance...

pronix:
hello
thanks for recommendations.
patch fixed for working in any vm provider.

gildegoma:
@pronix I'm closing this PR since I'v merged your changes after adding the following changes: e8cf9bb16828187412c4afd7b82c2476f09f8579

This shouldn't be a problem since `vagrant-libvirt` correctly set the `ssh_info.proxy_command` setting [only when communicating over SSH](https://github.com/vagrant-libvirt/vagrant-libvirt/blob/a963ff4c70f5e6ce5bc2589b7e98c1c0671e2ccc/lib/vagrant-libvirt/provider.rb#L72-L77). But please verify, and tell me if you face any troubles.

PR #7758, opened 2016-08-26
Add note about using a separate file for sudoers

rquadling:


chrisroberts:
Hi. Thanks for this update to the documentation. Could you add a note to include the caveat of sudoers.d being supported (just to cover the edge case of where it may not be enabled)? Thanks!

chrisroberts:
Thanks @rquadling !

PR #7778, opened 2016-09-05
Allow closing a Vagrant::Util::Subprocess's STDIN

justjake:
Previously, there was no way to close the STDIN stream of a subprocess,
so commands that read from stdin in a subprocess would hang forever,
such as `/bin/sh -s`. If one tried to close the stdin, the
IO.select() call in Subprocess#execute would raise an error for calling
select() on a closed IO.

Here's a concrete example of a command that needs to close STDIN to work
properly:

```ruby
script = SOME_VERY_LONG_STRING
command = %w(ssh foo.example.com /bin/sh -s foo bar)
result = ::Vagrant::Util::Subprocess.execute(*command) do |type, data_or_io|
  if type == :stdin
    data_or_io.write(script)
    data_or_io.write("\n")
    data_or_io.close
    next
  end

  puts "Remote: #{data_or_io}"
end
```

### DIFF
--- a/lib/vagrant/util/subprocess.rb
+++ b/lib/vagrant/util/subprocess.rb
@@ -144,10 +144,11 @@ module Vagrant
         # Record the start time for timeout purposes
         start_time = Time.now.to_i
 
+        open_readers = [stdout, stderr]
+        open_writers = notify_stdin ? [process.io.stdin] : []
         @logger.debug("Selecting on IO")
         while true
-          writers = notify_stdin ? [process.io.stdin] : []
-          results = ::IO.select([stdout, stderr], writers, nil, 0.1)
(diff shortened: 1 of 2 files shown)


chrisroberts:
:+1: This looks good. Could you add some tests to cover the behavior?

justjake:
@chrisroberts My PR originally did not include tests because no tests for `Vagrant::Util::Subprocess` existed.

I have added tests for STDIN-related behavior, especially behavior that I change in this PR. These tests depend on `echo` and `cat` coreutils; and the tests are marked as "pending" unless Vagrant::Util::Which can find these commands.

justjake:
@chrisroberts bump

PR #7793, opened 2016-09-13
[core] Remove bundler usage for plugin management

chrisroberts:
Remove bundler usage for internal plugin management.


lib/vagrant/pre-rubygems.rb
```diff
@@ -1,34 +0,0 @@
...
-
-if defined?(Bundler)
-  require "bundler/shared_helpers"
-  if Bundler::SharedHelpers.in_bundle? && !Vagrant.very_quiet?
-    puts "Vagrant appears to be running in a Bundler environment. Your "
-    puts "existing Gemfile will be used. Vagrant will not auto-load any plugins"
-    puts "installed with `vagrant plugin`. Vagrant will autoload any plugins in"
-    puts "the 'plugins' group in your Gemfile. You can force Vagrant to take over"
-    puts "with VAGRANT_FORCE_BUNDLER."
-    puts
```
gildegoma: I bring my two cents here, after I've just noticed that the plugins auto-loading is no longer disabled when running Vagrant from a Bundler environment. This is "breaking" my Vagrant installation from source, as Vagrant now tries to load all the plugins defined in `~/.vagrant.d/plugins.json`, but fails to find them in such context.

Thanks to the new `VAGRANT_DISABLE_PLUGIN_INIT` environment variable, I can force to disable the plugins initialisation, which is fair enough for working on the Vagrant project itself.

@chrisroberts In order to avoid misunderstanding situations for people wanting to contribute to the Vagrant project, I guess it would probably be worth to

* consider disabling the plugin auto-loading _by default_ when running in a "development" context (or be sure to not conflict with plugins installed in an official Vagrant setup).
* update the related [README.md](https://github.com/mitchellh/vagrant/blob/master/README.md#contributing-to-vagrant) section, especially the note copied below

> **NOTE:** By default running Vagrant via `bundle` will disable plugins.
This is necessary because Vagrant creates its own private Bundler context
(it does not respect your Gemfile), because it uses Bundler to manage plugin
dependencies.

What do you think? Sorry if I missed something obvious, as I haven't taken time to fully understand the changes brought here. I merely wanted to report a "contributor experience" feedback. Hope it helps and makes sense for you...

Ma27:
this seems to break the installation of `vagrant-r10k`.

Whenever I run `vagrant plugin install vagrant-r10k` with Vagrant 1.9 I get this:

``` code
/opt/vagrant/embedded/lib/ruby/2.2.0/rubygems/dependency.rb:315:in `to_specs': Could not find 'faraday_middleware' (>= 0) among 45 total gem(s) (Gem::LoadError)
```

lib/vagrant/pre-rubygems.rb
chrisroberts: @gildegoma Hi! Thanks very much for the feedback. I did miss that section in the README file and will need to update it. But to the bigger issue around the behavior of package installation vs source installation around plugins, I'm really happy to get this feedback. 

I'm trying to make the behavior of Vagrant from a package installation context and a source installation running via Bundler context function in mostly the same way. This unexpected breakage was "expected" from an implementation point of view, but not intentional from a usability point of view for contributors. My apologies!

I would like for plugins to remain auto loading regardless of the running context (so I'll definitely update that portion of the README, and the relevant section in the website docs). I am curious on thoughts around context specific plugins files and installations. It was something I was considering originally, and perhaps might be worthwhile. This would allow switching from a package installation context to a source installation context without needing to run a `plugin repair` or `plugin expunge --reinstall`.

Any more thoughts/feedback is very welcome. Cheers!

gabriel403:
I get something similar with `vagrant-librarian-puppet`

```
/opt/vagrant/embedded/lib/ruby/2.2.0/rubygems/dependency.rb:315:in `to_specs': Could not find 'faraday' (< 1.0, >= 0.7.4) among 45 total gem(s) (Gem::LoadError)
```

Ma27:
yeah, it seems as the new installation destruct all vagrant plugins if they use runtime dependencies (both rely on the faraday gem)

Ma27:
ping @chrisroberts is this reproducable from your side?

chrisroberts:
@Ma27 yep, and has been fixed in #8068

Ma27:
cool, thank you :)

PR #7802, opened 2016-09-15
[windows] Use subprocess for safe_exec on windows

chrisroberts:
Detects Windows platform within `SafeExec.exec` and uses `Subprocess` instead of `fork`. Related to #6938. 

I am curious if there's actually a need for two code paths here. Always running through `Subprocess` when using `Kernel.exec` directly fails seems like a more consistent implementation. Any thoughts?


chrisroberts:
@sethvargo updated

PR #7813, opened 2016-09-20
Fix non upstart linux nfs mounts

rudineirk:
I ran into a very weird bug when trying to run openwrt (x86) machines on vagrant, they locked when mounting nfs, even when it mounted successfully.

I discovered it was because the `/sbin/init` binary in openwrt ignores the `--version` parameter and open a menu, stoping the box creation process.

This fix should work for any non upstart (and the ones that use upstart) Linux distros.


chrisroberts:
:+1: Thanks!

PR #7836, opened 2016-09-27
add the `salt_call_args` option

jarondl:
This option allows you to pass additional arguments to `salt-call`
during state execution, (similar to `install_args` in installation).
This is useful, for example, for `file-root` and `pillar-root`.

I have also taken the opportunity to reorganize the option string building of
`salt-call` in a similar way to `install`. Instead of multiple methods
that were called once, there is one simple method.

### DIFF
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -29,6 +29,7 @@ module VagrantPlugins
       attr_accessor :log_level
       attr_accessor :masterless
       attr_accessor :minion_id
+      attr_accessor :salt_call_args
 
       ## bootstrap options
       attr_accessor :temp_config_dir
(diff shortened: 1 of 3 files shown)


aneeshusa:
I'm not a fan of having 'salt_call_args' as a single string - it would be better to pass an array to avoid having to guess at word splitting. Even better would be explicit `file_root` and `pillar_root` arguments in the Vagrant Salt provisioner (for our use case in servo/saltfs, where we need to pass those parameters).

jarondl:
My first version had explicit `file_root` and `pillar_root`, but then I looked at https://docs.saltstack.com/en/latest/ref/cli/salt-call.html and saw the amount of other options.
I do not care this way or the other, both are rather simple to implement.

briancain:
Closing in favor of #8927

PR #7712, opened 2016-08-13
fix #7489 preventing winrm connection leakage

mwrock:
This ensures that all winrm commands open a new connection and close it after the command. While the previous shell code is "mostly" leak proof, it keeps a connection open for each `WinRMShell` instance and does not close them until the vagrant process exits. Thats generally a more efficient approach and reduces connection startup overhead for multiple commands. In the case of Windows 2008 R2, the default maximum open shells is 5. So if one uses multiple vagrant plugins, it may be easy to exceed this maximum and errors will ensue.

The cost savings of openning new connections really does not outweigh the error scenarios possible here on Windows 2008R2 (and Windows 7).

PR #7837, opened 2016-09-28
fix winrm connection leak

sambarrie:


chrisroberts:
This was addressed and fixed via #7712. Thank you for the PR!

PR #7848, opened 2016-09-29
guests/linux: Place ethernet devices at start of device list

### LINKED ISSUE #7844: 1.8.6: Vagrant private network is incorrectly configured due to sorting of interface names if Docker is installed
### Vagrant version
1.8.6

### Host operating system
ArchLinux

### Guest operating system
CentOS 7

### Debug output
```
DEBUG network_interfaces: Unsorted list: ["enp0s3", "enp0s8", "docker0"]
DEBUG network_interfaces: Sorted list: ["docker0", "enp0s3", "enp0s8"]
DEBUG ssh: Uploading: /var/folders/7b/7mhr441d32q9t25f7r_c30h00000gp/T/vagrant-redhat-configure-networks20160928-13678-1s0qhn4 to /tmp/vagrant-network-entry-enp0s3-1475102022-0
DEBUG ssh: Re-using SSH connection.
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute: # Down the interface before munging the config file. This might
# fail if the interface is not actually set up yet so ignore
# errors.
/sbin/ifdown 'enp0s3' || true

# Move new config into place
mv '/tmp/vagrant-network-entry-enp0s3-1475102022-0' '/etc/sysconfig/network-scripts/ifcfg-enp0s3'

# Bring the interface up
ARPCHECK=no /sbin/ifup 'enp0s3'
 (sudo=true)
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute: echo; printf $SSH_AUTH_SOCK (sudo=false)
DEBUG ssh: Exit status: 0
 INFO ssh: Setting SSH_AUTH_SOCK remotely: /tmp/ssh-s1ZuUuAtJS/agent.6490
DEBUG ssh: Sending SSH keep-alive...
DEBUG ssh: Sending SSH keep-alive...
DEBUG ssh: Sending SSH keep-alive...
DEBUG ssh: Sending SSH keep-alive...
```

### Expected behavior
Vagrant should determine the adapter to network interface mapping correctly when Docker is installed and configure the correct interface as part of the private network.

### Actual behavior
The change in https://github.com/mitchellh/vagrant/pull/7705 can cause Vagrant to incorrectly map adapters to interfaces if a private network is configured by Vagrant and Docker is installed on the Vagrant box. This results in `vagrant up` hanging indefinitely.

I believe this is what is happening (I might have diagnosed the root cause incorrectly and I may have butchered some of the terminology... sorry):
- With the Virtualbox provider plugin, Vagrant creates a dictionary of network adapters which includes a NAT adapter by default (see https://github.com/mitchellh/vagrant/blob/master/plugins/providers/virtualbox/config.rb#L80). 
- Much later the Virtualbox provider plugin maps the adapters to the interfaces based on their indexes (see https://github.com/mitchellh/vagrant/blob/master/plugins/providers/virtualbox/config.rb#L80). For example, the 2nd adapter gets mapped to 2nd interface. 
- Previously the order of the interfaces was based on the order they appeared in the `ip addr` command. The interfaces are now ordered alphabetically.
- The first interface is always mapped to the NAT adapter 
- If the VM has an interface with a name that is alphabetically ordered before the actual interface that corresponds to the NAT adapter Vagrant incorrectly maps that interface to the NAT adapter.
- The interface corresponding to the private network is then mapped to the interface corresponding to the NAT adapter and everything goes down hill from there.

For example:

Previously the list of network interfaces would be something like:
`["enp0s3", "enp0s8", "docker0"]`

The first interface (`enp0s3`) was mapped to the NAT adapter and the second interface (`enp0s8`) was mapped to the adapter connected to the private network.

The list of network interfaces is now sorted so the list is now:
`["docker0", "enp0s3", "enp0s8"]`

The `docker0` interface is now mapped to the NAT adapter and `enp0s3` is now mapped to the adapter connected to the private network. When Vagrant overrides the configuration for the `enp0s3` interface with the configuration meant for `enp0s8`, the VM is unable to reach the outside.

`docker0` is created and configured by Docker by default. I think I might be able to work around the problem by configuring docker to create the interface with a different name. I think running Docker inside a Vagrant box might be a common use case though?

### Steps to reproduce
1. Download a Vagrant box with Docker installed
2. Configure the Vagrantfile to create a private network
3. `vagrant up` hangs indefinitely

### References
- https://github.com/mitchellh/vagrant/issues/7668

chrisroberts:
Always provide sorted ethernet devices at the start of the network devices list.

Fixes #7844

### DIFF
--- a/plugins/guests/linux/cap/network_interfaces.rb
+++ b/plugins/guests/linux/cap/network_interfaces.rb
@@ -2,6 +2,11 @@ module VagrantPlugins
   module GuestLinux
     module Cap
       class NetworkInterfaces
+        # Valid ethernet device prefix values.
+        # eth - classic prefix
+        # en  - predictable interface names prefix
+        ETHERNET_PREFIX = ["eth", "en"]
+
         @@logger = Log4r::Logger.new("vagrant::guest::linux::network_interfaces")
 
         # Get network interfaces as a list. The result will be something like:
@@ -15,10 +20,16 @@ module VagrantPlugins
             s << data if type == :stdout
           end
           ifaces = s.split("\n")
+          eth_prefix = nil
           @@logger.debug("Unsorted list: #{ifaces.inspect}")
           # Break out integers from strings and sort the arrays to provide
           # a natural sort for the interface names
           ifaces = ifaces.map do |iface|
+            unless eth_prefix
+              eth_prefix = ETHERNET_PREFIX.detect do |prefix|
+                iface.start_with?(prefix)
+              end
+            end
             iface.scan(/(.+?)(\d+)/).flatten.map do |iface_part|
               if iface_part.to_i.to_s == iface_part
                 iface_part.to_i
@@ -28,6 +39,13 @@ module VagrantPlugins
             end
           end.sort.map(&:join)
           @@logger.debug("Sorted list: #{ifaces.inspect}")
+          # Extract ethernet devices and place at start of list
+          if eth_prefix
+            eth_start = ifaces.index{|iface| iface.start_with?(eth_prefix) }
+            eth_end = ifaces.rindex{|iface| iface.start_with?(eth_prefix) }
+            ifaces.unshift(*ifaces.slice!(eth_start, eth_end - 1))
+            @@logger.debug("Ethernet preferred sorted list: #{ifaces.inspect}")
+          end
           ifaces
         end
       end
--- a/plugins/guests/linux/cap/network_interfaces.rb
+++ b/plugins/guests/linux/cap/network_interfaces.rb
@@ -5,7 +5,7 @@ module VagrantPlugins
         # Valid ethernet device prefix values.
         # eth - classic prefix
         # en  - predictable interface names prefix
-        ETHERNET_PREFIX = ["eth", "en"]
+        POSSIBLE_ETHERNET_PREFIXES = ["eth".freeze, "en".freeze].freeze
 
         @@logger = Log4r::Logger.new("vagrant::guest::linux::network_interfaces")
 
@@ -25,8 +25,8 @@ module VagrantPlugins
           # Break out integers from strings and sort the arrays to provide
           # a natural sort for the interface names
           ifaces = ifaces.map do |iface|
-            unless eth_prefix
-              eth_prefix = ETHERNET_PREFIX.detect do |prefix|
+            if eth_prefix.nil?
+              eth_prefix = POSSIBLE_ETHERNET_PREFIXES.detect do |prefix|
                 iface.start_with?(prefix)
               end
             end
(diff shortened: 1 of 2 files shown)


plugins/guests/linux/cap/network_interfaces.rb
```diff
@@ -2,6 +2,11 @@ module VagrantPlugins
   module GuestLinux
     module Cap
       class NetworkInterfaces
+        # Valid ethernet device prefix values.
+        # eth - classic prefix
+        # en  - predictable interface names prefix
+        ETHERNET_PREFIX = ["eth", "en"]
```
sethvargo: We should freeze this.

sethvargo: And possibly name it something more descriptive like `POSSIBLE_ETHERNET_PREFIXES`

test/unit/plugins/guests/linux/cap/network_interfaces_test.rb
```diff
@@ -44,5 +44,18 @@
...
+      result = cap.network_interfaces(machine)
+      expect(result).to eq(["eth0", "eth1", "eth2", "bridge0", "docker0"])
+    end
+
+    it "sorts ethernet devices discovered with predictable network interfaces naming first in list" do
+      expect(comm).to receive(:sudo).and_yield(:stdout, "enp0s8\ndocker0\nenp0s3\nbridge0\nenp0s5")
+      result = cap.network_interfaces(machine)
+      expect(result).to eq(["enp0s3", "enp0s5", "enp0s8", "bridge0", "docker0"])
+    end
+
```
sethvargo: Extra newline here

plugins/guests/linux/cap/network_interfaces.rb
```diff
@@ -15,10 +20,16 @@ def self.network_interfaces(machine, path = "/sbin/ip")
             s << data if type == :stdout
           end
           ifaces = s.split("\n")
+          eth_prefix = nil
           @@logger.debug("Unsorted list: #{ifaces.inspect}")
           # Break out integers from strings and sort the arrays to provide
           # a natural sort for the interface names
           ifaces = ifaces.map do |iface|
+            unless eth_prefix
```
sethvargo: I know it's Ruby, but neither @mitchellh or I are a fan of the "unless" and much prefer `if !` or `if .nil?` since it mentally maps better to the Golang folks who might look at this code in some future.

chrisroberts:
@sethvargo updated

lhernanz:
Hi team,
after applying the new version of plugins/guests/linux/cap/network_interfaces.rb to an stock 1.6 version, the original problem is indeed fixed (the configuration process completes), but the algorithm is not working as intended. This is what I get:

DEBUG network_interfaces: Unsorted list: ["enp0s3", "enp0s8", "docker0"]
DEBUG network_interfaces: Sorted list: ["docker0", "enp0s3", "enp0s8"]
DEBUG network_interfaces: Ethernet preferred sorted list: ["enp0s3", "docker0", "enp0s8"]

The outcome is that the private network configuration is applied to docker0 instead of enp0s8.

Regards.

chrisroberts:
@lhernanz Hi and thanks for the update! There have been a few other cases pointed out as well that need some adjustment. I'll be pushing out a new PR with expanded specs and an update shortly. Cheers!

Issue #7706, opened 2016-08-12
Vagrant not compatible with HyperV export from Windows 10 / Server 2016 + Extraction Bug


phawxby:
### Description

**Updated**
This issue appears to cover two bugs:
1. Vagrant is not compatible with exports from HyperV on Windows 10 and Server 2016 after [changes to the config format by Microsoft](https://technet.microsoft.com/en-us/windows-server-docs/compute/hyper-v/what-s-new-in-hyper-v-on-windows#a-name-bkmk-config-a-virtual-machine-configuration-file-format). 
2. There is a problem with the version of bsdtar included with Vagrant causing boxes packaged with zip not to extract properly. This has previously been referenced in GH-7068

My rambling comments below show me figuring this out.

**Old**
~~Either there is a bug within the vagrant box unpackaging or the [instructions](https://www.vagrantup.com/docs/hyperv/boxes.html#packaging-the-box) on how to package a HyperV box are incorrect. I cannot find a way to package my box without it generating errors during unpackaging. I have also found other HyperV boxes with the same unpackaging issues.~~

```
paul@CRAZYIVAN:/mnt/s/Virtual Machines/Trusty-Tahr/Export$ tar cvzf ../Trusty-Tahr.box ./*
./metadata.json
./Vagrantfile
./Virtual Hard Disks/
./Virtual Hard Disks/Trusty Tahr.vhdx
./Virtual Machines/
./Virtual Machines/0FD24F92-0B85-45A9-A275-21CAAA2C822F.vmcx
./Virtual Machines/0FD24F92-0B85-45A9-A275-21CAAA2C822F.VMRS
paul@CRAZYIVAN:/mnt/s/Virtual Machines/Trusty-Tahr/Export$
```
### Vagrant version

```
C:\Users\z269cjbb\Documents\GitHub\xerox-drupal>vagrant -v
Vagrant 1.8.5
```
### Host operating system

Windows 10 Enterprise x64. Version 1607, build 14393.51
### Guest operating system

Ubuntu 14 x64
### Vagrantfile

``` ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :
memory = 4096
cpus = 2

Vagrant.configure(2) do |config|
  config.vm.box = "phawxby/trusty64"
  # config.vm.box = "withinboredom/Trusty64" # A different broken box

  config.vm.provider "hyperv" do |hv|
    hv.memory = memory
    hv.cpus = cpus
  end
end
```
### Debug output

[...]

Issue #6102, opened 2015-08-06
VMCX file format 


Novakov:
I'm not sure where I should create issue regarding Hyper-V provider, so this is the same issue as https://github.com/MSOpenTech/vagrant-windows-hyperv/issues/73

After installing Windows 10 I wanted to play with vagrant and hyper-v. After creating a box, doing `vagrant up` failed with error:

```
C:/HashiCorp/Vagrant/embedded/gems/gems/i18n-0.7.0/lib/i18n/config.rb:92:in `block in missing_interpolation_argument_handler': missing interpolation argument :name in "The box you're using with the Hyper-V provider ('%{name}')\nis invalid. A Hyper-V box should contain both a\n\"Virtual Machines\" and a \"Virtual Hard Disks\" folder that are\ncreated as part of exporting a Hyper-V machine.\n\nWithin these directories, Vagrant expects to find the\nvirtual machine configuration as well as the root hard disk.\n\nThe box you're attempting to use is missing one or both of\nthese directories or does not contain the files expected. Verify\nthat you added the correct box. If this problem persists,\nplease contact the creator of the box for assistance." ({:_key=>:box_invalid, :_namespace=>"vagrant_hyperv.errors"} given) (I18n::MissingInterpolationArgument)
        from C:/HashiCorp/Vagrant/embedded/gems/gems/i18n-0.7.0/lib/i18n/interpolate/ruby.rb:29:in `call'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/i18n-0.7.0/lib/i18n/interpolate/ruby.rb:29:in `block in interpolate_hash'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/i18n-0.7.0/lib/i18n/interpolate/ruby.rb:21:in `gsub'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/i18n-0.7.0/lib/i18n/interpolate/ruby.rb:21:in `interpolate_hash'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/i18n-0.7.0/lib/i18n/interpolate/ruby.rb:17:in `interpolate'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/i18n-0.7.0/lib/i18n/backend/base.rb:152:in `interpolate'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/i18n-0.7.0/lib/i18n/backend/base.rb:41:in `translate'
[...]

PR #7854, opened 2016-10-02
Add support for vmcx (Hyper-V binary config format) to the Hyper-V import script

bbrala:
Added an import script that reads the new vmcx format with powershell. It uses native commands to read the format and create the machine the same way as it does right now with the XML's. This way there is no need anymore to hack an XML to get Hyper-V boxes to work.

As Hyper-V support seems to get close in packer, we really need this small feature added so we can start using those boxes as quickly as possible.

This should fix: #7706 and #6102

### DIFF
--- a/plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
+++ b/plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
@@ -22,7 +22,6 @@
 $Dir = Split-Path $script:MyInvocation.MyCommand.Path
 . ([System.IO.Path]::Combine($Dir, "utils\write_messages.ps1"))
 
-# load the config from the vmcx and make a copy for editing, use TMP path so we are sure there is no vhd at the destination
 $VmProperties = @{
     Path = $vm_config_file 
     SnapshotFilePath   = Join-Path $data_path 'Snapshots'
@@ -31,7 +30,6 @@ $VmProperties = @{
 }
 
 $vmConfig = (Compare-VM -Copy -GenerateNewID @VmProperties)
-
 
 $generation = $vmConfig.VM.Generation
 
@@ -67,19 +65,16 @@ if (!$memory) {
     $configMemory = Get-VMMemory -VM $vmConfig.VM
     $dynamicmemory = $configMemory.DynamicMemoryEnabled
 
-    # Memory values need to be in bytes
     $MemoryMaximumBytes = ($configMemory.Maximum)
     $MemoryStartupBytes = ($configMemory.Startup)
     $MemoryMinimumBytes = ($configMemory.Minimum)
-}
-else {
+} else {
     if (!$maxmemory){
         $dynamicmemory = $False
         $MemoryMaximumBytes = ($memory -as [int]) * 1MB
         $MemoryStartupBytes = ($memory -as [int]) * 1MB
         $MemoryMinimumBytes = ($memory -as [int]) * 1MB
-    }
-    else {
+    } else {
         $dynamicmemory = $True
         $MemoryMaximumBytes = ($maxmemory -as [int]) * 1MB
         $MemoryStartupBytes = ($memory -as [int]) * 1MB
@@ -87,11 +82,9 @@ else {
     }
 }
 
-
 if (!$switchname) {
     $switchname = (Get-VMNetworkAdapter -VM $vmConfig.VM).SwitchName
 }
-
 
 Connect-VMNetworkAdapter -VMNetworkAdapter (Get-VMNetworkAdapter -VM $vmConfig.VM) -SwitchName $switchname
 Set-VM -VM $vmConfig.VM -NewVMName $vm_name -MemoryStartupBytes $MemoryStartupBytes
@@ -123,13 +116,12 @@ if ($generation -ne 1) {
 
 $report = Compare-VM -CompatibilityReport $vmConfig
 
-# Stop if there is incompatibilities which would fail anyhow.
+# Stop if there are incompatibilities
 if($report.Incompatibilities.Length -gt 0){
     Write-Error-Message $(ConvertTo-Json $($report.Incompatibilities | Select -ExpandProperty Message))
     exit 0
 }
 
-# Differencing disk
 if($differencing_disk){
     # Get all controller on the VM, first scsi, then IDE if it is a Gen 1 device
     $controllers = Get-VMScsiController -VM $vmConfig.VM
@@ -148,7 +140,6 @@ if($differencing_disk){
             }
         }
     }
-    
 }
 
 Import-VM -CompatibilityReport $vmConfig
--- a/plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
+++ b/plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
@@ -18,8 +18,6 @@
     [string]$differencing_disk=$null
 )
 
-"$($data_path)/Snapshots"
-
 # Include the following modules
 $Dir = Split-Path $script:MyInvocation.MyCommand.Path
 . ([System.IO.Path]::Combine($Dir, "utils\write_messages.ps1"))
@@ -124,9 +122,11 @@ if ($generation -ne 1) {
 }
 
 $report = Compare-VM -CompatibilityReport $vmConfig
+
+£ Stop if there is incomatibilities which would fail anyhow.
 if($report.Incompatibilities.Length -gt 0){
-$report.Incompatibilities
-    Write-Error-Message ConvertTo-Json $report.Incompatibilities
+    Write-Error-Message $(ConvertTo-Json $($report.Incompatibilities | Select -ExpandProperty Message))
+    exit 0
 }
 
 # Differencing disk
@@ -150,7 +150,7 @@ if($differencing_disk){
     }
     
 }
- 
+
 Import-VM -CompatibilityReport $vmConfig
 
 $vm_id = (Get-VM $vm_name).id.guid
--- a/plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
+++ b/plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
@@ -18,7 +18,7 @@ $Dir = Split-Path $script:MyInvocation.MyCommand.Path
 . ([System.IO.Path]::Combine($Dir, "utils\write_messages.ps1"))
 
 # load the config from the vmcx and make a copy for editing, use TMP path so we are sure there is no vhd at the destination
-$vmConfig = (Compare-VM -Copy -Path $vm_config -GenerateNewID -VhdDestinationPath $env:Temp)
+$vmConfig = (Compare-VM -Copy -Path $vm_config_file -GenerateNewID -VhdDestinationPath $env:Temp)
 
 $generation = $vmConfig.VM.Generation
 
--- a/plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
+++ b/plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
@@ -123,7 +123,7 @@ if ($generation -ne 1) {
 
 $report = Compare-VM -CompatibilityReport $vmConfig
 
-£ Stop if there is incomatibilities which would fail anyhow.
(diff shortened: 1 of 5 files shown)


bbrala:
Can you provide the vagrant file? I've tried a few and didn't run into problems.

irab:
My vagrant file is super minimal, just:

  config.vm.box = "ira/leap"
end

plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
```diff
@@ -0,0 +1,167 @@
...
+    [string]$auto_start_action=$null,
+    [string]$auto_stop_action=$null
+)
+
+# Include the following modules
+$Dir = Split-Path $script:MyInvocation.MyCommand.Path
+. ([System.IO.Path]::Combine($Dir, "utils\write_messages.ps1"))
+
+# load the config from the vmcx and make a copy for editing, use TMP path so we are sure there is no vhd at the destination
+$vmConfig = (Compare-VM -Copy -Path $vm_config -GenerateNewID -VhdDestinationPath $env:Temp)
```
bbrala: Ok, I might've did an oops with the laat commit there. I'll check when I have a pc available.

irab:
There also seems to be another weird problem but it is probably to do with my linux vagrant box:

Using "vagrant up" to start a this Debian Jessie VM creates a VM but it can't boot from HDD due to "no x64-based uefi boot loader was found"

But if I import the VM from the downloaded box files using the Hyper-V GUI it boots fine... probably Vagrant default import settings messing with the import.

bbrala:
I will investigate that box, see what is missing in the import.

bbrala:
Ok, i've looked into it, and it seems it is missing a part of the firmware (what used to be the bios). It should boot from an EFI partition, but that is not imported. I'll have a look how one would go about importing that.

bbrala:
Well, that is fun. I can't find a way to manually set an EFI file to boot from. The file is contained in the firmware, but since we create a whole new VM, it is not present there. Will try again soon to see if i missed something, i'm kind of tired right now. But if not, i might need to use a real copy to get the VM up and running, which should be doable, but required a bit of a different approach.

irab:
Looks like there are [issues with importing any Gen 2 Linux VHDs](https://blogs.msdn.microsoft.com/virtual_pc_guy/2015/02/11/copying-the-vhd-of-a-generation-2-linux-vmand-not-booting-afterwards/). 

Tried the 'fix' below on original VM

grub-install --target=x86_64-efi --efi-directory=/boot/efi --no-nvram --removable

Recreated a box file, re-did "vagrant up --provider hyperv" and VM now boots perfectly. Thanks for the getting the import working!!

bbrala:
Hmm, that issue is because of recreating the whole VM instead of making a copy. Although recreating is how it works right now, i'm not entirely sure it is the best way.

Pros would be the fact you have more control on what features are enabled/set, if we would be copying the VM you never really know what features are enabled. This might make support a bit harder since you have less control.

For now, glad it works. :)

bbrala:
There is already a mention of this problem in the docs it seems, so no real loss of functionality if it doesn't support EPI in firmware.

https://www.vagrantup.com/docs/hyperv/boxes.html

bbrala:
For testing I made a gist explaining how to get vagrant working from source on Windows.

https://gist.github.com/bbrala/7e96be8a4e26f14834e8bbc21d246c38

Just checkout the branch in the repository after running the commands in the gist.

bbrala:
Ok, couldn't resist, the whole New-VM seems flaky sometimes so I used the proper PowerShell command `Compare-VM`  to create the new VM.

- Implemented the differencing disk for vmcx.
- Disk is now copied by Hyper-V instead of Ruby for machines with the new config.
- EFI Firmware now works for machines since it is a real copy. Compare-VM will report incompatibilities should they be found, but i didn't really find anything that could be incompatible with a host.

Only thing i'm slightly worried about is how Windows 8 Pro will perform. But i feel this PR is done.

bbrala:
@chrisroberts it would be awesome to get this merged soon, contrary to the package pr i can't release this as a plugin. And it will enable me and others to start using hyper v boxes exported from recent windows versions. 

Is there anything I can do to help more to get this going?

b33pr:
I pulled these in and did a test - after I made the change in https://github.com/bbrala/vagrant/pull/2 everything worked. Thanks for all of your work on this

bbrala:
Thank you for having a look. The condition seems fine like that indeed, good catch. Regarding bsdtar I'd like some extra info.

Does bsdtar give an error when you try to package? And what version of vagrant do you have installed? What version of windows? Did you use the gist as configuration? And you apparently merged both branches I made? Also how large is the hard disk that fails? 

(Sent by mobile)

b33pr:
bsdtar was dumping an error 'the file seemed to change' but the packager
didn't notice. Running win 10 'anniversary edition' Did use the gist, I
think, did merge both branches, and absurdly large - like 13gb.

bbrala:
@b33pr i tried 4 different vagrant hyper v boxes from hashicorp atlas. 

1. vagrant up
2. vagrant package
3. vagrant box add / init / up

Haven't run into any problems...

DavidCzadilek:
@bbrala first off, thank you for your great work! build vagrant-dev according to your instructions and now importing hyper-v boxes works for the most part.

Ran into some furhter problems and i suppose it might have something to do with the fact that i was trying to use a 2nd generation hyper-v vm.
The Problem is, that no OS is detected on the VM. I've attached a screenshot with the difference in settings of the original and the vagrant-created vm in hyper-v-manager.
pasted image at 2016_10_19 02_40 pm

Seems to have something to do with setting the bootloader..
I'll be trying Gen1 for now.

Info:
* using vagrant-dev according to https://gist.github.com/bbrala/7e96be8a4e26f14834e8bbc21d246c38 & current master
* windows 10, hyper-v, all updates installed (not the win-insider ones)
* vm based on ubuntu server 14.04.5 x64 iso from ubuntu website
* hyper-v integration tools installed (by default), and added to boot-sequence, according to http://www.serverwatch.com/server-tutorials/installing-and-activating-hyper-v-linux-integration-services.html

bbrala:
Thanks for having a look @DavidCzadilek. Looking at the imported VM it seems you packaged it with a custom XML which isn't needed anymore with this PR. Vagrant will look in the box file if there is a XML file and if so import it the old way. The old way doesn't support gen 2. You can do two things, remove the XML so there is only the binary config and repackage. Or also use the second PR i made to package it through vagrant. (https://github.com/mitchellh/vagrant/pull/7867).

DavidCzadilek:
Thanks for the reply @bbrala!
I'm packaging manually according to https://www.vagrantup.com/docs/hyperv/boxes.html
My final box file (`*.tar.gz` -> `*.box`) holds the `Virtual Hard Disk` folder containing the hard drive `*.vhdx` file, the `Virtual Machines` folder containing a `*.vmcx` and a `*.VMRS` file, as well as a `metadata.json` with only the provider setting:
`{"provider":"hyperv"}`
Thats it - nothing else in the box.

bbrala:
That is pretty weird. I'll have a go again and see what happens. It should copy the firmware...

bbrala:
Did you tar on linux?

DavidCzadilek:
Tried with both 7zip and Ubuntu bash on win10.
Havent worked much with powershell so far, especially scripting, but while patching I noticed that in the XML version of the import script there's a block for boot devices (https://github.com/mitchellh/vagrant/blob/master/plugins/providers/hyperv/scripts/import_vm.ps1#L86) which is missing entirely in the vmcx import script.

Tried to add it from the XML version, but that didnt seem to have an effect.. its possible i did something wrong though, since i have no experience with PS scripts (no debug errors though).

bbrala:
Since the new script makes a copy from the machine itself, it copies the boot-state also. I've just exported a VM with the same image and settings as you and going to test if it will add it properly later tonight.

bbrala:
Ok, i did the same as you described.

1. New Hyper V VM generation 2.
2. Install ubuntu server
3. Export 
4. tar with windows bash

Then i added the box to my vagrant install and loaded it up. It seems like it was a perfect copy with the bios in place.

importedboxes

I can use it without a problem... I did use this [article from Microsoft](
https://technet.microsoft.com/en-us/windows-server-docs/compute/hyper-v/supported-ubuntu-virtual-machines-on-hyper-v) to install the hyperv stuff though, it didn't report IP when i used your link (although i mightve missed something)

DavidCzadilek:
Thanks for the test - the manual installation of services as described in the [article from Microsoft](https://technet.microsoft.com/en-us/windows-server-docs/compute/hyper-v/supported-ubuntu-virtual-machines-on-hyper-v) was exactly what I was missing. Everything up and running smoothly now! 👍

I'll be making another box based on Ubuntu 16.x using Packer now. Will keep you updated if I run into any  issues or find out more info worth sharing here.

bbrala:
Awesome, glad it works as intended.

bbrala:
Made two [Hyper V 2nd generation boxes](https://atlas.hashicorp.com/bbrala), built at work then repackaged at home with Vagrant without any issues.

bbrala:
@chrisroberts it seems the PR for packer is slotted for the next release, which is 15 november. It would be a shame if this spawns new Hyper V generation 2 machines which do not work on Vagrant.

taliesins:
@bbrala you can select the generation to use.

Whats the issue with generation 2 and vagrant? Is it simply that not all oses support generation 2?

bbrala:
@taliesins BIggest problem with vagrant is that the new format for machines (vmcx) isn't supported, it cannot read that format. The wat it is fixed currectly, people use a custom XML with the configuration of the machine, but this means, no EFI bios is possible. 

So generation 2 is mostly broken because the default options (EFI bios) isn't supported at all.

b33pr:
What's the status here?

bbrala:
I'd really like to know if this realisticly will make it into 1.9 otherwise I will make it a plugin until release. But rather not do that if it will make it in.

krasiyosifov:
I wanted to thank you @bbrala for your work! After fighting for two days with the problem of unable to import my own hyperv box into latest version of vagrant, I found your solution. 
To try your solution I had to install vagrant from source of your pull request. When I tested again to import the box, everything worked as expected with no errors! 
Now I'm happy with my setup and hoping that fix to be included in the official release.
If you need some assistance I can help you with testing various scenarios, because I want this fix to be released officially as soon as possible. I will leave the vagrant installation from the latest code from github pull request until it is officially released and working.
Thanks again!

plugins/providers/hyperv/scripts/import_vm_vmcx.ps1
```diff
@@ -0,0 +1,154 @@
...
+if ($generation -ne 1) {
+    Set-VMFirmware -VM $vmConfig.VM -EnableSecureBoot (Get-VMFirmware -VM $vmConfig.VM).SecureBoot
+}
+
+$report = Compare-VM -CompatibilityReport $vmConfig
+
+# Stop if there are incompatibilities
+if($report.Incompatibilities.Length -gt 0){
+    Write-Error-Message $(ConvertTo-Json $($report.Incompatibilities | Select -ExpandProperty Message))
+    exit 0
```
chrisroberts: Should this be returning back a non-zero status to indicate import failure?

bbrala: Hmm, sematicly, you are correct. But what i understood from how the code of Vagrant works it just parses the output and handles the error and prints the error in a nice way. Think the default error handling is a bit more rough in its messages. 

That is how i understood the code at least :)

Relevant code: https://github.com/bbrala/vagrant/blob/8744caebcf50095e808233fd7282849115eddacb/plugins/providers/hyperv/driver.rb#L20-L40

chrisroberts: :+1:

bbrala:
Awesome, glad to see it slotted for 1.9 :)

chrisroberts:
@bbrala I'm looking to have it included in 1.9.1

bbrala:
@chrisroberts don't forget to close https://github.com/mitchellh/vagrant/issues/6102

chrisroberts:
:+1:

PR #7867, opened 2016-10-05
Package Hyper-V boxes

bbrala:
First draft of packaging of Hyper-V boxes. Uses the same logic as the Virtualbox package command./

Need to refactor the `package_*.rb` files since right now they are copies of the virtualbox versions, might just put them in general and extend them.

Currently in my dev enviroment i run into the problem that bsdtar is not doing its job. It doesn't tar all the files for some weird reason when they are given as seperate arguments.

From the log:
`"C:\\HashiCorp\\Vagrant\\embedded\\gnuwin32\\bin/bsdtar.EXE", "-czf", "G:/laragon/www/devlocal/package.box", "./Vagrantfile", "./vagrant_private_key", "./Virtual Hard Disks", "./Virtual Machines"]`
- [x] Refactor package_*.rb files to remove duplicate code.
- [x] Make bsdtar work as intented
- [x] Clean up :)

Add or remove `--base` switch to select any existing virtual machine?

### DIFF
--- a/lib/vagrant/action/general/package_setup_files.rb
+++ b/lib/vagrant/action/general/package_setup_files.rb
@@ -1,0 +1,51 @@
+module Vagrant
+  module Action
+    module General
+      class PackageSetupFiles
+        def initialize(app, env)
+          @app = app
+
+          env["package.include"] ||= []
+          env["package.vagrantfile"] ||= nil
+        end
+
+        def call(env)
+          files = {}
+          env["package.include"].each do |file|
+            source = Pathname.new(file)
+            dest   = nil
+
+            # If the source is relative then we add the file as-is to the include
+            # directory. Otherwise, we copy only the file into the root of the
+            # include directory. Kind of strange, but seems to match what people
+            # expect based on history.
+            if source.relative?
+              dest = source
+            else
+              dest = source.basename
+            end
+
+            # Assign the mapping
+            files[file] = dest
+          end
+
+          if env["package.vagrantfile"]
+            # Vagrantfiles are treated special and mapped to a specific file
+            files[env["package.vagrantfile"]] = "_Vagrantfile"
+          end
+
+          # Verify the mapping
+          files.each do |from, _|
+            raise Vagrant::Errors::PackageIncludeMissing,
+                  file: from if !File.exist?(from)
+          end
+
+          # Save the mapping
+          env["package.files"] = files
+
+          @app.call(env)
+        end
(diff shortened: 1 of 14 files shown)


bbrala:
Everything is working now, exporting a Hyper-V box and then importing it with the PR #7854. Quite awesome :)

Just need to move a few actions to General and then it seems quite fine. I could also copy the tests for Vagrant package to check at least the correct calls.

bbrala:
@chrisroberts I have a small question. I was refactoring, but it means some of the packaging code will move into the lib folder. I'm not sure if that fits the architecture. Could you provide some feedback on this?

https://github.com/bbrala/vagrant/compare/hyperv-package...bbrala:hyperv-package-refactor?expand=1

bbrala:
For testing I made a gist explaining how to get vagrant working from source on Windows.

https://gist.github.com/bbrala/7e96be8a4e26f14834e8bbc21d246c38

Just checkout the branch in the repository after running the commands in the gist.

chrisroberts:
@bbrala Skimming through the refactor branch, I don't see any glaring issues. I'm very much in favor of isolating common and duplicated behavior, so :+1:. I will have a more in depth look at that branch (as well as this entire PR) this afternoon. This looks really great!

bbrala:
Ok cool! Thanks for looking into it. I will refactor the other bits the same way then tonight (its 17:23 here).

bbrala:
Ok, it wasn't much effort to change. Just can't test it at work. The `--base` thing is maybe a slight problem. But it might not be very easy to fix that. There is some hardcoded stuff there for VirtualBox which is kinda ugly :) 

Not sure how i would go about fixing that since it means checking what virtualisation is used on that machine. Perhaps check if VirtualBox is enabled on the machine. Although you could have both installed... 

https://github.com/mitchellh/vagrant/blob/f577ca3d232f470bd5eabd2ca24ebd8d10492374/plugins/commands/package/command.rb#L54-L54

>  # XXX: This whole thing is hardcoded and very temporary. The whole
        # `vagrant package --base` process is deprecated for something much
        # better in the future. We just hardcode this to keep VirtualBox working
        # for now.

bbrala:
Thank! I've been using HyperV instead of VirtualBox on Windows 10 for a while now. The performance gain is insane. Next week a the first few developers at @swisnl will move to Hyper V :) 

Hopefully things will go smoothly, this will at least make repackaging easy ^^

Issue #7728, opened 2016-08-17
Incorrect `vagrant ssh-config` output  while working with `--host xxx`


guessi:
### Vagrant version
Vagrant 1.8.5

### Host operating system
Fedora 24

### Guest operating system
CentOS 7

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure(2) do |config|
  config.vm.box = "centos/7"

  instances = [
    'node1',
    'node2',
  ]

  instances.each do |box|
    config.vm.define box
  end
end

```

### Debug output
SKIPPED

### Expected behavior
vagrant ssh-config with `--host` option should display "Host" correctly

### Actual behavior
vargrant ssh-config with `--host` option show "Host" field incorrectly

```
$ vagrant ssh-config 
Host node1
  HostName 127.0.0.1
  User vagrant
  Port 2222
  UserKnownHostsFile /dev/null
  StrictHostKeyChecking no
  PasswordAuthentication no
  IdentityFile /home/xxxxxx/centos7/.vagrant/machines/node1/virtualbox/private_key
  IdentitiesOnly yes
  LogLevel FATAL

Host node2
  HostName 127.0.0.1
  User vagrant
  Port 2200
  UserKnownHostsFile /dev/null
  StrictHostKeyChecking no
  PasswordAuthentication no
  IdentityFile /home/xxxxxx/centos7/.vagrant/machines/node2/virtualbox/private_key
  IdentitiesOnly yes
  LogLevel FATAL

$ vagrant ssh-config --host node1
Host node1
  HostName 127.0.0.1
  User vagrant
  Port 2222
  UserKnownHostsFile /dev/null
  StrictHostKeyChecking no
  PasswordAuthentication no
  IdentityFile /home/xxxxxx/centos7/.vagrant/machines/node1/virtualbox/private_key
  IdentitiesOnly yes
  LogLevel FATAL

Host node1
  HostName 127.0.0.1
  User vagrant
  Port 2200
  UserKnownHostsFile /dev/null
  StrictHostKeyChecking no
  PasswordAuthentication no
  IdentityFile /home/xxxxxx/centos7/.vagrant/machines/node2/virtualbox/private_key
  IdentitiesOnly yes
  LogLevel FATAL

$ vagrant ssh-config --host node2
Host node2
  HostName 127.0.0.1
  User vagrant
  Port 2222
  UserKnownHostsFile /dev/null
[...]

PR #7877, opened 2016-10-07
Fix incorrect ssh-config output when '--host' is defined

guessi:
Fixes: #7728

### DIFF
--- a/plugins/commands/ssh_config/command.rb
+++ b/plugins/commands/ssh_config/command.rb
@@ -32,6 +32,10 @@ module VagrantPlugins
           ssh_info = machine.ssh_info
           raise Vagrant::Errors::SSHNotReady if ssh_info.nil?
 
+          if options[:host] && options[:host].to_s != machine.name.to_s
+            next
+          end
+
           variables = {
             host_key: options[:host] || machine.name || "vagrant",
             ssh_host: ssh_info[:host],


plugins/commands/ssh_config/command.rb
```diff
@@ -32,6 +32,11 @@ def execute
           ssh_info = machine.ssh_info
           raise Vagrant::Errors::SSHNotReady if ssh_info.nil?
 
+          if not options[:host].nil? and
+             not options[:host].to_s.eql?(machine.name.to_s)
+            next
+          end
+
```
chrisroberts: This can be reduced to:

```ruby
next if options[:host] != machine.name
```

guessi: thanks, updated.

guessi: Hi,
I've try to simplify it into `next if options[:host] != machine.name`, but failed,
see [Travis-CI Here](https://travis-ci.org/mitchellh/vagrant/builds/165839731#L335-L339)

the reason I put `not options[:host].nil?` in condition block is trying to avoid returning empty string if `--host XXX` not match

thanks for your quick review :+1:

chrisroberts: Ah, right, when no `--host` is provided, all should be printed, so only `next`ing when `options[:host]` is not `nil`. Still getting my :coffee: in. :smiley: 

Since code style in general prefers symbols over keywords, perhaps:

```ruby
if options[:host] && options[:host].to_s != machine.name.to_s
  next
end
```

and thanks for the quick followup!

guessi: Ha, fixed,
actually I'm thinking in shell + python :nerd_face:

guessi:
@chrisroberts 
if the MR is accepted, it still need to complete the unit test for the case with/without `--host` flag,
but unfortunately I'm not able to handle that part, may need to create another MR to finish them.  :sob:

chrisroberts:
@guessi No worries. I can fill in those tests. Thanks so much!

PR #7881, opened 2016-10-09
Make ansible-playbook command configurable

gonzolino:
The ansible-playbook command is currently hardcoded for the ansible and
ansible_local provisioners. This patch adds the config option
playbook_command to allow the user to change the command.

This may come in handy for people who want to use some kind of wrapper around ansible-playbook.
For example, I want to use vagrant to build an OpenStack test environment using openstack-ansible (see https://github.com/openstack/openstack-ansible). openstack-ansible installs a script at `/usr/local/bin/openstack-ansible` which acts as a wrapper around ansible-playbook and sets some environment variables and adds some arguments to ansible-playbook.
Having the possibility to specify openstack-ansible as playbook_command would make provisioning much more comfortable in this case.

### DIFF
--- a/plugins/provisioners/ansible/config/base.rb
+++ b/plugins/provisioners/ansible/config/base.rb
@@ -4,6 +4,7 @@ module VagrantPlugins
       class Base < Vagrant.plugin("2", :config)
 
         GALAXY_COMMAND_DEFAULT = "ansible-galaxy install --role-file=%{role_file} --roles-path=%{roles_path} --force".freeze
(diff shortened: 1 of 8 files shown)


gonzolino:
@gildegoma I updated the documentation and added a unit test for the playbook_command option.
Thanks for the suggestions!

gildegoma:
Thank you Daniel for updating so quickly. I merged it, with following addition: a55590b. I hope it's okay for you.

Issue #7885, opened 2016-10-11
[BUG] Vagrant unable to handle (appropriately) http 416 response code


akamensky:
### Vagrant version

`1.8.4`
### Host operating system

MacOS 10.11.6
### Guest operating system

Any
### Vagrantfile

``` ruby
# Not related
```
### Steps to reproduce
1. Start downloading a vagrant box with `vagrant box update`, realize that network is insanely slow to the server and stop the download. 
2. Check name of tmp file for box under ~/.vagrant.d/tmp/box***
3. Download box file (as shown in wget progress) from another/faster network and place it in ~/.vagrant.d/tmp/  with the previous tmp file name (from step 2)
4. Attempt to continue box update with downloaded file by calling `vagrant box update` one more time
### Expected behavior

Vagrant has the complete file and is able to handle HTTP 416 response (which clearly says that entire file has been downloaded already) and proceeds further
### Actual behavior

Vagrant is unable to handle HTTP 416 response and takes it as if it was real error (that is not an error btw).
### Comment

That is so ridiculous that 1 - vagrant box download speed is only reasonable from US, from anywhere else it is insanely slow and takes about 3 hours to download 277 MB box and 2 - that vagrant has no way to add box version from file, can only add new box (which will loose relation to previous version) with already downloaded file...

akamensky:
Same is on `1.8.6`. See attached screenshot.

![screen shot 2016-10-11 at 5 43 27 pm](https://cloud.githubusercontent.com/assets/985805/19266011/3f4cbab2-8fda-11e6-9db8-9883c5502f82.png)

akamensky:
Added PR for this issue with a patch. Tested on named box with "Steps to reproduce" and works as expected.

snipertomcat:
Heres a quick solution to this issue (well it worked for me anyway):

- `cd ~/.vagrant.d`
- `rm -rf tmp/*`
- `rm -rf data/machine-index/*`
- re-run your vagrant box command (i.e. `vagrant box add laravel/homestead` or whatever)

I think it has something to do with a partial download because mine didn't start acting up until a download got messed up part of the way through. After that I get the 416 error no matter what I do. Clearing out the 2 directories listed above did the trick for me.

GarethOates:
Any updates on this issue. It seems to still be a problem.

briancain:
Fixed via #9729 🎉

PR #7886, opened 2016-10-11
Issue #7885 Don't raise error if response is 416

akamensky:
Vagrant downloader should handle response codes appropriately. HTTP response code 416 means that the requested bytes are out of range, which in case of curl --continue-at or wget --continue means that the entire file has already been downloaded.

The chosen place for patch is justified by the fact that we should not raise ANY error if response code is 416.

### DIFF
--- a/lib/vagrant/util/downloader.rb
+++ b/lib/vagrant/util/downloader.rb
@@ -204,7 +204,7 @@ module Vagrant
           parts    = result.stderr.split(/\n*curl:\s+\(\d+\)\s*/, 2)
           parts[1] ||= ""
           if parts[1].include? "416"
-            # All good actually. 416 means there is no mory bytes to download
+            # All good actually. 416 means there is no more bytes to download
           else
             raise Errors::DownloaderError,
               code: result.exit_code,
--- a/lib/vagrant/util/downloader.rb
+++ b/lib/vagrant/util/downloader.rb
@@ -203,9 +203,13 @@ module Vagrant
           @logger.warn("Downloader exit code: #{result.exit_code}")
           parts    = result.stderr.split(/\n*curl:\s+\(\d+\)\s*/, 2)
           parts[1] ||= ""
-          raise Errors::DownloaderError,
-            code: result.exit_code,
-            message: parts[1].chomp
+          if parts[1].include? "416"
+            # All good actually. 416 means there is no mory bytes to download
+          else
+            raise Errors::DownloaderError,
+              code: result.exit_code,
+              message: parts[1].chomp
+          end
         end
 
         result


lib/vagrant/util/downloader.rb
```diff
@@ -203,9 +203,13 @@ def execute_curl(options, subprocess_options, &data_proc)
           @logger.warn("Downloader exit code: #{result.exit_code}")
           parts    = result.stderr.split(/\n*curl:\s+\(\d+\)\s*/, 2)
           parts[1] ||= ""
-          raise Errors::DownloaderError,
-            code: result.exit_code,
-            message: parts[1].chomp
+          if parts[1].include? "416"
+            # All good actually. 416 means there is no mory bytes to download
```
glennpratt: s/mory/more

More importantly, I'm not sure this is true. I received a 416 on an incomplete file when my machine crashed.

```
$ vagrant up
==> default: Box 'puppetlabs-precise64' could not be found. Attempting to find and install...
    default: Box Version: >= 0
==> default: Box file was not detected as metadata. Adding it directly...
==> default: Adding box 'puppetlabs-precise64' (v0) for provider: virtualbox
    default: Downloading: http://puppet-vagrant-boxes.puppetlabs.com/ubuntu-server-12042-x64-vbox4210.box
==> default: Box download is resuming from prior download progress
    default: Progress: 47% (Rate: 1217k/s, Estimated time remaining: 0:00:41)

Session Restored

Last login: Wed Oct 26 15:06:58 on ttys002
default is not running. Please start this with docker-machine start default

$ vagrant up
==> default: Box 'puppetlabs-precise64' could not be found. Attempting to find and install...
    default: Box Version: >= 0
==> default: Box file was not detected as metadata. Adding it directly...
==> default: Adding box 'puppetlabs-precise64' (v0) for provider: virtualbox
    default: Downloading: http://puppet-vagrant-boxes.puppetlabs.com/ubuntu-server-12042-x64-vbox4210.box
==> default: Box download is resuming from prior download progress
An error occurred while downloading the remote file. The error
message, if any, is reproduced below. Please fix this error and try
again.

The requested URL returned error: 416
```

akamensky: @glennpratt This is true. You can refer to HTTP response codes specification.

416 response code is returned when "requested starting byte is out of range". In this case client sends an offset (in bytes) starting with which it wants to receive data, if http server supports this mode, then it 1 - checks the size of files and if the size is more than specified offset then 2 - seek to the offset and starts streaming from it. This is how "resume download" implemented in all cases. 416 code is returned when 1 is not satisfiable (meaning specified offset is out of byte range of actual file size). If you get 416 on "not completely" downloaded file, this only means that client sent offset that is outside of range - either client bug or client received garbage that made file size same/bigger than the one on the server.

I am just following specifications of HTTP(S) protocol here.



> s/mory/more

A minor typo ;)

akamensky: Moreover I actually tested changes in the code as per my issue #7885. In that case file was completely and correctly downloaded but box was not added, because 416 returned which is treated currently as download failure. With this patch it unpacks the file and box is successfully added.

akamensky: Which is fine because then it would still fail on unpacking if file is it complete or corrupted. 

But for the case which I experienced in referenced issue it would actually make it work

akamensky: @glennpratt Else need to switch from curl to wget which handles this status code appropriately (it exits with 0 code when 416 is returned, unlike curl which delegates handling of that to user level)

briancain:
Thank you very much for the pull request! I've rebased this and simplified the code a bit here: #9729

PR #7902, opened 2016-10-16
providers/docker: Docker provider build from repo

landswellsong:
Docker has an [option to build a container from a remote git repository](https://docs.docker.com/engine/reference/commandline/build/#/git-repositories) allowing to specify branches, commits, tags etc. Vagrant's docker provider only expects the `build_dir` property to be a host directory. I've added an another `git_repo` property which allows to accomplish git builds with docker, bypassing the valid directory check and the part where Vagrant syncs the directory to docker host machine. Seems to work flawlessly at my setup. Docu and tests included.

PS: I'm not a Ruby coder so I'd very much love if you could review it throughfully.


landswellsong:
I wonder if somebody could update on the status of this? Is there a problem with my code? I thought it was a very trivial feature.

briancain:
Thank you so much for your contribution, and apologies for how long it took for us to address it! I have rebased it, and fixed it up (preserving your commits) in a new pull request here! https://github.com/hashicorp/vagrant/pull/10221 Thanks again.

landswellsong:
Thanks for the update!

Issue #7901, opened 2016-10-16
rsync installation by vagrant broken on netbsd


ThomasWaldmann:
### vagrant version
Vagrant 1.8.6 (from .deb package)

### Host
Ubuntu Linux 16.04

### Guest
NetBSD 7.0 (but also affects 5.x)

### Vagrantfile (excerpt)
```ruby
...
Vagrant.configure(2) do |config|
  # use rsync to copy content to the folder
  config.vm.synced_folder ".", "/vagrant/borg/borg", :type => "rsync", :rsync__args => ["--verbose", "--archive", "--delete", "-z"]
  # do not let the VM access . on the host machine via the default shared folder!
  config.vm.synced_folder ".", "/vagrant", disabled: true
...
```

### Expected behavior:
vagrant should install rsync into the netbsd box.

### Actual behavior
vagrant silently fails installing rsync.

```
tw@tux:~/w/borg$ vagrant up netbsd64
Bringing machine 'netbsd64' up with 'virtualbox' provider...
==> netbsd64: Box 'netbsd/NetBSD-7.0' could not be found. Attempting to find and install...
    netbsd64: Box Provider: virtualbox
    netbsd64: Box Version: >= 0
==> netbsd64: Loading metadata for box 'netbsd/NetBSD-7.0'
    netbsd64: URL: https://atlas.hashicorp.com/netbsd/NetBSD-7.0
==> netbsd64: Adding box 'netbsd/NetBSD-7.0' (v1.0.0) for provider: virtualbox
    netbsd64: Downloading: https://atlas.hashicorp.com/netbsd/boxes/NetBSD-7.0/versions/1.0.0/providers/virtualbox.box
==> netbsd64: Successfully added box 'netbsd/NetBSD-7.0' (v1.0.0) for 'virtualbox'!
==> netbsd64: Importing base box 'netbsd/NetBSD-7.0'...
==> netbsd64: Matching MAC address for NAT networking...
==> netbsd64: Checking if box 'netbsd/NetBSD-7.0' is up to date...
==> netbsd64: Setting the name of the VM: borg_netbsd64_1476632923766_33244
==> netbsd64: Fixed port collision for 22 => 2222. Now on port 2200.
==> netbsd64: Clearing any previously set network interfaces...
==> netbsd64: Preparing network interfaces based on configuration...
    netbsd64: Adapter 1: nat
    netbsd64: Adapter 2: hostonly
==> netbsd64: Forwarding ports...
[...]

PR #7922, opened 2016-10-24
guests/netbsd: Use full path to pkg_add command

chrisroberts:
Fixes: #7901

### DIFF
--- a/plugins/guests/netbsd/cap/rsync.rb
+++ b/plugins/guests/netbsd/cap/rsync.rb
@@ -8,9 +8,11 @@ module VagrantPlugins
 
         def self.rsync_install(machine)
           machine.communicate.sudo(
-            'PKG_PATH="http://ftp.NetBSD.org/pub/pkgsrc/packages/NetBSD/' \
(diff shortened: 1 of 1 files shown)


plugins/guests/netbsd/cap/rsync.rb
```diff
@@ -10,7 +10,7 @@ def self.rsync_install(machine)
           machine.communicate.sudo(
             'PKG_PATH="http://ftp.NetBSD.org/pub/pkgsrc/packages/NetBSD/' \
             '`uname -m`/`uname -r | cut -d. -f1-2`/All" ' \
-            'pkg_add rsync')
+            '/usr/sbin/pkg_add rsync')
```
sethvargo: iirc, this broke one of the other BSDs that inherits from netbsd 🐸

chrisroberts: Hrm, okay. I'll take another crack at it then and just adjust the PATH so it gets picked up.

chrisroberts:
@sethvargo updated

PR #7931, opened 2016-10-26
Puppet environment variables

chrisroberts:
Completes #7252 which did not include actual usage of configuration defined environment variables.

Fixes: #2270

### DIFF
--- a/plugins/provisioners/puppet/config/puppet.rb
+++ b/plugins/provisioners/puppet/config/puppet.rb
@@ -13,6 +13,7 @@ module VagrantPlugins
         attr_accessor :manifests_path
         attr_accessor :environment
         attr_accessor :environment_path
+        attr_accessor :environment_variables
         attr_accessor :module_path
         attr_accessor :options
         attr_accessor :synced_folder_type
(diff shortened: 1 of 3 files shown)


website/source/docs/provisioning/puppet_apply.html.md
```diff
@@ -52,6 +52,9 @@ available below this section.
 * `environment_path` (string) - Path to the directory that contains environment
   files on the host disk.
 
+* `environment_variables` (hash) - A hash of data to be set as environment
```
sethvargo: Could we provide an example here? or document that it's string -> string?

plugins/provisioners/puppet/config/puppet.rb
```diff
@@ -85,6 +87,9 @@ def finalize!
             if @manifest_file == UNSET_VALUE
               @manifest_file = nil
             end
+            if @environment_variables == UNSET_VALUE
+              @environment_variables = nil
```
sethvargo: I think, if you set this to `{}`, you'll have a better day. Then you don't need any nil checks - you can just .each it all the way 🏡

PR #7939, opened 2016-10-28
Rewrite linux/nfs_cleanup for security and multi-user, fixes #7938

### LINKED ISSUE #7938: nfs_cleanup security race and permissions
The current `nfs_cleanup` method for Linux has two problems:
1. The temporary file is created in an insecure manner with a race before using sudo privileges to write back to `/etc/exports`
2. It doesn't work properly for multiple users on a single system, because the file `/tmp/exports` isn't removed. As a result, after user1 runs vagrant, then user2 tries but can't write to `/tmp/exports` because it's owned by user1, so previously-added vagrant NFS exports are lost as the users toggle back and forth.

For the security race, there's a trivial attack where a separate unprivileged user on the host can add arbitrary entries to `/etc/exports`. I can provide details as necessary.
### Vagrant version

Vagrant 1.8.6 (all versions)
### Host operating system

Fedora 24 (any Linux)
### Guest operating system

any

agriffis:
Avoid using a temporary file, rather do the substitution in Ruby and
write /etc/exports directly.

### DIFF
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -93,17 +93,25 @@ module VagrantPlugins
         def self.nfs_cleanup(id)
           return if !File.exist?("/etc/exports")
 
-          user = Regexp.escape(Process.uid.to_s)
-          id   = Regexp.escape(id.to_s)
-
           # Only use "sudo" if we can't write to /etc/exports directly
           sudo_command = ""
           sudo_command = "sudo " if !File.writable?("/etc/exports")
 
-          # Use sed to just strip out the block of code which was inserted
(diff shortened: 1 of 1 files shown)


agriffis:
Note the BSD version of `nfs_cleanup` doesn't suffer from the same problem because it doesn't use a temporary file. However it still [calls out to sed](https://github.com/mitchellh/vagrant/blob/3deed353aede705457677ea6ee49a81cef640742/plugins/hosts/bsd/cap/nfs.rb#L165) and could probably be replaced wholesale by the code in this PR which avoids OS differences.

agriffis:
Superceded by #7947

PR #7947, opened 2016-10-30
Refactor host linux nfs capability

chrisroberts:
This builds on #7939 and refactors the implementation to utilize the StringBlockEditor and unify how modifications are applied to the /etc/exports file within the capability.

Fixes: #7938

### DIFF
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -877,6 +877,14 @@ en:
         the issues below and execute "vagrant reload":
 
         %{output}
+      nfs_exports_failed: |-
+        Vagrant failed to install an updated NFS exports file. This may be
+        due to overly restrictive permissions on your NFS exports file. Please
+        validate them and try again.
+
+        command: %{command}
+        stdout: %{stdout}
+        stderr: %{stderr}
       nfs_cant_read_exports: |-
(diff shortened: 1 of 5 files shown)


templates/locales/en.yml
```diff
@@ -877,6 +877,14 @@ en:
         the issues below and execute "vagrant reload":
 
         %{output}
+      nfs_exports_failed: |-
+        Vagrant failed to install an updated NFS exports file. This may be
+        due to overly restrictive permissions on your NFS exports file. Please
+        validate them and try again.
```
agriffis: It seems like the only case where this can happen is if /etc/exports isn't readable by the user. Writing, chowning and moving are optionally prefixed by sudo. If reading optionally used sudo as well, then it would cover the case where /etc/exports is unreadable by normal users.

chrisroberts: Hi! You make a good point on the content reading, which I will update to shell out being consistent with the `chown` and `mv`. The edge case this error is mainly in use for would be where the file is detected as readable or accessible to the user (and sudo is thus not used) and the permissions on the file is changed between the check and the actual operation (like the `mv`).

mitchellh:
This looks great! Awesome that its tested now too.

PR #7978, opened 2016-11-09
Feature command box prune

bbrala:
As discussed in #6863 with @chrisroberts a clean branch for the command. Silly rebase. :)

# Box prune

**Command: `vagrant box prune`**

This command removes old versions of installed boxes. If the box in currently in use vagrant will ask you if you to confirm.

## Options

* `--provider PROVIDER` - The specific provider type for the boxes to destroy.

* `--dry-run` - Only print the boxes that would be removed.

* `--name NAME` - The specific box name to check for outdated versions.

* `--force` - Destroy without confirmation even when box is in use.

Fixes: #4412 #5633

### DIFF
--- a/plugins/commands/box/command/prune.rb
+++ b/plugins/commands/box/command/prune.rb
@@ -1,0 +1,128 @@
+require 'optparse'
+
+module VagrantPlugins
+  module CommandBox
+    module Command
+      class Prune < Vagrant.plugin("2", :command)
+        def execute
+          options = {}
+          options[:force] = false
+          options[:dry_run] = false
+
+          opts = OptionParser.new do |o|
+            o.banner = "Usage: vagrant box prune [options]"
+            o.separator ""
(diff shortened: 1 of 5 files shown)


bbrala:
cool Nice to see it merged! I'll update the plugin soon to work until next version number :)

TomyLobo:
Nice, I could really use this feature right now.
Thanks to the author and merger for such a quick (and the great timing) .
I just transformed our self-hosted boxes into versioned boxes, so there will be some garbage to collect using this feature in the future :)

bbrala:
Until this get released just use the [plugin i built](https://github.com/swisnl/vagrant-remove-old-box-versions), it is the same code, just extracted as a plugin since it took 10 months to get merged. Couldn't wait :)

alexserver:
Thanks for making our life easier !

Issue #6604, opened 2015-11-27
request: ability to force provision in any state


avdd:
If I know the state of the VM, I can force a (re-)provision:
- destroyed: `up`
- halted: `up --provision`
- suspended: `resume; provision` (`up --provision` does not provision!)
- running: `provision` (`up --provision` does nothing)

If I want to script this, I could parse the output of `status`, but that's sloooow.

It would be nice to have some way of forcing a provision from any state.  Ideally `up --provision` would always work, but perhaps for backwards compat a flag could be added to `provision` to ensure the VM is running, e.g. `provision --up`

TomyLobo:
````bash
if vagrant resume; then
    vagrant provision
else
    vagrant provision || vagrant up --provision
fi
````

Could possibly be shortened to:
````bash
vagrant resume; vagrant provision || vagrant up --provision
````

Another way:
````bash
vagrant up --no-provision; vagrant provision
````

chrisroberts:
See #7979 for notes on next implementation

TomyLobo:
Why not just fix `up --provision` to provision no matter the previous state of the machine?
In fact, that's exactly what you implemented as the new "provision --up" command.

chrisroberts:
Hi there,

It looks like this has been resolved within a previously shipped version of Vagrant so I am now closing this issue. If the original issue was not fully resolved, please reopen this issue or create a new one.

Cheers!

PR #7979, opened 2016-11-09
Add `--up` option to provision command

chrisroberts:
This option will attempt to transition a machine in a non-running state
to a running state, if possible, prior to provisioning.

Fixes: #6604

### DIFF
--- a/plugins/commands/provision/command.rb
+++ b/plugins/commands/provision/command.rb
@@ -9,7 +9,6 @@ module VagrantPlugins
 
       def execute
         options = {}
-        options[:provision_types] = nil
 
         opts = OptionParser.new do |o|
           o.banner = "Usage: vagrant provision [vm-name] [--provision-with x,y,z]"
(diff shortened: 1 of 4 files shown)


plugins/commands/provision/command.rb
```diff
@@ -27,6 +30,20 @@ def execute
...
+            when :not_created, :poweroff
+              machine.action(:up)
+            when :paused, :saved
+              machine.action(:resume)
+            else
+              raise Vagrant::Errors::ProvisionAutoUpFailure.new(
+                machine_name: machine.name,
+                machine_state: machine.state.short_description
+              )
+            end
```
mitchellh: This actually isn't a safe way to do this since providers can return arbitrary state IDs. 

Instead, I would fix this in the provider level with `action_provision` bringing up the machine if necessary.... thoguhts?

chrisroberts: The only downside I see is that the providers will need to be updated instead of the functionality being intrinsic (more of a concern for non-builtin providers). But given the arbitrary state IDs, it will have to be pushed to the providers. Will give this another consideration and re-implement with a more provider focused implementation.

Issue #6323, opened 2015-09-25
Option for checksum when using remote shell provisioner


howardroark:
Hey!

It would be great if a source hash could be supplied when using a remote file as the path of the shell provision option.

Is this something that the core team thinks would be a worthwhile addition?

Thanks!

sethvargo:
Hi @howardroark 

Thank you for opening an issue. It is unclear what you mean by "remote file as the path of the shell provision option", but I presume you mean something like this:

```ruby
Vagrant.configure("2") do |config|
  config.vm.provision "shell",
    inline: "/path/to/the/script/already/on/the/guest.sh"
end
```

The recommended way to specify the shell is to either add a shebang to the file that you upload the guest, or specify the runner:

```ruby
Vagrant.configure("2") do |config|
  config.vm.provision "shell",
    inline: "/bin/sh /path/to/the/script/already/on/the/guest.sh"
end
```

Does that answer your question?

howardroark:
Basically you can specify a URL for the `path` option of the `shell` provisioner.  It would be nice if you could also specify a hash like md5 to verify the integrity of the downloaded file before running it.

sethvargo:
Oh I see - you want a checksum attribute. Sounds like a reasonable thing.

howardroark:
Yeah!  I guess that was a bit confusing :P  It is really handy to load "bootstrap" scripts direct from GitHub's servers.  Checking a sum would make it a little more secure.  

Thanks!

PR #7985, opened 2016-11-10
Add md5 and sha1 checksum support to Downloader.

chrisroberts:
Allows checksum validation on downloaded files via Util::Downloader
using MD5 and/or SHA1 checksums. This also integrates checksum validation
support with the shell provisioner for downloaded remote files.

Fixes: #6323

### DIFF
--- a/lib/vagrant/util/downloader.rb
+++ b/lib/vagrant/util/downloader.rb
@@ -1,7 +1,8 @@
 require "uri"
 
 require "log4r"
-
+require "digest/md5"
+require "digest/sha1"
 require "vagrant/util/busy"
 require "vagrant/util/platform"
 require "vagrant/util/subprocess"
@@ -17,6 +18,12 @@ module Vagrant
       #
       #     Vagrant/1.7.4 (+https://www.vagrantup.com; ruby2.1.0)
       USER_AGENT = "Vagrant/#{VERSION} (+https://www.vagrantup.com; #{RUBY_ENGINE}#{RUBY_VERSION})".freeze
+
+      # Supported file checksum
+      CHECKSUM_MAP = {
+        :md5 => Digest::MD5,
+        :sha1 => Digest::SHA1
+      }.freeze
 
       attr_reader :source
       attr_reader :destination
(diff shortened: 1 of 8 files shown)


lib/vagrant/util/downloader.rb
```diff
@@ -178,6 +191,47 @@ def head
...
+      #
+      # @param source [String] Source of file
+      # @param path [String, Pathname] local file path
+      # @param checksums [Hash] User provided options
+      # @option checksums [String] :md5 Compare MD5 checksum
+      # @option checksums [String] :sha1 Compare SHA1 checksum
+      # @return [Boolean]
+      def validate_download!(source, path, checksums)
+        CHECKSUM_MAP.each do |type, klass|
+          if checksums[type]
```
sethvargo: Should we add some kind of log message here, like:

> Validating checksum (md5) for X, expected: xxx actual: xxx

Could help in future debugging.

lib/vagrant/util/downloader.rb
```diff
@@ -178,6 +191,47 @@ def head
...
+      end
+
+      # Generate checksum on given file
+      #
+      # @param digest_class [Class] Digest class to use for generating checksum
+      # @param path [String, Pathname] Path to file
+      # @return [String] hexdigest result
+      def checksum_file(digest_class, path)
+        digester = digest_class.new
+        File.open(path.to_s, 'rb') do |file|
```
sethvargo: This is 1024 _bytes_, which seems really small. The [core method](https://github.com/ruby/ruby/blob/9ec0f8b7cb49c538f2bed77ec1506d59119f4bf1/ext/digest/lib/digest.rb#L49-L57) uses 16384...

Also, there is a core method for this 😄

templates/locales/en.yml
```diff
@@ -734,6 +734,13 @@ en:
       downloader_interrupted: |-
         The download was interrupted by an external signal. It did not
         complete.
+      downloader_checksum_error: |-
+        The calculated checksum of the requested file does not match the expected
+        checksum!
```
sethvargo: Could we include some information about the type of checksum in here (md5 vs sha256)?

chrisroberts:
@sethvargo updated

PR #8011, opened 2016-11-16
Plugins internal state

chrisroberts:
Provides better update support due to delayed activation which prevents overly restrictive constraints being applied. Includes better internal management to clean non-required gems as needed. Some refactoring cleanup to reduce code duplication. More logging output.

### DIFF
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -29,7 +29,7 @@ argv.each_index do |i|
       ENV['VAGRANT_NO_PLUGINS'] = "1"
     end
 
-    if arg == "plugin" && ["repair", "expunge"].include?(argv[i+1])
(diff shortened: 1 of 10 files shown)


bin/vagrant
```diff
@@ -29,7 +29,7 @@ argv.each_index do |i|
       ENV['VAGRANT_NO_PLUGINS'] = "1"
     end
 
-    if arg == "plugin" && ["repair", "expunge"].include?(argv[i+1])
+    if arg == "plugin" && argv[i+1] != "list"
```
sethvargo: Do we have to worry about being out of bounds with the +1?

chrisroberts: It will just return `nil`

PR #8033, opened 2016-11-23
Hyper-V VMCX: Setting memory too low or high makes the import script fail.

bbrala:
When you set the memory of a HyperV machine to something lower than the original range it fails. 

This is because it sets the current memory before adjusting the range, which makes the command invalid.


bbrala:
@chrisroberts Found a little bug in the import script. :)

bbrala:
Thanks :)

PR #8050, opened 2016-11-29
Add documentation for ssl_peer_verification

matt-richardson:
I was searching for how to use a self-signed certificate with winrm, but I only found this mentioned in a PR.


chrisroberts:
Fixed via #8102

Issue #7994, opened 2016-11-11
vagrant up  failing for fedora 24


tzvetkoff:
OS: `macOS Sierra 10.12.1`
Vagrant: `1.8.7`
Ruby: `2.3.1p112` (RVM)

I was trying to start my freshly installed Fedora box and I got this:

```
[w00t@phoenix(~/redacted)]$ vagrant up
Bringing machine 'default' up with 'vmware_fusion' provider...
==> default: Cloning VMware VM: 'tzvetkoff/fedora-24'. This can take some time...
==> default: Verifying vmnet devices are healthy...
==> default: Preparing network adapters...
==> default: Starting the VMware VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 172.16.26.143:22
    default: SSH username: vagrant
    default: SSH auth method: private key
    default:
    default: Vagrant insecure key detected. Vagrant will automatically replace
    default: this with a newly generated keypair for better security.
    default:
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!
==> default: Forwarding ports...
    default: -- 22 => 2222
==> default: Configuring network adapters within the VM...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

# Restart network
service network restart


Stdout from the command:

Restarting network (via systemctl):  [FAILED]


Stderr from the command:

Job for network.service failed because the control process exited with error code. See "systemctl status network.service" and "journalctl -xe" for details.
```

Here are the logs from the VM:
```
[vagrant@localhost ~]$ journalctl -xe
Nov 11 05:35:29 localhost.localdomain audit[1970]: USER_CMD pid=1970 uid=1000 auid=1000 ses=3 subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 msg='cwd="/home/vagrant" cmd=7365727669
Nov 11 05:35:29 localhost.localdomain audit[1970]: CRED_REFR pid=1970 uid=0 auid=1000 ses=3 subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 msg='op=PAM:setcred grantors=pam_env,pam_
Nov 11 05:35:29 localhost.localdomain sudo[1970]: pam_systemd(sudo:session): Cannot create session: Already running in a session
[...]

PR #8052, opened 2016-11-29
Fix `service network restart` on RHEL-7 / Fedora

tzvetkoff:
RHEL-7 / Current Fedora versions tend to use NetworkManager for configuring the networks, and `service network restart` might fail.
If the `NetworkManager` service is running, we should restart it, otherwise we try restarting `network`.

The issue was introduced in `v1.8.7`.

See #7994.

### DIFF
--- a/plugins/guests/redhat/cap/change_host_name.rb
+++ b/plugins/guests/redhat/cap/change_host_name.rb
@@ -31,8 +31,12 @@ module VagrantPlugins
                 sed -i'' '1i 127.0.0.1\\t#{name}\\t#{basename}' /etc/hosts
               }
 
-              # Restart network
-              service network restart
+              # Restart network (through NetworkManager if running)
+              if service NetworkManager status 2>&1 | grep -q running; then
+                service NetworkManager restart
+              else
+                service network restart
+              fi
             EOH
           end
         end
--- a/plugins/guests/redhat/cap/configure_networks.rb
+++ b/plugins/guests/redhat/cap/configure_networks.rb
@@ -36,7 +36,7 @@ module VagrantPlugins
 
             # Add the new interface and bring it back up
             final_path = "#{network_scripts_dir}/ifcfg-#{network[:device]}"
-            commands << <<-EOH.gsub(/^ */, '')
+            commands << <<-EOH.gsub(/^ {14}/, '')
               # Down the interface before munging the config file. This might
               # fail if the interface is not actually set up yet so ignore
               # errors.
@@ -48,9 +48,13 @@ module VagrantPlugins
             EOH
           end
 
-          commands << <<-EOH.gsub(/^ */, '')
-            # Restart network
-            service network restart
+          commands << <<-EOH.gsub(/^ {12}/, '')
+            # Restart network (through NetworkManager if running)
+            if service NetworkManager status 2>&1 | grep -q running; then
+              service NetworkManager restart
+            else
+              service network restart
+            fi
           EOH
 
           comm.sudo(commands.join("\n"))


joelpittet:
This seems to have caused a regression. I reversed the patch and vagrant started to work again.  See: https://github.com/mitchellh/vagrant/issues/8096

karlkfi:
Just because NetworkManager is enabled and running does not mean that all network interfaces are using NetworkManager. Interfaces may be configured with `NM_CONTROLLED=no` (like the vagrant configured host-only interface). In this case, on CentOS 7.2  `service network restart` is required.

karlkfi:
How/when does `service network restart` fail on RHEL-7 / Fedora?

joelpittet:
@chrisroberts Can you let us know a bit more detail around this change?

mvolhontseff:
Research this issue further, and found a similar problem documented in https://bugzilla.redhat.com/show_bug.cgi?id=1264024.

This was for virtual networking on Xen; however, the symptoms are very similar.  If you look at https://github.com/mitchellh/vagrant/issues/7994, @tzvetkoff includes these log lines:

Nov 11 05:35:29 localhost.localdomain NetworkManager[651]: <info>  [1478835329.8763] audit: op="connection-activate" uuid="3b3413b9-a44d-3c69-a12a-8a925a2e7bbd" name="ens33" result="fail" reas
Nov 11 05:35:29 localhost.localdomain network[1992]: Bringing up interface ens33:  Error: Connection activation failed: No suitable device found for this connection.
Nov 11 05:35:29 localhost.localdomain network[1992]: [FAILED]

To troubleshoot the above, the original reporter of the bug, Jan Stancek, had enabled NM debugging, and had noticed more detail regarding this; i.e.:

Sep 17 09:01:52 dhcp47-156 NetworkManager[614]: <debug> [1442494912.079108] [platform/nm-platform.c:2896] log_link(): platform: signal: link changed: 2: eth0 <DOWN;broadcast,multicast> mtu 1500 arp 1 ethernet init addrgenmode eui64 addr 52:56:00:00:00:31 driver vif
Sep 17 09:01:52 dhcp47-156 NetworkManager[614]: <debug> [1442494912.079274] [platform/nm-platform-utils.c:69] ethtool_get(): ethtool: Request failed: Operation not supported
Sep 17 09:01:52 dhcp47-156 NetworkManager[614]: <debug> [1442494912.086196] [platform/nm-platform.c:2896] log_link(): platform: signal: link changed: 1: lo <UP,LOWER_UP;loopback,up,running,lowerup> mtu 65536 arp 772 loopback init addrgenmode eui64 addr 00:00:00:00:00:00 driver unknown

It was discovered that the virtual interface was reporting a permanent address of 00:00:00:00:00:00, which should have been rejected as invalid.  There was a patch introduced to reject this addy, and "fall back" to the original temp address that was reported.

This may not be the same bug, but @tzvetkoff should enable NM debugging in his vmware env, etc to properly troubleshoot what's going on.

tzvetkoff:
The problem is that, at some point, we're having both `NetworkManager`-managed interfaces and `network`-managed (configured by Vagrant itself)

I was mistaken since my Fedoras were only NetworkManager-based.

But since `configure_networks.rb` is not checking error codes / exit status, easy (dirty) fix would be having both

``` bash
service network restart
service NetworkManager restart
```

in the mix. I'll pay my 2 cents for some error handling, alas - in the near future...

mmihalev:
@tzvetkoff your PR pissed a lot of CentOS users. This caused a major bug in Vagrant (#8096) ... but  I suppose you already know it. Why did you mention RHEL-7 if u tested it only on Fedora?

Issue #8067, opened 2016-12-01
Embedded OpenSSL requires SSL_CERT_FILE to be set on Linux


randomvariable:
Embedded OpenSSL requires SSL_CERT_FILE to be set on Linux

### Vagrant version
Vagrant 1.9

### Host operating system
```
NAME=Fedora
VERSION="24 (Workstation Edition)"
ID=fedora
VERSION_ID=24
```

### Guest operating system
N/A

### Vagrantfile
N/A

### Debug output
https://gist.github.com/randomvariable/051d8514b6d9d5bdfa9c6e673f73d26e

### Expected behavior
What should have happened?
```
Installing the 'vagrant-libvirt' plugin. This can take a few minutes...
Building native extensions.  This could take a while...
Fetching: formatador-0.2.5.gem (100%)
Fetching: excon-0.54.0.gem (100%)
Fetching: fog-core-1.43.0.gem (100%)
Fetching: fog-xml-0.1.2.gem (100%)
Fetching: fog-json-1.0.2.gem (100%)
Fetching: fog-libvirt-0.3.0.gem (100%)
Fetching: vagrant-libvirt-0.0.36.gem (100%)
Installed the plugin 'vagrant-libvirt (0.0.36)'!
```

### Actual behavior
What actually happened?
```
Installing the 'vagrant-libvirt' plugin. This can take a few minutes...
/opt/vagrant/embedded/lib/ruby/2.2.0/rubygems/dependency.rb:315:in `to_specs': Could not find 'fog-core' (>= 0) among 45 total gem(s) (Gem::LoadError)
Checked in 'GEM_PATH=/opt/vagrant/embedded/gems', execute `gem env` for more information
        from /opt/vagrant/embedded/lib/ruby/2.2.0/rubygems/specification.rb:1311:in `block in activate_dependencies'
        from /opt/vagrant/embedded/lib/ruby/2.2.0/rubygems/specification.rb:1300:in `each'
        from /opt/vagrant/embedded/lib/ruby/2.2.0/rubygems/specification.rb:1300:in `activate_dependencies'
        from /opt/vagrant/embedded/lib/ruby/2.2.0/rubygems/specification.rb:1282:in `activate'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.9.0/lib/vagrant/bundler.rb:297:in `block in activate_solution'
[...]

PR #8068, opened 2016-12-01
Detect load failure within solution set and retry if found

chrisroberts:
Installation solution sets in 2.2.5 can end up out of order (not seen
in 2.3.1) causing LoadErrors when the specification is in the solution
set during validation. This detects the missing spec within the solution
and if found will move spec to the start of the solution set and retry
solution activation.

### DIFF
--- a/lib/vagrant/bundler.rb
+++ b/lib/vagrant/bundler.rb
@@ -330,6 +330,20 @@ module Vagrant
         # again, just reverse order on failure and attempt again.
         if retried
           @logger.error("Failed to load solution set - #{e.class}: #{e}")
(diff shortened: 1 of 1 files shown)


randomvariable:
This does fix the dependency resolution part of #8067

chris1984:
tested this on ubuntu 16.04 and this solved my issue with trying to get the fog-core gem

bassrock:
👍  For a patch release soon

Ma27:
can confirm that it works again with vagrant 1.9.1. Thanks a lot @chrisroberts

Issue #7983, opened 2016-11-09
Hyper-V provider inaccurately reporting WinRM address as 0.0.0.0:5985 randomly


bshurts:
I'm seeing an issue with the vagrant hyperv provider where the WinRM address is reported as 0.0.0.0:5985 occasionally. This causes provisioning to fail randomly due to vagrant trying to WinRM with a bad address.

### Vagrant version
1.8.6

### Host operating system
Windows 10

### Guest operating system
Windows Server 2012 R2 on Hyper-V

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# All Vagrant configuration is done below. The "2" in Vagrant.configure
# configures the configuration version (we support older styles for
# backwards compatibility). Please don't change it unless you know what
# you're doing.
Vagrant.configure("2") do |config|
  # The most common configuration options are documented and commented below.
  # For a complete reference, please see the online documentation at
  # https://docs.vagrantup.com.

  # Every Vagrant development environment requires a box. You can search for
  # boxes at https://atlas.hashicorp.com/search.
  config.vm.box = "win2012r2-aio-20160915"

  # Disable automatic box update checking. If you disable this, then
  # boxes will only be checked for updates when the user runs
  # `vagrant box outdated`. This is not recommended.
  # config.vm.box_check_update = false

  # Create a forwarded port mapping which allows access to a specific port
  # within the machine from a port on the host machine. In the example below,
  # accessing "localhost:8080" will access port 80 on the guest machine.
  # config.vm.network "forwarded_port", guest: 80, host: 8080

  # Create a private network, which allows host-only access to the machine
  # using a specific IP.
  # config.vm.network "private_network", ip: "192.168.33.10"

  # Create a public network, which generally matched to bridged network.
  # Bridged networks make the machine appear as another physical device on
  # your network.
  # config.vm.network "public_network"

  # Share an additional folder to the guest VM. The first argument is
  # the path on the host to the actual folder. The second argument is
  # the path on the guest to mount the folder. And the optional third
  # argument is a set of non-required options.
  # config.vm.synced_folder "../data", "/vagrant_data"

  # Provider-specific configuration so you can fine-tune various
  # backing providers for Vagrant. These expose provider-specific options.
  # Example for VirtualBox:
  #
  # config.vm.provider "virtualbox" do |vb|
  #   # Display the VirtualBox GUI when booting the machine
  #   vb.gui = true
  #
[...]

PR #8073, opened 2016-12-02
HyperV inaccurate winrm address - issue-7983

bshurts:
Helper now throws WinRMNotReady exception is host ip is reported as an empty string

This was causing issues when provisioning win2012 r2 boxes in hyper-v. Sometimes hyper-v would report and empty string as the host address #in the winrm data to vagrant then vagrant would try to winrm to the machine with an ip of 0.0.0.0. This would cause sporadic failures when provisioning.

See #7983 for more information.

### DIFF
--- a/plugins/communicators/winrm/helper.rb
+++ b/plugins/communicators/winrm/helper.rb
@@ -33,7 +33,7 @@ module VagrantPlugins
         return addr if addr
 
         ssh_info = machine.ssh_info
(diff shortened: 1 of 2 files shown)


bshurts:
@chrisroberts is there something I can do to get this merged in? It's been sitting here for quite a while. I'm happy to fix any issues if it's not up to standards.

PR #8091, opened 2016-12-06
Add link to Atlas signup form

ruudk:
The Atlas website doesn't have a Signup button anywhere.

Every new developer at our company asked me "how do I actually signup for Atlas?"

To make it easier for everybody, I changed the url so that developers can start right away :)


briancain:
Closing this because it's now vagrant cloud, and also it has a sign up button!

Issue #8090, opened 2016-12-06
Vagrant 1.9 fails with plug ins that use Env.with_clean_env 


taliesins:
When spinning up a new environment that makes use of Berkshelf plugin I receive the following exception:
```
C:/HashiCorp/Vagrant/embedded/lib/ruby/site_ruby/2.2.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- ../bundler (LoadError)
	from C:/HashiCorp/Vagrant/embedded/lib/ruby/site_ruby/2.2.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.0/lib/vagrant/util/env.rb:1:in `<top (required)>'
	from D:/Services/.vagrant.d/gems/2.2.5/gems/vagrant-berkshelf-5.1.1/lib/vagrant-berkshelf/helpers.rb:62:in `berks'
	from D:/Services/.vagrant.d/gems/2.2.5/gems/vagrant-berkshelf-5.1.1/lib/vagrant-berkshelf/action/check.rb:33:in `berkshelf_version_check!'
	from D:/Services/.vagrant.d/gems/2.2.5/gems/vagrant-berkshelf-5.1.1/lib/vagrant-berkshelf/action/check.rb:16:in `call'
	from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.0/lib/vagrant/action/warden.rb:34:in `call'
	from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.0/lib/vagrant/action/builtin/config_validate.rb:25:in `call'
	from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.0/lib/vagrant/action/warden.rb:34:in `call'
	from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.0/lib/vagrant/action/builtin/handle_box.rb:56:in `call'
	from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.0/lib/vagrant/action/warden.rb:34:in `call'
	from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.0/lib/vagrant/action/builder.rb:116:in `call'
[...]

PR #8094, opened 2016-12-07
Remove direct bundler usage within Env util

chrisroberts:
Fixes: #8090


taliesins:
@chrisroberts thank you for jumping onto these issues so quickly! We all appreciate it.

Issue #7626, opened 2016-07-22
Undocumented configs for winrm


javadevmtl:
### Vagrant version

1.8.4
### Host operating system

Ubuntu 14.04
### Guest operating system

Windows Server 2008 with nsidc/vSphere plugin

The VM is launched and sysprep kicks in which causes the Windows VM to take longer to boot and it also reboots a couple times until sysprep is done.

The winrm communicator fails with No route to host, with a retry every 2 seconds based on what is seen from the INFO logs. But the VM in vSphere is up and running coretcly. A second, manual call after checking if sysprep completed, vagrant up call marks the machine as ready.

Google brought me to the original WinRB plugin. config.winrm.max_tries seems to do the trick. None of the "timeout" configs helped.

```
    config.winrm.max_tries = 100 <--- This worked
    config.winrm.timeout = 300 <--- doesn't seem to do anything
```

 config.winrm.max_tries is not documented in the official docs. Maybe it can be documented with a tip indicating what the default is and to bump it up if syspreping?

Thanks

dragon788:
Nice find! The WinRM timeout only affects how long Vagrant waits for the WinRM service (via plugin) to respond before failing.

chrisroberts:
Fixed via #8102

Issue #7953, opened 2016-11-01
Using file provioner with windows host fails on big files


StefanSchoof:
### Vagrant version
1.8.6

### Host operating system
windows 10 (1511)

### Guest operating system
windows 2016

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

VAGRANTFILE_API_VERSION = "2"

# linked clone feature requires 1.8
Vagrant.require_version ">= 1.8" 
Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  config.vm.box = "mwrock/Windows2016"
  config.vm.guest = :windows
  config.vm.communicator = "winrm"

  config.vm.provider "virtualbox" do |v|
    v.linked_clone = true
    v.customize ["modifyvm", :id, "--clipboard", "bidirectional"]
  end

  config.vm.provision "file", source: "14393.0.160715-1616.RS1_RELEASE_SERVER_EVAL_X64FRE_EN-US.ISO", destination: "C:\14393.0.160715-1616.RS1_RELEASE_SERVER_EVAL_X64FRE_EN-US.ISO"
end
```

### Debug output
https://gist.github.com/StefanSchoof/9dc34ecb613e51e9e22383c1247a7736

### Expected behavior
The big file (4,95 GB) should be uploaded or a message which is the maximal allowed file size for the file Provisioner.

### Actual behavior
A WMI Error:
```
The WS-Management service cannot 
process the request. This user is allowed a maximum number of 1500 concurrent operations, which has been exceeded. 
Close existing operations for this user, or raise the quota for this user.
```

### Steps to reproduce
1. Download win2016 eval iso http://care.dlservice.microsoft.com/dl/download/1/6/F/16FA20E6-4662-482A-920B-1A45CF5AAE3C/14393.0.160715-1616.RS1_RELEASE_SERVER_EVAL_X64FRE_EN-US.ISO
2. place iso next to vagrant file
3. `vagrant up`

chrisroberts:
Fixed via #8102

Issue #7717, opened 2016-08-15
[WinRM::FS::Core::FileTransporter] Upload failed


milanoid:
### Vagrant version

Vagrant 1.8.5
### Host operating system

Windows 10 with VirtualBox 5.0.26 r108824
### Guest operating system

Windows Server 2012 R2
### Vagrantfile

``` ruby
Vagrant.configure("2") do |config|
  config.vm.box = "my-box"
  config.vm.provision "file", source: "my_script.cmd", destination: "my_script.cmd"
end
```
### Debug output

https://gist.github.com/milanoid/a341572b96f8e9a9b5821817cfefe30d
### Expected behavior

File copied over to guest.
### Actual behavior

Upload error.
### Steps to reproduce
1. `vagrant reload --provision --debug`

git-jiby-me:
@chrisroberts the problem here is the powershell scrips in winrm-fs, they use types like `System.Security.Cryptography.MD5CryptoServiceProvider` and `System.Security.Cryptography.FromBase64Transform` which are not available in windows nano. I was able to manually tweak these scripts in my machine, to use alternatives and provisioning was working. 

Can we update `winrm` and `winrm-fs` to the latest versions from upstream, as there are changes in how it handles file uploads there, and then i could test with the latest `winrm-fs` version and submit patches for the powershell scripts if necessary

chrisroberts:
Related: #8102

chrisroberts:
Fixed via #8102

Issue #7526, opened 2016-06-27
Plugin winrm-fs is out of date and contains bugs (currently 0.3.2, latest version is 0.4.3)


TheFrogDaddy:
### Vagrant version

1.8.4
### Host operating system

OSX El Capitan
### Guest operating system

Windows 10
### Expected behavior

winrm-fs plugin be a recent version >= 4.0
### Actual behavior

winrm-fs plugin is out of date and contains bugs which preventing provisioning of a windows box after installing chocolatey
### References

This was my original issue, the devs of winrm-fs have informed me that the latest versions resolves my issue.

https://github.com/chocolatey/choco/issues/834

StefanScherer:
I manually updated winrm-fs 0.4.3 in vagrant 1.8.4 with [these steps](https://github.com/mitchellh/vagrant/issues/7476#issuecomment-229887350) and I can work with it for a while. Only a minor issue is that some parts of the WinRM output is now visisble:

<img width="622" alt="bildschirmfoto 2016-07-01 um 10 33 31" src="https://cloud.githubusercontent.com/assets/207759/16516257/5d81cd3c-3f77-11e6-9155-8c9e31d1f23a.png">

TheFrogDaddy:
Thanks @StefanScherer , I'll give those steps a go.

StefanScherer:
@TheFrogDaddy but be warned as Seth mentioned in #7476 there may be new issues.
Yesterday I got a reproducible error creating a Windows Domain Controller with a PowerShell provision script that the WinRM login failed.

I now have switched back to Vagrant 1.8.1 without any patches to it and using the embedded winrm-1.3.6 and winrm-fs 0.2.3.

Ping me if someone needs this scenario and provision scripts to test against newer winrm/winrm-fs versions.

dragon788:
@StefanScherer @TheFrogDaddy This may be an issue in the underlying WinRb/WinRM gem, not sure if it is patched in the version you mention above as there is still an open issue, https://github.com/WinRb/winrm-fs/issues/18.

chrisroberts:
Hi there,

It looks like this has been resolved within a previously shipped version of Vagrant so I am now closing this issue. If the original issue was not fully resolved, please reopen this issue or create a new one.

Cheers!

PR #8102, opened 2016-12-09
Refactor winrm communicator to use latest winrm gems and v2 api

mwrock:
This updates the winrm dependencies to the latest versions. The latest versions address several bugs and introduce some enhancements. Here are the highlights

* Powershell commands use the PSRP protocol which lifts the 8k command line limitation and adds compatibility with Windows Nano server.
* Due to the above, winrm-fs is orders of magnitude faster and more reliable
* The elevated shell via scheduled task logic now lives in winrm-elevated which is also used by other projects
* Updated winrm configuration documentation.

cc @sneal 

Signed-off-by: Matt Wrock <matt@mattwrock.com>


mwrock:
Fixes #7626, #7717, #7526, #8050, #7476, #7953

PR #8119, opened 2016-12-14
Harden VMCX support

bbrala:
Vagrant now checks if your Windows version has support for VMCX and then uses the VMCX file.

The reason for this is because the support for [HyperV in Packer is now merged](https://github.com/mitchellh/packer/pull/2576), but, it always creates an XML file, I assume for backwards compatibility, which does make sense especially for Windows versions before 10 which are still used. Vagrant should prefer using the newer format when supported, even when the old XML format is available.

The check consists of checking if the Windows version is greater or equal to 10, since Windows 10, and Windows Server 2016 both use the binary format, and they won't change back to the old config in future releases I assume.

To test, I loaded up about 5 of the HyperV boxes on Atlas with 1.9.1 and with this branch. Everything went fine. In order to make possible future debugging easier I added the type of config that is loaded to the debug log.

### DIFF
--- a/plugins/providers/hyperv/action/import.rb
+++ b/plugins/providers/hyperv/action/import.rb
@@ -38,18 +38,18 @@ module VagrantPlugins
           config_type = nil
           vm_dir.each_child do |f|
             if f.extname.downcase == '.xml'
+              @logger.debug("Found XML config...")
               config_path = f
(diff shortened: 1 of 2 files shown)


bbrala:
Welcome back @chrisroberts , hope you a good holiday :)

Issue #8096, opened 2016-12-08
Private network comes up with no IPv4 address without manual restart


rupert654:
### Vagrant version
I'm running Vagrant 1.8.6. This is not the latest version but I can't run 1.8.7 due to #8024 , nor 1.90 due to #8088 .

### Host operating system
Windows 7

### Guest operating system
CentOS 7

### Vagrantfile
```ruby
Vagrant.configure(2) do |config|
  config.vm.box = 'puppetlabs/centos-7.2-64-puppet-enterprise'

  config.vm.define :master do |master|
    master.vm.network :private_network, ip: '10.20.1.10'
  end

  config.vm.define :node do |node|
    node.vm.network :private_network, ip: '10.20.1.11'
  end
end
```

### Expected behavior
Each VM should have assigned the IP addresses specified in the Vagrantfile.

### Actual behavior
No IP address appears to be assigned. However, an IP address is assigned after restarting.

### Steps to reproduce
1. Log into one of the nodes
`vagrant ssh master`

2. Verify that the IP address is actually being configured
```
cat /etc/sysconfig/network-scripts/ifcfg-enp0s8
#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
NM_CONTROLLED=no
BOOTPROTO=none
ONBOOT=yes
IPADDR=10.20.1.10
NETMASK=255.255.255.0
DEVICE=enp0s8
PEERDNS=no
#VAGRANT-END
```

3. But no IP address is actually assigned (note there is no inet value):
```
ifconfig -a
enp0s8: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet6 fe80::a00:27ff:fe8b:416d  prefixlen 64  scopeid 0x20<link>
        ether 08:00:27:8b:41:6d  txqueuelen 1000  (Ethernet)
        RX packets 66  bytes 21226 (20.7 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 65  bytes 11106 (10.8 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```

4. Now exit and restart
```
exit
vagrant halt master
vagrant up master
```

5. Check the IP again and its working (note the inet value is not present and correct):
```
vagrant ssh master
ifconfig -a
enp0s8: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
[...]

Issue #8024, opened 2016-11-20
Network Error after upgrade to 1.8.7


nrganeshbabu:
### Vagrant version
Vagrant 1.8.7

### Host operating system
Windows 10 Pro version 1511

### Guest operating system
centos-70-x86_64

### Vagrantfile
```ruby
# Copy-paste your Vagrantfile here
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
	#config.ssh.private_key_path = "C:\\Users\\gb\\centos-7.0-x86_64\\vagrant_private_key.ppk"	
	#config.ssh.forward_agent = true	
	config.ssh.username = "vagrant"
    config.ssh.password = "vagrant"
	config.ssh.insert_key = false

  # Multi Machine config
  
	config.vm.define "master" do |master|
	master.vm.box = "centos7"
	master.vm.network "private_network", ip: "192.168.33.11", :adapter => 2	
	master.vm.hostname = "master"
	end
  
	config.vm.define "web1" do |web1|
	web1.vm.box = "centos7"
	web1.vm.network "private_network", ip: "192.168.33.12", :adapter => 2
	web1.vm.hostname = "web1"
	end
  
    config.vm.define "db" do |db|
	db.vm.box = "centos7"
	db.vm.network "private_network", ip: "192.168.33.13", :adapter => 2	
	db.vm.hostname = "db"
	end
   
  config.ssh.insert_key = false  
  config.vm.box = "centos7"
 
  #config.vm.network "private_network", ip: "192.168.33.10"
  
  
   config.vm.provider "virtualbox" do |vb|
	#vb.gui = true
     vb.memory = 4096
	 vb.cpus = 2
   end  
end

```

Please note, if you are using Homestead or a different Vagrantfile format, we
may be unable to assist with your issue. Try to reproduce the issue using a
vanilla Vagrantfile first.

### Debug output
==> master: Configuring and enabling network interfaces...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

# Down the interface before munging the config file. This might
# fail if the interface is not actually set up yet so ignore
# errors.
/sbin/ifdown 'enp0s8'
# Move new config into place
mv -f '/tmp/vagrant-network-entry-enp0s8-1479653162-0' '/etc/sysconfig/network-scripts/ifcfg-enp0s8'
# attempt to force network manager to reload configurations
nmcli c reload || true

# Restart network
service network restart

[...]

PR #8121, opened 2016-12-14
guests/redhat: Properly configure for NetworkManager

chrisroberts:
Adds guest detection for NetworkManager and properly configures
interface based on NetworkManager being present or absent.

Fixes: #8096 #8024


mikefaille:
I think `/etc/init.d/network restart` can manage network with and without network manager config. My explanation : https://github.com/mitchellh/vagrant/pull/8148

chrisroberts:
Fixed via #8148

Issue #8126, opened 2016-12-15
vagrant up fails with: Invalid argument - ruby_setenv() (Errno::EINVAL)


daveespo:
When `Vagrant::Util::Env.with_original_env` is called, it horks a 

Invalid argument - ruby_setenv() (Errno::EINVAL)

I've tracked this down to https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/util/env.rb#L9 and when dumping the contents of `Vagrant.original_env` to the console, I see that it has a blank string key (the complete stack trace is in the gist linked below):

````
Vagrant.original_env
[...]

Issue #7857, opened 2016-10-03
Error `Invalid argument - ruby_setenv (Errno::EINVAL)` on all vagrant command


marema31:
### Vagrant version
1.8.6

### Host operating system
Windows 7 64bits / Cygwin

### Expected behavior
At least a clearer error message -like the one on this [commit](https://github.com/mitchellh/vagrant/commit/1405395d813159a998a64fb82f9bd375786fc154)

### Actual behavior
Since update to 1.8.6 (from 1.8.4), every vagrant command does the following:
vagrant box list
C:/HashiCorp/Vagrant/embedded/gems/gems/bundler-1.12.5/lib/bundler.rb:18:in `replace': Invalid argument - ruby_setenv(BUNDLE_ORIG_PATH) (Errno::EINVAL)
        from C:/HashiCorp/Vagrant/embedded/gems/gems/bundler-1.12.5/lib/bundler.rb:18:in `<module:Bundler>'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/bundler-1.12.5/lib/bundler.rb:15:in `<top (required)>'
        from C:/HashiCorp/Vagrant/embedded/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:128:in `require'
        from C:/HashiCorp/Vagrant/embedded/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:128:in `rescue in require'
        from C:/HashiCorp/Vagrant/embedded/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:39:in `require'
        from C:/HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.8.6/bin/vagrant:69:in `<main>'

By removing some environment variables, this problem disapear (as in Issue #6017

### References
- GH-6017

ArthurZ:
In my case the issue appeared after I installed an update or new software.
So I fixed it by removing some what I deemed un-necessary entries in the System Path Environment Variable, then when it became = 1,888 chars Vagrant was able to start.

PR #8127, opened 2016-12-15
omits empty ENV key when updating

daveespo:
Addresses #8126

### DIFF
--- a/lib/vagrant/util/env.rb
+++ b/lib/vagrant/util/env.rb
@@ -6,7 +6,7 @@ module Vagrant
         if defined?(::Bundler) && defined?(::Bundler::ORIGINAL_ENV)
           ENV.replace(::Bundler::ORIGINAL_ENV)
         end
-        ENV.update(Vagrant.original_env)
+        ENV.update(Vagrant.original_env.delete_if {|k,v| k.nil? || k == ""})
         yield
       ensure
         ENV.replace(original_env.to_hash)


skoshy:
Making this change on Windows helped me overcome issue #7857, thanks @daveespo!

AlissonPQ:
Thanks @daveespo . I believe this fix adds a bit of an extra protection, but I'm not sure the community will accept it, since it kind of conceals a more serious problem. I found @BackSlasher in depth explanation here http://blog.backslasher.net/running-external-ruby-code-from-vagrant.html and his solution to the problem was added to the 1.9 milestone, here #7047 .

daveespo:
@AlissonPQ -- I don't entirely care about the nature of the fix ... I provided one solution but as long as the PR by @BackSlasher you reference also fixes the problem, that's fine. What's more frustrating is that both this PR and the one you reference have been open for a long time (6 weeks for this one, ONE YEAR for #7047) and there are a lot of frustrated Windows users out there that have to hand-patch vagrant to get their VMs launched

There's good reason to be cautious about accepting a fix, but not at the expense of leaving the app broken for months. Accepting my short-term, short-sighted fix would provide a lot of relief since it seems like #7047 is doomed to never get merged.

AlissonPQ:
True. I haven't used vagrant in six months or so and I was quite surprised
to find it broken this time around when I downloaded 1.9.x. Funny fact is I
downloaded an older version of both vagrant and chef thinking that would do
it, but it only worked a few times and the issue started again (don't ask
me why). Maybe they are trying to track down exactly what's causing the
problem and release a definitive solution. Who knows?

Your fix is a standard way of adding extra security to calls that reference
strings, it eliminates the symptoms. They could accept it and open an issue
to track down the real problem. I agree that leaving the software broken
for windows users is very lame.

chrisroberts:
Fixed via #8198 with test coverage. Thanks for this PR!

daveespo:
Thank you for merging

Issue #7619, opened 2016-07-21
Setting for the ansible provisioner to specify the location of generated inventory file


discopatrick:
This is a suggested enhancement, that I'm willing to contribute to.

I use the vagrant ansible provisioner. My projects usually begin with me using the vagrant boxes only, at which time I set the generated inventory file at .vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory as my inventory (in ansible.cfg). This is because I often like to run my playbooks manually via `ansible-playbook`, rather than automatically via `vagrant up` or `vagrant provision`.

Later in the project, I begin to add remote servers into my ansible project. What this usually means is creating my remote server inventory files in a hosts/ directory, and then changing my ansible.cfg setting to point to this new directory. However, I sometimes want to run playbooks against both my vagrant and remote hosts, and so I still need the information in the generated inventory file. The problem is, you can't specify multiple inventory locations in ansible - you can only specify a single file or directory.

Therefore, the enhancement I'm suggesting is that there should be a setting for the vagrant ansible provisioner that allows you to specify the location of your generated inventory file - which I would then set to my ansible hosts/ dir.

For perspective, I have also come up 4 other possible solutions:

1. Duplicate the generated file into my hosts/ dir - not ideal, as it's not very DRY, and my duplicate won't update itself if the information changes.
2. Create a symlink inside my hosts/ dir pointing to the generated file - this has worked well in most cases, but there is an edge case where running the same ansible project on a remote host via ansible pull will complain that the symlinked file doesn't exist (because `vagrant provision` hasn't been run there).
3. Write a dynamic inventory script (placed in my hosts/ dir) that observes the generated file and returns the correct values based on that.
4. Extend ansible to allow for multiple inventory directories (it currently only allows you to set one file or dir).

I'm posting here to gauge interest and get feedback on this possible enhancement. I'm happy to get stuck in myself and produce a pull request, if the vagrant team would be open to including this feature. I don't have a lot of experience with ruby, but it seems quite a simple feature to get started with, and I'd be happy to follow guidance from the team to make sure the code is of acceptable quality.

cmarchent:
I have a similar problem; but have existing inventories for real deployments made with ansible that I wan to be able to re-create as vagrant images (to test before we deploy to Live). The ansible inventories contain key information like the groups that the host belongs to meaning that we need to go through a lengthy and error prone process of copying the inventory config into a Vagrant file - and given the number of deployments we are working with; this needs automation.

[...]

Issue #7552, opened 2016-07-07
ansible_ssh_host deprecated


DerekTBrown:
So, two concerns:

1) As of the latest version of Ansible, the ansible_ssh_host variable is now deprecated in favor of ansible_host, this needs to be updated.

2) ansible_host has broader reaching applications than just ssh (people are using ansible_host as a standin for the servers' DNS or IP Address).   Therefore, it is not always best that Vagrant sets this to be 127.0.0.1, for instance, the case where the user has a private network configured. 

Therefore, I think Vagrant needs an ansible_host variable, which defaults to 127.0.0.1, but can be easily overridden.

gildegoma:
Thank you Derek for reporting these two points.

The first point is already tracked as part of #6570. I hope to find some free time to work on these changes as of August-September, but help via pull request would be great (even for the resolution of some parts of all the points listed in #6570 :)

For your second point, the IP Address and SSH port set by the `ansible` provisioner depends on the address returned by the machine provider. The IP address is typically `127.0.0.1` for the VirtualBox provider which uses port-forwarding to provide SSH access, but it is not the case with many other providers (obviously for Cloud providers, but also for local machines or containers provided by libvirt, VMware, Parallels, Docker, ...). 

The _alternative_ when using a private network and VirtualBox can be to rely on an Ansible static inventory file. This requires to maintain two copies of the IP addresses (one in Vagrantfile and one in the Ansible inventory), which is not ideal but acceptable in many situations.

I'd like to close this issue, but if you think we should further investigate about how to improve the point 2, please provide more details about your specific use case. Thanks again!

DerekTBrown:
@gildegoma in regard to the static inventory, that was my thought initially as well, but this has an issue in that you cannot use Vagrant's dynamically created private key (which is stored in a different directory based on execution).  That's why just a simple if statement which would allow the user to overwrite the ansible_ssh_host or ansible_host if set in the hostvars would be ideal.

gildegoma:
> you cannot use Vagrant's dynamically created private key

The pattern for the private key path is: `.vagrant/machines/<machine-name>/<machine-provider>/private_key`. I don't see why you cannot specify these paths in your static inventory.
Note that depending of your context, you also can keep the default insecure key: `config.ssh.insert_key = false`.

> a simple if statement which would allow the user to overwrite the ansible_ssh_host or ansible_host if set in the hostvars would be ideal

[...]

PR #8131, opened 2016-12-17
Allow ansible host and port to be overridden

cmarchent:
Re #7619

When using ansible provisioning with static inventories it is not possible to provision unless you know the IP address or forwarded ssh port for for the VM (and hence can hardcode them). To resolve this I extend the concept used for ansible.force_remote_user to include remote_host and remote_port; to pass these items from the vagrants ssh_config down via ansible extra-vars overriding defaults or those specified in the inventory. 
The result is that I can rapidly replicate and of my deployed configurations into a Vagrant host.


gildegoma:
Thank you @cmarchent for this pull request. I'll try to take some time before End of Year to review it.

I think this PR would also resolve needs requested in #7552 (to be confirmed).

cmarchent:
One thing for consideration; I cannot see a circumstance where the user would want to force either ip or port separately; perhaps the ansible.force_remote_host and ansible.force_remote_port should be replaced with the single option ansible.force_remote_socket?

gildegoma:
@cmarchent good point about "host + port" (~socket) combination.

Please have a look at https://github.com/mitchellh/vagrant/issues/7552#issuecomment-268369996, where I am even more convinced that we should combine ideas and needs, to build an elegant solution.

chrisroberts:
As this was part of a larger issue discussed in other various issues and never gained more traction I'm going to close this up. Cheers!

PR #8092, opened 2016-12-07
Fix quotes

ephemeralsnow:
We use boot2docker.

Succeed in Vagrant 1.8.6.
```
DEBUG guest: Trying: tinycore
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute: cat /etc/issue | grep 'Core Linux' (sudo=false)
DEBUG ssh: Exit status: 0
 INFO guest: Detected: tinycore!
```

Fails in Vagrant 1.9.0 and 1.8.7.
```
DEBUG guest: Trying: tinycore
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute: if test -r /etc/os-release; then
source /etc/os-release && test xCore Linux = x$ID && exit
fi
if test -x /usr/bin/lsb_release; then
/usr/bin/lsb_release -i 2>/dev/null | grep -qi Core Linux && exit
fi
if test -r /etc/issue; then
cat /etc/issue | grep -qi Core Linux && exit
fi
exit 1
 (sudo=false)
DEBUG ssh: stderr: sh: Linux: unknown operand

DEBUG ssh: stderr: grep: Linux: No such file or directory

DEBUG ssh: Exit status: 1
```

Corrected execution result.
```
DEBUG guest: Trying: tinycore
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute: if test -r /etc/os-release; then
source /etc/os-release && test 'xCore Linux' = "x$ID" && exit
fi
if test -x /usr/bin/lsb_release; then
/usr/bin/lsb_release -i 2>/dev/null | grep -qi 'Core Linux' && exit
fi
if test -r /etc/issue; then
cat /etc/issue | grep -qi 'Core Linux' && exit
fi
exit 1
 (sudo=false)
DEBUG ssh: Exit status: 0
 INFO guest: Detected: tinycore!
```

PR #8134, opened 2016-12-19
Fix broken OS detection for IDs with space

artem-sidorenko:
Guest detection logic was changed in GH-7887, however it breaks if
`GUEST_DETECTION_NAME` contains a space.

Example:
```bash
$ if test -r /etc/os-release; then source /etc/os-release && test xLinux Mint = x$ID && exit; fi
bash: test: too many arguments
```


chrisroberts:
Fixed via #8092. Thanks!

PR #8135, opened 2016-12-19
Fix grepping for GUEST_DETECTION_NAME with spaces

infernix:
Some GUEST_DETECTION_NAME definitions use white spaces but the grep commands do not quote the variable while checking. 

That causes guest detection to fail for those OSes since grep uses arguments after the witespace as files instead of parsing stdin.


chrisroberts:
Fixed via #8092. Thanks!

Issue #7988, opened 2016-11-10
vagrant uses incorrect name of network interface for bridged interface


vinod-990:
Dear Developers,

This is continuation of bug that I reported in https://github.com/mitchellh/vagrant/issues/7858 . I have updated to vagrant 1.8.7 and still had the same issue:

```
INFO global: Vagrant version: 1.8.7
INFO global: Ruby version: 2.2.5
INFO global: RubyGems version: 2.4.5.1
...
DEBUG guest: Searching for cap: network_interfaces
DEBUG guest: Checking in: redhat
DEBUG guest: Checking in: linux
DEBUG guest: Found cap: network_interfaces in linux
 INFO guest: Execute capability: network_interfaces [#<Vagrant::Machine: RHEL-7.2 (VagrantPlugins::ProviderVirtualBox::Provider)>] (redhat)
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute: /sbin/ip -o -0 addr | grep -v LOOPBACK | awk '{print $2}' | sed 's/://' (sudo=true)
DEBUG ssh: Exit status: 0
DEBUG network_interfaces: Unsorted list: ["eth0", "eth1", "eth0.1@eth0"]
DEBUG network_interfaces: Sorted list: ["eth0", "eth0.1@eth0", "eth1"]
DEBUG network_interfaces: Ethernet preferred sorted list: ["eth0", "eth0.1@eth0", "eth1"]
DEBUG ssh: Uploading: /tmp/vagrant-redhat-configure-networks20161110-29035-19ldhw3 to /tmp/vagrant-network-entry-eth0.1@eth0-1478782550-0
DEBUG ssh: Re-using SSH connection.
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute: # Down the interface before munging the config file. This might
# fail if the interface is not actually set up yet so ignore
# errors.
/sbin/ifdown 'eth0.1@eth0'
# Move new config into place
mv -f '/tmp/vagrant-network-entry-eth0.1@eth0-1478782550-0' '/etc/sysconfig/network-scripts/ifcfg-eth0.1@eth0'
# attempt to force network manager to reload configurations
nmcli c reload || true

....
Job for network.service failed because the control process exited with error code. See "systemctl status network.service" and "journalctl -xe" for details.

ERROR vagrant: /opt/vagrant/embedded/gems/gems/vagrant-1.8.7/plugins/communicators/ssh/communicator.rb:240:in `execute'
[...]

Issue #7935, opened 2016-10-27
Installation of docker causes network failure after reboot


chingc:
Hi, I'm experiencing a network connection issue when I have docker installed in the guest.  The problem is that I lost my network connection if I were to shutdown or reboot the guest.

I'm using Vagrant 1.8.6 on OS X 10.11.6 (El Capitan).

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|

  config.vm.box = "ubuntu/trusty64"
  config.vm.network "public_network"
  config.vm.provision "docker"

end
```

### Steps to Reproduce
1. `vagrant up --provider=virtualbox` (when asked about the bridged network I choose the first option, for me it's `en0: Wi-Fi (AirPort)`
2. `vagrant ssh`
3. `ping google.com` (pings are successful)
4. Cancel ping and exit out of the VM
5. `vagrant halt`
6. `vagrant up` (when asked about the bridged network I choose the same option as before)
7. `vagrant ssh`
8. `ping google.com` (pings fail with `connect: Network is unreachable`)

### Expected
Ping should be successful both times.

### Debug output
1. [vagrant up (initial)](https://gist.github.com/chingc/e7fb0f80568763f24e8562ff58a58180)
2. [vagrant halt](https://gist.github.com/chingc/d0c4986ceb838b5c8f3fa9b31984770f)
3. [vagrant up](https://gist.github.com/chingc/466a080deb7c43e82fff5bbb9883bbc2)

### Notes
I've tried this using the latest debian/jessie64 as the base box and get similar results.

The issue seems to have something to do with the `public_network` setting since commenting out that line gives the expected results, but I need this machine to be using a bridged adapter.

sgehrig:
Seems to be related to #7858, #7458 and/or #7876 and should be fixed with PR #7866.

chingc:
Thanks for the update!  Looking forward to `1.8.7`.  I'll close the ticket.

chingc:
Just tried `v1.8.7` and I can no longer reproduce this issue. 👍 👍

Issue #7923, opened 2016-10-24
Tries to assign VM wrong IP addresss


tleyden:
### Vagrant version

Fails on 1.8.6
Works on 1.8.1
### Host operating system

OSX El Capitan
### Guest operating system

CentOS7
### Vagrantfile

``` ruby

Vagrant.configure(2) do |config|

  host_ip = {
    "couchbase-server" => "192.168.34.10",    
    "sync-gateway1" => "192.168.34.11",
    "sync-gateway2" => "192.168.34.12",
    "nginx" => "192.168.34.13",    
    "sync-gateway3" => "192.168.34.14",    
  }

  host_ip.each do |host_name, ip|
      config.vm.define host_name do |host|
        host.vm.box = "couchbase-mobile-training"
        host.vm.hostname = host_name
        host.vm.network "private_network", ip: ip
        host.vm.synced_folder ".", "/tmp/sync", disabled: true
        host.vm.provider "virtualbox" do |vb|
            # Customize the amount of memory on the VM:
            vb.memory = "2048"

            # Allow vm to send data via VPN
            vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
        end
      end
  end

end
```
### Debug output

https://gist.github.com/a3269e1b9faed297887cf52ec86a37f9
### Expected behavior

It should start up the VM cluster without errors
### Actual behavior

The VM startup fails with this output:

https://gist.github.com/tleyden/ab07fc01f68dddc4db64de05d7387f2c#file-stdout

More logs from the VM available here: https://gist.github.com/tleyden/ab07fc01f68dddc4db64de05d7387f2c
### Steps to reproduce
1. `wget http://cbmobile-bucket.s3.amazonaws.com/training-virtual-machines/couchbase-mobile-training.box`
2. Save `Vagrantfile` listed in this ticket
3. `vagrant up`

mikefaille:
@tleyden

mikefaille:
@tleyden Could you upgrade Vagrant to 1.8.7 ?

chrisroberts:
Hi there,

It looks like this has been resolved within a previously shipped version of Vagrant so I am now closing this issue. If the original issue was not fully resolved, please reopen this issue or create a new one.

Cheers!

Issue #7876, opened 2016-10-07
Hang at `Configuring and enabling network interfaces...` when using private_network with a static IP


karlkfi:
### Vagrant version
1.8.6

### Host operating system
Mac OS 10.10.5

### Guest operating system
CentOS 7

### Host-only Network
```bash
network_name="$(VBoxManage hostonlyif create | grep "successfully created" | sed "s/Interface '\(.*\)' was successfully created/\1/")"
VBoxManage hostonlyif ipconfig --ip 192.168.65.1 "${network_name}"
```

### Vagrantfile
```ruby
Vagrant.configure(2) do |config|
  config.vm.define "dcos-centos-virtualbox" do |vm_cfg|
    vm_cfg.vm.box = 'mesosphere/dcos-centos-virtualbox'
    vm_cfg.vm.box_url = 'http://downloads.dcos.io/dcos-vagrant/metadata.json'
    vm_cfg.vm.box_version = '0.7.0'

    vm_cfg.vm.provider "virtualbox" do |v, override|
      v.name = "dcos-centos-virtualbox"
      v.cpus = 2
      v.memory = 2048
      v.customize ['modifyvm', :id, '--natdnshostresolver1', 'on']
      override.vm.network :private_network, ip: '192.168.65.10'
    end
  end
end
```

### Expected behavior

1. Machine should initially come up with a DHCP IP (enp0s3).
2. Vagrant should add a new iface (enp0s8) to be a Static IP iface (on the host-only network).

### Actual behavior

1. Machine initially comes up with a DHCP IP (enp0s3).
2. Vagrant hangs executing the script to reconfigure the network interfaces.

Last ui output is `Configuring and enabling network interfaces...`

Vagrant seems to be hanging inside of the redhat-specific version of `configure_networks`. 

Using a monkeypatch to replace configure_networks and add logging, I was able to determine that `/etc/sysconfig/network-scripts/ifcfg-enp0s3` is being reconfigured from a dhcp iface to a static IP iface. Then the dchp iface is being brought down while vagrant is still connected. Vagrant then hangs - the ssh command never returns.

### Steps to reproduce
```
vagrant up
```

### References

Sounds like it might be related to https://github.com/mitchellh/vagrant/pull/7866

karlkfi:
[...]

PR #8148, opened 2016-12-22
Revert "Fix `service network restart` on RHEL-7 / Fedora"

mikefaille:
Fix theses bugs : 
https://github.com/mitchellh/vagrant/issues/8142 https://github.com/mitchellh/vagrant/issues/8096

For RHEL7 and Fedora 25 `/etc/init.d/network restart` already  shutdown interfaces and restart NM if needed.

In start() :
```
        if [ "$(LANG=C nmcli -t --fields running general status 2>/dev/null)" = "running" ]; then
                nmcli connection reload
        fi

```

In stop() :

```
       for i in $vpninterfaces $xdslinterfaces $bridgeinterfaces $vlaninterfaces $remaining; do
                unset DEVICE TYPE
                (. ./ifcfg-$i
                if [ -z "$DEVICE" ] ; then DEVICE="$i"; fi

                if ! check_device_down $DEVICE; then
                   action $"Shutting down interface $i: " ./ifdown $i boot
                   [ $? -ne 0 ] && rc=1
                fi
                )
        done
```
Where $remaining include all "others" interfaces including eth*

It work for `NM_CONTROLLED=X` where X = yes or no 

This reverts commit 166d10d4e16836e908be0bd94c2f671cac2b38b4.

### DIFF
--- a/plugins/guests/redhat/cap/change_host_name.rb
+++ b/plugins/guests/redhat/cap/change_host_name.rb
@@ -31,12 +31,8 @@ module VagrantPlugins
                 sed -i'' '1i 127.0.0.1\\t#{name}\\t#{basename}' /etc/hosts
               }
 
-              # Restart network (through NetworkManager if running)
-              if service NetworkManager status 2>&1 | grep -q running; then
-                service NetworkManager restart
-              else
-                service network restart
-              fi
+              # Restart network
+              service network restart
             EOH
           end
         end
--- a/plugins/guests/redhat/cap/configure_networks.rb
+++ b/plugins/guests/redhat/cap/configure_networks.rb
@@ -36,7 +36,7 @@ module VagrantPlugins
 
             # Add the new interface and bring it back up
             final_path = "#{network_scripts_dir}/ifcfg-#{network[:device]}"
-            commands << <<-EOH.gsub(/^ {14}/, '')
+            commands << <<-EOH.gsub(/^ */, '')
               # Down the interface before munging the config file. This might
               # fail if the interface is not actually set up yet so ignore
               # errors.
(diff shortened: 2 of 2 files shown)


karlkfi:
Works for me.

We still don't have an explanation of why the change this reverts was made in the first place. But that shouldn't block the fix tho, imo. Even with an explanation, that change wasn't a valid, well tested fix without side effects.

mikefaille:
@karlkfi Here the original pull request : https://github.com/mitchellh/vagrant/pull/8052

karlkfi:
Yes, I know. I asked the same thing there and got no answer.

mikefaille:
I see. The link on original pull request could help to trace changes.

karlkfi:
Hmm, you're right. I don't remember seeing that link before. 

I wonder if it's the same as #7923. There have been a rash of networking bugs in the last few Vagrant releases. Both issues could have been caused by one of the iface sorting bugs.

karlkfi:
recent network interface issues: #7844 #7858 #7988 #7935 #7668 #7876 #7458

mikefaille:
As I understand, network interfaces must be manages appropriately depending process owner (ex : vagrant or docker for bridge). It's why we short interfaces. Isn't ?

At least, in issues mentioned initially in my pull request we really need to fix them fast cause I think it cause more damages than we think. And, network configuration in this pull request is clean and predictable as Debian config : https://github.com/mikefaille/vagrant/blob/96611341a96d7d19fdade5556a110b22c6add22b/plugins/guests/debian/cap/configure_networks.rb

jairojunior:
Only saw this PR now. #8166 is also related to this.

mikefaille:
Thank you @chrisroberts for this follow up.

alexivkin:
still broken in 1.9.1.
The actual fix is to add 
              /sbin/ifup '#{network[:device]}'
right after 
              nmcli c reload || true
in 
/plugins/guests/redhat/cap/configure_networks.rb

mikefaille:
> still broken in 1.9.1. 
@alexivkin It makes sense. My fixe was based on 1.9.1. Maybe, you should wait for Vagrant 1.9.2 https://github.com/mitchellh/vagrant/commit/8e7c267b3f008fdb906366cf2b1ff6efb822c4ea or use Vagrant 1.9.0 https://releases.hashicorp.com/vagrant/1.9.0/

meekrosoft:
It seems like this is also a bug on cents/7 - should i I open a new issue for this?

mikefaille:
@meekrosoft Nope :-) Since you got the same bugs as RHEL and RHEL == Centos, I prefer to modify my issue description.

PR #8158, opened 2016-12-28
Quote ansible host variables

ekohl:
This allows using spaces in hash arguments.

I'm unsure if this is the proper fix or if it's needed in more places. Somebody more familiar with Vagrant and Ansible should check this.

### DIFF
--- a/plugins/provisioners/ansible/provisioner/base.rb
+++ b/plugins/provisioners/ansible/provisioner/base.rb
@@ -148,7 +148,7 @@ module VagrantPlugins
           end
           s = nil
           if vars.is_a?(Hash)
-            s = vars.each.collect{ |k, v| "#{k}=#{v}" }.join(" ")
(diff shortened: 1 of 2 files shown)


gildegoma:
Closing as this fix is not necessary.

You can easily add the required quotes in your variable definitions. For example:

```ruby
          ansible.host_vars = {
            'machine1' => {"http_port" => 8080, comments: "'once upon a time...'"}
          }
```

and you'll get something like this in the generated inventory:

```
machine1 ansible_ssh_host=... http_port=8080 comments='once upon a time...'
```

That said, I agree that's not so obvious. For this reason I added (in 9493901e0cb5f93c2cf020ff07e9a98a182b3007) an example in the [docs](https://www.vagrantup.com/docs/provisioning/ansible_common.html#host_vars) and in the unit tests. I hope this will help people in the future.

ekohl:
Thanks, it makes sense when you point it out and glad you documented it. This is better.

PR #8172, opened 2017-01-04
Fix linked clone backwards compatibility example

mattkenn4545:
WIth the release of 1.9 the regex is invalid


chrisroberts:
Docs have been updated. Thanks for the PR!

PR #8173, opened 2017-01-05
Adding diff option for Ansible provisioner

jtyr:
This patch is adding support for the `--diff` command line option for the Ansible provisioner.


gildegoma:
@jtyr Thank you for this proposition. For a simple Ansible option (with no arguments) like `--diff`, I recommend to simply rely on the multi-purpose [`raw_arguments` option](https://www.vagrantup.com/docs/provisioning/ansible_common.html#raw_arguments). 

As a "maintenance rule of thumb": The Vagarnt Ansible provisioners are not thought to proxy all the possible options in Ansible (that would be lot of effort for very little gains). 

For these reason, I am going to close this PR. I hope you agree + understand. If not feel free to comment further...

PR #8188, opened 2017-01-12
Warn about changes to CWD on every machine action

Ferdy89:
This is an attempt to implement https://github.com/mitchellh/vagrant/issues/3921 @chrisroberts could you please take a look?

Whenever the path where the machine was first created changes, Vagrant will now show just one warning when an action is run on the machine. The idea is that if a user copies the machine over to a different directory with the intention of running two different machines, this warning will now help the user determine how to make that work.

This PR lacks tests, I'll add those once we can agree on a correct implementation.

### DIFF
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -170,6 +170,8 @@ module Vagrant
       # Extra env keys are the remaining opts
       extra_env = opts.dup
 
+      check_cwd
+
       # Create a deterministic ID for this machine
       vf = nil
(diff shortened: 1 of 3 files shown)


Ferdy89:
@chrisroberts sorry for the delay! I wrote a test for the feature, let me know how that looks

briancain:
@Ferdy89 - thank you so much for your contribution! I've picked up the work and created a new pull request to get merged here: https://github.com/mitchellh/vagrant/pull/8651

PR #8193, opened 2017-01-12
adding Cumulus as guest plugin

8bitdad:
This is to add Cumulus Linux as a guest to Vagrant.

### DIFF
--- a/plugins/guests/cumulus/cap/change_host_name.rb
+++ b/plugins/guests/cumulus/cap/change_host_name.rb
@@ -1,0 +1,44 @@
+module VagrantPlugins
+  module GuestDebian
+    module Cap
+      class ChangeHostName
+        def self.change_host_name(machine, name)
+          comm = machine.communicate
+
+          if !comm.test("hostname -f | grep '^#{name}$'", sudo: false)
+            basename = name.split(".", 2)[0]
+            comm.sudo <<-EOH.gsub(/^ {14}/, '')
+              # Set the hostname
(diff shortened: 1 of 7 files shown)


chrisroberts:
This sounds like a reasonable idea, but this has low chance of being a priority for us anytime soon, so we'd rather close this issue than have it stagnate into the foreseeable future. The best way to get this implemented would be to submit a PR. If you do so, we'd welcome this work. If you want to discuss how to go about doing this, I'd be happy to help with that.

Again, thank you for suggesting this and it _is_ a reasonable idea. But, since this is a feature request, it requires some work to add it to the project and we don't plan on doing that soon. But as this is also an open source project we welcome contributions. Thanks!

MaxPeal:
@8bitdad can you sign the CLA?

8bitdad:
CLA is signed, but is not updating.

briancain:
Hi @8bitdad - apologies for how long it took for us to address this PR. A few questions:

Is there a reason why this is needed over the default debian guest? It also looks like the majority of these guest methods are copied from the existing guest plugin, when they could just inherit from the parent debian guest class instead of copy/pasting the code. And is there something specific to Cumulus that prevents it from using the Debian guest plugin?

briancain:
Hey there - I'm going to close this from a lack of response. Thanks again for opening a pull request with Vagrant. If this is something that someone has time for in the future, please feel free to open a new pull request again. Thanks!

Issue #8214, opened 2017-01-18
SMB credential are not used.


SaschaMzH:
### Vagrant version
Vagrant 1.9.1

### Host operating system
Windows 7

### Guest operating system
Ubuntu 

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "kaorimatz/ubuntu-16.04-amd64"

  config.vm.synced_folder "c:\\projects\\", "/c/projects", type: "smb"
end
```

### Expected behavior
Use the credential which are typed into the console during the `vagrant up`.

### Actual behavior
The credentials which are type into the console are not used. Furthermore the local smb share is read and writeable for everyone.

### Steps to reproduce
1. `vagrant up`
2. Look into the newtork settings from the created smb share.

chrisroberts:
Hi there,

It looks like this has been resolved within a previously shipped version of Vagrant so I am now closing this issue. If the original issue was not fully resolved, please reopen this issue or create a new one.

Cheers!

PR #8215, opened 2017-01-18
Use a generic login mechanism for the SMB shares.

SaschaMzH:
This is a patch for the issue #8214 to use the smb share.
It propose a username (this could also be a domain plus username) and asked for the password.
The generated smb share on the host is using this credential to setup the user rights correctly.


chrisroberts:
Hi there,

Cheers!

PR #8244, opened 2017-02-01
Delete empty keys from ENV

maurodx:
An ENV with a value with empty key "" stops vagrant: the solution i've found is to remove all key/value pairs with a key length of 0 chars.


chrisroberts:
Closing as dup of #8198. Thanks for the PR!

PR #8289, opened 2017-02-16
Add proper path translation for WSL (Windows Subsystem for Linux) installations

### LINKED ISSUE #7731: Use Linux Subsystem on Windows
### Vagrant version

Vagrant 1.8.5
### Host operating system

Windows 10.0.14393
### Guest operating system

Ubuntu 14.04.5 LTS
### Expected behavior

Vagrant should be able to use Linux Subsystem on Windows 10 😎
Instead, it expects cygwin/mingw/git-bash. It's much more awesome, more up-to-date, more native, and well integrated. It makes total sense to use it when installed, especially considering Vagrant appears to be made exclusively for Linux and then ported (or something) to Windows.
### Actual behavior

This error ☹

```
`ssh` executable not found in any directories in the %PATH% variable. Is an SSH client installed? Try installing Cygwin, MinGW or Git, all of which contain an SSH client. Or use your favorite SSH client with the following authentication information shown below:
```
### Steps to reproduce
1. Install the Windows 10 summer upgrade to build 14393 (also referred to as Version 1607)
2. Try running `vagrant ssh` on any vagrant box. It doesn't matter at all which one.
### References

Just to be absolutely clear, here's how to get Linux Subsystem (commonly referred to as Bash for Windows) and to clarify which is the one I mean:
http://www.howtogeek.com/249966/how-to-install-and-use-the-linux-bash-shell-on-windows-10/

jcherniak:
This patch allows you to run vagrant inside of Ubuntu on WSL if you do the following (inspiration from https://github.com/Microsoft/BashOnWindows/issues/733):

1. Be running a build of Windows that supports running Windows programs inside of WSL (build 14951 or higher).
2. Install the WINDOWS version of Virtualbox.
3. Symlink VBoxManage to the expected path in Linux:
`sudo ln -s "/mnt/c/Program Files/Oracle/VirtualBox/VBoxManage.exe" /usr/bin/VBoxManage`.

This patch provides the proper path translation for WSL's mounting scheme, similar to the cygwin one I took inspiration from.

This should fix  #7731 or at least much of it as well as Microsoft/BashOnWindows#733

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -33,6 +33,22 @@ module Vagrant
           end
         end
 
+        def wsl?
+          return @_wsl if defined?(@_wsl)
+          @_wsl = -> {
+            platform = File.read('/proc/sys/kernel/osrelease')
+            return platform.include?("Microsoft")
+          }.call
+          return @_wsl
+        end
+
+        [:darwin, :bsd, :freebsd, :linux, :solaris].each do |type|
+          define_method("#{type}?") do
+            platform.include?(type.to_s)
+          end
+        end
+
+
         def windows?
           return @_windows if defined?(@_windows)
           @_windows = %w[mingw mswin].any? { |t| platform.include?(t) }
@@ -105,6 +121,16 @@ module Vagrant
             end
           end
 
+          if wsl?
+            begin
+              path = path.gsub("\\", "/")
+              drive = path.chr.downcase
+              path = '/mnt/' + path.chr.downcase + '/' + path.sub(drive + ":", "")
+
+              return path
+            end
+          end
+
           # Sometimes cygpath isn't available (msys). Instead, do what we
           # can with bash tricks.
           process = Subprocess.execute(
@@ -120,6 +146,24 @@ module Vagrant
         #
         # @return [String]
         def cygwin_windows_path(path)
+          if wsl?
+            if path.start_with?("/mnt")
+              path = path.gsub('/mnt/', '')
+              home = path.chr.upcase + ":"
+              path.slice!(0)
+            else
+              process = Subprocess.execute(
+                "/mnt/c/Windows/System32/cmd.exe",
+                "/C \"echo %LocalAppData%\""
+              )
+              home = process.stdout.chomp.gsub("\"", "") + "\\Lxss"
+            end
+
+            path = home + path.gsub("/", "\\").gsub("\"", "")
+
+            return path
+          end
+
           return path if !cygwin?
 
           # Replace all "\" with "/", otherwise cygpath doesn't work.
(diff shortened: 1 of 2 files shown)


lib/vagrant/util/platform.rb
```diff
@@ -33,6 +33,22 @@ def cygwin?
           end
         end
 
+        def wsl?
+          return @_wsl if defined?(@_wsl)
+          @_wsl = -> {
+            platform = File.read('/proc/sys/kernel/osrelease')
+            return platform.include?("Microsoft")
+          }.call
+          return @_wsl
```
chrisroberts: Any reason for using a `proc` here? Should be able to reduce this and add an existence check as well:

```ruby
if @_wsl.nil?
  if File.exist?("/proc/sys/kernel/osrelease")
    platform = File.read("/proc/sys/kernel/osrelease")
    @_wsl = platform.include?("Microsoft")
  else
    @_wsl = false
  end
end
@_wsl

File: lib/vagrant/util/platform.rb [hunk 2]
```diff
@@ -33,6 +33,22 @@ def cygwin?
...
+        end
+
+        [:darwin, :bsd, :freebsd, :linux, :solaris].each do |type|
+          define_method("#{type}?") do
+            platform.include?(type.to_s)
+          end
+        end
```
chrisroberts: Looks like an accidental copy here.

lib/vagrant/util/platform.rb
jcherniak: Honest answer: this is the first time I've written Ruby code and I copied it from above :)

robwiss:
After this patch I still have an issue with boxes that are stored on the local filesystem and accessed via relative path. e.g.
```
  config.vm.box = "../../packer/my_box.box"
end
```

When I try to bring the machine up I get an error:
```
$ vagrant up
/opt/vagrant/embedded/gems/gems/vagrant-1.9.3/lib/vagrant/util/platform.rb:24: warning: Insecure world writable dir /mnt/c in PATH, mode 040777
==> default: Box '../../packer/my_box.box' could not be found. Attempting to find and install...
    default: Box Version: >= 0
==> default: Box file was not detected as metadata. Adding it directly...
==> default: Adding box '../../packer/my_box.box' (v0) for provider: virtualbox
    default: Downloading: ../../packer/my_box.box
An error occurred while downloading the remote file. The error
message, if any, is reproduced below. Please fix this error and try
again.

Couldn't open file /C:/Users/username/project/packer/my_box.box
```

It seems to be related to the downloader method in `lib/vagrant/action/builtin/box_add.rb` which uses `Util::Platform.cygwin_windows_path` when in this case a windows path breaks things.

Making the following change to line 390 of that file fixes the issue for me but I don't know if it's a good fix:
```
if !Util::Platform::wsl?
    file_path = Util::Platform.cygwin_windows_path(file_path)
end
```

jcherniak:
Fixed by #8570

PR #8296, opened 2017-02-20
Add hint for .verbose-attribute

florianb:
Since the verbose-attribute isn't explicitly mentioned in the salt-provisioner it is hard to find out it can (and must) be set to get the result of the salt highstate-command.


florianb:
I have to thank! 👍

PR #8302, opened 2017-02-25
Website typography update

jasoncostello:
:wave: This PR is part of a larger intiative by the HashiCorp design team to create a more consistent experience across OSS project sites. These commits update the typography of the homepage and update the logo. The changes proposed are:

• Replace instances of Museo Sans with Open Sans and Klavika fonts
• Replace the logotype with Klavika
• In the Docs, the only change is Headings are set in Open Sans, to match our other docs
• Unique to this site, I've modified the design of the header. The Vagrant graphic is really cool but the `<h1>` being positioned on top of it, dark text on a medium value image is hard to read and may present accessibility issues. Recent HashiCorp design styles avoid placing type over busy graphics. I thought it'd be best to let the graphic be positioned unobstructed, so I moved the `<h1>` below it and onto one line. I resized (barely perceivable) things so the header is virtually the same height.
• A few mobile tweaks to the header to keep the proportions feeling right and everything fitting

**It looks like this**

screencapture-localhost-4567-1487895413522

cc/ @captainill


jasoncostello:
@chrisroberts Thanks ✨

PR #8308, opened 2017-02-27
Update snapshot.html.md

nathanielks:
Updated docs to specify the use of the `vm-name` when restoring a snapshot.


nathanielks:
All in a days work, @chrisroberts!

PR #8310, opened 2017-02-28
guests/esxi: Added public_key capability

tsmolka:
I have been experimenting with ESXi guests with Vagrant and implemented `insert_public_key` and `remove_public_key` capabilities. 

At least on my ESXi v6.5 there is `AuthorizedKeysFile /etc/ssh/keys-%u/authorized_keys` specified in `/etc/ssh/sshd_config`. This is the reason why `SSH_DIR=...` one-liner is used. Otherwise it's very similar to linux guest capability. 

I have successfully tested with Vagrant v1.8.6 on Windows against ESXi v6.5 and haven't experienced any issues.


tsmolka:
I will also add simple unit test to cover this cap.

PR #8312, opened 2017-02-28
Update vagrant gem dependencies

myoung34:
rspec is locked to 3.5.0 in the lockfile but the gemspec  requires `~>2.14.0`:

```
bundle install
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Fetching dependency metadata from https://rubygems.org/
You have requested:
  rspec ~> 2.14.0

The bundle currently has rspec locked at 3.5.0.
Try running `bundle update rspec`

If you are updating multiple gems in your Gemfile at once,
try passing them all to `bundle update`

```

also gems like test-kitchen and inspec which end up wanting `net-scp ~>1.2` and `1.1` is way out of date making it incompatible.

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency "log4r", "~> 1.1.9", "< 1.1.11"
   s.add_dependency "net-ssh", "~> 3.0.1"
   s.add_dependency "net-sftp", "~> 2.1"
-  s.add_dependency "net-scp", "~> 1.1.0"
+  s.add_dependency "net-scp", "~> 1.2.0"
   s.add_dependency "rb-kqueue", "~> 0.2.0"
   s.add_dependency "rest-client", ">= 1.6.0", "< 3.0"
   s.add_dependency "wdm", "~> 0.1.0"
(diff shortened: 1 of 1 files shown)


chrisroberts:
Hi there. Thanks for this PR! Before the constraint on rspec can be relaxed some work must be done to make the tests within Vagrant compatible with the newer versions of rspec. There is no Gemfile.lock checked into the repository so running a `bundle update` locally should resolve the issue you are encountering.

Cheers!

myoung34:
@chrisroberts that's not possible given how plugins work is it?

```
source 'http://rubygems.org'

group :development do
  gem 'vagrant', :git => 'git://github.com/myoung34/vagrant.git'
  gem 'pry'
  gem 'byebug'
  gem 'pry-byebug'
end

group :testing do
  gem 'rspec-its'
  gem 'rspec'
  gem 'rake'
  gem 'kitchen-vagrant'
  gem 'kitchen-inspec'
  gem 'test-kitchen'
end

group :plugins do
  gem 'vagrant-ovirt4', :path => '.'
end
```

There's no way to do any sort of update, and these libraries are incompatible  with net-scp 1.1

PR #8329, opened 2017-03-04
Update vagrant gem dependencies

myoung34:
Alot of other gems (train, kitchen, etc depend on net-scp ~>1.2.0) and will not work without this one being bumped. net-scp 1.1.0 is _extremely_ old.


geerlingguy:
Yay!

PR #8337, opened 2017-03-06
Fix vagrant bash completion when path have spaces

pmarques:
# Problem

Using bash completion in a path that have spaces, for instance _*/User/pmaques/My_ Vagrant/*_ will fails with some lines of folder not found.


chrisroberts:
Thanks @pmarques !

PR #8386, opened 2017-03-17
Added ability to set env variable VAGRANT_USER_AGENT in order to over…

Purple90:
…ride the default USER_AGENT var


gschaetz:
I see this in quite a few other projects.  If you look at .gitconfig and npm they both have this ability to override.  Some proxy's block by user agent and this is one of the only ways to get around that.

PR #8393, opened 2017-03-21
Prevent other provider install attempts when explicit provider given

### LINKED ISSUE #8389: Docs are confusing about VAGRANT_DEFAULT_PROVIDER vs --provider flag
### Vagrant version
Vagrant 1.8.7 and 1.9.2

### Host operating system
OSX, but likely everywhere

### Guest operating system
Irrelevant

### Document 1
https://www.vagrantup.com/docs/providers/basic_usage.html

> Vagrant Up
> Once a provider is installed, you can use it by calling vagrant up with the --provider flag. This will force Vagrant to use that specific provider. No other configuration is necessary!
> 
> In normal day-to-day usage, the --provider flag is not necessary since Vagrant can usually pick the right provider for you. More details on how it does this is below.
>
> If you specified a --provider flag, you only need to do this for the up command. Once a machine is up and running, Vagrant is able to see what provider is backing a running machine, so commands such as destroy, suspend, etc. do not need to be told what provider to use.Once you vagrant destroy existing environments, this will take effect.

### Document 2
https://www.vagrantup.com/docs/other/environmental-variables.html

> VAGRANT_DEFAULT_PROVIDER
> This configures the default provider Vagrant will use.
>
> This normally does not need to be set since Vagrant is fairly intelligent about how to detect the default provider. By setting this, you will force Vagrant to use this provider for any new Vagrant environments. Existing Vagrant environments will continue to use the provider they came up with. > Once you vagrant destroy existing environments, this will take effect.

### Vagrantfile

This should work with any Vagrantfile, but here is a bare-bones example
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :
  config.vm.box = "debian/jessie64"
end
```

### Debug output

Buried in ~/.bash_profile or some place like that:
```
export VAGRANT_DEFAULT_PROVIDER=aws
```

Then run this command:
```
$ vagrant up --provider virtualbox

AWS Provider:
* An AMI must be configured via "ami" (region: #{region})
```

### Expected behavior
I expect the above command to override the default set by the environment variable and use the "virtualbox" provider. 

### Actual behavior
It used the default provider ("aws") as specified by the environment variable and failed

### Steps to reproduce
1. This errors: `VAGRANT_DEFAULT_PROVIDER=aws vagrant up --provider virtualbox`
2. This works: `VAGRANT_DEFAULT_PROVIDER=virtualbox vagrant up`

We have a large number of engineers that have set `VAGRANT_DEFAULT_PROVIDER=aws` in their dotfiles and would prefer not to force all of them to learn about this odd behavior.

chrisroberts:
Fixes #8389

### DIFF
--- a/plugins/commands/up/command.rb
+++ b/plugins/commands/up/command.rb
@@ -128,7 +128,7 @@ module VagrantPlugins
         # First create a set of all the providers we need to check for.
         # Most likely this will be a set of one.
         providers = Set.new
-        with_target_vms(names) do |machine|
+        with_target_vms(names, provider: provider) do |machine|
           # Check if we have this machine in the index
           entry    = @env.machine_index.get(machine.name.to_s)
 
--- a/test/unit/plugins/commands/up/command_test.rb
+++ b/test/unit/plugins/commands/up/command_test.rb
@@ -1,0 +1,65 @@
+require File.expand_path("../../../../base", __FILE__)
+
+require Vagrant.source_root.join("plugins/commands/up/command")
+
+describe VagrantPlugins::CommandUp::Command do
+  include_context "unit"
+
+  let(:argv)     { [] }
+  let(:vagrantfile_content){ "" }
+  let(:iso_env) do
+    env = isolated_environment
+    env.vagrantfile(vagrantfile_content)
+    env.create_vagrant_env
+  end
+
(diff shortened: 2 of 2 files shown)


WyseNynja:
Thank you for the fast fix!

Issue #8391, opened 2017-03-21
vagrant ansible 2.3.0 regression


shoreflyer:
It seems vagrants ansible provisioner isn't ready for ansible's 2.3.0 release.

### Vagrant version
1.9.2

### Host operating system
macOS 10.12.3 (16D32)

### Guest operating system
ubuntu/xenial64        (virtualbox, 20170320.0.0)

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

VAGRANTFILE_API_VERSION = "2"
MACHINE_HOSTNAME = "oceans"


$script = <<SCRIPT
apt-get install python nginx -y
SCRIPT

Vagrant.configure(VAGRANTFILE_API_VERSION) do |cfg|
  cfg.vm.box = "ubuntu/xenial64"
  cfg.vm.network "private_network", ip: "192.168.33.20"
  cfg.vm.synced_folder ".", "/vagrant"

  cfg.vm.provider :virtualbox do |vb|
    vb.customize ["modifyvm", :id, "--name", MACHINE_HOSTNAME]
    vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
    vb.customize ["modifyvm", :id, "--memory", 1024]
    vb.customize ["modifyvm", :id, "--cpus", 2]
    vb.customize ["modifyvm", :id, "--ioapic", "on"]
  end

  cfg.vm.provision "shell", inline: $script

  cfg.vm.provision "ansible" do |a|
    a.playbook = "site.yml"
    a.sudo = true
    a.extra_vars = { ansible_ssh_user: 'ubuntu' }
  end
end
```
### Debug output
https://gist.github.com/shoreflyer/9be5342bb4c569de913483dddd8d2ce5

### Expected behaviour
Completed provisioning or error within the ansible playbook.

### Actual behaviour
Errors out with the following:
```
...
==> default: Running provisioner: ansible...
    default: Running ansible-playbook...
 [WARNING]: provided hosts list is empty, only localhost is available


PLAY [all] *********************************************************************
skipping: no hosts matched

PLAY RECAP *********************************************************************
...
```

gildegoma:
[...]

PR #8170, opened 2017-01-04
Allow optional pip args via optional `pip_args`

jamescarr:
When using `install_mode = pip` for the version we're currently locked to, it fails due to some compilation errors on dependent libraries. This actually isn't needed since we have wheels for the dependencies in our own internal resolver but vagrant doesn't provide a good way to define `--index-url`. This PR resolves that. 

```
  n.vm.provision :ansible_local do |ansible|
    ansible.version = "2.1.1.0"
    ansible.install_mode = "pip"
    ansible.pip_args = "--install-url https://pypi.internal"
    ansible.playbook = "main.yml"
    ansible.provisioning_path = "/vagrant/ansible"
    ansible.galaxy_roles_path = THIRD_PARTY_ROLES_DIR
    ansible.galaxy_role_file = "#{ANSIBLE_ROOT_DIR}/third-party-roles.yml"
    ansible.galaxy_command = "ansible-galaxy install --role-file=%{role_file} --roles-path=%{roles_path} #{GALAXY_FORCE_OPTION}"
    ansible.sudo = true
    ansible.skip_tags = ['configure']
    ansible.extra_vars = ANSIBLE_INSTALL_VARS_ABSOLUTE
    ansible.raw_arguments =  ANSIBLE_RAW_ARGUMENTS
  end

```

PR #8405, opened 2017-03-24
Add the "pip_args" option and a new "pip_args_only" install mode

gildegoma:
This is first the "final" review and follow up of GH-8170.

With , I also propose the addition of `pip_args_only` installation mode, which was originally motivated by the following use case:

```ruby
config.vm.provision "ansible_local" do |ansible|
  ansible.playbook = "playbook.yml"
  ansible.install_mode = "pip_args_only"
  ansible.pip_args = "http://releases.ansible.com/ansible/ansible-2.3.0.0-0.1.rc1.tar.gz"
end
```
**Attention:** Don't use `2.3.0.0-0.1.rc1` with Vagrant, as inventory management is broken in Ansible 2.3.0RC1 (#8391)

Note that I first though about an even more generic install mode (kind of "custom" allowing to execute any command the user want). But I preferred the `pip_args_only`, with the following argumentation:
- "full custom setup" is easily and better achieved with an additional [`shell`](https://www.vagrantup.com/docs/provisioning/shell.html) provisioner configuration.
- Provide the ability to take full control on "pip" is a good thing, but "advanced pip users" will still appreciate the "pip auto setup" goody and a sweet integration into `Vagrantfile`.

@jamescarr @brandongalbraith could you please give me your thoughts on all this? Many thanks in advance ❤️ 💙 💚 ❗️

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/arch/ansible_install.rb
+++ b/plugins/provisioners/ansible/cap/guest/arch/ansible_install.rb
@@ -8,7 +8,7 @@ module VagrantPlugins
           module AnsibleInstall
 
             def self.ansible_install(machine, install_mode, ansible_version, pip_args)
-              if install_mode == :pip
+              if install_mode != :default
                 raise Ansible::Errors::AnsiblePipInstallIsNotSupported
               else
                 machine.communicate.sudo "pacman -Syy --noconfirm"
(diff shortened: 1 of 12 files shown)


jamescarr:
@gildegoma :+1: Really like this approach.

brandongalbraith:
@gildegoma This is 💯  Really appreciate it!

PR #8432, opened 2017-04-01
Use exportfs for NFS management on linux hosts

chrisroberts:
This picks up the original work from #3076 and finalizes the implementation. Removes management of the `/etc/exports` file and utilizes the `exportfs` command for managing exported file systems on NFS. This changeset also fixes a subtle issue where unexpected pruning could occur when using multiple providers with NFS synced folders on the same host.


chrisroberts:
One thing to note with this current implementation is that the UID information previously preserved (due to shared global file) is now lost. This means unintended pruning may occur in multi-user setups.

PR #8443, opened 2017-04-05
Add chef_install(ed?) guest capability for freebsd

anandsuresh:


anandsuresh:
Not sure of the priority for supporting FreeBSD as a guest, but I would be happy to help in anyway I can to get that done, if someone here can point me in the right direction.

Issue #5017, opened 2014-12-18
IdentitiesOnly=yes ssh argument not passed by ansible provisioner


shuhaowu:
Version: 1.7.1

This argument is passed in `vagrant ssh` but not `vagrant provision`. Causing issues with provisioning when the host have 5+ ssh keys.

Edit: I only observed in the ansible provisioning. Not sure about others.

trevorr:
In this case (`config.ssh.keys_only = true`), shouldn't `ansible_ssh_common_args='-o IdentitiesOnly=yes'` also be included in the generated inventory? The current fix works for the provisioner, but not when running Ansible manually using the generated inventory.

PR #8467, opened 2017-04-11
fix `config.ssh.keys_only = false` for ansible

MartinNowak:
Also see #5017.


MartinNowak:
Done

MartinNowak:
Done @chrisroberts

PR #8493, opened 2017-04-18
change path to bash to work in ubuntu

user-xy:
due to the fact that the vagrant box uses ubuntu, the path for the bash was wrong. Copy and paste the script didn't work. Following this guide with the hashicorp/precise64 box causes the error, that script was not loaded. With this change the script loads and apache is installed.


chrisroberts:
Hi there. I have tested the `/usr/bin/env` usage and have found no problems. It may have simply been a cut and paste error. I'm going to close this PR now, but if you continue to see this error and can provide some information so that I can reproduce, I will be happy to reopen this. Thanks!

PR #8399, opened 2017-03-22
Use 127.0.0.1 for host IP when unset and 0.0.0.0 is not available

chrisroberts:
Fixes #8395

PR #8494, opened 2017-04-18
host_ip while forwarding

user-xy:
without setting the host_ip paramter an error-message appears (requested address is not valid in its context) on "vagrant up".
setting the host_ip to 127.0.0.1 does the trick.


chrisroberts:
Thanks for this PR but the root cause of the issue was fixed by #8399. Cheers!

PR #8503, opened 2017-04-20
Unescape url to provide actual local path when adding box

### LINKED ISSUE #6825: box add [name] [path] does not recognise metadata.json file on Windows when [path] contains a space
This is happening on 1.8.1, Windows. Related to https://github.com/mitchellh/vagrant/issues/3201

I have two folders, "C:/boxes" and "C:/box es", each containing an identical "metadata.json" file, describing the versions of a base box. In Windows, on the command line, running the following commands results in only the path without a space working: 

```
C:\>vagrant box add test "C:\boxes\metadata.json"
==> box: Loading metadata for box 'C:\boxes\metadata.json'
    box: URL: file://C:/boxes/metadata.json
The box you're adding has a name different from the name you
requested. [this is fine, it means it has read the json file correctly]
```

```
C:\>vagrant box add test "C:\box es\metadata.json"
==> box: Box file was not detected as metadata. Adding it directly...
==> box: Adding box 'test' (v0) for provider:
    box: Unpacking necessary files from: file://C:/box%20es/metadata.json
    box: Progress: 100% (Rate: 254/s, Estimated time remaining: 0:00:01)
The box failed to unpackage properly. Please verify that the box
file you're trying to add is not corrupted and try again. The
output from attempting to unpackage (if any):

bsdtar.EXE: Error opening archive: Unrecognized archive format
```

This is obviously some issue to do with the escaping of spaces into %20s, which is required for URI handling, but then there appears to be some issue in doing a File.open() on those URIs with escaped spaces. I'd be willing to guess the problem happens in https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/action/builtin/box_add.rb .

There doesn't appear to be any other way of escaping the space from the command line (like adding the %20, or %%20, or ^%20, or whatever other methods of escaping) since these all get further escaped for the URI.

chrisroberts:
Fixes #6825

### DIFF
--- a/lib/vagrant/action/builtin/box_add.rb
+++ b/lib/vagrant/action/builtin/box_add.rb
@@ -478,6 +478,7 @@ module Vagrant
             url ||= uri.opaque
             #7570 Strip leading slash left in front of drive letter by uri.path
             Util::Platform.windows? && url.gsub!(/^\/([a-zA-Z]:)/, '\1')
+            url = URI.unescape(url)
 
             begin
               File.open(url, "r") do |f|


hiyangguo:
👍

PR #8505, opened 2017-04-21
Fixes automatic configuration of networking on amazon linux guests

kevinsperrine:
AWS doesn't have the NetworkManager package, so nmcli isn't available and I'd been running into tons of issues with NFS failing to connect to my OSX Sierra host. I copied the configure_networks capability from version 1.8.6 (since it was the last version not using nmcli) and added this capability for an amazon guest, now my NFS issues are gone in version 1.9.3

Signed-off-by: Kevin <kperrine@gmail.com>


chrisroberts:
Hi there. Thanks for this PR. I'm going to close this in favor of #8531 to reduce the amount of duplication for rhel derivative guests. Cheers!

PR #8510, opened 2017-04-21
Fixed Admin Verification for HyperV

johnrizzo1:
Fixed admin test to verify that you are running in an elevated shell, not that you are in the administrators group since that is not required.

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -51,9 +51,9 @@ module Vagrant
           return @_windows_admin if defined?(@_windows_admin)
 
           @_windows_admin = -> {
-            ps_cmd = "[System.Security.Principal.WindowsIdentity]::GetCurrent().Groups | ForEach-Object { if ($_.Value -eq 'S-1-5-19'){ Write-Host 'true'; break }}"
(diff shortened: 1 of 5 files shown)


DrowningElysium:
I can confirm that this fixes the admin issues I have when trying to create smb shares

mwrock:
This does indeed fix the issue where the hyperv provider errors stating that the user lacks admin rights when they in fact do. It currently just checks if the user is a system account and that would be very unlikely unless vagrant is running as a service.

I would remove the second commit from this PR. Its completely unrelated and IMO unnecessary since the `bridge` setting should accomplish this.

invisibleaxm:
fwiw, I stumbled into this issue, manually applied the fix and i can also confirmed that it worked.

wickedviking:
Regarding the VMSwitch handling, unless this could be updated to handle mutli-homed scenarios, it would better to leave it as is or update the current network processes to handle mutliple adapters/switches.

User case, setting up a Router VM to bridge two networks (local host private and lab for example) to create an isolated environment that "borrows" internet access from another network.

mwrock:
It should at least be implemented to not break current Vagrantfiles that use the `bridge` to specify a switch. For clarity I think having a separate PR is for the best.

chrisroberts:
Extracted the admin check to isolate change in #8548. Thank you for the PR!

johnrizzo1:
Thank you.  I didn't intend for both to be a part of the pull request.

Regards,

John

PR #8515, opened 2017-04-24
Update version etcd in configure_networks.rb

weivall:
If enable etcd2.service in coreos and build vagrant box using packer.io (coreos_*.iso) we are going to have etcd.service running after vagrant up.
As etcd is deprecated it is better to stick to etcd2


briancain:
Hey there! Thanks for opening a pull request with Vagrant. It looks like now there are several ways to start the `etcd` binary on coreOS: https://coreos.com/etcd/docs/latest/platforms/container-linux-systemd.html. `etc2` will be removed soon it looks like. Sorry it took us so long to look at this issue! I'm sure at the time this was correct.

As far as I can tell however, the recommended way to bring up a coreOS guest with Vagrant is using the project they have: https://coreos.com/os/docs/latest/booting-on-vagrant.html https://github.com/coreos/coreos-vagrant. From my testing, it looks like they have their own plugin for setting up networking with Vagrant, and this guest capability never actually gets called.

In any case, I'm going to go ahead and close this. Thanks again!

PR #8517, opened 2017-04-24
Update is_port_open.rb

vbrh-immalle:
Solves https://github.com/mitchellh/vagrant/issues/3031 (again)


jbonhag:
For future reference, the discussion for this PR starts at the following comment: https://github.com/hashicorp/vagrant/issues/3031#issuecomment-288570525

PR #8524, opened 2017-04-25
Reserves IP address 3-50 for static IP private network.

Maxwell1987:
When using private network, it is necessary to use the hybrid dhcp and static, so keep the x.x.x.3-50 for static.


briancain:
I'm going to go ahead and close this. Thanks again for your pull request! ❤️

PR #8525, opened 2017-04-25
8468 - make more virtualbox commands retryable

samueljc:
Issue: https://github.com/mitchellh/vagrant/issues/8468

A lot of vboxmanage commands are flakey and frequently cause
bringing multiple machines up at once to fail, especially when
the host system is under heavy load. Most commands are also safe
to retry and just result in a no-op, so we can simply add
'retryable' to a lot of existing calls. For the others we need to
do a little bit of cleanup or reevaluate the parameters before
trying again.

### DIFF
--- a/plugins/providers/virtualbox/driver/version_5_0.rb
+++ b/plugins/providers/virtualbox/driver/version_5_0.rb
@@ -17,19 +17,23 @@ module VagrantPlugins
         end
 
         def clear_forwarded_ports
-          args = []
-          read_forwarded_ports(@uuid).each do |nic, name, _, _|
-            args.concat(["--natpf#{nic}", "delete", name])
-          end
+          retryable(on: Vagrant::Errors::VBoxManageError, tries: 3, sleep: 1) do
+            args = []
+            read_forwarded_ports(@uuid).each do |nic, name, _, _|
+              args.concat(["--natpf#{nic}", "delete", name])
+            end
 
-          execute("modifyvm", @uuid, *args) if !args.empty?
+            execute("modifyvm", @uuid, *args) if !args.empty?
+          end
         end
 
         def clear_shared_folders
-          info = execute("showvminfo", @uuid, "--machinereadable", retryable: true)
-          info.split("\n").each do |line|
-            if line =~ /^SharedFolderNameMachineMapping\d+="(.+?)"$/
-              execute("sharedfolder", "remove", @uuid, "--name", $1.to_s)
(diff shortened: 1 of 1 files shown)


plugins/providers/virtualbox/driver/version_5_0.rb
```diff
@@ -17,19 +17,23 @@ def initialize(uuid)
         end
 
         def clear_forwarded_ports
-          args = []
-          read_forwarded_ports(@uuid).each do |nic, name, _, _|
-            args.concat(["--natpf#{nic}", "delete", name])
-          end
+          retryable(on: Vagrant::Errors::VBoxManageError, tries: 3, sleep: 1) do
```
wasosa: Looks like `read_forwarded_ports()` is already calling `execute()` with `retryable: true`:
```
execute("showvminfo", uuid, "--machinereadable", retryable: true)
```

Could it be that failures happen in the other `execute()` call below?

wasosa: The `args` could become invalid after the first retry, so retrying the whole block makes more sense. Same pattern follows in the rest of this. Nevermind :D LGTM!

marek-obuchowicz:
What is status of this PR? The bug affects us heavily and it seems that fix is already developed. It would be great if that could be merged/released anytime soon

briancain:
Closing this via #8951

wasosa:
Hi Brian! Glad to see this  was finally closed. I'm curious though, why wasn't this just merged as-is and future work based on master instead of rolling it into a new pull request?

briancain:
Hi @wasosa - I was just about to leave another comment on here...

The reason why I opened a new PR was because a couple of the changes you made in your patch needed some attention. We wanted to get all of the retryable stuff in though soon because we're working on the next vagrant release (which is soon!) so that's why there was a new pull request. It was also paired with some similar retryable changes as well. I didn't want to request changes on this PR and let this sit without getting some of the work into the next release!

As for the feedback...

The `create_snapshot` progress updates were missing variable initializations and thus resulted in a stacktrace when ever you tried to take a snapshot of a vm, and then even when I added them it didn't seem to work and I didn't see a progress bar like I do with the delete snapshot. If you'd like, I'd be happy to check out a new PR with that functional!

Also the `forward_ports` change ended up invoking the virtualbox command for each port forwarded for every machine rather than doing it all at once so I put that back to how it functioned before which seems more optimal. If that change was intentional please let me know in case I am missing something.

Hope that clears things up, thank you! 👍 🎉

PR #8531, opened 2017-04-26
guests/rhel: Update network configuration

chrisroberts:
Properly detects NetworkManager on guest as well as devices controlled
by NetworkManager. Provides configuration option to enable/disbale
NetworkManager control on devices.


curiositycasualty:
@chrisroberts Thank you for this!

PR #8555, opened 2017-05-03
Fix dockerfile parameter handling

nervo:
Fix https://github.com/mitchellh/vagrant/issues/7914


chrisroberts:
Hi there. Thanks for this pull request! The argument should be split into two items pushed onto `args` so things like spaces in the path and such don't cause breakage. I have the modification included in #8576 so I'm going to close this out. Again, thanks for the PR!

PR #8558, opened 2017-05-04
RFC: VAGRANT_PREFERRED_PROVIDERS

agriffis:
This PR comes from discussion at https://bugzilla.redhat.com/show_bug.cgi?id=1444492

Fedora distributes and supports the libvirt and docker providers, and does not distribute or support Virtualbox. This implies a different set of priorities from those in upstream Vagrant, meaning the priorities called out by the individual provider plugins.

With the existing code in Vagrant, the only way for Fedora to express system-level provider priorities is to set `VAGRANT_DEFAULT_PROVIDER`, but that has the undesired effect of short-circuiting the rest of Vagrant's [nuanced algorithm for choosing a provider](https://www.vagrantup.com/docs/providers/basic_usage.html#default-provider):

> 1. The --provider flag on a vagrant up is chosen above all else, if it
>    is present.
> 
> 2. If the VAGRANT_DEFAULT_PROVIDER environmental variable is set, it
>    takes next priority and will be the provider chosen.
> 
> 3. Vagrant will go through all of the config.vm.provider calls in the
>    Vagrantfile and try each in order. It will choose the first
>    provider that is usable. For example, if you configure Hyper-V, it
>    will never be chosen on Mac this way. It must be both configured
>    and usable.
> 
> 4. Vagrant will go through all installed provider plugins (including
>    the ones that come with Vagrant), and find the first plugin that
>    reports it is usable. There is a priority system here: systems that
>    are known better have a higher priority than systems that are
>    worse. For example, if you have the VMware provider installed, it
>    will always take priority over VirtualBox.
> 
> 5. If Vagrant still has not found any usable providers, it will error.

This pull request adds `VAGRANT_PREFERRED_PROVIDERS` which is a comma-separated ordered list of preferred providers. If the environment variable is unset, then the code works exactly as before. However if the variable is set, then it implies two additional steps in the algorithm:

> 2.5. Vagrant will go through all of the config.vm.provider calls in
>      the Vagrantfile and try each in order. It will choose the first
>      provider that is usable and listed in VAGRANT_PREFERRED_PROVIDERS.
> 
> 3.5. Vagrant will go through VAGRANT_PREFERRED_PROVIDERS and find the
>      first plugin that reports it is usable.

The individual commits in this PR are as follows:

1. Rewrite `Environment#default_provider` to implement Vagrant's existing algorithm in a more straightforward manner. The existing code does math to recalculate priorities and finally sort based on them, but that approach doesn't lend itself to an easy comparison between the algorithm and the code, and it's hard to update the code for tweaks to the algorithm. The new code tries to smooth this out.

2. Add support for `VAGRANT_PREFERRED_PROVIDERS`. Because of the changes in the first commit, this is very straightforward.

3. Add new tests for the new functionality. This is in a separate commit so it's clear that the existing (and extensive) tests continued to work unmodified.

In general, I think this PR sheds light on the existing questionable approach to provider priorities. With the existing setup, provider plugins have to be aware of their own place in the pecking order, creating an implicit coupling between plugins, and making it hard for a user or distro (or Vagrant) to change priorities except by patching the provider plugins. I'm not sure that `VAGRANT_PREFERRED_PROVIDERS` is the perfect solution, but it seems to work well for this purpose.

### DIFF
--- a/lib/vagrant/environment.rb
+++ b/lib/vagrant/environment.rb
@@ -319,6 +319,9 @@ module Vagrant
 
       # Implement the algorithm from
       # https://www.vagrantup.com/docs/providers/basic_usage.html#default-provider
+      # with additional steps 2.5 and 3.5 from
+      # https://bugzilla.redhat.com/show_bug.cgi?id=1444492
+      # to allow system-configured provider priorities.
       #
       # 1. The --provider flag on a vagrant up is chosen above all else, if it is
       #    present.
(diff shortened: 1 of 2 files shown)


agriffis:
@chrisroberts Any thoughts about this? Is Vagrant upstream interested at all in moving provider policy out of the plugins and into Vagrant proper? And if not, would you consider a PR to merge the first commit in the series, since it clarifies the code and makes downstream patching easier?

chrisroberts:
Hey @agriffis. This PR looks good from a glance. I was waiting on digging into it until I had cut a release so I'm going to be reviewing and test driving this today and tomorrow.

agriffis:
@chrisroberts I'm just back from a trip so checking in. How's it going?

agriffis:
@chrisroberts Ping?

chrisroberts:
@agriffis pong

agriffis:
Woo! Thanks @chrisroberts!

PR #8570, opened 2017-05-08
Add support for running within WSL

### LINKED ISSUE #7731: Use Linux Subsystem on Windows

chrisroberts:
Fixes #7731

### DIFF
--- a/website/source/docs/other/wsl.html.md
+++ b/website/source/docs/other/wsl.html.md
@@ -1,0 +1,118 @@
+---
+layout: "docs"
+page_title: "Vagrant and Windows Subsystem for Linux"
+sidebar_current: "other-wsl"
+description: |-
+  An overview of using Vagrant on Windows within the Windows Subsystem
+  for Linux.
+---
+
+# Vagrant and Windows Subsystem for Linux
+
+Windows has recently introduced a new feature called the Windows Subsystem
+for Linux (WSL). This is a beta feature available in developer mode on recent
+releases of Windows 10. It is important to note that this feature is still
+in _beta_ on Windows, and Vagrant support should be considered _alpha_.
+
+<div class="alert alert-warning">
+  <strong>Warning: Advanced Topic!</strong> Using Vagrant within the Windows
+  Subsystem for Linux is an advanced topic that only experienced Vagrant users
+  who are reasonably comfortable with Windows, WSL, and Linux should approach.
+</div>
+
+
+# Installation
+
+Installation requires WSL, Ubuntu on Windows, and Vagrant. Read on for installation
+instructions for each item.
+
+## Windows Subsystem for Linux and Ubuntu on Windows
+
+First install the Windows Subsystem for Linux, followed by Ubuntu on Windows. This guide
+from Microsoft walks through the process:
+
+* https://msdn.microsoft.com/en-us/commandline/wsl/install_guide
+
+## Vagrant Installation
+
+Vagrant _must_ be installed within Ubuntu on Windows. Even though the `vagrant.exe`
+file can be executed from within the WSL, it will not function as expected. To
+install Vagrant into the WSL, follow these steps:
+
+* Download the 64-bit Debian package from the downloads page.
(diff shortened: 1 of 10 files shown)


Log1x:
This will make many Windows developers running CU and Insider very happy. :) 

Thank you.

website/source/docs/other/wsl.html.md
```diff
@@ -0,0 +1,118 @@
...
+Windows system. In some cases, a user may be using Vagrant in a regular
+Windows environment, and then transition to using Vagrant within the
+WSL. Using Vagrant within the WSL will appear to be isolated from
+the Windows system. A new `VAGRANT_HOME` directory will be created within
+the WSL (meaning all boxes will require re-downloading). Vagrant will also
+lose the ability to control Vagrant managed machines within Windows (due
+to user ID mismatches).
+
+Vagrant supports enabling user access to provide seamless behavior and
+control between Vagrant on Windows and Vagrant on WSL. By setting the
```
mwhooker: > +Vagrant supports enabling user access to provide seamless behavior and
> +control between Vagrant on Windows and Vagrant on WSL.

I think I understand what it says (that with this magic flag you can let one environment control the other), but the wording is slightly confusing to me.

website/source/docs/other/wsl.html.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+It is important to note that file permissions cannot be enforced when Vagrant
+modifies the Windows file system. It is for this reason that you must explicitly
+enable this functionality with the express knowledge of the implication. If you
+are unsure of how this may affect your system, do not enable this feature.
+
+## Using Docker
+
+The docker daemon cannot be run in side the Windows Subsystem for Linux. However,
```
mwhooker: `s/in side/inside`

website/source/docs/other/wsl.html.md
```diff
@@ -0,0 +1,118 @@
...
+the daemon _can_ be run on Windows and accessed by Vagrant while running in the
+WSL. Once docker is installed and running on windows, export the following
```
mwhooker: capitalize Windows

leighmcculloch:
Woah – 🎉 – which version of vagrant is this slated for?

kylemacfarlane:
Does this work with VMWare which usually requires admin access? I don't expect it to work but the updated docs aren't explicit.

chrisroberts:
@kylemacfarlane I'm wrapping up the vmware support and it will be available in the next plugin release.

ideal2545:
I am so excited for this!!!! WSL is really really really growing

MatheusRV:
which version is this solution included? tks

MatheusRV:
tks this correct https://github.com/Microsoft/BashOnWindows/issues/733#issuecomment-279107617 to me

PR #8596, opened 2017-05-17
Update environmental-variables.html.md

tonynguyen68:


QWp6t:
There's nothing wrong with how they had it. Periods and commas go inside quotes according to most (or all?) American style guides.

Not that I really care. Just saying this isn't actually an issue.

PR #8618, opened 2017-05-25
guests/smartos: Add/fix various guest capabilities

jen20:
This commit adds/changes the following for SmartOS guests:

- modifies the "Halt" capability to use /usr/sbin/poweroff in preference to /usr/sbin/shutdown with parameters, and modifies the associated test.

- adds an "InsertPublicKey" capability and tests.

- adds a "RemovePublicKey" capability and tests.

With this commit applied, the vast majority of typical Vagrant workflow is available to SmartOS global zone guests (provided NFS mounts are used rather than VMWare shared folders).


jen20:
(Updated to support both global and non-global zone guests).

PR #8619, opened 2017-05-25
Enforce unique snapshot names

### LINKED ISSUE #7810: vagrant snapshots don't have to have unique names
### Vagrant version

Vagrant 1.8.5
### Host operating system

Mac OS X 10.12.0 (GM)
### Guest operating system

ubuntu/trusty64
### Expected behavior

Repeated snapshots of a box with the same snapshot name should either fail, or overwrite the previous snapshot
### Actual behavior

Two snapshots were created with the same name
### Steps to reproduce
1. vagrant up # on a box
2. vagrant snapshot mysnap
3. vagrant snapshot mysnap
4. vagrant snapshot list

briancain:
Prior to this commit, the vagrant snapshot plugin would save snapshots
with existing names which lead to duplicate snapshot names being saved.
This commit fixes that by checking to see if the given snapshot name
already exists and if so, fails telling the user the given snapshot name
already exists. If a user passes a --force flag, vagrant will first
delete the existing snapshot, and take a new one with the given name.

Fixes #7810

### DIFF
--- a/plugins/commands/snapshot/command/save.rb
+++ b/plugins/commands/snapshot/command/save.rb
@@ -6,6 +6,7 @@ module VagrantPlugins
       class Save < Vagrant.plugin("2", :command)
         def execute
           options = {}
+          options[:force] = false
 
           opts = OptionParser.new do |o|
             o.banner = "Usage: vagrant snapshot save [options] [vm-name] <name>"
@@ -16,6 +17,10 @@ module VagrantPlugins
             o.separator ""
             o.separator "Snapshots are useful for experimenting in a machine and being able"
             o.separator "to rollback quickly."
+
+            o.on("-f", "--force", "Replace snapshot without confirmation") do |f|
+              options[:force] = f
+            end
           end
 
           # Parse the options
@@ -28,7 +33,22 @@ module VagrantPlugins
 
           name = argv.pop
           with_target_vms(argv) do |vm|
-            vm.action(:snapshot_save, snapshot_name: name)
+            if !vm.provider.capability?(:snapshot_list)
+              vm.ui.info(I18n.t("vagrant.commands.snapshot.not_supported"))
+              next
+            end
+
+            snapshot_list = vm.provider.capability(:snapshot_list)
+
+            if !snapshot_list.include? name
+              vm.action(:snapshot_save, snapshot_name: name)
+            elsif options[:force]
+              # not a unique snapshot name
+              vm.action(:snapshot_delete, snapshot_name: name)
+              vm.action(:snapshot_save, snapshot_name: name)
+            else
+              raise Vagrant::Errors::SnapshotConflictFailed
+            end
           end
 
           # Success, exit status 0
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -1156,6 +1156,8 @@ en:
         (probably for one of your shared folders). This is an extremely rare
         error case and most likely indicates an unusual configuration of the
         guest system. Please report a bug with your Vagrantfile and debug log.
+      snapshot_force: |-
+        You must include the `--force` option to replace an existing snapshot.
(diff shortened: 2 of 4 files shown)


plugins/commands/snapshot/command/save.rb
```diff
@@ -28,7 +33,19 @@ def execute
 
           name = argv.pop
           with_target_vms(argv) do |vm|
-            vm.action(:snapshot_save, snapshot_name: name)
+            # TODO: remove this comment before merge....
+            # do I need to check if the provider on the vm supports snapshots before doing this?
+            snapshot_list = vm.provider.capability(:snapshot_list)
```
briancain: @chrisroberts question about this here...should I check to see if the vms provider has the capability to take snapshots before calling this action? I noticed this step happens else where in a different snapshot command.

test/unit/plugins/commands/snapshot/command/save_test.rb
```diff
@@ -0,0 +1,97 @@
...
+          and_return(["test"])
+
+        expect(machine).to receive(:action).twice do |name, opts|
+          if name == :snapshot_delete
+            expect(name).to eq(:snapshot_delete)
+            expect(opts[:snapshot_name]).to eq("test")
+          else
+            expect(name).to eq(:snapshot_save)
+            expect(opts[:snapshot_name]).to eq("test")
+          end
```
briancain: OK, this passes, but it's awful. Is there a better way to test that a machine is expected to have two actions?

plugins/commands/snapshot/command/save.rb
chrisroberts: @briancain Yeah, as there are providers that may not have the capability it should gracefully error.

test/unit/plugins/commands/snapshot/command/save_test.rb
chrisroberts:
```ruby
expect(machine).to receive(:action).with(:snapshot_delete, snapshot_name: "test")
expect(machine).to receive(:action).with(:snapshot_save, snapshot_name: "test")
```

templates/locales/en.yml
```diff
@@ -1732,6 +1732,8 @@ en:
           No pushed snapshot found!
 
           Use `vagrant snapshot push` to push a snapshot to restore to.
+        force: |-
+          You need to specify --force to replace an existing snapshot.
```
chrisroberts: > You must include the `--force` option to replace an existing snapshot

plugins/commands/snapshot/command/save.rb
```diff
@@ -16,6 +17,10 @@ def execute
             o.separator ""
             o.separator "Snapshots are useful for experimenting in a machine and being able"
             o.separator "to rollback quickly."
+
+            o.on("-f", "--force", "Remove snapshot without confirmation.") do |f|
```
chrisroberts: > Replace snapshot without confirmation.

briancain:
@chrisroberts - Updated PR based on feedback!

plugins/commands/snapshot/command/save.rb
```diff
@@ -28,7 +33,22 @@ def execute
...
+            snapshot_list = vm.provider.capability(:snapshot_list)
+
+            if !snapshot_list.include? name
+              vm.action(:snapshot_save, snapshot_name: name)
+            elsif options[:force]
+              # not a unique snapshot name
+              vm.action(:snapshot_delete, snapshot_name: name)
+              vm.action(:snapshot_save, snapshot_name: name)
+            else
+              vm.ui.warn(I18n.t("vagrant.commands.snapshot.force"))
```
chrisroberts: This should raise an exception, forcing a non-zero exit code providing feedback that the snapshot was not taken, rather than issuing a warning.

PR #8653, opened 2017-06-06
Clean up vagrant snapshot restore/delete error messages

### LINKED ISSUE #7188: Vagrant snapshot restore exit code is zero when snapshot does not exist
### Vagrant version

```
# vagrant -v
Vagrant 1.8.1
```
### Host operating system

Ubuntu 14.04.3 LTS
### Guest operating system

N/A
### Vagrantfile

Vanilla
### Debug output

N/A, as I believe this is functioning as designed.
### Expected behavior

Exit code of `vagrant snapshot restore foo` should be nonzero when the `foo` snapshot does not exist.
### Actual behavior

Exit code of `vagrant snapshot restore test` is zero when the `foo` snapshot does not exist. 
### Steps to reproduce
1. Create a vanilla vagrant configuration: `vagrant init hashicorp/precise64`
2. Attempt to restore a snapshot that does not exist: `vagrant snapshot restore test`
3. Retrieve the exit code: `echo $?`

For example:

```
# vagrant init hashicorp/precise64
A `Vagrantfile` has been placed in this directory. You are now
ready to `vagrant up` your first virtual environment! Please read
the comments in the Vagrantfile as well as documentation on
`vagrantup.com` for more information on using Vagrant.
# echo $?
0
# false
# echo $?
1
# vagrant snapshot restore test
==> default: VM not created. Moving on...
# echo $?
0
```

Original discovery occurred when attempting something like this:

```
vagrant snapshot restore test || vagrant up && vagrant snapshot save test
```

briancain:
This commit adds some better handling around the snapshot restore
command for the virtualbox provider. If a user attempts to restore from
a vm that does not exist, instead of exiting 0 it will raise an
exception saying the virtual machine has not been created yet.
Addtionally, if a user attempts to restore from a snapshot id that does
not exist, instead of printing a complicated exception from the
virtualbox cli tool, it prints a more useful error message telling the
user that the snapshot does not exist.

Fixes #7188

### DIFF
--- a/plugins/commands/snapshot/command/delete.rb
+++ b/plugins/commands/snapshot/command/delete.rb
@@ -23,7 +23,19 @@ module VagrantPlugins
 
           name = argv.pop
           with_target_vms(argv) do |vm|
-            vm.action(:snapshot_delete, snapshot_name: name)
+            if !vm.provider.capability?(:snapshot_list)
+              raise Vagrant::Errors::SnapshotNotSupported
+            end
+
+            snapshot_list = vm.provider.capability(:snapshot_list)
+
+            if snapshot_list.include? name
+              vm.action(:snapshot_delete, snapshot_name: name)
+            else
+              raise Vagrant::Errors::SnapshotNotFound,
+                snapshot_name: name,
+                machine: vm.name.to_s
+            end
           end
 
           # Success, exit status 0
--- a/lib/vagrant/errors.rb
+++ b/lib/vagrant/errors.rb
@@ -656,6 +656,10 @@ module Vagrant
       error_key(:snapshot_force)
     end
 
+    class SnapshotNotFound < VagrantError
+      error_key(:snapshot_not_found)
+    end
+
     class SnapshotNotSupported < VagrantError
       error_key(:snapshot_not_supported)
     end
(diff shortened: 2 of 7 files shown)


briancain:
Not totally happy yet with how this behaves. It wasn't super clear to me how I could clear the buffer so that the error message lives on its own line rather than coming after the `Progress: 0%`. 

https://github.com/mitchellh/vagrant/blob/master/plugins/providers/virtualbox/action/snapshot_restore.rb#L13-L17

Looks like that starts when the `provider.driver.restore_snapshot` function is invoked...any ideas there?

```
brian@localghost:vagrant-sandbox % be vagrant snapshot restore virtualbox foo                           ±[●●][master]

==> virtualbox: Restoring the snapshot 'foo'...
Progress: 0%This machine does not have any snapshots named foo.
```

briancain:
Ok, cleaning this up and putting the validation in the snapshot restore command rather than deeper down in the action class. Need to write tests now for `snapshot/restore`

plugins/commands/snapshot/command/delete.rb
```diff
@@ -23,7 +23,19 @@ def execute
...
+
+            snapshot_list = vm.provider.capability(:snapshot_list)
+
+            if snapshot_list.include? name
+              vm.action(:snapshot_delete, snapshot_name: name)
+            else
+              raise Vagrant::Errors::SnapshotNotFound,
+                snapshot_name: name,
+                machine: vm.name.to_s
+            end
```
chrisroberts: :ok_hand:

PR #8659, opened 2017-06-08
fix `docker-compose` check on `PATH`

HarryWeppner:
Although `docker-compose` was executable and on the `PATH` the `docker` provider outputted an error indicating that it couldn't find it.

```
Vagrant has been instructed to use to use the Compose driver for the
Docker plugin but was unable to locate the `docker-compose` executable.
Ensure that `docker-compose` is installed and available on the PATH.
```

`vagrant` incorrectly checked for `vagrant-compose` instead of `docker-compose`.


HarryWeppner:
Fixes #8660

PR #8676, opened 2017-06-13
Catch pip_args in FreeBSD's and SUSE's ansible_install

andreaso:
While neither the FreeBSD provisioner nor the SUSE provisioner support
installing Ansible using pip their ansible_install methods still get
called with that fourth argument. The result being these errors when
Vagrant tries to install Ansible.

    /opt/vagrant/embedded/gems/gems/vagrant-1.9.5/plugins/provisioners/ansible/cap/guest/freebsd/ansible_install.rb:10:in `ansible_install': wrong number of arguments (4 for 3) (ArgumentError)
    /opt/vagrant/embedded/gems/gems/vagrant-1.9.5/plugins/provisioners/ansible/cap/guest/suse/ansible_install.rb:9:in `ansible_install': wrong number of arguments (4 for 3) (ArgumentError)

The Arch provider, it too without pip support, already catches the
pip_args argument this way.

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/freebsd/ansible_install.rb
+++ b/plugins/provisioners/ansible/cap/guest/freebsd/ansible_install.rb
@@ -7,7 +7,7 @@ module VagrantPlugins
         module FreeBSD
           module AnsibleInstall
 
-            def self.ansible_install(machine, install_mode, ansible_version)
+            def self.ansible_install(machine, install_mode, ansible_version, pip_args)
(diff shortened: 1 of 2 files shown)


gildegoma:
Merged as f6a9dcdd6f2581d495c5d7f1eaef0ad495620901.

Thank you very much @andreaso. This was unfortunately missed in #8170 and #8405 (that's what we get with uncovered parts of dynamic programming language 😨).

PR #8679, opened 2017-06-13
website: Packer box version should be variable

justincampbell:
When using multiple builders, the post-processor could yield different results for the `{{timestamp}}` interpolation.

By using a variable instead, the version will be consistent across all builders.

https://github.com/hashicorp/packer/issues/4973


mwhooker:
👍

PR #8682, opened 2017-06-14
Add config option omnibus_url for chef provisioners

artem-sidorenko:
This option is useful for internal setups, where own customized
omnibus installation script is used (e.g. to get chef from a mirror)

### DIFF
--- a/test/unit/plugins/provisioners/chef/config/base_test.rb
+++ b/test/unit/plugins/provisioners/chef/config/base_test.rb
@@ -82,4 +82,17 @@ describe VagrantPlugins::Chef::Config::Base do
       expect(subject.installer_download_path).to be(nil)
     end
   end
+
+  describe "#omnibus_url" do
+    it "defaults to https://omnitruck.chef.io" do
+      subject.finalize!
+      expect(subject.omnibus_url).to eq("https://omnitruck.chef.io")
+    end
+
+    it "makes use of the configured url" do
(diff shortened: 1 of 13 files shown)


artem-sidorenko:
@chrisroberts any chance to get a review here?

test/unit/plugins/provisioners/chef/config/base_test.rb
```diff
@@ -82,4 +82,11 @@
       expect(subject.installer_download_path).to be(nil)
     end
   end
+
+  describe "#omnibus_url" do
+    it "defaults to https://omnitruck.chef.io" do
+      subject.finalize!
+      expect(subject.omnibus_url).to eq("https://omnitruck.chef.io")
+    end
```
briancain: Can we also get a test that asserts if the user sets it to something other than the default, it gets properly used?

artem-sidorenko: @briancain thanks! I added an additional test, so it looks now exactly like other chef options.

If you need/want more tests, could you please make a suggestion for implementation of this tests?
(I can't find any existing tests, which go completely through config options, capabilities, installer etc.)

briancain: Thanks! I think that test should be enough! :)

artem-sidorenko:
thanks for review!

PR #8693, opened 2017-06-15
Introduce more ssh options for machines

### LINKED ISSUE #7855: vagrant ssh should allow overriding of all ssh parameters
### Vagrant version

1.8.1
### Host operating system

macos
### Guest operating system

Windows 10
### Vagrantfile

Openssh for Windows does not work with Compression enabled, but there is no way to turn it off. There should never be any non-overridable default values imo.

briancain:
This commit allows the user to configure two additional options that
were previously not configurable: Compression and DSAAuthentication.
Each config option is set as a boolean, and if left out of the config
will default to its previous behavior which is included and set to
"yes". If the user explicitly sets it to false, it will not be included
as an ssh option.

Fixes #7855


briancain:
Updated to `dsa_authentication`. Still need to update the docs before this gets merged.

PR #8695, opened 2017-06-16
guests/smartos: Add/fix various guest capabilities

jen20:
This commit adds a variety of fixes for SmartOS guest support:

- Host name setting now works in the global zone and in non-global zones
- NFS now works in the global zone and the non-global zone.
- Tests are updated and moved to the (apparently) more modern style

cc: @sean-


chrisroberts:
Thanks @jen20 !

Issue #7371, opened 2016-05-31
request to use existing SMB mount to avoid admin privileges requirement


joshco:
Could you add a configuration option to have the guest mount a specified existing SMB share/mount point from the host?

The current approach (vagrant exposes the share) seems to be what requires admin privileges.  
Admin privileges means that any calling app must also be run as administrator, in my case that's RubyMine.  I can't run the app via it's vagrant unless it is also running as admin.  Of course since running something like vagrant as admin means that that VMs and settings don't show up under my user account.

Another configuration parameter such as "smb_share" alongside "smb_host" would allow this.

My setup
### Vagrant version

Vagrant 1.8.1
### Host operating system

Windows 10 x64, mingw
### Guest operating system

Linux Ubuntu Trusty64

cyberis:
+1 Vagrant is most annoying to run on Windows when it requires admin privs during vagrant up. I am limited to cmd.exe when usually I run vagrant from git or cygwin bash when on Windows but can't because of priv elevation.

Perni1984:
+1 for the idea.

@cyberis: slightly off-topic, but you can start git cmd with Administrator privileges (Ctrl + Shift + Click on Launch Icon) to circumvent the priv. escalation. You still have to put in your admin password when you connect over smb inside a shared folder though.

joshco:
@Perni1984 your OT is true, but the challenge is that with the requirement for admin priv, one can no longer call vagrant from the growing set of applications that support it.  For me, that includes RubyMine.  Running that as admin just to be able to debug through a vagrant is a major pita. 

I've started just executing an SMB mount command inside the vagrant such as:::
````
mount -t cifs -o uid=1000, **credential info**,vers=3.02 //192.168.10.1/vb2 /vagrant
````
This mounts the SMB mount over the top of the original vbox file share, which seems to make things faster without breaking things. (so far)

dhandeo:
The windows share already exists for the host machine, or can be over network. It would be convenient to define the mounting in the Vagrantfile

krystan:
Still and issue and for us a very annoying one we cannot use vb shared drives as they do not support symlinks which breaks npm when installing back to a mapped drive, we need the node_modules on the host so people can use them to view source in editors or we could just let them reside on the Linux host. Would be good to be able to auto mount in this fashion as described in  this issue

chrisroberts:
Hi there,

This behavior was enabled via the SMB updates in #9294 so I'm going to close this up. 

Cheers!

PR #8698, opened 2017-06-18
smb synced folders: auto-elevate

valhallasw:
This removes the requirement to start `vagrant` as elevated user. Instead, an elevation prompt is shown during `vagrant up`. This elevation prompt is shown twice if the mount already exists (once to remove it, once to make a new one). This is comparable to what happens when network adapters are created and initialized.

The rest of the VM then runs completely as the regular user, which makes it much easier to use tooling that expects this (e.g. IntelliJ expects `vagrant status` to work when run as regular user).

### DIFF
--- a/plugins/synced_folders/smb/scripts/set_share.ps1
+++ b/plugins/synced_folders/smb/scripts/set_share.ps1
@@ -9,7 +9,7 @@ Param(
 $ErrorAction = "Stop"
 
 if (net share | Select-String $share_name) {
-  net share $share_name /delete /y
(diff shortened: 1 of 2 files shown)


valhallasw:
This is an alternative fix for #7371.

valhallasw:
@chrisroberts: does this still make sense with your latest updates to the SMB code (00fa50c296bda4ea708d2791caae09d627494f27)?

valhallasw:
From testing, I see the current implementation nicely asks for elevation, so this issue seems resolved.

PR #8699, opened 2017-06-19
Fix vagrant_cwd warnings

briancain:
Prior to this commit, if a user ran a vagrant command within a subdir,
it would warn about the cwd changing which is not actually the case.
This commit adds an additional check to see if vagrant is being invoked
within a subdirectory so that it doesn't warn the user.

### DIFF
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -577,13 +577,13 @@ module Vagrant
       end
 
       if vagrant_cwd.nil?
-        File.write(vagrant_cwd_filepath, @env.cwd)
-      elsif vagrant_cwd != @env.cwd.to_s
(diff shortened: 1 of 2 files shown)


lib/vagrant/machine.rb
```diff
@@ -577,13 +577,13 @@ def check_cwd
       end
 
       if vagrant_cwd.nil?
-        File.write(vagrant_cwd_filepath, @env.cwd)
-      elsif vagrant_cwd != @env.cwd.to_s
+        File.write(vagrant_cwd_filepath, @env.root_path)
+      elsif vagrant_cwd != @env.root_path.to_s &&
```
chrisroberts: Extraneous `&&` here

briancain: Oops! Good catch :)

PR #8722, opened 2017-06-26
Add post-install provisioner to docker provisioner

### LINKED ISSUE #7139: docker provisioning fails when /etc/default/docker is installed
### Vagrant version

1.8.1
### Host operating system

Ubuntu trusty (14.04LTS)
### Guest operating system

Ubuntu trusty (14.04LTS)
### Expected behavior

Vagrant should install docker without exploding when following the recommendations in the docker provisioner documentation, namely that the best way to configure your docker is to populate `/etc/default/docker` prior to running the docker provisioner.
### Actual behavior

Kaboom!  Fails on prompt as to what to do with the existing vs new `/etc/default/docker`s.

...tail end of `vagrant up`:

```
+ sh -c sleep 3; apt-get update; apt-get install -y -q docker-engine

Configuration file '/etc/default/docker'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** docker (Y/I/N/O/D/Z) [default=N] ? dpkg: error processing package docker-engine (--configure):
 EOF on stdin at conffile prompt
Errors were encountered while processing:
 docker-engine
E: Sub-process /usr/bin/dpkg returned an error code (1)
```
### Steps to reproduce
1. Create `/etc/default/docker` as the provisioner docs recommend.
2. Use the docker provisioner.
3. `vagrant up`

briancain:
Prior to this commit, if a user attempted to configure
`/etc/default/docker` through vagrant prior to installation, the package
manager would not override an existing configuration and installing
docker would then fail. This commit fixes this by introducing a
`post_install_provisioner` that allows users to define a provisioner
block that will run after docker has been installed, allowing users to
configure `/etc/default/docker` how they want.

Fixes #7139

### DIFF
--- a/plugins/provisioners/docker/config.rb
+++ b/plugins/provisioners/docker/config.rb
@@ -4,9 +4,11 @@ module VagrantPlugins
   module DockerProvisioner
     class Config < Vagrant.plugin("2", :config)
       attr_reader :images
+      attr_accessor :post_install_provisioner
 
       def initialize
         @images = Set.new
+        @post_install_provisioner = nil
 
         @__build_images   = []
         @__containers     = Hash.new { |h, k| h[k] = {} }
@@ -36,6 +38,15 @@ module VagrantPlugins
 
       def pull_images(*images)
         @images += images.map(&:to_s)
+      end
+
+      def post_install_provision(name, **options, &block)
+        # Abort
+        raise DockerError, :wrong_provisioner if options[:type] == "docker"
+
+        proxy = VagrantPlugins::Kernel_V2::VMConfig.new
+        proxy.provision(name, **options, &block)
+        @post_install_provisioner = proxy.provisioners.first
       end
 
       def run(name, **options)
--- a/plugins/provisioners/docker/installer.rb
+++ b/plugins/provisioners/docker/installer.rb
@@ -8,10 +8,12 @@ module VagrantPlugins
       # This handles verifying the Docker installation, installing it if it was
       # requested, and so on. This method will raise exceptions if things are
       # wrong.
+      # @return [Boolean] - false if docker cannot be detected on machine, else
+      #                     true if docker installs correctly or is installed
       def ensure_installed
         if !@machine.guest.capability?(:docker_installed)
           @machine.ui.warn(I18n.t("vagrant.docker_cant_detect"))
-          return
+          return false
         end
 
         if !@machine.guest.capability(:docker_installed)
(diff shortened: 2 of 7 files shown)


plugins/provisioners/docker/installer.rb
```diff
@@ -26,6 +26,8 @@ def ensure_installed
         if @machine.guest.capability?(:docker_configure_vagrant_user)
           @machine.guest.capability(:docker_configure_vagrant_user)
         end
+
+        return true
```
chrisroberts: Drop the `return` here.

plugins/provisioners/docker/installer.rb
```diff
@@ -11,7 +11,7 @@ def initialize(machine)
       def ensure_installed
```
chrisroberts: Include some method docs with explanation of return value meanings.

briancain:
@chrisroberts - thanks! Updated the based on the two comments.

plugins/provisioners/docker/config.rb
```diff
@@ -4,9 +4,11 @@ module VagrantPlugins
   module DockerProvisioner
     class Config < Vagrant.plugin("2", :config)
       attr_reader :images
+      attr_accessor :post_install_provisioner
 
       def initialize
         @images = Set.new
+        @post_install_provisioner = []
```
chrisroberts: Should either remove the `@post_install_provisioner` initialization or just set it to `nil`

plugins/provisioners/docker/provisioner.rb
```diff
@@ -19,7 +19,14 @@ def provision
         @logger = Log4r::Logger.new("vagrant::provisioners::docker")
 
         @logger.info("Checking for Docker installation...")
-        @installer.ensure_installed
+        if @installer.ensure_installed
```
chrisroberts: This conditional should also include a check to confirm a post install provisioner has been defined.

test/unit/plugins/provisioners/docker/config_test.rb
```diff
@@ -137,4 +138,25 @@
...
+      mock_provisioners = [mock_provisioner]
+
+      allow(VagrantPlugins::Kernel_V2::VMConfig).to receive(:new).
+        and_return(prov)
+      allow(prov).to receive(:provision).and_return(mock_provisioners)
+      allow(prov).to receive(:provisioners).and_return(mock_provisioners)
+
+      subject.post_install_provision("myprov", :inline=>"echo 'hello'")
+      expect(subject.post_install_provisioner).to eq(mock_provisioner)
+    end
```
chrisroberts: Should include a test with no `post_install_provision` defined to validate correct behavior.

PR #8724, opened 2017-06-27
Set encoding when reading/writing cwd file

chrisroberts:


templates/locales/en.yml
```diff
@@ -166,9 +166,9 @@ en:
 
       %{list}
     moved_cwd: |-
-      This machine used to live in %{old_wd} but it's now at %{current_wd}.
-      Please change the name of the machine if you want to run it as a different
-      machine.
+      This machine used to live in %{old_wd} but it's now at %{current_wd}. Depending
```
briancain: I think I'd place the `Depending` here on a new line. It kind of looks funky on my machine after the working dir gets printed since it ends up being on its own line

screen shot 2017-06-27 at 9 05 38 am

PR #8725, opened 2017-06-27
Rebuild command and arguments before exec on Windows

### LINKED ISSUE #8690: vagrant 1.9.4 and 1.9.5 cannot handle spaces in a vm.name
### Vagrant version
1.9.5
1.9.4

### Host operating system
Windows 7 Professional

### Guest operating system
Linux, all varieties (tested on centos-6 and debian8)

### Vagrantfile
```# -*- mode: ruby -*-
# vi: set ft=ruby :

  config.vm.box = "puppetlabs/centos-6.6-64-nocm"

  vm_name = "spaces test"
  config.vm.define vm_name

     vb.gui = true
     vb.memory = "1024"
     vb.name = vm_name
  end

end
```

### Debug output
I'm sorry, I can't spend more hours doing the uninstall/reboot/install/reboot thing to break my work computer again, get you the debug, then repeat with a good version so I can get some work done.

This one is pretty obvious.

### Expected behavior
all vagrant commands such as 'vagrant ssh' should work

### Actual behavior
```$ vagrant ssh
Warning: Identity file C:/Users/vds/Desktop/vag/vmbuilder/vms/.vagrant/machines/spaces not accessible: No such file or directory.
vagrant@127.0.0.1's password:
```
What's going on seems to be the vm.name not being quoted correctly while constructing the pathname to the machines subdirectory, which is 'spaces test' (with embedded space) if you do a 'ls' in cygwin.

Note that VirtualBox 'does' support spaces in a vm name just fine, and that previous versions of vagrant used for several years (1.9.3, 1.9.2, 1.9.1, 1.9.0, 1.8.0, and 1.7.4 at a minimum) all worked just fine.   The functionality broke in 1.9.4 and it's still broken in 1.9.5

### Steps to reproduce
1. install 1.9.5 or 1.9.4 of vagrant
2. install the latest Virtual Box (I use Version 5.1.22 r115126)
3. do a 'vagrant up' which will work
4. do a 'vagrant ssh' and you will see the error above

### References
I previously asked on the vagrant google group, but got a non-answer blaming ruby for some reason, which is clearly incorrect.  This has worked fine for 2+ years on vagrant/virtualbox on windows.

chrisroberts:
Flat command can cause issues with arguments. Force common encoding of all strings before passing to exec.

Fixes #8690

### DIFF
--- a/lib/vagrant/util/safe_exec.rb
+++ b/lib/vagrant/util/safe_exec.rb
@@ -41,14 +41,11 @@ module Vagrant
             end
           else
             if Vagrant::Util::Platform.windows?
-              @@logger.debug("Converting command and arguments to single string for exec")
-              @@logger.debug("Command: `#{command.inspect}` Args: `#{args.inspect}`")
-              full_command = "#{command} #{args.join(" ")}"
-              @@logger.debug("Converted command: #{full_command}")
-              Kernel.exec(full_command)
-            else
-              Kernel.exec(command, *args)
+              # Re-generate strings to ensure common encoding
+              command = "#{command}".force_encoding("UTF-8")
+              args = args.map{|arg| "#{arg}".force_encoding("UTF-8") }
             end
+            Kernel.exec(command, *args)
           end
         rescue *rescue_from
           # We retried already, raise the issue and be done


lib/vagrant/util/safe_exec.rb
```diff
@@ -41,14 +41,11 @@ def self.exec(command, *args)
...
             if Vagrant::Util::Platform.windows?
-              @@logger.debug("Converting command and arguments to single string for exec")
-              @@logger.debug("Command: `#{command.inspect}` Args: `#{args.inspect}`")
-              full_command = "#{command} #{args.join(" ")}"
-              @@logger.debug("Converted command: #{full_command}")
-              Kernel.exec(full_command)
-            else
-              Kernel.exec(command, *args)
+              # Re-generate strings to ensure common encoding
+              command = "#{command}".force_encoding("UTF-8")
```
briancain: Should we have the logger produce a message about forcing UTF-8 encoding?

chrisroberts: Certainly wouldn't hurt

PR #8740, opened 2017-06-29
Invoke tee from specific path

### LINKED ISSUE #6913: BSD host's NFS exports editing uses tee from $PATH
I am running Vagrant 1.7.3 on OS X 10.11 with gnu coreutils from homebrew (using the `gnubin` $PATH hack), and was wondering why my sudoers entries from the Vagrant documentation were not having an effect: I was still prompted despite having the exact /etc/sudoers entries from the documentation on my system.

Console.app however reveals that Vagrant is picking the wrong `tee` program:

```
1/22/16 12:04:28.489 AM sudo[24588]:      asf : TTY=ttys001 ; PWD=/Users/asf/Hacks/mytool ; USER=root ; COMMAND=/usr/local/opt/coreutils/libexec/gnubin/tee -a /etc/exports
```

(my `$PATH` with the gnubin hack is `/usr/local/opt/coreutils/libexec/gnubin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin`)

 this doesn't seem optimal, and in the worst case could also lead to a silly code-exec situation where a downloaded malicious Vagrantfile overrides `$PATH` to point to their own `tee` and then runs programs as root on a local machine when the oblivious user enters their password at the prompt (this seems far-fetched, but still ... ugh).

I believe the fix would be to specify an explicit path on https://github.com/mitchellh/vagrant/blob/master/plugins/hosts/bsd/cap/nfs.rb#L114, so `/usr/bin/tee` (as specified in the docs). Happy to submit a pull request if you agree (:

briancain:
This commit updates the use of `tee` to use an explicit path when
invoked rather than relying on where `tee` has been defined in a users
PATH.

Fixes #6913

### DIFF
--- a/plugins/hosts/bsd/cap/nfs.rb
+++ b/plugins/hosts/bsd/cap/nfs.rb
@@ -111,7 +111,7 @@ module VagrantPlugins
             line = Vagrant::Util::ShellQuote.escape(line, "'")
             system(
               "echo '#{line}' | " +
-              "#{sudo_command}tee -a /etc/exports >/dev/null")
+              "#{sudo_command}/usr/bin/tee -a /etc/exports >/dev/null")
           end
 
           # We run restart here instead of "update" just in case nfsd


REBELinBLUE:
Out of interest, what is the reason for this? Confused me for a while today when I upgraded vagrant and my sudoers file stopped working (because previously it was using the version from $PATH so I had my file setup for that). Just wondered why this one is explicit whilst `nfsd` and `sed` are allowed to resolve from $PATH rather than being explicit?

briancain:
Hello @REBELinBLUE - if you are curious about the change, check out the originally filed issue for this PR here https://github.com/mitchellh/vagrant/issues/6913. Thanks!

REBELinBLUE:
Thanks! I could have sworn when I clicked the link in the first reply on Saturday the linked issue was empty. That is exactly the setup I had with brew, I guess at some point and I had figured out why sudoers wasn't working and changed it 😂 Would it not be the same problem if you've installed `sed` via homebrew then? i.e. using `/usr/local/bin/sed` instead of `/usr/bin/sed` (which is indeed how I have it setup)

briancain:
@REBELinBLUE - that is likely true, however I don't see a place in Vagrant where it uses `sed` like how it uses `tee` (or at all really). Do you see a place in the code that's using `sed`?

Thanks!

REBELinBLUE:
When it removes the NFS export https://github.com/mitchellh/vagrant/blob/master/plugins/hosts/bsd/cap/nfs.rb#L176-L186

briancain:
@REBELinBLUE - thanks! I'd recommend opening a new issue similar to the `tee` issue so that we can work on it. Thanks!

PR #8756, opened 2017-07-06
Only rsync-auto current working dir with docker provider

### LINKED ISSUE #5160: Docker: rsync-auto to only watch current vagrant folder and not all projects on proxy VM
I have a setup where I am using Vagrant to spin up a Docker host using the `vagrant_vagrantfile` link and then I have multiple "projects" that are using this same Docker host. I am using rsync to get files into the Docker host, and rsync-auto to keep pushing files into the host.

What I have noticed is that when I run `vagrant rsync-auto` in on of the projects, it starts the rsync watchers across ALL projects that are using the Docker host.

I'm assuming this is because the rsync-auto command is using the shared folder settings of the host, and not just looking at the shared folder settings of the project it is in.

I am suggesting a change to that. I think it makes more sense to only start the rsync watcher for the project I am in, as a day could come when I have a dozen projects all sharing the same Docker host and watching all of them at the same time could really bog things down.

briancain:
Prior to this commit, when users invoked the `rsync-auto` command using
the docker provider with boot2docker, vagrant would rsync all known
containers using the boot2docker vm rather than the current working dir.
This commit updates that behavior to ensure that only the current
working dirs vagrant machines will be rsynced.

Fixes #5160

### DIFF
--- a/plugins/synced_folders/rsync/command/rsync_auto.rb
+++ b/plugins/synced_folders/rsync/command/rsync_auto.rb
@@ -48,6 +48,9 @@ module VagrantPlugins
           paths = {}
           ignores = []
           with_target_vms(argv) do |machine|
+            next if machine.state.id == :not_created
+            cwd = machine.env.cwd.to_s
+
             if machine.provider.capability?(:proxy_machine)
               proxy = machine.provider.capability(:proxy_machine)
               if proxy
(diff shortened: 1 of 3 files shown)


plugins/synced_folders/rsync/command/rsync_auto.rb
```diff
@@ -48,6 +48,9 @@ def execute
           paths = {}
           ignores = []
           with_target_vms(argv) do |machine|
+            next if machine.state.id == :not_created
```
briancain: FYI, this does not seem to be an issue with normal machines, as later on the sync_folders function ends up returning an empty hash so nothing is synced...but with docker machines since they both are technically the same "machine", it would attempt to sync docker containers that hadn't been created yet.

PR #8775, opened 2017-07-12
Added exception if chef.node_path is defined on Vagrantfile but the directory does not exist locally

lucky-sideburn:
If you chef.node_path does not exist locally, the Chef provision generates an error related a chef_server_url instead say the real reason:

==> myvm: Resource Not Found:
==> myvm: -------------------
==> myvm: The server returned a HTTP 404. This usually indicates that your chef_server_url is incorrect.

I wrote a couple of lines for manage this error and warn the user for creating the folder locally.

### DIFF
--- a/plugins/provisioners/chef/provisioner/chef_solo.rb
+++ b/plugins/provisioners/chef/provisioner/chef_solo.rb
@@ -87,7 +87,8 @@ module VagrantPlugins
                 key         = Digest::MD5.hexdigest(local_path)
                 remote_path = "#{guest_provisioning_path}/#{key}"
               else
-                @machine.ui.warn(I18n.t("vagrant.provisioners.chef.cookbook_folder_not_found_warning",
+                appended_folder = "cookbooks" if appended_folder.nil?
+                @machine.ui.warn(I18n.t("vagrant.provisioners.chef.#{appended_folder}_folder_not_found_warning",
                                        path: local_path.to_s))
                 next
               end
--- a/plugins/provisioners/chef/provisioner/chef_zero.rb
+++ b/plugins/provisioners/chef/provisioner/chef_zero.rb
@@ -35,7 +35,6 @@ module VagrantPlugins
           upload_encrypted_data_bag_secret
           setup_json
           setup_zero_config
-          verify_chef_nodes_folder
           run_chef_zero
           delete_encrypted_data_bag_secret
         end
@@ -99,12 +98,6 @@ module VagrantPlugins
             if !@machine.communicate.test("test -d #{folder}", sudo: true)
               raise ChefError, :missing_shared_folders
             end
-          end
-        end
-
-        def verify_chef_nodes_folder
-          if !File.exists? @config.nodes_path[0][1]
-            raise ChefError, :missing_chef_node_folder
           end
         end
 
--- a/plugins/provisioners/chef/provisioner/chef_zero.rb
+++ b/plugins/provisioners/chef/provisioner/chef_zero.rb
@@ -103,7 +103,7 @@ module VagrantPlugins
         end
 
         def verify_chef_nodes_folder
-          if not File.exists? @config.nodes_path[0][1]
+          if !File.exists? @config.nodes_path[0][1]
             raise ChefError, :missing_chef_node_folder
           end
         end
--- a/plugins/provisioners/chef/provisioner/chef_zero.rb
+++ b/plugins/provisioners/chef/provisioner/chef_zero.rb
@@ -35,6 +35,7 @@ module VagrantPlugins
           upload_encrypted_data_bag_secret
           setup_json
           setup_zero_config
+          verify_chef_nodes_folder
           run_chef_zero
           delete_encrypted_data_bag_secret
         end
@@ -98,6 +99,12 @@ module VagrantPlugins
             if !@machine.communicate.test("test -d #{folder}", sudo: true)
               raise ChefError, :missing_shared_folders
             end
+          end
+        end
+
+        def verify_chef_nodes_folder
+          if not File.exists? @config.nodes_path[0][1]
+            raise ChefError, :missing_chef_node_folder
           end
         end
 
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -2192,8 +2192,6 @@ en:
         running_solo_again: "Running chef-solo again (failed to converge)..."
         running_zero: "Running chef-client (local-mode)..."
         running_zero_again: "Running chef-client (local-mode) again (failed to converge)..."
-        missing_chef_node_folder: |-
-          Nodes folder that Chef requires is missing in this Vagrant project
         missing_shared_folders: |-
           Shared folders that Chef requires are missing on the virtual machine.
           This is usually due to configuration changing after already booting the
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -1579,6 +1579,8 @@ en:
           Missing required value for `chef.nodes_path'.
         environment_path_missing: |-
           Environment path not found: %{path}
+        nodes_path_missing: |-
+          Path specified for `nodes_path` does not exist: %{path}
         environment_path_required: |-
           When 'environment' is specified, you must provide 'environments_path'.
         cookbooks_path_missing: |-
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -2192,6 +2192,8 @@ en:
         running_solo_again: "Running chef-solo again (failed to converge)..."
         running_zero: "Running chef-client (local-mode)..."
         running_zero_again: "Running chef-client (local-mode) again (failed to converge)..."
+        missing_chef_node_folder: |-
+          Nodes folder that Chef requires is missing in this Vagrant project
         missing_shared_folders: |-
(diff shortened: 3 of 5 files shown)


lucky-sideburn:
Hi guys! Can this PR be useful?

briancain:
Hey @lucky-sideburn - I tried to reproduce this error and instead it just gave me a warning, and was able to continue on provisioning the vm with chef:

```ruby
  config.vm.define "chef" do |chef|
    chef.vm.box = "bento/ubuntu-16.04"
    chef.vm.provider :virtualbox

    chef.vm.provision :chef_solo do |c|
      c.add_recipe "test"
      c.nodes_path = "/not/real"
    end
  end
```

I saw a warning, which could be improved:

```
==> chef: The cookbook path '/not/real' doesn't exist. Ignoring...
```

But otherwise I didn't get the error you mentioned. Do you have a Vagrantfile you can share that will reproduce this? Thanks!

lucky-sideburn:
Hi briancain!

Thank you for your reply..

I think because you are using chef_solo. If I use chef_solo I have the warning correctly:

==> vmgitlab: The cookbook path '/Users/eugenio/WORK/MEDIOLANUM/cookbooks/vm_gitlab/chef/nodesccc' doesn't exist. Ignoring...

but with chef_zero I have error related a resource not found (it is correct from a Chef point of view, but maybe Vagrant should warn the user before the provisioning)

==> vmgitlab: Resource Not Found:
==> vmgitlab: -------------------
==> vmgitlab: The server returned a HTTP 404. This usually indicates that your chef_server_url is incorrect

Below, my Vagrantfile:

VAGRANTFILE_API_VERSION = '2'.freeze

Vagrant.require_version '>= 1.5.0'

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|

  config.vm.hostname = 'cookbook-omnibus-gitlab'

  config.omnibus.chef_version = :latest

  config.vm.box = 'my-centos7-base.box'

  config.vm.provision 'shell',
                      inline: "sed -i 's/proxy.*//g' /etc/yum.conf"

  config.vm.network :private_network, type: 'dhcp'



  config.berkshelf.berksfile_path = './Berksfile'

  config.berkshelf.enabled = true

  roles_path = './chef/roles'
  databags_path = './chef/data_bags'

  #fake path
  nodes_path = './chef/foooooobar'

  config.vm.define 'vmgitlab' do |vmgitlab|
    vmgitlab.vm.provision :chef_zero do |chef|
      chef.roles_path = roles_path
      chef.nodes_path = nodes_path
      chef.run_list = ['role[gitlab]']
    end
  end

end

briancain:
@lucky-sideburn - ah ok. I switched to `:chef_zero` and I still get a warning, but now I see the error from chef as described. Pull request review incoming.

plugins/provisioners/chef/provisioner/chef_zero.rb
```diff
@@ -101,6 +102,12 @@ def verify_shared_folders(folders)
           end
         end
 
+        def verify_chef_nodes_folder
+          if !File.exists? @config.nodes_path[0][1]
+            raise ChefError, :missing_chef_node_folder
+          end
+        end
```
briancain: OK, I think instead of doing this, you should instead add a new validation under `chef_zero`'s config validate that ensures each element in the `nodes_path` array exists on disk, and if not, appends a new error message to that `errors` array (make sure it works for the case where it's either just a `string` or `array`, since it looks like it can be both according to the docs). Similar to how it works when the config setting is missing completely. The class is [here](https://github.com/mitchellh/vagrant/blob/master/plugins/provisioners/chef/config/chef_zero.rb#L80-L82). I'd put that method under the highlighted if-statement there.

Also PS don't forget to [add a new test](https://github.com/mitchellh/vagrant/blob/master/test/unit/plugins/provisioners/chef/config/chef_zero_test.rb#L105-L110)! :)

templates/locales/en.yml
```diff
@@ -2192,6 +2192,8 @@ en:
         running_solo_again: "Running chef-solo again (failed to converge)..."
         running_zero: "Running chef-client (local-mode)..."
         running_zero_again: "Running chef-client (local-mode) again (failed to converge)..."
+        missing_chef_node_folder: |-
+          Nodes folder that Chef requires is missing in this Vagrant project
```
briancain: I think this would read clearer if it included the missing folder in question. Then someone would be able to know which folder is missing. Something like:

```
nodes_path_missing: |-
  The following `nodes_path` is not present: %{path}
```

Then I'd move the error key to here, next to the other chef config error keys: https://github.com/mitchellh/vagrant/blob/master/templates/locales/en.yml#L1578

Then when you iterate over each possible `nodes_path` array in the `#validate` config method, you can raise the exception like so:

```
errors << I18n.t("vagrant.config.chef.nodes_path_missing", path: nodes_path)
```

lucky-sideburn:
Hi Brian,

thanks ;)
Just for trying.. I wrote code for chef_zero's config validation and unit test as you suggested!

Since Chef Zero provisioning does not produce errors if at least one of the folders specified in nodes_path exists, I decided to produce error only if there are no existing folders totally. I tried w/ array and string for nodes_path in my Vagrantfile and sounds good.

ps.
My provision generate "The cookbook path '/foobar' doesn't exist. Ignoring..." (/foobar is not cookbooks_path).
In plugins/provisioners/chef/provisioner/chef_solo.rb expanded_folders() produce cookbook_folder_not_found_warning for all kind of folder. May be can useful distinguish by type?

briancain:
hey @lucky-sideburn thanks for the updates! Just an FYI, I'm in the process of moving to a new place, so I won't be able to take a look at your PR this week. Just letting you know that I haven't dropped off :)

briancain:
Hey @lucky-sideburn , the updated changes look good!

On your last comment...

> ps.
My provision generate "The cookbook path '/foobar' doesn't exist. Ignoring..." (/foobar is not cookbooks_path).
In plugins/provisioners/chef/provisioner/chef_solo.rb expanded_folders() produce cookbook_folder_not_found_warning for all kind of folder. May be can useful distinguish by type?

Can you explain what you mean here?

lucky-sideburn:
Hi Brian,

thank you for merging ;)

In plugins/provisioners/chef/provisioner/chef_solo.rb I am watching is this code:
 
```
             if File.exist?(local_path)
                # Path exists on the host, setup the remote path. We use
                # the MD5 of the local path so that it is predictable.
                key         = Digest::MD5.hexdigest(local_path)
                remote_path = "#{guest_provisioning_path}/#{key}"
              else
                @machine.ui.warn(I18n.t("vagrant.provisioners.chef.cookbook_folder_not_found_warning",
                                       path: local_path.to_s))
                next
              end
```

machine.ui.warn use "vagrant.provisioners.chef.cookbook_folder_not_found_warning" even for nodes and roles paths.

This is my Vagrantfile and warnings produced:

```
  config.vm.define 'vmgitlab' do |vmgitlab|
    vmgitlab.vm.provision :chef_zero do |chef|
      chef.roles_path = "/path/not/real2"
      chef.nodes_path = ['/tmp','/path/not/real1']
      #chef.run_list = ['role[gitlab]']
      chef.cookbooks_path = "."
    end
  end
```

```
==> vmgitlab: The cookbook path '/path/not/real2' doesn't exist. Ignoring...
==> vmgitlab: The cookbook path '/path/not/real1' doesn't exist. Ignoring... 

```

Thanks!

briancain:
@lucky-sideburn oh I see, so the change could be improving that warning message about what directory actually exists and what kind of folder it is? That change makes sense to me! If you wanna push that up that would be great, thank you!

plugins/provisioners/chef/provisioner/chef_solo.rb
```diff
@@ -87,7 +87,7 @@ def expanded_folders(paths, appended_folder=nil)
                 key         = Digest::MD5.hexdigest(local_path)
                 remote_path = "#{guest_provisioning_path}/#{key}"
               else
-                @machine.ui.warn(I18n.t("vagrant.provisioners.chef.cookbook_folder_not_found_warning",
+                @machine.ui.warn(I18n.t("vagrant.provisioners.chef.#{appended_folder}_folder_not_found_warning",
```
briancain: This is great! I think I'd like one last change. In the case that `appended_folder` was not provided and is nil, we should default to `cookbook_folder` so that users don't see an exception from the i18n library. So basically

```ruby
else
  appended_folder = "cookbook_folder" if appended_folder.nil?
  @machine.ui.warn(I18n.t("vagrant.provisioners.chef.#{appended_folder}_folder_not_found_warning",
    path: local_path.to_s))
end
```

briancain:
@lucky-sideburn excellent! I had one last comment before merging. Should be a straight forward fix :) Once you push that up, I'll be happy to merge! Thanks again for your contribution.

plugins/provisioners/chef/provisioner/chef_solo.rb
lucky-sideburn: Done!

lucky-sideburn:
I pushed up last changes, I have made some rebase for squashing some commit!

briancain:
@lucky-sideburn - thanks so much!

PR #8790, opened 2017-07-18
winssh communicator: fix public key insertion

rgl:
we need to retain the authorized_keys file acl permissions because the [OpenSSH v0.0.17.0](https://github.com/PowerShell/Win32-OpenSSH/releases/tag/v0.0.17.0) enforces them by default (and refuses to work without the correct acl).

### DIFF
--- a/plugins/guests/windows/cap/public_key.rb
+++ b/plugins/guests/windows/cap/public_key.rb
@@ -50,6 +50,7 @@ module VagrantPlugins
           end
           comm.upload(keys_file.path, remote_upload_path)
           keys_file.delete
+          comm.execute("Set-Acl \"#{remote_upload_path}\" (Get-Acl \"#{remote_authkeys_path}\")", shell: "powershell")
           comm.execute("move /y \"#{remote_upload_path}\" \"#{remote_authkeys_path}\"", shell: "cmd")
         end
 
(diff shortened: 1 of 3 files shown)


rgl:
@chrisroberts take a look now

plugins/guests/windows/cap/public_key.rb
```diff
@@ -6,81 +6,57 @@ module Cap
...
+            modify_authorized_keys machine do |keys|
+              keys.delete(contents)
+            end
           else
             raise Vagrant::Errors::SSHInsertKeyUnsupported
           end
         end
 
-        def self.winssh_insert_public_key(machine, contents)
+        def self.modify_authorized_keys(machine)
```
chrisroberts: Please keep the method signature and parameters the same. There's no need to rename the method and it removes the context that it is specifically for winssh. Thanks!

rgl: rename it to `winssh_modify_authorized_keys(machine)`?

the parameters cannot stay the same because that method is not really the same as `winssh_insert_public_key`. its used to implement the old `winssh_insert_public_key` and `winssh_remove_public_key`.

chrisroberts:
Closing as this has been merged into master. Thanks!

PR #8805, opened 2017-07-20
add note about groups for sudoers

kikitux:
Improve documentation.

Fix https://github.com/mitchellh/vagrant/issues/8801

### DIFF
--- a/website/source/docs/synced-folders/nfs.html.md
+++ b/website/source/docs/synced-folders/nfs.html.md
@@ -143,6 +143,11 @@ For \*nix users, make sure to edit your `/etc/sudoers` file with `visudo`. It pr
 
 All of the snippets below require Vagrant version 1.7.3 or higher.
 
+<div class="alert alert-warning" role="alert">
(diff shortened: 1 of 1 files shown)


website/source/docs/synced-folders/nfs.html.md
```diff
@@ -143,6 +143,11 @@ For \*nix users, make sure to edit your `/etc/sudoers` file with `visudo`. It pr
 
 All of the snippets below require Vagrant version 1.7.3 or higher.
 
+<div class="alert alert-warning" role="alert">
+  <strong>Use the apropiate group for your user</strong> The following examples may
```
briancain: Replace `apropiate` with `appropriate`

briancain: With the last sentence, I think we can be a little more helpful here. Perhaps we could say something like:

> Depending on how your machine is configured, you might need to use a different admin group than the ones listed in the examples below.

PR #8817, opened 2017-07-25
use ubuntu/xenial64 instead of hashicorp/precise64 for tutorial

DwordPtr:
I started with the tutorial today and this could trip up new users. Since precise's repos aren't up anymore its impossible to install updates from ubuntu or do much in the way of a normal ubuntu machine. 

I imagine using the current it would be helpful for most users.

This project looks really cool so far!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ IMPROVEMENTS:
   - guests/bsd: Invoke `tee` with explicit path [GH-8740]
   - guests/smartos: Guest updates for host name and nfs capabilities [GH-8695]
   - guests/windows: Add public key capabilities for WinSSH communicator [GH-8761]
+  - providers/virtualbox: Filter machine IPs when preparing NFS settings [GH-8819]
 
 BUG FIXES:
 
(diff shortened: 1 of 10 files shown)


edmorley:
Would the `bento/ubuntu-16.04` boxes be a better choice than `ubuntu/xenial64`? I've heard that the latter doesn't follow all of the Vagrant box creation best practices...

mightyfree:
Agreed. As per official Vagrant documentation (https://www.vagrantup.com/docs/boxes.html), the `hashicorp` and `bento` boxes are the only two officially-recommended box sets. Additionally, Vagrant says of `ubuntu` boxes:

> These boxes only support Virtualbox and do not provide an ideal experience for most users. If you encounter issues with these boxes, please try the Bento boxes instead.

I agree with @edmorley, the `bento/ubuntu-16.04` box is the way to go.

briancain:
Hi all,

We are currently working on building and releasing a full set of HashiCorp supported vagrant boxes. These should be released soon so be on the lookout! I'm going to go ahead and close this though, thanks!

PR #8839, opened 2017-07-31
Validate powershell prior to powershell use

chrisroberts:
Adds powershell validation to ensure powershell is available on
the PATH and checks powershell version to ensure meets the
defined minimum powershell version.

### DIFF
--- a/lib/vagrant/util/powershell.rb
+++ b/lib/vagrant/util/powershell.rb
@@ -8,8 +8,15 @@ module Vagrant
     # This is primarily a convenience wrapper around Subprocess that
     # properly sets powershell flags for you.
     class PowerShell
+      # NOTE: Version checks are only on Major
+      MINIMUM_REQUIRED_VERSION = 3
+
+      # @return [Boolean] powershell executable available on PATH
       def self.available?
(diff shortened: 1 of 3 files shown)


lib/vagrant/util/powershell.rb
```diff
@@ -8,15 +8,23 @@ module Util
...
     # properly sets powershell flags for you.
     class PowerShell
+      # NOTE: Version checks are only on Major
+      MINIMUM_REQUIRED_VERSION = 3
+
+      # @return [Boolean] powershell executable available on PATH
       def self.available?
-        !!Which.which("powershell")
+        if !defined?(@_powershell_available)
+          @_powershell_available = !!Which.which("powershell")
```
briancain: Isn't `!!Which.which("powershell")` equivalent to `Which.which("powershell")`?

briancain: Oh, perhaps the `!!` is turning what `which` returns, `nil`, into a boolean?

chrisroberts: @briancain yes, exactly that. it's just forcing a boolean type.

PR #8844, opened 2017-08-02
fix guest-linux private_key bug

srockstyle:
Default  can't "vagrant up virtual machine".
 So I added chmod permission authorized_keys

### DIFF
--- a/plugins/guests/linux/cap/public_key.rb
+++ b/plugins/guests/linux/cap/public_key.rb
@@ -53,6 +53,7 @@ module VagrantPlugins
               result=$?
             fi
             rm -f '#{remote_path}'
+            chmod 0600 ~/.ssh/authorized_keys
             exit $result
           EOH
         end


plugins/guests/linux/cap/public_key.rb
```diff
@@ -53,6 +53,7 @@ def self.remove_public_key(machine, contents)
               result=$?
             fi
             rm -f '#{remote_path}'
+            chmod 0600 ~/.ssh/authorized_keys
```
briancain: Hey there! Thanks for your contribution. I have a question....

It looks like this step is already happening and chmod'ing authorized_keys in the `if` block above. Do you have a reproduction case where it's failing, requiring this change? Thank you!

srockstyle: Hi briancain. Thank you for reply.

It seems to have been fixed by this change ,
https://github.com/mitchellh/vagrant/commit/a6760dd8e7743e048cb2f38c474e05889356e8ac
But the change seems to have been deleted now.Any change you intended? Please let me know if possible

briancain: Hmm, it looks to still be there on the master branch, as far as I can tell. Check it out [here](https://github.com/mitchellh/vagrant/blob/a6760dd8e7743e048cb2f38c474e05889356e8ac/plugins/guests/linux/cap/public_key.rb#L57)

srockstyle: Oh, Is that so.I will try again and report again. Thank you!

srockstyle: Hi briancain

I tried to check master branch, it but it seems to have not been fix yet. 
[here](https://github.com/mitchellh/vagrant/blob/master/plugins/guests/linux/cap/public_key.rb#L54)

I am sorry if I look at a different branch.

srockstyle: It is not reflected even in the latest version(v1.9.7) of the feeling I saw.Is this okay with this?
[here](https://github.com/mitchellh/vagrant/blob/v1.9.7/plugins/guests/linux/cap/public_key.rb#L54)
Sorry for my misunderstanding.

briancain: Ah, you are right. I'm not sure why I was looking at an old commit on github but it doesn't seem to be there on master.

briancain: It's still in this line though:

```ruby
mv ~/.ssh/authorized_keys.tmp ~/.ssh/authorized_keys && chmod 0600 ~/.ssh/authorized_keys
```

At line 56. This should still be chmod'ing the correct file. Are you finding that's not the case? Do you have any vagrant logs to share?

Thanks!

srockstyle: I think that it may work if that code is in normal circumstances. but I sometimes use a tool that uses vagrant as a plugin. for example  [test-kitchen](https://github.com/test-kitchen/kitchen-vagrant) and so.At such times, it often happened that an error occurred because the condition could not be satisfied(
I was watching the log, but I have erased it last time and it is not at hand. Sorry).

I think that it is not harm to do this change.

briancain: If possible, could you please provide the debug output for when it fails? Thanks!

chrisroberts:
Hey there,

Issue #8830, opened 2017-07-28
Ability to use powershell command in elevated mode.


wiltaylor:
Please note that the Vagrant issue tracker is reserved for bug reports and
enhancements. For general usage questions, please use the Vagrant mailing list:
https://groups.google.com/forum/#!forum/vagrant-up. Thank you!

### Vagrant version
Vagrant 1.9.7

### Host operating system
Windows 10 Enterprise x64 1703

### Guest operating system
Windows 2012 R2 Enterprise

### Debug output
Provide a link to a GitHub Gist containing the complete debug output:
https://www.vagrantup.com/docs/other/debugging.html. The debug output should
be very long. Do NOT paste the debug output in the issue, just paste the
link to the Gist.

### Expected behavior
Ability to access SMB shared folders while using the "vagrant powershell" command to connect to virtual machine.

### Actual behavior
Because of the way winrm works you get access denied. RDP works as expected. To work around this the shell provisioner has an privilged option which uses the task scheduler to run the command. This bypasses the issue.

### Steps to reproduce
1. run "vagrant powershell"
2. cd c:\vagrant

### Suggestion:
Add --Elevated commandline option to allow execution under the same method as the provisioner.

PR #8845, opened 2017-08-02
Added changes to allow elevated PowerShell commands from cli

wiltaylor:
This pull request is in to resolve ticket #8830

Added an --elevate / -e option for the Powershell command. This allows you to run commands as elevated like the shell provisioners use.

This is particularly useful for use with the smb synced folders as windows will not let you access them regularly with just plain a plain winrm connection.

### DIFF
--- a/plugins/commands/powershell/command.rb
+++ b/plugins/commands/powershell/command.rb
@@ -13,6 +13,10 @@ module VagrantPlugins
       def execute
         options = {}
 
+        comops = {
+            elevated:    false
+        }
+
         opts = OptionParser.new do |o|
           o.banner = "Usage: vagrant powershell [-- extra powershell args]"
 
@@ -22,6 +26,10 @@ module VagrantPlugins
 
           o.on("-c", "--command COMMAND", "Execute a powershell command directly") do |c|
             options[:command] = c
+          end
+
+          o.on("-e", "--elevate", "Execute command elevated using task scheduler") do |c|
+            comops[:elevated] = true
           end
         end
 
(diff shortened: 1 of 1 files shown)


plugins/commands/powershell/command.rb
```diff
@@ -51,14 +59,14 @@ def execute
...
-              machine.ui.detail(data) if type == :stdout
-            end
-            if out_code == 0
-              machine.ui.success("Command: #{options[:command]} executed succesfully with output code #{out_code}.")
-            end
-            next
+            out_code = machine.communicate.execute(options[:command].dup, comops) do |type,data|
+            machine.ui.detail(data) if type == :stdout
+          end
+          if out_code == 0
```
briancain: Based on these changes, won't `out_code` always be nil here? It seems like the only changes that should of been made here was adding `comops` to the `execute` function, but everything else should of stayed the same?

plugins/commands/powershell/command.rb
```diff
@@ -23,6 +27,10 @@ def execute
           o.on("-c", "--command COMMAND", "Execute a powershell command directly") do |c|
             options[:command] = c
           end
+
+          o.on("-e", "--elevate", "Execute command elevated using task scheduler") do |c|
+            comops[:elevated] = true
```
briancain: Is there a reason why this couldn't just be apart of the `options` hash?

chrisroberts:
Thanks for this PR! Since we haven't heard back from you I'm going to close this up in favor of #10528 which includes the ability to run elevated commands and also fixes up when some requirements are imposed as well as adds test coverage. Cheers!

PR #8873, opened 2017-08-10
Handle calling reboot detection during an existing shutdown

taliesins:
If shutdown is called during an existing shutdown then exception code 1115 is throw.


taliesins:
@briancain what do we need to do to get this merged? its a relatively small PR.

briancain:
Closing this since it has already been implemented via #8874. Thanks!

PR #8874, opened 2017-08-10
Handle chef provisioner reboot request

taliesins:
Handle Chef provisioner requests to reboot. A request to reboot is not a convergence failure, so don't treat it like a convergence failure.

Wait for the machine to reboot before trying to continue the provisioning.


taliesins:
@briancain what do we need to do to get this merged? its a relatively small PR.

chrisroberts:
@taliesins The wait was wanting to introduce this change after the 2.0 release. I'm looking to include it in the next release. Thanks!

PR #8876, opened 2017-08-10
command/login: Add description to created token

justincampbell:
This adds a prompt for a token description, which is now supported in
Vagrant Cloud. Pressing enter on the prompt uses the default description
of `"Vagrant login"`.

    $ vagrant login
    In a moment we will ask for your username and password to HashiCorp's
    Vagrant Cloud. After authenticating, we will store an access token locally on
    disk. Your login details will be transmitted over a secure connection, and
    are never stored on disk locally.

    If you do not have an Vagrant Cloud account, sign up at
    https://www.vagrantcloud.com

    Vagrant Cloud Username: justincampbell
    Password (will be hidden):
    Token description (Defaults to "Vagrant login"):
    You are now logged in.
    $

Which created a token with the default description of "Vagrant login":

Entering a description:

    Token description (Defaults to "Vagrant login"): Justin's MacBook Pro

### DIFF
--- a/plugins/commands/login/client.rb
+++ b/plugins/commands/login/client.rb
@@ -40,15 +40,24 @@ module VagrantPlugins
       # Login logs a user in and returns the token for that user. The token
       # is _not_ stored unless {#store_token} is called.
       #
-      # @param [String] user
-      # @param [String] pass
+      # @param [String] username_or_email
+      # @param [String] password
+      # @param [String] description
       # @return [String] token The access token, or nil if auth failed.
-      def login(user, pass)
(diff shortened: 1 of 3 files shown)


justincampbell:
It would be super-cool to default to the username@hostname, but that seems like a lot of work to do correctly cross-platform

```bash
$ echo "Vagrant login from $(whoami)@$(hostname)"
Vagrant login from Justin@Justins-MBP.home
```

justincampbell:
Added a commit which uses `Socket.gethostname`:

```
Token description (Defaults to "Vagrant login from Justins-MBP.home"):
```

PR #8895, opened 2017-08-16
Introduce extra_args setting for ssh configs

### LINKED ISSUE #8697: Feature Request: Add config.ssh.ssh_options
I would like to have a `config.ssh.ssh_options` to allow adding extra ssh arguments from within the Vagrantfile. It is already possible to pass extra options manually by running `vagrant ssh -- [ssh_options here]`.

Ultimately I want this so that I can add localhost forwarded ports like `-L 127.1.2.7:8888:127.1.2.7:8888`.

The vagrant-multi-putty plugin has an ssh_options config option that does what I want, but uses putty instead of ssh. See https://github.com/nickryand/vagrant-multi-putty#example-multiple-tunnels

### Vagrant version
1.9.3 (reading the source, this feature is not implemented in master)

### Host operating system
Linux (Gentoo)

### Guest operating system
Ubuntu or CentOS or other linux distros

### Vagrantfile
simple example of what I would like:
```ruby
  config.ssh.ssh_options = ["-L", "127.1.2.7:8008:127.1.2.7:8008", "-R", "127.1.2.7:8009:127.1.2.7:8009"]
end
```

or a more complex example:
```ruby
  config.ssh.ssh_options = []
  ports = 61000..62000
  ports.each do |port|
    config.ssh.ssh_options += ["-L", "127.1.2.7:#{port}:127.1.2.7:#{port}"]
  end
end
```

Note that these examples cannot be accomplished with `config.vm.network` because [virtualbox](https://www.virtualbox.org/ticket/13971) (and probably other providers) cannot forward to localhost IPs (127.0.0.1/8). Though a forwarding mechanism could be added to solve my needs, a more generic ssh_options should allow for a greater variety of potential use cases.

### Debug output
Not applicable to a feature request

### Expected behavior
Allow me to specify additional ssh options from within a Vagrantfile. 

### Actual behavior
I can't specify additional options from within the Vagrantfile

### Steps to reproduce
not applicable

### References
Nothing to see here.

briancain:
Prior to this commit, there was no way to add additional ssh arguments
within a Vagrantfile for a given vagrant machine. This commit introduces
a new option extra_args that allows users to pass in a single argument
or an array of flags that will be added onto the ssh command.

 Fixes #8697

### DIFF
--- a/lib/vagrant/util/ssh.rb
+++ b/lib/vagrant/util/ssh.rb
@@ -167,6 +167,12 @@ module Vagrant
         #
         # Without having extra_args be last, the user loses this ability
         command_options += ["-o", "ForwardAgent=yes"] if ssh_info[:forward_agent]
+
+        # Note about :extra_args
+        #   ssh_info[:extra_args] comes from a machines ssh config in a Vagrantfile,
+        #   where as opts[:extra_args] comes from running the ssh command
+        command_options += Array(ssh_info[:extra_args]) if ssh_info[:extra_args]
+
         command_options.concat(opts[:extra_args]) if opts[:extra_args]
 
         # Build up the host string for connecting
--- a/test/unit/vagrant/machine_test.rb
+++ b/test/unit/vagrant/machine_test.rb
@@ -793,6 +793,17 @@ describe Vagrant::Machine do
         it "paranoid should be default" do
           expect(instance.ssh_info[:paranoid]).to be(false)
         end
+        it "extra_args should be nil" do
+          expect(instance.ssh_info[:extra_args]).to be(nil)
+        end
+        it "extra_args should be set" do
+          instance.config.ssh.extra_args = ["-L", "127.1.2.7:8008:127.1.2.7:8008"]
+          expect(instance.ssh_info[:extra_args]).to eq(["-L", "127.1.2.7:8008:127.1.2.7:8008"])
+        end
+        it "extra_args should be set as an array" do
+          instance.config.ssh.extra_args = "-6"
+          expect(instance.ssh_info[:extra_args]).to eq("-6")
+        end
         it "keys_only should be overridden" do
           instance.config.ssh.keys_only = false
           expect(instance.ssh_info[:keys_only]).to be(false)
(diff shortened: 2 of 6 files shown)


briancain:
FYI @chrisroberts - I'm flexible on the actual name of this new config option. I used `extra_args` because that is what was used internally with the `vagrant ssh` command (and perhaps `config.ssh.ssh_options` might confuse users into thinking that's their only way of configuring ssh).

test/unit/vagrant/machine_test.rb
```diff
@@ -793,6 +793,13 @@ def detect?(machine)
         it "paranoid should be default" do
           expect(instance.ssh_info[:paranoid]).to be(false)
         end
+        it "extra_args should be nil" do
+          expect(instance.ssh_info[:extra_args]).to be(nil)
+        end
+        it "extra_args should be set" do
+          instance.config.ssh.extra_args = ["-L" "127.1.2.7:8008:127.1.2.7:8008"]
+          expect(instance.ssh_info[:extra_args]).to eq(["-L" "127.1.2.7:8008:127.1.2.7:8008"])
```
chrisroberts: Add commas to make multi item array

lib/vagrant/util/ssh.rb
```diff
@@ -167,6 +167,18 @@ def self.exec(ssh_info, opts={})
...
+
+        # Note about :extra_args
+        #   ssh_info[:extra_args] comes from a machines ssh config in a Vagrantfile,
+        #   where as opts[:extra_args] comes from running the ssh command
+        if ssh_info[:extra_args]
+          if ssh_info[:extra_args].kind_of?(Array)
+            command_options += ssh_info[:extra_args]
+          else
+            command_options += [ssh_info[:extra_args]]
+          end
```
chrisroberts: Reduce to:

```ruby
command_options += Array(ssh_info[:extra_args])
```

test/unit/vagrant/machine_test.rb
```diff
@@ -793,6 +793,13 @@ def detect?(machine)
...
+        end
```
chrisroberts: Include test with string value being set into config that validates proper conversion to array

briancain: Hmm, the conversion does not happen until the `util/ssh` class, but I think I can at least add a test to make sure it sets it in the single string case.

Issue #8907, opened 2017-08-22
Solus Linux Support


mattandersen:
### Vagrant version
Vagrant 1.9.8.dev

### Host operating system
Solus Linux

### Guest operating system
Any

### Expected behavior
When starting a guest with NFS synced folders, the exports file should be modified correctly and the NFS service should be reloaded successfully.

### Actual behavior
The service is not reloaded, so exports are not available to the guest. This is due to using the fallback linux host which does not attempt to reload services through systemd.

### Proposed Change
Add support for the through a host plugin. See related PR for an implementation.

chrisroberts:
Fixed via #8938. Cheers!

PR #8908, opened 2017-08-22
Add support for Solus Linux

mattandersen:
Provide NFS service and share support for Solus Linux. Based on the
Arch and Fedora hosts.

See issue #8907.


ikeydoherty:
Not that this is really anything to do with me, but I did see this on the interwebs. Shouldn't you be detecting the Linux distribution based on the `os-release` file, and not tertiary assets like `/etc/solus-release` ?

Ideally any distro detection should first look for:

```
    /etc/os-release
    /usr/lib/os-release
```

(most of the world is stuck to using /etc/ for now because tools haven't taken the cascading priority into account)

The `ID` field in os-release is what matters most:

```ini
ID="solus"
```

This is the more robust and standard solution nowadays

mattandersen:
I was just working from the current pattern, but you're quite right. If there is something defined by freedesktop.org that we can rely on, we should. I've updated to reflect that. I am not a Ruby programmer by trade, so I hope it is satisfactory.

chrisroberts:
@mattandersen Hi! Thank you taking the time to create this PR. I was waiting for #8938 to land before reviewing your PR so I could make update recommendations based on the new updates. However, it looks like the updates within the linux guest nfs capability cover the need this PR is fulfilling for solus. As such, it doesn't really make sense to create explicit guest support for solus at this time. However, if there is a need for customized guest capabilities for solus in the future please don't hesitate. Your original PR and updates were great, and I'm sorry that it ended up getting covered by code not yet merged. Thanks again.

Cheers!

mattandersen:
No worries, that was the correct solution all along but I figured something like that was bit much for me to tackle at this point. 👍

PR #8912, opened 2017-08-24
Add IPv6 network template for OpenBSD machines

tolbrino:
The template for ipv6 was missing so far, which resulted in an error when trying to use IPv6 for a machine.


tolbrino:
I rebased these changes on the latest master, since the PR hasn't been touched in a while. I've used the change without issues since then.

tolbrino:
No problem.

PR #8913, opened 2017-08-24
Ansible: Add "compatibility_mode" and "*become*" common options; Move "version" to common options

### LINKED ISSUE #6570: Switch to recommended Ansible 2.x arguments with backwards compatible aliases
- `ask_become_pass` (aliased as `ask_sudo_pass`)
- `become` (aliased as `sudo`)
- `become_user` (aliased as `sudo_user`)
- some parameters in the generated inventory (`ansible_user`, `ansible_host` and `ansible_port`)

Deprecation warnings should also be activated for all these aliases.

Notes:
- The "become" parameters were completely introduced in Ansible 1.9
- The "not ssh specific" parameters (e.g. `ansible_user`) were introduced in Ansible 2.0
- The legacy parameters (e.g. ansible_ssh_user) will probably be removed in Ansible 2.2+.

Open Questions:
- Introduce a "strict" mode, where deprecated parameters are rejected. In such mode, all the settings behind the scene are following Ansible 2+ recommendations (e.g. inventory settings like `ansible_host` instead of `ansible_ssh_host`)
- The "strict" mode could be implemented with a "compatibility_level" option, which for now would support two values: `ansible-1.8` and `ansible-2.0`

Possible Roadmap:
1. Vagrant 1.8.x: backwards compatible introduction of new behaviors (with deprecation warnings)
2. Vagrant 1.9.0: breaking change 1: switch the `compatibility_level` default value to "ansible-2.0" to encourage users to upgrade.
3. Vagrant 2.0.0 (or 1.10.0): breaking change 2: fully remove support of "ansible-1.8" compatibility level

Note: Depending on the situation (ansible 1.9 and 2.x adoption and evolution by the time of releasing Vagrant 1.9), we also could consider introducing the two breaking changes at the same time (i.e. directly in Vagrant 1.9.0)

gildegoma:
Resolve #6570 

### Questions to Reviewer(s)

- [x] For sake of better understanding, should the compatibility mode "auto-detection" better be configured as `compatibility_mode = "auto"` (instead of using `nil` (undefined))? **Yes, done**
- [x] Should we improve the `compatibility_mode` option validation (i.e. rejects invalid values instead of the rather silent fallback to "auto-detection" behaviour)? **Yes, done**
- [x] As mentionned in the commit message, "auto-detection" slightly breaks backwards compatibility. Is it a problem for the Vagrant 2.0 milestone? **Accepted**
- [x] Should we introduce a "deprecation" util library to unify this process across _all_ Vagrant components? **Yes, but it will be introduced a little bit later...**

### Done

- [x] added `compatibility_mode` option
- [x] added `become` & co options and their `sudo` aliases
- [x] added `compatibility_mode` strict validation and introduced the `"auto"` (as default value)
- [x] Implement #8914 by the way
- [x] Add incompatibility sanity checks (i.e. Ansible current version must not conflict with the specified compatibility mode) 
- [x] Optimization: Reduce the number of `ansible` command executions
- [x] Should we better remove the two "programming error" safety checks around compatibility detection? **No, we keep and polish them**
- [x] Get some [reviews/approvals from the Ansible community](https://groups.google.com/d/msg/vagrant-up/ilrun_TouSc/j2aFDR4qAgAJ) about this compatibility mode approach. *Not many feedbacks, but at least no negative ones ;-)* Thanks @resmo 😍
- [x] CHANGELOG update (probably with breaking change note)


chrisroberts:
> For sake of better understanding, should the compatibility mode "auto-detection" better be configured as compatibility_mode = "auto" (instead of using nil (undefined))?

Yes, I think this is a good change and makes the usage much more clear. Having the config option default to `"auto"` would be great.

> Should we improve the compatibility_mode option validation (i.e. rejects invalid values instead of the rather silent fallback to "auto-detection" behaviour)?

Again I think this is a good change. Adding an error message about unsupported values for `compatibility_mode` when the configuration is finalized will help prevent unexpected behavior and provide clarity to the user on what is currently supported. Forcing the user to configure a valid value instead of silently resetting to `"auto"` also prevents the user from assuming the configuration value is having some effect.

> As mentionned in the commit message, "auto-detection" slightly breaks backwards compatibility. Is it a problem for the Vagrant 2.0 milestone?

I _think_ this is okay given the argument in the commit message. I am definitely open to counter arguments but I don't think the impact will be large and it is easy to resolve. One thing that I think would make the change easier is including output when the provisioner is starting its run that identifies what compatibility mode is currently in use along with the current detected version. 

> Should we introduce a "deprecation" util library to unify this process across all Vagrant components?

Yes. We have a utility module for deprecating commands, but a general use deprecation module is definitely something we should introduce to keep things consistent and clean. I will pull something together for that.

Let me know any feedback you have. This all looks great!

gildegoma:
Updated PR main description after adding 46a7333. More coming within the next days...

website/source/docs/provisioning/ansible_common.html.md
```diff
@@ -146,3 +185,16 @@ Some of these options are for advanced usage only and should not be used unless
...
+
+    This option is disabled by default.
+
+    When an Ansible version is defined (e.g. `"2.1.6.0"`), the Ansible provisioner will be executed only if Ansible is installed at the requested version.
+
+    When this option is set to `"latest"`, no version check is applied.
+
+    <div class="alert alert-info">
+      <strong>Tip:</strong>
+      With the `ansible_local` provisioner, it is currently possible to use this option to specify which version of Ansible must be automatically installed, but <strong>only</strong> in combination with the [**`install_mode`**](ansible_local.html#install_mode)  set to <strong>`:pip`</strong>.
```
gildegoma: @chrisroberts I've just noticed that it is totally okay for `middleman` rendering, to link HTML pages with the same directory without the whole path (e.g. `ansible_local.html#install_mode` instead of `/docs/provisioning/ansible_local.html#install_mode`).
If I'm correct, do you agree if I replace all these links (at least within the `/docs/provisioning/ansible*.html` scope) ?

gildegoma:
@chrisroberts @briancain I know this PR is insanely growing, to reach the "unreviewable" state ;-)
This is more like a public "work in progress" branch, so that Chris can look at the "deprecation" stuff and easily answer to my questions.
I also tried to engage the Ansible community to verify the upcoming documentation to ensure that the proposal is well designed (but so far haven't received any feedback).

From my point of view, this is almost ready to be merged into `master`, and I could do the "fine tuning" in separate PR/commits.

website/source/docs/provisioning/ansible_common.html.md
chrisroberts: @gildegoma All the documentation currently uses the full path. Would prefer to keep things using the full path for consistency throughout the documentation pages.

gildegoma: Agreed! I'll fix that one soon. Thanks for the quick reply!!

gildegoma:
@chrisroberts Beside the open tasks in the PR main description, this is ready for review. Most important: ensure that Ruby (and English) style is good enough.

templates/locales/en.yml
```diff
@@ -2364,16 +2364,23 @@ en:
...
+            The requested Ansible compatibility mode (%{compatibility_mode}) is in conflict with
+            the Ansible installation on your Vagrant %{system} system (currently: %{ansible_version}).
+            See https://docs.vagrantup.com/v2/provisioning/ansible_common.html#compatibility_mode
             for more information.
           config_file_not_found: |-
             `%{config_option}` does not exist on the %{system}: %{path}
           extra_vars_invalid: |-
             `extra_vars` must be a hash or a path to an existing file. Received: %{value} (as %{type})
+          no_compatibility_mode: |-
+            `compatibility_mode` must be correctly set (possible values: %{valid_modes}).
```
chrisroberts: > `compatibility_mode` must be a valid mode (possible values: %{valid_modes}).

website/source/docs/provisioning/ansible_common.html.md
```diff
@@ -17,6 +17,36 @@ These options get passed to the `ansible-playbook` command that ships with Ansib
 
 Some of these options are for advanced usage only and should not be used unless you understand their purpose.
 
+- `become` (boolean) - Cause Ansible to perform all the playbook tasks [as another user](http://docs.ansible.com/ansible/become.html), different from the one used to log into the guest system.
```
chrisroberts: > Perform all Ansible playbook tasks [as another user](http://docs.ansible.com/ansible/become.html), different from the user used to log into the guest system.

website/source/docs/provisioning/ansible_common.html.md
```diff
@@ -17,6 +17,36 @@ These options get passed to the `ansible-playbook` command that ships with Ansib
...
+
+    The default value is `false`.
+
+- `become_user` (string) - Set the default username to be used by the Ansible `become` [privilege escalation](http://docs.ansible.com/ansible/become.html) mechanism.
+
+    By default this option is not defined, and the Ansible default value (`root`) will be used.
+
```
chrisroberts: > By default this option is not set

Change of `defined` to `set`

website/source/docs/provisioning/ansible_common.html.md
```diff
@@ -17,6 +17,36 @@ These options get passed to the `ansible-playbook` command that ships with Ansib
...
+- `compatibility_mode` (string) - Set the **minimal** version of Ansible to be supported. Vagrant will use some parameters that are only compatible since the given version.
```
chrisroberts: > Vagrant will only use parameters that are compatible with the given version.

website/source/docs/provisioning/ansible_common.html.md
```diff
@@ -17,6 +17,36 @@ These options get passed to the `ansible-playbook` command that ships with Ansib
...
+
+    Possible values:
+
+    - `"auto"` _(Vagrant will automatically select the optimal compatibilty mode by checking the Ansible version currently available)_
+    - `"1.8"` _(Ansible versions prior to 1.8 should mostly work well, but some options might not be supported)_
+    - `"2.0"` _(The generated Ansible inventory will be incompatible with Ansible 1.x)_
+
+    By default this option is set to `"auto"`. If Vagrant is not able to detect any supported Ansible version, it will falls back on the compatibility mode `"1.8"` with a warning.
```
chrisroberts: > it will fall back to compatibility mode `"1.8"` with a warning.

website/source/docs/provisioning/ansible_common.html.md
```diff
@@ -17,6 +17,36 @@ These options get passed to the `ansible-playbook` command that ships with Ansib
...
+    Vagrant will error if the specified compatibility mode is incompatible with the current Ansible version.
+
+    <div class="alert alert-warning">
+      <strong>Attention:</strong>
+      Vagrant doesn't perform any validation between the `compatibility_mode` value and the value of the [`version`](#version) option.
+    </div>
+
+    <div class="alert alert-info">
+      <strong>Compatibility Note:</strong>
+      This option was introduced in Vagrant 2.0. Previous Vagrant versions behave like if this option was set to `"1.8"`.
```
chrisroberts: > This option was introduced in Vagrant 2.0. The behavior of previous Vagrant versions can be simulated by setting the `compatibility_mode` option to `"1.8"`.

plugins/provisioners/ansible/provisioner/base.rb
```diff
@@ -14,15 +15,78 @@ class Base < Vagrant.plugin("2", :provisioner)
...
+            config.compatibility_mode = Ansible::SAFE_COMPATIBILITY_MODE
+
+            @machine.env.ui.warn(I18n.t("vagrant.provisioners.ansible.compatibility_mode_not_detected",
+              compatibility_mode: config.compatibility_mode,
+              gathered_version: @gathered_version_stdout) +
+            "\n")
+          end
+
+          unless Ansible::COMPATIBILITY_MODES.slice(1..-1).include?(config.compatibility_mode)
+            raise "Programming Error: compatibility_mode must correctly set at this stage!"
```
chrisroberts: Raise custom exception type here with message defined within locales template.

plugins/provisioners/ansible/provisioner/base.rb
```diff
@@ -285,6 +349,39 @@ def verbosity_argument
           end
         end
 
+        private
+
+        def detect_compatibility_mode
+          if !@gathered_version_major || config.compatibility_mode != Ansible::COMPATIBILITY_MODE_AUTO
+            raise "Programming Error: detect_compatibility_mode() shouldn't have been called."
```
chrisroberts: Raise custom exception type here with message defined within locales template.

gildegoma:
Thank you @chrisroberts for the review. If not wrong, all the requested changes have been addressed now.

Below, my proposal for the CHANGELOG entry (please fix english wording at will to make it understandable).

I let you review it again, and do the merge + changelog update (if okay for you so).

-----------
CHANGELOG additions:

in the BREAKING CHANGES section:

  - Both Ansible provisioners are now capable to automatically set the `compatibility_mode` that best fits with the Ansible version in use. You may encounter some compatibility issue when upgrading If you were using Ansible 2.x **and** referring to the `_ssh`-prefixed variables present in the generated inventory (e.g. `ansible_ssh_host`).
  In this case, you can simply fix your Vagrant setup by setting `compatibility_mode = "1.8"`, or by migrating to the new variable names (e.g. `ansible_host`).

In the IMPROVEMENTS section:

  - provisioners/ansible(both): Add the `compatibility_mode` option, with auto-detection enabled by default [GH-8913, GH-6570]
  - provisioners/ansible: Add the `version` option to the host-based provisioner [GH-8913, GH-8914]
  - provisioners/ansible(both): Add the `become` and `become_user` options with deprecation of `sudo` and `sudo_user` options [GH-8913, GH-6570]
  - provisioners/ansible: Add the `ask_become_pass` option with deprecation of the `ask_sudo_pass` option [GH-8913, GH-6570]

chrisroberts:
@gildegoma Will merge and update changelog in the moring.

gildegoma:
@chrisroberts feature branch rebased because of recent e09848c. I kept the whole git history, as I find it relevant and informative. Note that unfortunately no Travis CI build has been started because of the `[ci skip]` in my last commit (I remember that once Travis was able to get back to the last commit to build in a push.... odd :thinking:).

Anyway, this is ready for merge. Can I let you the task to update the changelog (with more appropriate wording...)

Issue #8608, opened 2017-05-22
Docker provider tries to map volumes with host machine path format instead of the host_vm's


agusti-t:
### Vagrant version
1.9.5

### Host operating system
Windows 10

### Guest operating system
ubuntu/trusty64

### Vagrantfile (host)
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
	config.vm.boot_timeout = 120
	config.vm.box = "ubuntu/trusty64"
	config.vm.hostname = "docker-provider-host"
	config.vm.provider :virtualbox do |vb|
		vb.name = "docker-provider-host"
		vb.customize [
                "modifyvm", :id,
                "--memory", 8192,
		]
	end
	
	config.vm.network "private_network", ip: "192.168.33.60"
	
	config.ssh.username = "vagrant"
        config.ssh.password = "vagrant"
        config.ssh.insert_key = true
	
	# Disable sync of current folders
	config.vm.synced_folder ".", "/home/vagrant/sync", disabled: true

        $logs_folders = <<-LOGS_FOLDERS
		mkdir -p /var/log/my-application
	LOGS_FOLDERS
	config.vm.provision "logs_folders", type: "shell", inline: $logs_folders
	
	# Make it possible for the volumes mounted into the container to be written to
	# 	ref: https://goldmann.pl/blog/2014/07/18/logging-with-the-wildfly-docker-image/
	# Observe that since the above post the base docker image for Wildfly has changed
	# and now they use uid/gid 1000/1000 so first regular user, i.e. in a vagrant box
	# it is the vagrant user.
	# 	ref: https://github.com/jboss-dockerfiles/base/blob/master/Dockerfile
	$logs_folders_permissions = <<-LOGS_FOLDERS_PERMISSIONS
		chown -R vagrant:vagrant /var/log/my-application
	LOGS_FOLDERS_PERMISSIONS
	config.vm.provision "logs_folders_permissions", type: "shell", privileged: true, inline: $logs_folders_permissions

	# Install docker
	config.vm.provision "docker"
	
	# Add user vagrant to docker group
	$docker_group = <<-DOCKER_GROUP
		usermod -a -G docker vagrant
		restart docker
		newgrp docker
	DOCKER_GROUP
[...]

PR #8921, opened 2017-08-28
fix for issue #8608/Windows hosts

lookfwd:
Fixes issue #8608. The existing `windows?` expression applies only when it's a local
machine. When it's a VM, it's always set to `false`.


okorz001:
Any progress on this?

chrisroberts:
Thanks!

Issue #8925, opened 2017-08-29
Salt provisioner can't install Windows minions after version 2016.11.7


jonlabroad:
When attempting to use a Salt minion version past 2016.11.7, the provisioner cannot find the installer when downloading:
```
Downloading Salt minion installer Salt-Minion-2017.7.1-AMD64-Setup.exe
powershell.exe : Exception calling "DownloadFile" with "2" argument(s): "The remote server retur
    + CategoryInfo          : NotSpecified: (Exception calli...te server retur:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
ned an error: (404) Not Found."
At C:\tmp\bootstrap_salt.ps1:56 char:24
+ $webclient.DownloadFile <<<< ($url, $file)
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : DotNetMethodException

Installing Salt minion...
C:\tmp\salt.exe /S /minion-name=stable
The term 'C:\tmp\salt.exe' is not recognized as the name of a cmdlet, function,
 script file, or operable program. Check the spelling of the name, or if a path
 was included, verify that the path is correct and try again.
At C:\tmp\bootstrap_salt.ps1:67 char:18
+   C:\tmp\salt.exe <<<<  /S /minion-name=$minion | Out-Null
    + CategoryInfo          : ObjectNotFound: (C:\tmp\salt.exe:String) [], Com
   mandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

This is because Salt has changed their installer filenames to include the Python version after 2016.11.7: https://repo.saltstack.com/windows/

jonlabroad:
This is fixed by #8926

jbonhag:
Closing this issue because it was reported as fixed. 🎉

PR #8926, opened 2017-08-29
Support Windows Salt Minions greater than 2016.x.x

jonlabroad:
This is a fix for #8925 

The latest version of the Windows Salt Minion could not be used because Salt added the Python version to the filename, therefore the Salt provisioner could not find the minion installer.

The Python version is added to the filename if the version is >= 2017.x.x. The Python version can be changed using the python_version parameter (default=2).

Additionally, the Salt provisioner was adding Linux-only bootstrap script parameters to the Powershell script. The expected behavior is that the Linux-only parameters are ignored and not passed into the Powershell script at all.


jonlabroad:
Hi @briancain , I finally got around to adding those tests you requested.

fzk-rec:
When will a new Vagrant version be released? We run 2.0.0 and hit this exact issue :(

jonlabroad:
@fzk-rec 
In the meantime, a decent workaround is making a copy of the script (https://github.com/hashicorp/vagrant/blob/master/plugins/provisioners/salt/bootstrap-salt.ps1) and running it after the Salt provisioner:
`config.vm.provision "shell", path: "bootstrap-salt.ps1", args: "-Version 2017.7.1"`

briancain:
Hi @fzk-rec - we are planning for a release early next week!

fzk-rec:
Thx for the info guys!

PR #8927, opened 2017-08-29
Introduce salt_call_args and salt_args option for salt provisioner

briancain:
This config option for the salt provisioner allows you to pass
additional arguments to the salt-call executable.

This PR also introduces the `salt_args` config option to allow users to pass in additional flags to the `salt` tool.

This overrides #7836

### DIFF
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -24,6 +24,7 @@ module VagrantPlugins
       attr_accessor :log_level
       attr_accessor :masterless
       attr_accessor :minion_id
+      attr_accessor :salt_call_args
 
       ## bootstrap options
       attr_accessor :temp_config_dir
@@ -66,6 +67,7 @@ module VagrantPlugins
         @version = UNSET_VALUE
         @run_service = UNSET_VALUE
         @master_id = UNSET_VALUE
+        @salt_call_args = UNSET_VALUE
       end
 
       def finalize!
@@ -91,6 +93,7 @@ module VagrantPlugins
         @version            = nil if @version == UNSET_VALUE
         @run_service        = nil if @run_service == UNSET_VALUE
         @master_id          = nil if @master_id == UNSET_VALUE
+        @salt_call_args     = nil if @salt_call_args == UNSET_VALUE
 
         # NOTE: Optimistic defaults are set in the provisioner. UNSET_VALUEs
         # are converted there to allow proper detection of unset values.
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -25,6 +25,7 @@ module VagrantPlugins
       attr_accessor :masterless
       attr_accessor :minion_id
       attr_accessor :salt_call_args
+      attr_accessor :salt_args
 
       ## bootstrap options
       attr_accessor :temp_config_dir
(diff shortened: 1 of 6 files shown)


aneeshusa:
I don't currently have a concrete need for this, but it would be nice for this option to pass additional arguments to calls to `salt` as well as `salt-call` (maybe rename the option to `custom_args` or similar?).  I don't want to end up with adding `salt_call_args` now for `salt-call` and then adding e.g. `salt_args` later which only applies to `salt`, which would be confusing.

briancain:
Hey @aneeshusa - I can definitely add a way to pass additional arguments to the `salt` tool, however it should be separate from `salt-call` since they are two separate cli tools with different cli flags. I think it's probably more confusing to have a single way to configure both `salt` and `salt-tool` cli flags that depend on the `masterless` setting being set. With two separate config settings, it's explicit which tool the flags will be applied to. Thanks!

aneeshusa:
Ah, I forgot that they take some different arguments, that makes sense to me. Thanks for reviving this PR by the way!

briancain:
Sure thing! I'll push up an update soon with the additional config option.

plugins/provisioners/salt/config.rb
```diff
@@ -146,6 +152,18 @@ def validate(machine)
           errors << I18n.t("vagrant.provisioners.salt.must_accept_keys")
         end
 
+        if @salt_call_args
+          if !@salt_call_args.is_a?(Array)
```
chrisroberts: This can be reduced to a single check: `if !@salt_call_args.is_a?(Array)`

plugins/provisioners/salt/config.rb
```diff
@@ -146,6 +152,18 @@ def validate(machine)
...
+            errors << I18n.t("vagrant.provisioners.salt.args_array")
+          end
+        end
+
+        if @salt_args
+          if !@salt_args.is_a?(Array)
```
chrisroberts: Reduce as above.

plugins/provisioners/salt/provisioner.rb
```diff
@@ -198,6 +198,30 @@ def get_masterless
         return options
       end
 
+      # Append additional arguments to the salt command
+      def get_salt_args
```
chrisroberts: Can be reduced to: 

```ruby
Array(@config.salt_args).join(" ")
```

plugins/provisioners/salt/provisioner.rb
```diff
@@ -198,6 +198,30 @@ def get_masterless
...
+          @config.salt_args.each do |opt|
+            options += " #{opt}"
+          end
+        end
+
+        return options
+      end
+
+      # Append additional arguments to the salt-call command
+      def get_call_args
```
chrisroberts: Same reduction as above.

plugins/provisioners/salt/provisioner.rb
```diff
@@ -386,7 +410,7 @@ def call_highstate
               unless @config.masterless?
                 @machine.communicate.sudo("salt-call saltutil.sync_all")
               end
-              @machine.communicate.sudo("salt-call state.highstate --retcode-passthrough#{get_masterless}#{get_loglevel}#{get_colorize}#{get_pillar}", ssh_opts) do |type, data|
+              @machine.communicate.sudo("salt-call state.highstate --retcode-passthrough#{get_masterless}#{get_loglevel}#{get_colorize}#{get_pillar}#{get_call_args}", ssh_opts) do |type, data|
```
chrisroberts: Can you break these long lines like:

```ruby
@machine.communicate.sudo("salt-call state.highstate --retcode-passthrough#{get_masterless}" \
  "#{get_loglevel}#{get_colorize}#{get_pillar}#{get_call_args}", ssh_opts) do |type, data|
```

PR #8945, opened 2017-09-05
Remove duplicate export folders before writing /etc/exports

### LINKED ISSUE #4666: Double entries in /etc/exports
Sometimes Vagrant puts doubles entries in `/etc/exports`, causing nfsd to fail, for example:

```
$ cat /etc/exports
# VAGRANT-BEGIN: 1001 9e5c7f10-0b0e-4713-a87c-0638a78bdf62
"/home/martin/code/cust/dewoningbeheerder" 192.168.33.10(rw,no_subtree_check,all_squash,anonuid=1001,anongid=1001,fsid=2556391661)
# VAGRANT-END: 1001 9e5c7f10-0b0e-4713-a87c-0638a78bdf62
# VAGRANT-BEGIN: 1001 9e5c7f10-0b0e-4713-a87c-0638a78bdf62
"/home/martin/code/cust/dewoningbeheerder" 192.168.33.10(rw,no_subtree_check,all_squash,anonuid=1001,anongid=1001,fsid=2556391661)
# VAGRANT-END: 1001 9e5c7f10-0b0e-4713-a87c-0638a78bdf62
```

I have this maybe once every two weeks, for all projects (using Vagrant daily).

I've been getting in the habit of using `vagrant up --debug`, but for some reason the problem only seems to occur if I leave the `--debug` flag out (probably coincidence, since sometimes I forget), so I can't point you to the cause of these problems...

I'm using Vagrant 1.6.5 on Arch Linux. I also had this problem on Ubuntu 12.04, and has been around since at least March (since I've been using Vagrant).

briancain:
Prior to this commit, if you set up multiple folders to export with NFS
on linux with the exact same hostpath, the template used to write
/etc/exports would end up placing the same path with the same IP in
/etc/exports and cause an error preventing the folders from being
properly mounted. This commit fixes that by first looking at which
folders are being exported and if there are any duplicates. If so,
remove the duplicates and only export 1 hostpath folder. If these
duplicate folders have differing nfs linux options, an exception must be
thrown because we cannot assume which options the user intended to
export with.

Fixes #4666

### DIFF
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -29,6 +29,7 @@ module VagrantPlugins
           nfs_start_command = env.host.capability(:nfs_start_command)
 
           nfs_opts_setup(folders)
+          folders = folder_dupe_check(folders)
           output = Vagrant::Util::TemplateRenderer.render('nfs/exports_linux',
                                            uuid: id,
                                            ips: ips,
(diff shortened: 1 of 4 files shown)


plugins/hosts/linux/cap/nfs.rb
```diff
@@ -84,6 +85,38 @@ def self.nfs_prune(environment, ui, valid_ids)
...
+
+          # We need to check that each group key only has 1 value,
+          # and if not, check each nfs option. If all nfs options are the same
+          # we're good, otherwise throw an exception
+          export_groups.each do |path,group|
+            if group.size > 1
+              # if the linux nfs options aren't all the same throw an exception
+              group1_opts = group.first[:linux__nfs_options]
+              if group.map { |g|
+                g[:linux__nfs_options] == group1_opts }.include? false
```
chrisroberts: I think using `Array#all?` will make the conditional a little bit clearer:

```ruby
if !group.all?{|g| g[:linux__nfs_options] == group1_opts}
```

guysoft:
Hey people,
I am still getting this bug on version 2.0.3. Attaching log.
I confirm that each time I remove the duplicate line. Do ``vagrant destroy ; vagrant up`` it comes back again. Full report at https://github.com/hashicorp/vagrant/issues/9681 
[vagrant.log](https://github.com/hashicorp/vagrant/files/1935031/vagrant.log)

PR #8952, opened 2017-09-06
Update smb.html.md

TonyApuzzo:
Document a PowerShell one-liner to clean up all the existing SMB shares.

### DIFF
--- a/website/source/docs/synced-folders/smb.html.md
+++ b/website/source/docs/synced-folders/smb.html.md
@@ -92,8 +92,7 @@ elevated PowerShell. Note that you will lose access to the shared folders in
 any running VMs, so halt all Vagrant VMs before running this.
 
 ```PowerShell
-PS C:\> Get-WmiObject -Class Win32_Share | Where-Object { $_.Name -match "[0-9a-f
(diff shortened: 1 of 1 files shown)


briancain:
Hey there @TonyApuzzo - Were you encountering Vagrant leaving these shares behind when using smb? If so, that's likely a vagrant bug and we'd be happier to fix that instead of documenting this fix. Thanks!

TonyApuzzo:
Hi, yes, Vagrant does have that limitation, see the paragraph above my diff in the file:

> The primary limitation of SMB synced folders at the moment is that they are never pruned or cleaned up. Once the folder share is defined, Vagrant never removes it. To clean up SMB synced folder shares, periodically run net share in a command prompt, find the shares you do not want, then run net share NAME /delete for each, where NAME is the name of the share.

TonyApuzzo:
(I always seem to hit the "close" button accidentally...)

briancain:
Right, if that's the case I would rather fix that in Vagrant instead of documenting a work around. I recommend opening an issue with a Vagrantfile and we can go ahead and triage it and get it fixed. Thanks!

PR #8982, opened 2017-09-18
Mark the become features to be version 2.0

JerryVerhoef:
The become and become_user are introduced in version 2.0

### DIFF
--- a/website/source/docs/provisioning/ansible_common.html.md
+++ b/website/source/docs/provisioning/ansible_common.html.md
@@ -20,10 +20,18 @@ Some of these options are for advanced usage only and should not be used unless 
 - `become` (boolean) - Perform all the Ansible playbook tasks [as another user](http://docs.ansible.com/ansible/become.html), different from the user used to log into the guest system.
 
     The default value is `false`.
+    <div class="alert alert-info">
+      <strong>Compatibility Note:</strong>
+      This option was introduced in Vagrant 2.0. 
+    </div>
 
 - `become_user` (string) - Set the default username to be used by the Ansible `become` [privilege escalation](http://docs.ansible.com/ansible/become.html) mechanism.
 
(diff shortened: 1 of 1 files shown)


gildegoma:
@JerryVerhoef Thank you for this pull request. Indeed, the Vagrant documentation site corresponds to the latest release, and unfortunately there is no general format to indicate *for each option* in which Vagrant version it has been introduced (unlike in Ansible docs, for isntance). As this is not the case, so far I only marked this compatibility note for very special cases (e.g. the `compatibility_mode` in 2.0, or the [`force_remote_user`](https://www.vagrantup.com/docs/provisioning/ansible.html#force_remote_user) in 1.8) where vagrant _new behaviour_ may break things. But we didn't for the moment for backward compatible **additions** to the provisioner.

I understand your request and I like the idea to improve this aspect, but in the current state of the documentation I am not sure if it is a good idea to partially introduce such this kind of patches... But maybe as of Vagrant 2.x we could revisit this. BTW: If we opt to go forward with this proposal, it would be great to add the same notice for the [`ask_become_pass`](https://www.vagrantup.com/docs/provisioning/ansible.html#ask_become_pass) 😉

@chrisroberts Do you have any plan to refactor the way to handle this in the future? I mean for all the Vagrant changes, not only the (Ansible) provisioners.

chrisroberts:
@gildegoma There has been conversations about versioned documentation but nothing concrete has been decided upon yet. Documentation now always refers to the latest released version of Vagrant.

chrisroberts:
Thank you for this PR. I am going to close this now. Currently the Vagrant documentation is always for the currently released version of Vagrant. At some point in the future we hope to have versioned documentation but until that is actualized we will simply continue with the documentation being targeted at the currently released version.

Cheers!

PR #8992, opened 2017-09-23
provisioners/salt: Fixes timeout issue in salt bootstrapping for windows

### LINKED ISSUE #8991: provisioners/salt: Windows bootstrap unable to download salt binary
### Vagrant version
Vagrant 2.0.0 & 1.9.8

### Host operating system
macOs High Sierra

### Guest operating system
Windows 2008R2 X64

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

  net_ip = "192.168.50"
  lin_os = "bento/ubuntu-16.10"
  win_os = "ferventcoder/win2008r2-x64-nocm"

  # Salt Master
  config.vm.define "salt" do |server|
      server.vm.provider "virtualbox" do |vb|
          vb.memory = "2048"
          vb.cpus = 1
          vb.name = "salt"
      end
    server.vm.box = "#{lin_os}"
    server.vm.hostname = 'saltmaster.local'
    server.vm.network "private_network", ip: "#{net_ip}.10"
    server.vm.synced_folder "salt/roots/", "/srv/salt/", create: true
    
    server.vm.provision :salt do |salt|
      salt.install_master = true
      salt.no_minion = true
      salt.install_type = "stable"
    end
  end

  # WithMaster Windows Minion
  config.vm.define "windows" do |server|
    server.vm.provider "virtualbox" do |vb|
      vb.memory = "1024"
      vb.cpus = 1
      vb.name = "windows"
  end
    server.vm.box = "#{win_os}"
    server.vm.communicator = "winrm"
    server.vm.hostname = "windows"
    server.vm.network "private_network", ip: "#{net_ip}.11"
    server.vm.provision :salt do |salt|
      salt.minion_config = "salt/configs/win_minion"
      salt.run_highstate = true
      salt.minion_id = "windows"
    end
  end
end
```

### Debug output
To be able to provide the full debug output I would first have to understand what it contains. I cannot scan the full file for sensitive information right now.

### Expected behavior
Vagrant successfully provisions the salt environment.

### Actual behavior
Vagrant executes the bootstrap_salt.ps1 file. The file is unable to download the salt minion. The bootstrap_salt.ps1 file never executes. Vagrant never exits. The issue is swallowed, because there is no local debug output. 

### Steps to reproduce
1. Provide some fake salt config for the Vagrantfile above
2. Execute "vagrant up"

### Root Cause
This is caused by a network issue. When the bootstrapper attempts to download the file it received a 408: Request Timeout. When manually trying to execute the bootstrap later on it works. I can imagine that the bootstrap is executed when the network is not ready, yet. I will provide an approach to fix this in a little while.

### References
The issue seems to be related to #8925. Without the fix provided in #8925 newer minion version will not download. However, this issue was also swallowed for me and the vagrant call never returned

BenjaminSchiborr:
Fixes #8991.

1. When remote executing scripts one should always call powershell with
1a. "-NonInteractive", in order to prevent interactive prompts from leading to an endless waiting time for the script to return
1b. "-NoProfile", in order to prevent the loading of unknown custom profiles before execution of the script which could have unintended side effects
2. During my tests I constantly ran into 408 timeouts when downloading the salt binaries. I've prevented that by adding a simple retry mechanism and an error exit in case of multiple failures. Without this change the bootstrap script never returned (and neither did vagrant up)

### DIFF
--- a/plugins/provisioners/salt/bootstrap-salt.ps1
+++ b/plugins/provisioners/salt/bootstrap-salt.ps1
@@ -53,7 +53,27 @@ Write-Host "Downloading Salt minion installer Salt-Minion-$version-$arch-Setup.e
 $webclient = New-Object System.Net.WebClient
 $url = "https://repo.saltstack.com/windows/Salt-Minion-$version-$arch-Setup.exe"
 $file = "C:\tmp\salt.exe"
-$webclient.DownloadFile($url, $file)
+
+[int]$retries = 0
+Do {
+ try {
+    $retries++
+    $ErrorActionPreference='Stop'
+    $webclient.DownloadFile($url, $file)
+ } catch [Exception] {
+    if($retries -eq 3) {
+        $_
+        $_.GetType()
+        $_.Exception
+        $_.Exception.StackTrace
+        Write-Host
+        exit 1
+    }
+    Write-Warning "Retrying download in 2 seconds. Retry # $retries"
+    Start-Sleep -s 2
+    }
+}
+Until($retries -eq 3)
 
 
 # Install minion silently
--- a/plugins/provisioners/salt/bootstrap-salt.ps1
+++ b/plugins/provisioners/salt/bootstrap-salt.ps1
@@ -60,8 +60,9 @@ Do {
     $retries++
     $ErrorActionPreference='Stop'
     $webclient.DownloadFile($url, $file)
+    break
  } catch [Exception] {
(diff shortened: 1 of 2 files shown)


briancain:
Thanks for the PR @BenjaminSchiborr ! I've put this on the 2.0.2 milestone and assigned myself to give you a review soon.

PR #8997, opened 2017-09-25
Adding check for APIPA in winrm helper to fix #8996

### LINKED ISSUE #8996: Vagrant attempts to use APIPA for WinRM
When Windows cannot get a good IP it will create an APIPA (see https://en.wikipedia.org/wiki/Link-local_address#IPv4) for itself. This can happen due to networking issues, but can also just due to slowness from the DHCP server. 

When it's a networking issue like Windows can't get a good IP then vagrant will grab the APIPA from Hyper-V (I've only tested this on Hyper-V) and try to WinRM using an IP that will never work. It eventually fails which would be expected behavior, but it's a little awkward to vagrant spins for a while trying to WinRM with an IP that is clearly bad.

Sometimes it just so happens that vagrant asks for the IP address before Windows has a good one, and it'll get the APIPA from Hyper-V even though moments later a good IP is assigned to Windows. This is a more concerning edge case because technically there was nothing wrong with the server. Vagrant just happened to ask for the IP before Windows was ready.

### Vagrant version
2.0.0

Pretty sure this issue is in every version of vagrant, but I've seen it as early as 1.9.2 for sure.

### Host operating system
Windows 10

### Guest operating system
Windows Server 2012 R2

### Vagrantfile

Default vagrantfile

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# you're doing.
  # The most common configuration options are documented and commented below.
  # For a complete reference, please see the online documentation at
  # https://docs.vagrantup.com.

  # Every Vagrant development environment requires a box. You can search for
  # boxes at https://vagrantcloud.com/search.
  config.vm.box = "win2012r2-template-030"

  # Disable automatic box update checking. If you disable this, then
  # boxes will only be checked for updates when the user runs
  # `vagrant box outdated`. This is not recommended.
  # config.vm.box_check_update = false

  # Create a forwarded port mapping which allows access to a specific port
  # within the machine from a port on the host machine. In the example below,
  # accessing "localhost:8080" will access port 80 on the guest machine.
  # NOTE: This will enable public access to the opened port
  # config.vm.network "forwarded_port", guest: 80, host: 8080

  # Create a forwarded port mapping which allows access to a specific port
  # within the machine from a port on the host machine and only allow access
  # via 127.0.0.1 to disable public access
  # config.vm.network "forwarded_port", guest: 80, host: 8080, host_ip: "127.0.0.1"

  # Create a private network, which allows host-only access to the machine
  # using a specific IP.
  # config.vm.network "private_network", ip: "192.168.33.10"

  # Create a public network, which generally matched to bridged network.
  # Bridged networks make the machine appear as another physical device on
  # your network.
  # config.vm.network "public_network"

  # Share an additional folder to the guest VM. The first argument is
  # the path on the host to the actual folder. The second argument is
  # the path on the guest to mount the folder. And the optional third
  # argument is a set of non-required options.
  # config.vm.synced_folder "../data", "/vagrant_data"

  # Provider-specific configuration so you can fine-tune various
  # backing providers for Vagrant. These expose provider-specific options.
  # Example for VirtualBox:
  #
  # config.vm.provider "virtualbox" do |vb|
  #   # Display the VirtualBox GUI when booting the machine
  #   vb.gui = true
  #
  #   # Customize the amount of memory on the VM:
  #   vb.memory = "1024"
  # end
  #
  # View the documentation for the provider you are using for more
  # information on available options.

  # Enable provisioning with a shell script. Additional provisioners such as
  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
  # documentation for more information about their specific syntax and use.
  # config.vm.provision "shell", inline: <<-SHELL
  #   apt-get update
  #   apt-get install -y apache2
  # SHELL
end
```

### Debug output
Here's a gist of vagrant failing to WinRM when there is a clear networking issue. Vagrant tries to WinRM using the APIPA until it times out. I unfortunately don't have a gist of when there's networking slowness since it's sporadic, but the behavior is the same on the vagrant side.

https://gist.github.com/bshurts/634e3ae5d8fe24f209bb3e64206d2d16

### Expected behavior
Vagrant throws out the APIPA (`169.254.*.*`) and tries again for a valid IP

### Actual behavior
Vagrant accepts the APIPA and proceeds to attempt to WinRM with an unusable IP address

### Steps to reproduce
1. Create a vanilla internal network switch in Hyper-V
2. Vagrant up a Windows box in Hyper-V using the switch

bshurts:
See https://github.com/hashicorp/vagrant/issues/8996

### DIFF
--- a/plugins/communicators/winrm/helper.rb
+++ b/plugins/communicators/winrm/helper.rb
@@ -33,8 +33,13 @@ module VagrantPlugins
         return addr if addr
 
         ssh_info = machine.ssh_info
-        raise Errors::WinRMNotReady if !ssh_info || ssh_info[:host].to_s.empty?
+        raise Errors::WinRMNotReady if winrm_info_invalid?(ssh_info)
         return ssh_info[:host]
+      end
+
+      def self.winrm_info_invalid?(ssh_info)
+        invalid = (!ssh_info || ssh_info[:host].to_s.empty? || ssh_info[:host].to_s.match(/^169.254/))
+        return invalid
       end
 
       # Returns the port to access WinRM.
--- a/plugins/communicators/winrm/helper.rb
+++ b/plugins/communicators/winrm/helper.rb
@@ -38,7 +38,7 @@ module VagrantPlugins
       end
 
       def self.winrm_info_invalid?(ssh_info)
-        invalid = (!ssh_info || ssh_info[:host].to_s.empty? || ssh_info[:host].to_s.match(/^169.254/))
+        invalid = (!ssh_info || ssh_info[:host].to_s.empty? || ssh_info[:host].start_with?("169.254"))
         return invalid
       end
 
--- a/test/unit/plugins/communicators/winrm/helper_test.rb
+++ b/test/unit/plugins/communicators/winrm/helper_test.rb
@@ -48,6 +48,12 @@ describe VagrantPlugins::CommunicatorWinRM::Helper do
       expect { subject.winrm_address(machine) }.
         to raise_error(VagrantPlugins::CommunicatorWinRM::Errors::WinRMNotReady)
     end
+
+    it "raise an exception if it detects an APIPA" do
+      machine.stub(ssh_info: { host: "169.254.123.123" })
+      expect { subject.winrm_address(machine) }.
+        to raise_error(VagrantPlugins::CommunicatorWinRM::Errors::WinRMNotReady)
+    end
   end
 
   describe ".winrm_info" do
--- a/test/unit/plugins/communicators/winrm/helper_test.rb
+++ b/test/unit/plugins/communicators/winrm/helper_test.rb
@@ -148,4 +148,31 @@ describe VagrantPlugins::CommunicatorWinRM::Helper do
       expect(subject.winrm_port(machine)).to eq(2456)
     end
   end
+
+  describe ".winrm_info_invalid?" do
+    it "returns true if it can't detect a host" do
+      allow(machine).to receive(:ssh_info).and_return(nil)
+      expect(subject).to be_winrm_info_invalid(machine.ssh_info)
+    end
+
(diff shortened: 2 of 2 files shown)


bshurts:
It seems like the CI build just timed out. I don't see any tests failing. I don't think I have permissions to kick off a new build.

briancain:
@bshurts - yeah...that cell has a sticky test that seems to timeout every now and then. I kicked it and it seems to be passing now!

plugins/communicators/winrm/helper.rb
```diff
@@ -33,10 +33,15 @@ def self.winrm_address(machine)
...
 
         ssh_info = machine.ssh_info
-        raise Errors::WinRMNotReady if !ssh_info || ssh_info[:host].to_s.empty?
+        raise Errors::WinRMNotReady if winrm_info_invalid?(ssh_info)
         return ssh_info[:host]
       end
 
+      def self.winrm_info_invalid?(ssh_info)
+        invalid = (!ssh_info || ssh_info[:host].to_s.empty? || ssh_info[:host].to_s.match(/^169.254/))
+        return invalid
```
chrisroberts: can you update this to use `#start_with?` in place of the `#match`:

```ruby
!ssh_info || ssh_info[:host].to_s.empty? || ssh_info[:host].to_s.start_with?("169.254")
```

bshurts:
Oh right sorry. I added tests on the calling method, but not the new method. I'll add those.

bshurts:
Looks like another hung build.

briancain:
@bshurts - yep...sorry. We've got a ticket in our backlog for trying to solve this. For now I'll restart the cell! Thanks.

bshurts:
Is there anything else required for this PR? I believe I made all of the requested changes, but let me know if I need to do anything else.

Issue #9011, opened 2017-09-30
winssh communicator fails to configure guest network


benh57:
### Vagrant version
2.0.0

### Host operating system
Windows 7

### Guest operating system
Windows 10

### Vagrantfile

Setup is complicated as it uses windows requires setting up winssh on the guest vm first. Important parts:
```ruby
    config.vm.define "win10" do |win|
      win.vm.guest = 'windows'
      win.vm.network :forwarded_port, guest: 22, host: 2022, id: "winssh", auto_correct: true
      win.vm.communicator = 'winssh'
      win.ssh.private_key_path = 'vagrantkeys/id_rsa'
      win.vm.network "public_network"
end
```

### Debug Output

https://gist.github.com/benh57/5f086e231aa66b4d0aa84a7264faa73d

### Expected behavior
No errors.

### Actual behavior
winssh communicator fails to configure network with error message:
```
Stderr from the command:

.productversion.split(" " was unexpected at this time.

The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

((test-wsman).productversion.split(" ") | select -last 1).split("\.")[0]
```
### Steps to reproduce
1. Set up a basebox with Windows and Openssh authentication properly working. 
2. Attempt to use 'winssh' communicator with proper keys  and run 'vagrant up <box>'

### References

This was broken by https://github.com/hashicorp/vagrant/commit/728ec28f2d0434f446a1f662cd5765afe84ae95d#diff-25815bcd4f0802b6d2456af9be8e88fb

I have a fix for this and will submit PR shortly.

PR #9012, opened 2017-09-30
Fix guest network setup with winssh communicator

benh57:
winssh communicator is not functional currently as it fails to configure the guest network.

Fixes 
https://github.com/hashicorp/vagrant/issues/9011


benh57:
After rebuilding from scratch i got an error that DHCP was already configured from Powershell on bringing up my win10 VM.  But, i think that might be a Windows 10 thing, in any case i can submit a separate PR for it.

PR #9026, opened 2017-10-03
Hyper V Provider, handle multiple attached VHDs for XML configs

simonp22:
Was creating a Hyper V box where I needed to attach multiple VHDs.

These code changes copy all VHDs from the "Virtual Hard Disks" folder in the box, then loops through the drives in the XML config for the box and adds them in the correct order


bbrala:
@simonp22 i actually think that if you you a gen 2 virtual machine, it would already work, since it really clones the original and makes some adjustments instead of building one from an XML file manually.

briancain:
Hey there everyone! I'm going to go ahead and close this issue, as I have just made a pull request that takes advantage of the recent generic disk config options in Vagrant. You can find that PR here: #11541 It offers a lot of extra goodies such as resizing disks (similar to the VirtualBox implementation). Thanks!

Issue #8877, opened 2017-08-10
HyperV Powers all VMs previously running VMs on startup


wiltaylor:
### Vagrant version
1.9.7

### Host operating system
Windows 10 x64 Enterprise 1703

### Guest operating system
Any (Windows 10 x64 Enterprise 1703 in this case though)

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :
Vagrant.configure("2") do |config|
  config.vm.box = "Windows10x64"
  config.vm.network "public_network", :bridge => "NAT"

  config.vm.provider "hyperv" do |h|
    h.enable_virtualization_extensions = true
    h.ip_address_timeout = 600
    h.memory = 4096
    h.vm_integration_services = {
        guest_service_interface: true,
        heartbeat: true,
        key_value_pair_exchange: true,
        shutdown: true,
        time_synchronization: true,
        vss: true
    }

    config.vm.synced_folder ".", "/vagrant", type: "smb", smb_username: ENV["smb_username"], smb_password: ENV["smb_password"]
  end
end
```

### Expected behavior
If you reboot your host OS Virtual Machines stay powered  off till Vagrant up is called.

### Actual behavior
Virtual Machine boots when host system is booted.

### Steps to reproduce
1. Vagrant up to create vm.
2. Restart computer
3. Open HyperV console and notice VM is now running again.

Solution will be to set VM automatic startup behaviour to nothing instead of the default Automatically start if it was running when the service stopped.

bbrala:
You can configure `h.auto_start_action` and `h.auto_stop_action` in you configuration if you like. Although currently there is a bug in setting `auto_stop_action` on newer VMs (vmcx files. see PR #9000).

Weird thing is though it is currently missing from the documentation, but ill fix this right away.

bbrala:
@chrisroberts this can be closed i think.

chrisroberts:
This is resolved via #9872 

Cheers!

PR #9029, opened 2017-10-04
Add auto_start_action and auto_stop_action to docs.

bbrala:
As noted in #8877, auto start and stop action is missing from the docs. 

Please merge #9000 also though, since it is currently bugged for binary HyperV VM files.


bbrala:
Thanks :)

PR #9048, opened 2017-10-10
Command/rm

claymodel:
"vagrant box rm" added for "vagrant box remove"

Hence, the both command will be useful.


briancain:
Hey there @claymodel - thank you for opening a pull request with this enhancement! At the moment however, I think this would be better suited for a local alias rather than a fully supported command.

Thanks!

PR #9112, opened 2017-10-26
Attempt fetching salt bootstrap script with wget

boltronics:
On Debian installations, `fetch`, `curl` and `python` commands may not exist. Only `wget` is guaranteed as explained in the commit message.

This fixes the following error on minimal Debian images, and possibly other distributions and operating systems.
```
==> default: Running provisioner: salt...
Copying salt minion config to vm.
Checking if salt-minion is installed
salt-minion was not found.
Checking if salt-call is installed
salt-call was not found.
Using Bootstrap Options:  -X -F -c /tmp stable
Bootstrapping Salt... (this may take a while)
stdin: is not a tty

/tmp/bootstrap_salt.sh: 9: /tmp/bootstrap_salt.sh:
python: not found

Salt successfully configured and installed!
run_overstate set to false. Not running state.overstate.
Calling state.highstate... (this may take a while)
stdin: is not a tty

bash: line 4: salt-call: command not found

salt-call state.highstate --retcode-passthrough  --local --log-level=warning --force-color

Stdout from the command:



Stderr from the command:

stdin: is not a tty
bash: line 4: salt-call: command not found
$
```


boltronics:
That was quick! Thanks @briancain!

Issue #9114, opened 2017-10-26
Enhancement: No way to specifically set ANSIBLE_ROLES_PATH (separate from galaxy roles path)


james-portman:
### Vagrant version
Current/all

### Host operating system
All

### Guest operating system
All

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "ubuntu/xenial64"
  config.vm.provision "ansible" do |ansible|
    ansible.playbook = "test.yml"
    ansible.galaxy_role_file = "requirements.yml"
    ansible.galaxy_roles_path = "../"
  end
end
```

### Expected behavior
It would be great to have a separate parameter `ansible.roles_path` which was passed to the `ansible-playbook` command as `ANSIBLE_ROLES_PATH`.

### Actual behavior
Currently the single parameter `ansible.galaxy_roles_path` is used for both the `ansible-galaxy` and `ansible-playbook` commands.

james-portman:
The current behaviour is related to/caused by this change:
https://github.com/hashicorp/vagrant/pull/7420/files

I don't know why it was agreed to use a variable specifically for galaxy/ansible-galaxy for the ansible-playbook command

james-portman:
Potential fix:
https://github.com/hashicorp/vagrant/pull/9115

gildegoma:
Most Ansible projects (with or without Vagrant in the game), will store their roles in a `roles/` directory [relative to the playbook file](https://docs.ansible.com/ansible/latest/playbooks_reuse_roles.html#role-search-path). The galaxy roles are often installed in a distinct directory, and the addition #7420 allows this.

If for some reason, you want to use a different roles path, this is easily configured with a `roles_path` entry in an [Ansible configuration file](https://docs.ansible.com/ansible/latest/intro_configuration.html#roles-path) (e.g. a local `ansible.cfg`).

james-portman:
I will test again, but from what I can tell, using roles_path within ansible.cfg also affects ansible-galaxy command.
It is still not logical for you to call the parameter galaxy_roles_path but apply it to both the ansible-galaxy and also the ansible-playbook command.

gildegoma:
> It is still not logical for you to call the parameter galaxy_roles_path but apply it to both the ansible-galaxy and also the ansible-playbook command.

[...]

PR #9115, opened 2017-10-26
Adding Ansible playbook_roles_path setting

james-portman:
To be used only for when running ansible-playbook, not ansible-galaxy
Ref https://github.com/hashicorp/vagrant/issues/9114


gildegoma:
See discussion in #9114 for the moment. So far, I think it is not worthy to add this option.

gildegoma:
See arguments in #9114.

PR #9129, opened 2017-10-31
Restarting network (via systemctl):  [FAILED]

iurquiza:
**Fix the following error:**

`Stdout from the command:

/bin/hostnamectl
Restarting network (via systemctl):  [FAILED]

Stderr from the command:

Job for network.service failed. See 'systemctl status network.service' and 'journalctl -xn' for details.`


iurquiza:
Fixed here https://github.com/hashicorp/vagrant/commit/ff021fcab404c95e52566bfca4207da9c0101e01#diff-fd473b135e71706459b47b51cb86e249

PR #9131, opened 2017-11-01
Fix --output path with specified folder

### LINKED ISSUE #9088: vagrant package silently fails when --output path contains a slash
### Vagrant version

```console
$ vagrant -v
Vagrant 2.0.0
```

### Host operating system

macOS Sierra

### Guest operating system

erickoegel/dfly48

### Vagrantfile
```ruby
  config.vm.box = "erickoegel/dfly48"
  config.vm.provision "shell", path: "bootstrap.sh"
end
```

### bootstrap.sh

```sh
#!/bin/sh
sudo pkg update && \
    sudo pkg install -y rsync
```

### Expected behavior

Ideally, Vagrant checks to see if a `mcandre` directory exists. If not, Vagrant creates the directory and writes a `dragonfly-rsync.box` file inside the directory.

Or, Vagrant exits with a non-zero status and warns the user that the path would be invalid.

### Actual behavior

Vagrant takes several minutes to produce a box, then I have no idea where the box ends up, because no *.box files are produced.

### Steps to reproduce
1. `vagrant up`
2. `vagrant package --output mcandre/dragonfly-rsync.box`

As a workaround, I am careful to `--output` to `dragonfly-rsync.box`, and then import the box as `mcandre/dragonfly-rsync` to my vagrant box list.

bpietraga:
This pull request is in to resolve ticket

Fixes #9088

### DIFF
--- a/lib/vagrant/action/general/package.rb
+++ b/lib/vagrant/action/general/package.rb
@@ -73,6 +73,8 @@ module Vagrant
 
           self.class.validate!(env["package.output"], env["package.directory"])
 
+          package_with_folder_path if env["package.output"].include?(File::SEPARATOR)
+
           raise Errors::PackageOutputDirectory if File.directory?(fullpath)
 
           @app.call(env)
@@ -81,6 +83,16 @@ module Vagrant
           copy_include_files
           setup_private_key
           compress
+        end
+
+        def package_with_folder_path
+          folder_path = File.expand_path("..", @fullpath)
+          create_box_folder(folder_path) unless File.directory?(folder_path)
+        end
+
+        def create_box_folder(folder_path)
+          @env[:ui].info(I18n.t("vagrant.actions.general.package.box_folder", folder_path: folder_path))
+          FileUtils.mkdir_p(folder_path)
         end
 
         def recover(env)
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -2114,6 +2114,7 @@ en:
         package:
           packaging: "Packaging additional file: %{file}"
           compressing: "Compressing package to: %{fullpath}"
+          box_folder: "Creating new folder: %{folder_path}"
           output_exists: |-
             The specified file '%{filename}' to save the package as already exists. Please
             remove this file or specify a different file name for outputting.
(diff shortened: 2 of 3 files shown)


lib/vagrant/action/general/package.rb
```diff
@@ -72,6 +72,7 @@ def call(env)
           @env = env
 
           self.class.validate!(env["package.output"], env["package.directory"])
+          package_with_folder_path(fullpath) if output_inside_folder?(env)
```
briancain: Is this variable different from `@fullpath`? And if not...I think it can be accessed within that function without being passed in as a parameter.

lib/vagrant/action/general/package.rb
```diff
@@ -83,6 +84,20 @@ def call(env)
           compress
         end
 
+        def output_inside_folder?(env)
+          env["package.output"].include?('/')
```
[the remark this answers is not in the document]
briancain: Also, can we just use `@fullpath` here, I think? If we can just use the variable, I don't know if it needs to be a function.

lib/vagrant/action/general/package.rb
```diff
@@ -83,6 +84,20 @@ def call(env)
...
+        end
+
+        def package_with_folder_path(fullpath)
+          folder_path = File.expand_path('../', fullpath)
```
briancain: I think this can just be `File.expand_path('..', fullpath)` so that it works on windows.

lib/vagrant/action/general/package.rb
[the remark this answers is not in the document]
bpietraga: `env["package.output"]` returns `"mcandre/dragonfly-rsync.box"`
`@fullpath`returns `"/Users/bpietraga/mcandre/dragonfly-rsync.box"`

lib/vagrant/action/general/package.rb
bpietraga: Thanks! :) I agree. A bit over 80 lines way to go then:

```ruby
package_with_folder_path if env["package.output"].include?(File::SEPARATOR)
```

bpietraga:
@briancain Thank you for extensive code review. 🙂 
I submitted changes to the commit addressing things you found and wrote test for this functionality.
Also I've removed the WIP from commit name and PR.

Issue #9154, opened 2017-11-08
2.0.1 break nfs detection on debian Jessie


stevenh:
After updating to vagrant 2.0.1 all our VM's fail to start, complaining about "machine doesn't support NFS"

After debugging the issue the problem was caused by: https://github.com/hashicorp/vagrant/pull/8938

The issue is two fold:
1. The service is nfs-kernel-server and not nfs-server
2. The service doesn't include a unit file so using list-unit-files fails to find the service, list-units is what's needed

### Vagrant version
Vagrant 2.0.1

### Host operating system
Debian Jessie (8.9)

### Guest operating system
ubuntu

### Debug output
 INFO subprocess: Starting process: ["/bin/sh", "-c", "systemctl --no-pager --no-legend --plain list-unit-files --all --type=service | grep nfs-server.service"]
 INFO subprocess: Command not in installer, restoring original environment...

### Expected behavior
vagrant up succeeds

### Actual behavior
It appears your machine doesn't support NFS, or there is not an
adapter to enable NFS on this machine for Vagrant. Please verify
that `nfsd` is installed on your machine, and try again. If you're
on Windows, NFS isn't supported. If the problem persists, please
contact Vagrant support.

### Steps to reproduce
1. vagrant up for a vm which mounts an NFS volume

### References
The following PR caused this breakage: https://github.com/hashicorp/vagrant/pull/8938

chrisroberts:
Fixed via #9274

PR #9155, opened 2017-11-08
Fixed nfs detection on debian Jessie

stevenh:
The recent change to use systemctl to detect and control nfs broke compatibility with debian Jessie due to the use of the wrong service name and systemctl list command.

This change reverts the name change and switches to list-units from list-unit-files as there is no unit file for the service.

The issue was introduced by:
https://github.com/hashicorp/vagrant/pull/8938

This fixes:
https://github.com/hashicorp/vagrant/issues/9154

### DIFF
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -17,7 +17,7 @@ module VagrantPlugins
 
         def self.nfs_check_command(env)
           if Vagrant::Util::Platform.systemd?
-            "systemctl status --no-pager nfs-server.service"
+            "systemctl status --no-pager nfs-kernel-server.service"
           else
             "/etc/init.d/nfs-kernel-server status"
           end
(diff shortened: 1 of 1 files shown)


stevenh:
@chrisroberts could you confirm this works on your test environment as there's clearly some differences between our two test systems.

stevenh:
Reading around there's almost certainly more work that needs to be done on this as it looks like different flavours of Linux use different names, so one method simply isn't going to cut it.

mcandre:
> different flavours of Linux use different names

yup! cap edits should be made on the specific guests (e.g. debian) rather than the top-level linux guest.

stevenh:
So should we start be reverting the problem commit or would it be better to have a Linux level default service name or names which can then be overridden be different flavours.
Duplicating the code for each flavour seems like a poor option.

chrisroberts:
@stevenh Hi and thanks for the PR and report. I have a new PR that targets the naming problem that was previously introduced. It does some simple detection for the nfs service name, falling back on a default. The names are provided via methods, so guests can provide just the name only: #9274 

Going to close this PR up and move forward with #9274. Again, thanks for this PR and the report!

PR #9160, opened 2017-11-09
integrate required ruby version with rvm and friends

mcandre:
Make it easier for developers to switch between Vagrant and other Ruby projects, by documenting the required version of Ruby in a .ruby-version file, that plays well with chruby, IDE's, and other development tools.


chrisroberts:
Supported versions of ruby are defined within the gem specification file. If you would like to enable this locally you can simply keep the file in your local repository copy without interference (as it is ignored by git anyway). Cheers!

PR #9162, opened 2017-11-09
Suggest being able to be a machine name from the file name too.

k-oguma:
## TL;DR

- I want to be able to easily specify machine name etc. by loose argument.

## Description

- I'm a suggest to increase the degree of freedom so that the name of the machine can be obtained from the file name too.
- For instance, when reusing ansible playbook with one Vagrantfile, it is possible to specify it by tab completion (eg, bash_completion), so I think it will be easier to handle.

* Thing I want to do
     * I want to increase freedom of machine name and other specification by cmd argument
     * I want to make it executable by file specification
     * I want to be able to execute even Path specification
     * If the above is OK, designation becomes possible by tab completion, it will be easier
        - The expectation for higher reusability
* Do not do it.
     * Specify as an environment variable
        - It does not match the current specification of Vagrant, it becomes hard to handle because its method of usage greatly changes with other Vagrantfiles

#### Before

```
vagrant up foo
```
- playbooks/foo.yml execute.

```
vagrant up foo.yml
vagrant up playbooks/foo.yml
vagrant up playbooks/foo
```
- playbooks/foo.yml not execute.
   - The error is following
 　- `The machine with the name 'foo.yml' was not found configured for this Vagrant environment.`

#### After

- All succeeds.
- This commit is to solve this problem.

#### Details

- To Github comment
   - inb4 tl;dr

### DIFF
--- a/lib/vagrant/vagrantfile.rb
+++ b/lib/vagrant/vagrantfile.rb
@@ -42,6 +42,7 @@ module Vagrant
     # @return [Machine]
     def machine(name, provider, boxes, data_path, env)
       # Load the configuration for the machine
+      name = File.basename(name.to_s).delete(File.extname(name.to_s)).to_sym # Become a vmname if name from file path.
       results = machine_config(name, provider, boxes)
       box             = results[:box]
       config          = results[:config]
@@ -111,6 +112,7 @@ module Vagrant
     #   machine. See the main documentation body for more info.
     def machine_config(name, provider, boxes)
       keys = @keys.dup
+      name = File.basename(name.to_s).delete(File.extname(name.to_s)).to_sym # Become a vmname if name from file path.
 
       sub_machine = @config.vm.defined_vms[name]
       if !sub_machine
--- a/lib/vagrant/vagrantfile.rb
+++ b/lib/vagrant/vagrantfile.rb
@@ -42,7 +42,7 @@ module Vagrant
     # @return [Machine]
     def machine(name, provider, boxes, data_path, env)
       # Load the configuration for the machine
-      name = File.basename(name.to_s).delete(File.extname(name.to_s)).to_sym # Become a vmname if name from file path.
+      name = File.basename(name.to_s).sub(File.extname(name.to_s),'').to_sym # Become a vmname if name from file path.
       results = machine_config(name, provider, boxes)
       box             = results[:box]
       config          = results[:config]
@@ -112,7 +112,7 @@ module Vagrant
     #   machine. See the main documentation body for more info.
     def machine_config(name, provider, boxes)
       keys = @keys.dup
-      name = File.basename(name.to_s).delete(File.extname(name.to_s)).to_sym # Become a vmname if name from file path.
+      name = File.basename(name.to_s).sub(File.extname(name.to_s),'').to_sym # Become a vmname if name from file path.
 
       sub_machine = @config.vm.defined_vms[name]
       if !sub_machine
--- a/test/unit/vagrant/vagrantfile_test.rb
+++ b/test/unit/vagrant/vagrantfile_test.rb
@@ -401,4 +401,23 @@ describe Vagrant::Vagrantfile do
       expect(subject.primary_machine_name).to be_nil
     end
   end
+
+  describe "#machine_names_if_filename_and_path" do
+    it "returns the machine name single-VM by the file name" do
+      RSpec.describe [:foo] do
+        before do
+          ARGV.clear
+          ARGV.concat(['playbooks/foo.yaml'])
+          vm_name = File.basename(ARGV[1]).delete(File.extname(ARGV[1]))
+          expect { print 'foo' }.to output(vm_name).to_stdout
+
+          configure do |config|
+            config.vm.define vm_name
+          end
+
+          expect(subject.machine_names).to eq([:foo])
+        end
+      end
+    end
+  end
 end
--- a/test/unit/vagrant/vagrantfile_test.rb
+++ b/test/unit/vagrant/vagrantfile_test.rb
@@ -408,7 +408,7 @@ describe Vagrant::Vagrantfile do
         before do
           ARGV.clear
           ARGV.concat(['playbooks/foo.yaml'])
-          vm_name = File.basename(ARGV[1]).delete(File.extname(ARGV[1]))
+          vm_name = File.basename(ARGV[1]).sub(File.extname(ARGV[1]),'')
           expect { print 'foo' }.to output(vm_name).to_stdout
 
           configure do |config|


aim-oguma:
# Details

An example of increasing the reusability of Vagrantfile by specifying machine name and Ansible playbook by argument

Environment

```
.
├── Vagrantfile
├── ansible.cfg
├── playbooks
│   ├── bar.yml
│   └── foo.yml
└── roles
    └── myname
        └── tasks
            └── main.yml
```

It is organized by roles and playbooks, and it is easy to reuse.
To organize roles and playbooks with Directory, specify custom role path in ansible.cfg

ansible.cfg

```
[defaults]
roles_path = roles
```

machine name is `ARGV[1]`
`@vm_name = File.basename(ARGV[1]).delete(File.extname(ARGV[1])`

Specify machine name as an argument you can make it run the same ansible playbook as the specified name

Vagrantfile

```
@vm_name = File.basename(ARGV[1]).delete(File.extname(ARGV[1])

playbook = {
  :target => "playbooks/#{@vm_name}.yml"
}

  config.vm.box = "hashicorp/precise32"

  config.vm.define @vm_name if !@vm_name.nil?

  config.vm.provision "ansible" do |ansible|
    ansible.playbook = playbook[:target]
    ansible.limit = 'all'
  end
end
```

Ansible playbooks

```
% cat playbooks/foo.yml
---
- hosts: all
  roles:
    - role: myname

% cat playbooks/bar.yml
---
- hosts: all
  roles:
    - role: myname
```

This also makes Vagrant reusable for each playbook.

```
vagrant up foo --provision
vagrant up bar --provision
```

By doing this, you can run the following ansible playbook respectively

```
playbooks
├── bar.yml
└── foo.yml
```

## Problems

The following cannot specify and execute if it is less than

```
vagrant up playbooks/foo --provision
vagrant up playbooks/foo.yml --provision
vagrant up foo.yml --provision
```

It is caused by the following problem and can not be executed
`The machine with the name 'playbooks/foo.yml' was not found configured for
this Vagrant environment.`

This commit is to solve this problem.

aim-oguma:
### The local test in this patch.

```
% ~/work/repos/vagrant/exec/vagrant --version
Vagrant 2.0.2.dev
```
```
% vagrant_cmd="${HOME}/work/repos/vagrant/exec/vagrant up";for machine_playbook in {,playbooks/}foo{,.yml} {,playbooks/}bar{,.yml}; do echo "$machine_playbook start"; eval "$vagrant_cmd $machine_playbook --provision" | grep '"inventory_hostname":' ;done
foo start
    "inventory_hostname": "foo"
foo.yml start
    "inventory_hostname": "foo"
playbooks/foo start
    "inventory_hostname": "foo"
playbooks/foo.yml start
    "inventory_hostname": "foo"
bar start
    "inventory_hostname": "bar"
bar.yml start
    "inventory_hostname": "bar"
playbooks/bar start
    "inventory_hostname": "bar"
playbooks/bar.yml start
    "inventory_hostname": "bar"
```

aim-oguma:
Travis CI's following error.

https://travis-ci.org/hashicorp/vagrant/builds/299526337?utm_source=github_status&utm_medium=notification

Only ruby 2.3.5 happen.

```
    with a box name accidentally set as a URL
      displays a warning to the user
    with a box name containing invalid URI characters


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

However, in this commit, the box name is irrelevant.

aim-oguma:
Supplement

`File.basename(ARGV.first(2).last).delete(File.extname(ARGV.first(2).last))`
VS
`File.basename(ARGV[1]).delete(File.extname(ARGV[1])`

There is no big difference, but ARGV[1] gave faster results about 14% at nsec.
Therefore, ARGV [1] was adopted.

aim-oguma:
I also thought of specifying playbook and machine name in environment variables. (I referred to The 12-factor app)
There can easily convert it to machine name as follows.

```
# e.g,
@vm_name = File.basename(ENV['PLAYBOOK']).delete(File.extname(ENV['PLAYBOOK'])) 
playbook = {
  :target => "playbooks/#{@vm_name}.yml"
}
config.vm.define @vm_name
# ..<snip>
```

But I quitted.
It doesn't fit the specification of existing the Vagrant.
(eg, `vagrant ssh foo` to` PLAYBOOK = foo.yaml vagrant ssh`)

lib/vagrant/vagrantfile.rb
```diff
@@ -42,7 +42,7 @@ def initialize(loader, keys)
     # @return [Machine]
     def machine(name, provider, boxes, data_path, env)
       # Load the configuration for the machine
-      name = File.basename(name.to_s).delete(File.extname(name.to_s)).to_sym # Become a vmname if name from file path.
+      name = File.basename(name.to_s).sub(File.extname(name.to_s),'').to_sym # Become a vmname if name from file path.
```
aim-oguma: This change because of suspect that there is a possibility of deleting other unintended characters if it is a delete function.
I am sorry if my recognition was wrong.

briancain:
Hey @aim-oguma - the test failure in travis is a known sticky test that we're trying to fix. I've rekicked the job...but it shouldn't be related to your patch :)

aim-oguma:
@briancain 
Thank you very much! :-)

briancain:
Hello @k-oguma! Thanks so much for your contribution! After carefully reviewing the proposed changes and reading over the information you included in the pull request here, I am afraid I'm going to turn it down.

The first being the actual changes within `lib/vagrant/vagrantfile.rb` modifying the guest name. This change would break how guests are named with a period resulting in some surprising behavior. For example, if a user had named their guest `guest.local`, it would get automatically trimmed to be called `guest` for all Vagrant providers.

The second, and main reason, is I think you can achieve your desired workflow without any changes to Vagrant core. If you are interested in bringing up arbitrary guests named after playbooks on disk, I think this can be done through environment variables. For example:

```
PLAYBOOK=foo vagrant up
PLAYBOOK=playbook/bar.yml vagrant up
```

```ruby
@vm_name = File.basename(ENV['PLAYBOOK']).sub(File.extname(ENV['PLAYBOOK']),'')

playbook = {
  :target => "playbooks/#{@vm_name}.yml"
}

  config.vm.box = "hashicorp/precise32"

  config.vm.define @vm_name if !@vm_name.nil?

  config.vm.provision "ansible" do |ansible|
    ansible.playbook = playbook[:target]
    ansible.limit = 'all'
  end
end
```

Again, thank you so much for making a contribution! ❤️

k-oguma:
Hi @briancain 
Thank you for confirmation and consideration.Also, thank you for presenting sample code.

PR #9170, opened 2017-11-10
Fix #8620 by making cifs mount_option "sec=" configurable

### LINKED ISSUE #8620: SMB mount error(13): Permission denied
### Vagrant version
Vagrant 1.9.5

### Host operating system
Windows 7 (64-bit)

### Guest operating system
Ubuntu 16.04

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

  config.vm.box = "get-mca/dev"
  config.vm.box_url = "file://vm.json"
  config.ssh.forward_x11 = true
  config.ssh.forward_agent = true
  config.vm.box_check_update = false
  config.vm.synced_folder ".", "/vagrant"

    v.check_guest_additions = false
    v.functional_vboxsf     = false

    v.gui = false
    v.name = "getmca_dev"
    v.customize ["modifyvm", :id, "--memory", "1024"]
    v.customize ["modifyvm", :id, "--vram", "12"]
    v.customize ["modifyvm", :id, "--cpus", "1"]
    v.customize ["modifyvm", :id, "--cpuexecutioncap", "50"]
  end
end

```

### Debug output
==> default: Configuring proxy for Apt...
==> default: Configuring proxy for Docker...
==> default: Configuring proxy environment variables...
==> default: Mounting SMB shared folders...
    default: C:/cygwin64/home/200003581/GET-MCA/VirtualMachines/packer/example => /vagrant
Failed to mount folders in Linux guest. This is usually because
the "vboxsf" file system is not available. Please verify that
the guest additions are properly installed in the guest and
can work properly. The command attempted was:

mount -t cifs -o uid=1000,gid=1000,sec=ntlm,credentials=/etc/smb_creds_e18c37823b835d29d892174d053c9a9c //10.10.10.1/e18c37823b835d29d892174d053c9a9c /vagrant

The error output from the last command was:

mount error(13): Permission denied
Refer to the mount.cifs(8) manual page (e.g. man mount.cifs)

### Expected behavior
When the smb mount is attempted on the guest, it unconditionally inserts "sec=ntlm" 
However, this mode is not allowed in all environments.  The newer cifs package would default to ntlmssp, but is overridden in the vagrant smb plugin 

HashiCorp/Vagrant/embedded/gems/gems/vagrant-1.9.5/plugins/guests/linux/cap/mount_smb_shared_folder.rb

If I comment out the ntlm option
```
#  options[:mount_options] << "sec=ntlm"
```
or replace with ntlmssp:
```
  options[:mount_options] << "sec=ntlmssp"
```

I am able to establish the SMB share...  
The current code does not allow this option to be overridden the configuration.  So even if you supplied the sec setting as a mount option:

  config.vm.synced_folder ".", "/vagrant", mount_options:["sec=ntlmssp"]

you end up with both the request and the forced sec option:

mount -t cifs -o uid=1000,gid=1000,sec=ntlm,credentials=/etc/smb_creds_e18c37823b835d29d892174d053c9a9c //<MY_IP_ADDRESS>/e18c37823b835d29d892174d053c9a9c /vagrant

Which causes the mount to fail

It would be good to only insert sec=ntlm if the sec option was not specified in Vagrantfile mount_options.

### Actual behavior
What actually happened?

### Steps to reproduce
1.
2.
3.

### References
Are there any other GitHub issues (open or closed) that should be linked here?
For example:
- Possibly GH-8381 - similar to and perhaps the root cause but was closed for other reason
- GH-4823 - but the fix has not made it into the main branch (now marked closed).

TonyApuzzo:
Recent Linux kernels require sec=ntlmssp or similar and don't work with sec=ntlm anymore.

This PR makes it easy to override the old behavior while retaining backward compatibility.

If a `sec=...` option is specified in the `mount_options`, then use that, otherwise fall back to the old behavior of hard-coding `sec=ntlm`

Example:
```
config.vm.synced_folder ".", "/vagrant", { mount_options: ["sec=ntlmssp"] } 
```

If a top-level option of 'nosec' is true then do not add the default `sec=ntlm`, allowing the guest's default to be used.
Example:
```
config.vm.synced_folder ".", "/vagrant", { nosec: true } 
```

### DIFF
--- a/plugins/guests/linux/cap/mount_smb_shared_folder.rb
+++ b/plugins/guests/linux/cap/mount_smb_shared_folder.rb
@@ -24,7 +24,7 @@ module VagrantPlugins
           smb_password = options[:smb_password]
 
           options[:mount_options] ||= []
-          options[:mount_options] << "sec=ntlm"
+          options[:mount_options] << "sec=ntlm" unless options[:mount_options].grep(/^sec=/) || options.fetch(:nosec, false)
           options[:mount_options] << "credentials=/etc/smb_creds_#{name}"
 
           mount_options = "-o uid=#{mount_uid},gid=#{mount_gid}"


chrisroberts:
Hi there. Thanks for submitting this PR! Instead of just supporting the modification of a single mount option, I've opened #9366 to support properly overriding any default mount options from the `mount_options` attribute. I'm going to close this PR since #9366 will provide the desired behavior (and updates the default `sec` value). Thanks again for submitting this.

Cheers!

Issue #1484, opened 2013-03-22
Unrecognised config keys cause unclear error


alexjurkiewicz:
As per:

https://github.com/mitchellh/vagrant-aws/issues/8

PR #9173, opened 2017-11-10
Ansible Provisioner: Fix broken 'ask_sudo_pass' option

gildegoma:
This bug is a regression introduced in Vagrant 2.0.0. It hasn't been caught by unit tests because
Vagrant::Plugin::V2::Config [catches all invalid/bad configuration calls](https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/plugin/v2/config.rb#L70-L83)
and save them for generating error messages later.
The `ask_sudo_pass=(value)` method was therefore not interrupted when calling the undefined `show_deprecation_warning` method and the
`@ask_become_pass` attribute was (surprisingly) correctly set (allowing
the related unit tests to pass).

In order to avoid similar problem to happen again:
* the deprecation message output is now fully verified (e47deb7)
* we now ensure that the provisioner configurations used by the unit tests have no errors (8333090d1d24578bd33f4913f5ae2c0f8b2cf899)

Bug reported by @trombik (https://github.com/hashicorp/vagrant/commit/8834afbd8e6198b49b0a0ad2b3756c139a8d27b7#commitcomment-25424227)

Workaround: use the recommended `ask_become_pass` option instead.

Related Issues:
* GH-1484
* GH-8913

### DIFF
--- a/plugins/provisioners/ansible/config/host.rb
+++ b/plugins/provisioners/ansible/config/host.rb
@@ -16,7 +16,7 @@ module VagrantPlugins
         #
         alias :ask_sudo_pass :ask_become_pass
         def ask_sudo_pass=(value)
-          show_deprecation_warning 'ask_sudo_pass', 'ask_become_pass'
+          show_deprecation_info 'ask_sudo_pass', 'ask_become_pass'
           @ask_become_pass = value
         end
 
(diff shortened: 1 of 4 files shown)


gildegoma:
@chrisroberts Thanks for checking ❤️

PR #9202, opened 2017-11-17
Support for custom Vagrantfile templates

zachflower:
This pull request adds a `--template` option to the `init` command, allowing users to override the default `Vagrantfile` template with their own custom template. When omitted, `init` falls back to the default `Vagrantfile` or `Vagrantfile.min` templates. 

## Example

Given the following custom template (also included in the repository for use within the unit tests):

```erb
# -*- mode: ruby -*-
# vi: set ft=ruby :

# you're doing.
  config.vm.hostname = "vagrant.dev"  
  config.vm.box = "<%= box_name %>"
  <% if box_version -%>
  config.vm.box_version = "<%= box_version %>"
  <% end -%>

  <% if box_url -%>
  # The url from where the 'config.vm.box' box will be fetched if it
  # doesn't already exist on the user's system.
  config.vm.box_url = "<%= box_url %>"
  <% end -%>
end
```

Executing a standard `vagrant init` command with the `--template` option included will result in a new `Vagrantfile` using the provided template (both absolute and relative paths are supported):

kapture 2017-11-17 at 12 26 20

When the `--template` option is omitted, the `init` command falls back to the current behavior:

kapture 2017-11-17 at 12 26 54

### DIFF
--- a/lib/vagrant/errors.rb
+++ b/lib/vagrant/errors.rb
@@ -784,6 +784,10 @@ module Vagrant
       error_key(:vagrantfile_syntax_error)
     end
 
+    class VagrantfileTemplateNotFoundError < VagrantError
+      error_key(:vagrantfile_template_not_found_error)
+    end
+
     class VagrantfileWriteError < VagrantError
       error_key(:vagrantfile_write_error)
     end
--- a/plugins/commands/init/command.rb
+++ b/plugins/commands/init/command.rb
@@ -14,6 +14,7 @@ module VagrantPlugins
           force: false,
           minimal: false,
           output: "Vagrantfile",
+          template: nil
         }
 
         opts = OptionParser.new do |o|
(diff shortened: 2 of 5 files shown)


marek-obuchowicz:
Did you also consider allowing remote template?
Background info: it will be very helpful to publish template for software XXX on some public location, then just ask user who want to get a VM running to do: `vagrant init --template https://github.com/org/project/Vagrantfile.erb`

zachflower:
Hey @marek-obuchowicz!

I hadn't thought about allowing remote templates, but that's a great idea. The current implementation is _about_ as simple as it can get, so this would definitely add a little bit more complexity, but it's not by any means a new convention (`vagrant box add` pulls down a remote file, for example).

I'd be happy to take a crack at it, but the added complexity might warrant a larger conversation with the maintainers on a new issue. I can see the value, but it might not be a direction they want to move in.

marek-obuchowicz:
Thanks for response, @zachflower. I think those two topics should be seperated and if this PR is ready to be merged and released soon, it should not be delayed by the idea I've mentioned above. Let's make a new issue out of that. If there is any way I could support you (ie. by providing real life use case we have for this), I will do my best. If you create a follow up thread, please post the new issue link. And thanks a lot for nice feature!

PR #9214, opened 2017-11-23
Add ssh key forwarding

vincetse:
Add a commented-out ssh key forwarding configuration to the `Vagrantfile.erb` template.


chrisroberts:
Hi there,

Thanks for this PR. The Vagrantfile template only provides a small subset of available configurations and is simply a starting point with simple examples. Enabling key forwarding does not feel like it fits in this template and can be easily found in the Vagrant documentation for those looking for the functionality.

Thanks again for the PR. Cheers!

PR #9227, opened 2017-11-29
synced_folders/rsync: Use "staff" as a default group name on Darwin

legal90:
By default, on macOS (and OS X) boxes for Vagrant don't have "vagrant" group existing. 
We should use "staff" as a default group name while doing `chown`.

This PR fixes the problem with rsync on macOS guests. Reproduction scenario:

```ruby
# Vagrantfile

config.vm.provider "virtualbox" do |_, override|
  override.vm.synced_folder ".", "/vagrant", type: "rsync"
end
```

```
$ vagrant up --provider=virtualbox
<...>
    default: No guest additions were detected on the base box for this VM! Guest
    default: additions are required for forwarded ports, shared folders, host only
    default: networking, and more. If SSH fails on this machine, please install
    default: the guest additions and repackage the box to continue.
    default:
    default: This is not an error message; everything may continue to work properly,
    default: in which case you may ignore this message.
==> default: Rsyncing folder: /Users/legal/Workspace/vagrant-test-macos/ => /vagrant

find /vagrant '!' -type l -a '(' ! -user vagrant -or ! -group vagrant ')' -exec chown vagrant:vagrant '{}' +

Stdout from the command:



Stderr from the command:

find: -group: vagrant: no such group
```

### DIFF
--- a/plugins/synced_folders/rsync/helper.rb
+++ b/plugins/synced_folders/rsync/helper.rb
@@ -62,7 +62,7 @@ module VagrantPlugins
 
         # Folder options
         opts[:owner] ||= ssh_info[:username]
-        opts[:group] ||= ssh_info[:username]
+        opts[:group] ||= Vagrant::Util::Platform.darwin? ? "staff" : ssh_info[:username]
 
         # set log level
         log_level = ssh_info[:log_level] || "FATAL"


chrisroberts:
Hi there,

By default vagrant will choose the `vagrant` group for all synced folders types (not just rsync). This can be modified in the Vagrantfile using the `group` option for the defined synced folder: https://www.vagrantup.com/docs/synced-folders/basic_usage.html#group

If you don't want to have to manually define the group within your Vagrantfile, a better option would be to create the `vagrant` group and add the `vagrant` user to it when building the box (which is what I generally do).

Cheers!

PR #9253, opened 2017-12-08
Check SMB credentials before using them

LachlanArthur:
Check that the provided username/password are valid before using them to set up SMB shares.

This helps avoid having to reboot the VM just because the user messed up a password.


bbrala:
Nice one, that would really help out, people do tend to make a typo 👍

chrisroberts:
@LachlanArthur Thank you very much for this PR. The SMB implementation has been updated since you originally created this PR. I have pulled your changes into a new PR (#9365) and applied some updates so that it will work correctly in the new implementation. This PR is getting closed but your update will still be getting merged in. Again, thank you!

Cheers!

PR #9261, opened 2017-12-12
Disable if/netplugd when setting up a static ip on a gentoo guest using openrc.

vStone:
When ifplugd is installed, the process is backgrounded and we get the
wrong return value.


vStone:
Forgot to update the tests. Fixed that.

PR #9294, opened 2017-12-21
SMB synced folders enhancements

chrisroberts:
This enhancement updates the existing implementation of the SMB synced folders. Host specific capabilities have been extracted from the SMB synced folder plugin itself and isolated into proper host capabilities. Items of note in this PR:

* Added support for SMB host on macOS
* Support removal of created host SMB shares on destroy
* Prevent display of mount credentials within UI output and logging
* Removal of Administrator user requirement on Windows
  * Share setup/removal executed via powershell with UAC
* Added test coverage on SMB synced folder and new related host capabilities


karousn:
Any possible validation for this PR ?

bbrala:
@chrisroberts I think you should also update docs since passwords are no longer in the logs?

https://www.vagrantup.com/docs/other/debugging.html

bbrala:
I've tried this build on my machine. First issue i ran intro was an error that a share was already existing. Probably because i use a named share so it doesn't populate our shared folders with 100 shares :)

`override.vm.synced_folder "../", "/home/localcopy", smb_id: "projecten70hyperv", type: "smb", mount_options: ["mfsymlinks", "forceuid", "noperm", "file_mode=0777", "dir_mode=0777", "vers=3.0"]`

[Gist with the debug log](https://gist.github.com/bbrala/6a5c31ff4f338fbcb95404bf49c197a1)

bbrala:
Next error when trying to get my machine up is: `Vagrant: * Unknown configuration section 'smb'.`  Not sure what changed :)

[Gist with debug log](https://gist.github.com/bbrala/eb77b5680e4e3bd35fecb4c0305c75c4)

After this i tried without the `vagrant-triggers` plugin, which might not help. Then i got the share already exists error again. After removing all vagrant shares i tried again and got the same error regarding the configuration.

bbrala:
I don't have a Mac env available at the moment, but could get that up and running when at the Windows env is working properly. My windows version is Windows 1709 by the way.

chrisroberts:
@bbrala Can you provide a minimal Vagrantfile that produces the error output you saw (`Unknown configuration section`)?

chrisroberts:
@bbrala I had been cleaning up the macOS bits previously. Finished cleaning the Windows bits today and should fix up the errors you were seeing with already defined shares. If you are still seeing the unknown configuration section error, please do let me know and I'll get that fixed up before this gets merged. Thanks!

bbrala:
I'll have a go again to see, i couldn't reproduce the problem with a clean box yesterday, so it might be an oversight on my side. I'll have a look at the updated PR

bbrala:
Ok, lets see.

Vagrantfile:
```
# -*- mode: ruby -*-
# vi: set ft=ruby :
  config.vm.box = "generic/ubuntu1604"
  
  config.vm.provider "hyperv" do |v, override|
        override.vm.synced_folder "../", "/home/localcopy", smb_id: "projecten70hyperv", type: "smb", mount_options: ["mfsymlinks", "forceuid", "noperm", "file_mode=0777", "dir_mode=0777", "vers=3.0"]

        v.memory = 2048
        v.maxmemory = 2048
        v.cpus = 4
        v.vmname = 'devlocal'
        v.auto_start_action = 'Nothing'

        # Currently bugged: see https://github.com/hashicorp/vagrant/pull/9000
        # v.auto_stop_action = 'ShutDown'
    end
end
```

Running the following commands with no existing shares. (vagrant dev is a shortcut to run Vagrant from source (see [this gist](https://gist.github.com/bbrala/7e96be8a4e26f14834e8bbc21d246c38)).

```
vagrant-dev up
```

I get the smb configuration option error, see [the debug gist](https://gist.github.com/bbrala/87bb011e72a6b0885736eb4301db200d). I think it might be because i use the `config.vm.provider "hyperv" do |v, override|` statement? Which i think is to enable different settings for specific providers.

update: This is not the issue it seems, i get the error also if i put it in the `Vagrant.configure("2") do |config|` section.

Another thing, i feel that an error for an existing share it a bit much? Perhaps an warning is enough? I kinda want to reuse the share if possible, but right now even when the signature (folder/name) is the same it will just die. This makes it impossible to reuse the share, which is a bit annoying :)

Also when i destroy the machine the shares are not pruned, but this might be because it errors while upping so for now lets first find out the config error.

chrisroberts:
@bbrala Thanks for the Vagrantfile + debug gist. Everything should be sorted out now. Existing shares _should_ be fine to use as long as the host paths match. Paths weren't being expanded prior to comparison, but are now so it should fix that issue:

https://github.com/hashicorp/vagrant/pull/9294/files#diff-151f1bb6d235fed7af0acb396a7d7a8fR66

The unknown configuration section error wasn't being triggered with the virtualbox provider, but once I switched over to hyper-v it was easily re-producible. That has been fixed up with the addition of the config section:

https://github.com/hashicorp/vagrant/pull/9294/files#diff-d2a3685932a5e2479b246a5440179354R1

And the shares not being cleaned up was due to the hyper-v provider not including synced folder cleanup which was added here:

https://github.com/hashicorp/vagrant/pull/9294/files#diff-a0ae8c3bfc255930c24c4f5ae3f2a5a1R44

and 

https://github.com/hashicorp/vagrant/pull/9294/files#diff-a0ae8c3bfc255930c24c4f5ae3f2a5a1R148

Please do let me know if you encounter any other issues. Otherwise I'll get this merged :)

bbrala:
I've gone through your additions and checked on the above box and our dev env. Seems like all issues have been resolved. No more config errors, SMB get cleaned up nicely, and shared seem to get reused. :)

Well done! 👍

PR #9295, opened 2017-12-21
recognize solaris alternatives

mcandre:
As the Solaris project is dead, recognize that Solaris alternatives like DilOS and SmartOS feature a Solaris compatible kernel and capabilities.


mcandre:
Looks like the CI test stalled for some reason. Can it be rerun?

briancain:
I've rekicked the job @mcandre !

Issue #9298, opened 2017-12-24
WSL rootfs on newer OS builds is no longer "%LOCALAPPDATA%\lxss"


BR0kEN-:
### Vagrant version
Vagrant 2.0.2

### Host operating system
Windows 10 (version `1709`, build `16299.125`).

### Expected behavior
Vagrant is to detect current WSL container and compute a proper path to it.

### Actual behavior
Vagrant handles only the WSL that is installed via `lxrun`.

### References

- https://github.com/Microsoft/WSL/issues/2578
- https://github.com/berkshelf/vagrant-berkshelf/issues/323
- https://github.com/BR0kEN-/cikit/tree/issues/52/docs/vagrant/wsl
- https://github.com/RoliSoft/WSL-Distribution-Switcher/issues/69

### Additional information
Before the `16215` build of Windows, there was a way to install a WSL using `lxrun`. After mentioned build this way has been deprecated and, since then, it's recommended to use Microsoft Store for installing a WSL. Moreover, since that version, you can have multiple instances of Linux subsystem.

When the installation is done by `lxrun /install` then `rootfs` of a WSL will be in `%LOCALAPPDATA%\lxss` and Vagrant will operate correctly since its codebase relies on this location (https://github.com/hashicorp/vagrant/blob/v2.0.1/lib/vagrant/util/platform.rb#L286-L289).

![lxrun installation](https://github.com/BR0kEN-/cikit/raw/issues/52/docs/vagrant/wsl/images/16215-lxrun.png)

When the installation is done from the Microsoft Store, then `rootfs` of a WSL will be in `%LOCALAPPDATA%\Packages\<WSL_ID>\LocalState`. Pay your attention at the `<WSL_ID>` token in the path which is completely dynamic.

![installation from store](https://github.com/BR0kEN-/cikit/raw/issues/52/docs/vagrant/wsl/images/16215-store.png)

The path of the particular container can be obtained from the registry at `HKCU:\Software\Microsoft\Windows\CurrentVersion\Lxss` - a listing of installed subsystems.

For instance, using the following PowerShell script, the path of default subsystem can be gained.

```powershell
$WSLREGKEY="HKCU:\Software\Microsoft\Windows\CurrentVersion\Lxss"
$WSLDEFID=(Get-ItemProperty "$WSLREGKEY").DefaultDistribution
$WSLFSPATH=(Get-ItemProperty "$WSLREGKEY\$WSLDEFID").BasePath
[...]

PR #9300, opened 2017-12-25
#9298: Compute a correct path to the current WSL instance

BR0kEN-:
https://github.com/hashicorp/vagrant/issues/9298

### DIFF
--- a/lib/vagrant/errors.rb
+++ b/lib/vagrant/errors.rb
@@ -939,5 +939,9 @@ module Vagrant
     class WSLVirtualBoxWindowsAccessError < VagrantError
       error_key(:wsl_virtualbox_windows_access)
     end
+
+    class WSLRootFsNotFoundError < VagrantError
+      error_key(:wsl_rootfs_not_found_error)
+    end
   end
 end
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -274,9 +274,9 @@ module Vagrant
           if linux?
             # A simplest way to get the Linux distribution name.
             result = Subprocess.execute(
-                "python",
-                "-c",
-                "import platform;print(platform.linux_distribution()[0])"
+              "python",
+              "-c",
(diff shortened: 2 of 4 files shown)


BR0kEN-:
Just checked this on SUSE and it turned out that some tweaks were required.

The previous script for obtaining Linux distro:

```python
python -c "import platform;print(platform.linux_distribution()[0]"
```

And the new one:

```python
python -c "import platform;print(platform.linux_distribution()[0].split(' ')[0])"
```

wsl-multi

BR0kEN-:
The PR brings not an elegant way of solving the problem but, apparently, it works.

Checked on three WSL distros: SUSE Server (the new method `linux_distro` will return `SUSE`), openSUSE (`linux_distro` -> `openSUSE`) and Ubuntu (`linux_distro` -> `Ubuntu`).

Of course, as said in the https://github.com/Microsoft/WSL/issues/2578#issuecomment-353900973, there could be a custom distro that won't be handled in that way, but for now this approach is multi-environment at least.

BR0kEN-:
The implementation has been changed based on investigation I've described in https://github.com/Microsoft/WSL/issues/2578#issuecomment-354010141.

chrisroberts:
@BR0kEN- Thanks for this PR and the underlying investigation to create it! Once I have validated the functionality on a few versions of win10, I'll get it approved and merged.

temoor-gilmutdinov:
Hi there, checked it on Win10, version 10.0.16299.192. Works well.

BR0kEN-:
My tests, unfortunately, were on the same build of Win10.

chrisroberts:
@BR0kEN- Hi and thanks again for all the hard work you put into this PR. It is greatly appreciated. I'm going to close this PR in favor of #9525 which adds in support for the original lxrun generated WSL instance types and adds a bit of spec coverage. It is built directly on top of your changes, so even though this PR is being closed, your changes will still be getting merged.

Thanks again!

PR #9330, opened 2018-01-02
Update docs to use `Tee-Object` in PowerShell in Windows.

bbrala:
By using `Tee-Object` you can ouput the debug log to the console and a file, which is way easier to use than just outputting to a logfile. This makes debugging a lot easier.


briancain:
Thanks for your contribution! 🎉

PR #9341, opened 2018-01-05
Deprecate :paranoid in favor of :verify_host_key

### LINKED ISSUE #9062: :paranoid is deprecated, please use :verify_host_key. Supported values are exactly the same, only the name of the option has changed.
Please note that the Vagrant issue tracker is reserved for bug reports and
enhancements. For general usage questions, please use the Vagrant mailing list:
https://groups.google.com/forum/#!forum/vagrant-up. Thank you!

### Vagrant version

1.9.8, but code still exists

### Host operating system

```
$ sudo lsb_release -da


No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux unstable (sid)
Release:	unstable
Codename:	sid
```

### Guest operating system

Ubuntu 16.04

### Vagrantfile

Irrelevant

Yah so this warning is being output to stderr with newer debian updates. Preeety sure the SSH client got updated.

Heads up.

briancain:
As of `net-ssh` version 4.2.0, the key :paranoid has been deprecated in
favor of using :verify_host_key. This commit updates Vagrants ssh config
to use the new key, and deprecates the use of :paranoid.

Fixes #9062

### DIFF
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -438,7 +438,7 @@ module Vagrant
       info[:port] ||= @config.ssh.default.port
       info[:private_key_path] ||= @config.ssh.default.private_key_path
       info[:keys_only] ||= @config.ssh.default.keys_only
-      info[:paranoid] ||= @config.ssh.default.paranoid
+      info[:verify_host_key] ||= @config.ssh.default.verify_host_key
       info[:username] ||= @config.ssh.default.username
       info[:compression] ||= @config.ssh.default.compression
       info[:dsa_authentication] ||= @config.ssh.default.dsa_authentication
@@ -449,7 +449,7 @@ module Vagrant
       info[:host] = @config.ssh.host if @config.ssh.host
       info[:port] = @config.ssh.port if @config.ssh.port
       info[:keys_only] = @config.ssh.keys_only
-      info[:paranoid] = @config.ssh.paranoid
+      info[:verify_host_key] = @config.ssh.verify_host_key
       info[:compression] = @config.ssh.compression
       info[:dsa_authentication] = @config.ssh.dsa_authentication
       info[:username] = @config.ssh.username if @config.ssh.username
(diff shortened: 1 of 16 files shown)


lib/vagrant/machine.rb
```diff
@@ -449,7 +449,10 @@ def ssh_info
       info[:host] = @config.ssh.host if @config.ssh.host
       info[:port] = @config.ssh.port if @config.ssh.port
       info[:keys_only] = @config.ssh.keys_only
-      info[:paranoid] = @config.ssh.paranoid
+      info[:verify_host_key] = @config.ssh.verify_host_key || @config.ssh.paranoid
+      if @config.ssh.paranoid
+        @env.ui.warn(I18n.t("vagrant.config.ssh.paranoid_deprecated"))
+      end
```
briancain: Not really sure if this is the right place for the warning..but it might be? It's fairly noisy here however if a user still has the `:paranoid` setting in their config.

test/unit/plugins/commands/ssh_config/command_test.rb
```diff
@@ -124,15 +124,17 @@
...
+      allow(machine).to receive(:ssh_info) { ssh_info.merge(verify_host_key: true) }
 
       output = ""
       allow(subject).to receive(:safe_puts) do |data|
         output += data if data
       end
 
       subject.execute
+      puts "the output tho"
+      puts output
```
chrisroberts: Remove debug output

lib/vagrant/machine.rb
chrisroberts: The warning would be better to originate from the config. We can also set the `verify_host_key` value within the config from `paranoid` if it is set within the config as well (when finalizing) and fully remove the `paranoid` usage from this file.

test/unit/plugins/commands/ssh_config/command_test.rb
briancain: Oops!

briancain:
@chrisroberts - updated based on feedback!

PR #9351, opened 2018-01-09
Change to a temporary directory before downloading script files

### LINKED ISSUE #9350: Salt provisioner downloads salt-bootstrap.sh into the vagrant user's home directory, leaving confusing cruft
### Vagrant version
Vagrant 2.0.1

### Guest operating system
Ubuntu 14.04 & 16.04

### Expected behavior
bootstrap-salt.sh should be downloaded into a temporary directory where it will not confuse first time users.

### Actual behavior
bootstrap-salt.sh is placed in the vagrant user's home directory, making it appear that salt provisioning needs some manual action to complete.  This is a regression since revision 7a7506cd009c9fb3d5615b1a7a04b49deeb609b8 - before that revision the salt-bootstrap.sh file was not saved into the filesystem, but immediately piped onwards.

gclinch:
Avoids leaving confusing files in the working directory (probably ~vagrant).  Fixes #9350

### DIFF
--- a/plugins/provisioners/salt/bootstrap-salt.sh
+++ b/plugins/provisioners/salt/bootstrap-salt.sh
@@ -1,5 +1,7 @@
 #!/bin/sh -
 
+cd `mktemp --directory`
+
 # We just download the bootstrap script by default and execute that.
 if [ -x /usr/bin/fetch ]; then
     /usr/bin/fetch -o bootstrap-salt.sh https://raw.githubusercontent.com/saltstack/salt-bootstrap/stable/bootstrap-salt.sh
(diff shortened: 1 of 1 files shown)


plugins/provisioners/salt/bootstrap-salt.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh -
 
+cd `mktemp --directory`
```
briancain: Perfect! The only change here is we should use `mktemp -d` instead of the full flag, because some OS's only support the short version for some reason.

PR #9352, opened 2018-01-09
website: Fix incorrect URLs in some API examples

Phoenix09:


justincampbell:
Thanks for catching this!

PR #9353, opened 2018-01-09
Allow docker provisioner to run on windows

### LINKED ISSUE #9280: Windows docker capability missing
### Guest operating system
Windows

### Vagrantfile
```ruby
config.vm.define "api" do |api|
    api.vm.box = "StefanScherer/windows_2016_docker"
    api.vm.box_version = "2017.12.14"
    api.vm.provision "docker" do |d|
    end
  end
```
### Expected behavior
Expect docker provisioning to work on windows

### Actual behavior
Capability is missing to detect if daemon is running

### Steps to reproduce
1. use above vagrantfile snippit

### LINKED ISSUE #9339: usage of "rm -f" is not portable
https://github.com/hashicorp/vagrant/blob/9c299a2a357fcf87f356bb9d56e18a037a53d138/plugins/provisioners/docker/client.rb#L131

Parameter cannot be processed because the parameter name 'f' is ambiguous. Possible matches include: -Filter -Force.

elimin8r:
fixes #9339 
remove whitespace causing winrm filter from working

fixes #9280 
Check to see if docker is running

also fixes issue when file does not exist on windows and it is attempted to be removed, it no longer causes an error.

### DIFF
--- a/plugins/provisioners/docker/client.rb
+++ b/plugins/provisioners/docker/client.rb
@@ -127,9 +127,8 @@ module VagrantPlugins
       def create_container(config)
         args = container_run_args(config)
 
-        @machine.communicate.sudo %[
-          rm -f #{config[:cidfile]}
-          docker run #{args}
+        @machine.communicate.sudo %[rm -f #{config[:cidfile]}
+                                    docker run #{args}
         ]
 
         sha  = Digest::SHA1.hexdigest(args)
--- a/plugins/provisioners/docker/client.rb
+++ b/plugins/provisioners/docker/client.rb
@@ -128,8 +128,9 @@ module VagrantPlugins
         args = container_run_args(config)
 
         @machine.communicate.sudo %[rm -f #{config[:cidfile]}
-                                    docker run #{args}
-        ]
+                                  ]
+        @machine.communicate.sudo %[docker run #{args}
+                                  ]
 
         sha  = Digest::SHA1.hexdigest(args)
         container_data_path(config).open("w+") do |f|
--- a/plugins/provisioners/docker/client.rb
+++ b/plugins/provisioners/docker/client.rb
@@ -132,7 +132,6 @@ module VagrantPlugins
           docker run #{args}
         ]
 
-        name = container_name(config)
         sha  = Digest::SHA1.hexdigest(args)
         container_data_path(config).open("w+") do |f|
           f.write(sha)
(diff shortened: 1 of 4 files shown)


plugins/provisioners/docker/client.rb
```diff
@@ -127,12 +127,11 @@ def container_args_changed?(config)
...
         args = container_run_args(config)
 
-        @machine.communicate.sudo %[
-          rm -f #{config[:cidfile]}
-          docker run #{args}
-        ]
+        @machine.communicate.sudo %[rm -f #{config[:cidfile]}
+                                  ]
+        @machine.communicate.sudo %[docker run #{args}
+                                  ]
```
briancain: It seems like we could still make this a single "command", so that vagrant isn't connecting twice to run the commands separately.

```ruby
 @machine.communicate.sudo %[rm -f #{config[:cidfile]}; docker run #{args}]
```

Maybe?

plugins/provisioners/docker/client.rb
```diff
@@ -127,12 +127,11 @@ def container_args_changed?(config)
...
 
-        name = container_name(config)
```
briancain: I assume this was deleted because it doesn't look to be used?

test/unit/plugins/communicators/winrm/command_filter_test.rb
```diff
@@ -63,9 +63,9 @@
 
     it 'filters out rm commands' do
       expect(subject.filter('rm /some/dir')).to eq(
-        "rm /some/dir -force")
+                                                  "if (Test-Path /some/dir) {Remove-Item /some/dir -force}")
       expect(subject.filter('rm -f /some/dir')).to eq(
-        "rm /some/dir -force")
+                                                     "if (Test-Path /some/dir) {Remove-Item /some/dir -force}")
```
briancain: We should clean up the whitespace here.

plugins/communicators/winrm/command_filters/rm.rb
```diff
@@ -31,7 +31,7 @@ def filter(command)
           if recurse
             ret_cmd = "rm #{dir} -recurse -force"
           else
-            ret_cmd = "rm #{dir} -force"
+            ret_cmd = "if (Test-Path #{dir}) {Remove-Item #{dir} -force}"
```
briancain: Does this mean that the logic branch above this also needs to test if `dir` exists before removing it?

briancain:
Oh also, it would be super great if you could rebase this down to one commit please. Thank you!

briancain:
Closing in favor of https://github.com/hashicorp/vagrant/pull/9754. Thanks!

PR #9366, opened 2018-01-13
Support SMB mount option overrides

chrisroberts:
This also updates the default `sec` mount option from `ntlm` to `ntlmssp`.

### DIFF
--- a/plugins/guests/linux/cap/mount_smb_shared_folder.rb
+++ b/plugins/guests/linux/cap/mount_smb_shared_folder.rb
@@ -25,16 +25,19 @@ module VagrantPlugins
           # Ensure password is scrubbed
           Vagrant::Util::CredentialScrubber.sensitive(smb_password)
 
-          options[:mount_options] ||= []
+          mnt_opts = []
           if machine.env.host.capability?(:smb_mount_options)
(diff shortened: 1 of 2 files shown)


plugins/guests/linux/cap/mount_smb_shared_folder.rb
```diff
@@ -76,6 +79,21 @@ def self.mount_smb_shared_folder(machine, name, guestpath, options)
...
+          o_kv = Hash[overrides.map{|item| item.split("=", 2) }]
+          merged = {}.tap do |opts|
+            (b_kv.keys + o_kv.keys).uniq.each do |key|
+              opts[key] = o_kv.fetch(key, b_kv[key])
+            end
+          end
+          merged.map do |key, value|
+            [key, value].compact.join("=")
+          end
+        end
```
briancain: We should have some tests around this function

chrisroberts: This provides implicit testing https://github.com/hashicorp/vagrant/pull/9366/files/b1c38c282e13550dead1bfa9805e6a60a3487a4a#diff-8ea7a458713d53670287098e3ad15a5bR96

but some explicit tests can be provided too.

PR #9369, opened 2018-01-15
Add Vagrantfile configuration option to register sensitive values

chrisroberts:


website/source/docs/vagrantfile/vagrant_settings.html.md
```diff
@@ -21,3 +21,7 @@ Vagrant. By default this is `:detect`, which causes Vagrant to auto-detect
 the host. Vagrant needs to know this information in order to perform some
 host-specific things, such as preparing NFS folders if they're enabled.
 You should only manually set this if auto-detection fails.
+
+`config.vagrant.sensitive` - (string, array) - Value or list of values that
+should not be displayed in Vagrant's output. Value(s) will be removed from
+Vagrant's normal UI output as well as logger output.
```
briancain: One thing that isn't clear to me with these docs is what values are allowed inside this setting. Are they just options within the `config` namespace? So it might look like:

```ruby
config.vagrant.sensistive = ["config.ssh.password", "config.winrm.password"]
```

Or is it something else? It might be good to have an example of what people would want to mask here.

chrisroberts: No, these are explicit string values to be scrubbed.

```
config.vagrant.sensitive = ["my-password", "my-token", "etc.."]
```

Issue #7392, opened 2016-06-06
Salt provisioner does not work if ca-certificates and Python are missing from base box


babilen:
### Vagrant version

```
$ vagrant --version
Vagrant 1.8.1
```
### Host operating system

```
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux stretch/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
### Vagrantfile

``` ruby
VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|

  config.vm.box = "debian/jessie64"
  config.vm.define :salttest do |salttest|
    salttest.vm.hostname = "salttest-debian-jessie.test"

    salttest.vm.provision :salt do |salt|
      salt.install_type = "stable"
      salt.install_master = true
     end
  end
end
```

If you test this, please use the `libvirt` provider as the `debian/jessie64` base box for virtualbox contains both `python` and `ca-certificates`.
### Expected behavior

I expected the `salt-minion` and `salt-master` packages to be installed.
### Actual behavior

They were not installed. If you copy the bootstrap script that comes with vagrant into the basebox the following happens:
### Steps to reproduce
1. Install https://github.com/vagrant-libvirt/vagrant-libvirt with `vagrant plugin install vagrant-libvirt`
2. Use aforementioned Vagrantfile and run `vagrant up --provider=libvirt`
3. Run `vagrant ssh` and `sudo -i`
4. See output of `dpkg -l salt-master salt-minion`
5. Run `apt-get update ; apt-get install python`
6. Run `/tmp/bootstrap_salt.sh -M stable` and observe:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.7/urllib.py", line 87, in urlopen
    return opener.open(url)
  File "/usr/lib/python2.7/urllib.py", line 213, in open
    return getattr(self, name)(url)
  File "/usr/lib/python2.7/urllib.py", line 443, in open_https
    h.endheaders(data)
  File "/usr/lib/python2.7/httplib.py", line 997, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python2.7/httplib.py", line 850, in _send_output
    self.send(msg)
[...]

PR #9386, opened 2018-01-19
bootstrap-salt.sh improvements

boltronics:
Should address #7392 while maintaining POSIX compatibility.

Tested on Debian Stretch with `ksh` and `dash` shells using `curl`, `wget` and `python` downloader options. I don't have a *BSD environment handy to test the `fetch` command, but I found a man page online and believe it should now work too.

To test, simply replace `bootstrap-salt.sh` from this PR with that on an existing installation and provision new guests with Salt.

If a download command fails (ie. `curl`, `wget`, etc.) and there is an option to retry with a local CA file, this script will retry.

Note the CA certificate in `GITHUB_CA_FILE` file will require updating at some point, however even when the cert becomes old and invalid, the situation won't be any worse than it is currently. In fact we would still be better off because when re-provisioning a machine the old Salt bootstrap script may still be around - and this will now attempt to use the old file if found as a last resort (with a warning printed).

Edit: Updated my PR to also address the related issue #9350.

### DIFF
--- a/plugins/provisioners/salt/bootstrap-salt.sh
+++ b/plugins/provisioners/salt/bootstrap-salt.sh
@@ -1,12 +1,10 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
 # Download the bootstrap script and execute it.
 
 # Move an existing bootstrap-salt.sh out of the way, if present.
-function rename_old_bootstrap()
+rename_old_bootstrap()
 {
-    local BOOTSTRAP_LOCAL_OLD="$(mktemp /tmp/bootstrap-salt-old.XXXXXXXX)"
-
     if [ -s "${BOOTSTRAP_LOCAL}" -a -e "${BOOTSTRAP_LOCAL_OLD}" ]
     then
         mv -f "${BOOTSTRAP_LOCAL}" "${BOOTSTRAP_LOCAL_OLD}"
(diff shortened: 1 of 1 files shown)


briancain:
Hello @boltronics, thanks so much for this pull request!

While this bash script is very clean (and even uses functions! :tada:) we should take advantage of Ruby since we have it available to us. I'd rather not add more bash to Vagrant. The [salt provisioner](https://github.com/hashicorp/vagrant/blob/master/plugins/provisioners/salt/provisioner.rb#L247-L259) is the exact place for this. If we go forward with replacing this file in Ruby however, there's still a bit more feedback I'd want to give on this feature before going forward.

For issue #7392, this is actually a problem with the base box itself rather than something Vagrant should be handling. We should not be maintaining a CA Cert for GitHub inside of Vagrant. If a box does not have this package, the maintainers should include it in the box itself.

For issue #9351, I recently merged a pull request that creates a temp directory to download the salt bootstrap into. We could preserve that functionality, or clean up the script once the provisioner is finished in Ruby.

I'm going to go ahead and close this for now, but if you'd like to make a new PR that replaces the `bootstrap-salt.sh` script with something in Ruby to download salts bootstrap script that would be great! Again, thanks so much for contributing to Vagrant! ❤️

boltronics:
Hi @briancain. Thanks for looking at my PR.

It's a shame that I'll need to rebuild a bunch of images to include ca-certificates. I was really hoping to avoid that. If we cannot include the necessary public certificate it in Vagrant, it would be very useful if we could at least have an option to allow the user's Vagrantfile to include it, or otherwise reference the certificate from there as an external file.

While I could probably come up with a solution to mostly do the same thing in Ruby, I'm not so confident that I could do it "the Ruby way" as Ruby is a language I rarely touch. It would probably also take me four times as long to write, so not super eager to do it currently. :smile:

PR #9389, opened 2018-01-20
Add warning message for VirtualBox shared folders option.

chrisroberts:
Display warning to user about the VirtualBox SharedFoldersEnableSymlinksCreate
option with link to VirtualBox documentation and instructions on how the
setting can be disabled.


mwb5:
How does one disable this warning when SharedFoldersEnableSymlinksCreate must be enabled and we are using trusted boxes?

PR #9398, opened 2018-01-25
More explicit Solaris 11 and inherit SmartOS from Solaris

drscream:
SmartOS is a distribution of illumos which is the successor of OpenSolaris before it was closed down by Oracle. SmartOS as distribution is similar to OmniOS which is already inherit by solaris.

The problem here is that SmartOS and OmniOS are between Solaris 10 and Solaris 11 so the `uname -sr` already return 5.11 also if it's not the same as Oracle Solaris 11. For that reason we need an more explicit check on solaris11 guests, so checking the `/etc/release` file make more sense.

### DIFF
--- a/plugins/guests/solaris11/guest.rb
+++ b/plugins/guests/solaris11/guest.rb
@@ -8,7 +8,7 @@ module VagrantPlugins
   module GuestSolaris11
     class Guest < Vagrant.plugin("2", :guest)
       def detect?(machine)
-        machine.communicate.test("uname -sr | grep 'SunOS.*5\\.11'")
+        machine.communicate.test("grep 'Solaris 11' /etc/release")
       end
     end
   end
--- a/plugins/guests/smartos/plugin.rb
+++ b/plugins/guests/smartos/plugin.rb
@@ -11,7 +11,7 @@ module VagrantPlugins
         Config
       end
 
-      guest(:smartos)  do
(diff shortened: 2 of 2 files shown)


plugins/guests/solaris11/guest.rb
```diff
@@ -8,7 +8,7 @@ module VagrantPlugins
   module GuestSolaris11
     class Guest < Vagrant.plugin("2", :guest)
       def detect?(machine)
-        machine.communicate.test("uname -sr | grep 'SunOS.*5\\.11'")
```
chrisroberts: I am unsure what compatibility impacts this modification will create for other solaris guests. From first glance this seems like it may easily break guests that currently are in working order.

drscream: If they are real Solaris 11 guests it will not make any impact because the information is also provided in `/etc/release` which show the information that it is a real Solaris 11 which somehow is not true with `uname -sr` especially for SmartOS, OmniOS and other distributions.

I made this change especially to be sure vagrant detect the SmartOS machine as `smartos` guest and not as `solaris11` guest. First of all to have the valid `smartos` guest features enabled the second was that some commands used in the `solaris11` guest network configuration do not work on SmartOS (or OmniOS).

chrisroberts: @drscream :+1: thanks for the explanation!

PR #9404, opened 2018-01-26
Debian: Point hostname to 127.0.1.1 in /etc/hosts

### LINKED ISSUE #9403: Debian: Wrong IP address for hostname in /etc/hosts
### Vagrant version
* Experienced with all versions ``>1.8.1``, including ``2.0.1``
* Does NOT happen on ``1.8.1``

### Host operating system
Ubuntu 16.04, also experienced on Ubuntu 17.10

### Guest operating system
Debian jessie official image (``debian/jessie64``)

### Expected behavior
The hostname entry in /etc/hosts point to ``127.0.1.1``

### Actual behavior
The hostname entry in /etc/hosts point to ``127.0.0.1``

### Investigation

I traced the issue back to the following commit: https://github.com/hashicorp/vagrant/commit/dbb2d992785c7f8cf23bd04f1dbc1e4c1dbe41f4

It erroneously sets the hostname to ``127.0.0.1``. In Debian, ``127.0.0.1`` is supposed to point to ``localhost``, and the hostname itself points to ``127.0.1.1``

### Workaround

Fix the /etc/hosts file manually

### Fix

We are currently working on a fix for vagrant and will report back.

hakoerber:
Closes #9403

### DIFF
--- a/plugins/guests/debian/cap/change_host_name.rb
+++ b/plugins/guests/debian/cap/change_host_name.rb
@@ -18,7 +18,11 @@ module VagrantPlugins
 
               # Prepend ourselves to /etc/hosts
               grep -w '#{name}' /etc/hosts || {
-                sed -i'' '1i 127.0.0.1\\t#{name}\\t#{basename}' /etc/hosts
+                if grep -w '^127\\.0\\.1\\.1' /etc/hosts ; then
+                  sed -i'' 's/^127\\.0\\.1\\.1\\s.*$/127.0.1.1\\t#{name}\\t#{basename}/' /etc/hosts
+                else
+                  sed -i'' '1i 127.0.1.1\\t#{name}\\t#{basename}' /etc/hosts
+                fi
               }
 
               # Update mailname


hakoerber:
Well, there is no pattern to match here. The ``sed`` command simply prepends a line. Or do you mean something else?

briancain:
@hakoerber I think @chrisroberts was assuming that it updated the existing `127.0.1.1` entry in `/etc/hosts`, but the `sed` command actually just adds a new line entirely for the new hostname. Perhaps it should actually append to the existing entry, rather than a new line?

```
vagrant@hellothere:~$ cat /etc/hosts
127.0.1.1	hellothere	hellothere
127.0.0.1	localhost
127.0.1.1	jessie.vagrantup.com	jessie
...
...
```

briancain:
@hakoerber - ok, after talking it over I think instead we should be replacing the line `127.0.1.1` if it exists, with the passed in hostname set by the Vagrantfile rather than prepending it to the top of the hosts file. Let me know if that makes sense...thanks!

hakoerber:
Ah, now I get it.

So instead of two entries for ``127.0.1.1``, we only want one, right? And this entry should incorporate both the existing names (``jessie`` etc) and the one set by vagrant, right?

I will implement it as soon as possible!

briancain:
@hakoerber yes! But I think we can probably get rid of the existing names laid down by the box. I think we can assume if the user has explicitly set a hostname in the config, the other ones are probably not valid?

hakoerber:
Yep, that would make more sense. No point in having multiple boxes named ``jessie``.

Will try to get that done today.

plugins/guests/debian/cap/change_host_name.rb
```diff
@@ -18,7 +18,11 @@ def self.change_host_name(machine, name)
 
               # Prepend ourselves to /etc/hosts
               grep -w '#{name}' /etc/hosts || {
-                sed -i'' '1i 127.0.0.1\\t#{name}\\t#{basename}' /etc/hosts
+                if grep -w '^127\.0\.1\.1' /etc/hosts ; then
+                  sed -i'' 's/^127\.0\.1\.1\s/127.0.0.1\\t#{name}\\t#{basename}/' /etc/hosts
```
briancain: I don't think this `sed` command actually works...I couldn't get it to work anyway. I think it should actually be:

```ruby
sed -i 's/^127\.0\.1\.1.*/127.0.1.1\\t#{name}\t#{basename}/g' /etc/hosts
```

hakoerber:
I'm sorry, I pushed untested stuff yesterday. Didn't pay attention. I will fix the ``sed`` command and test it (as soon as I get VirtualBox running ...)

hakoerber:
So, I now updated the merge request. I tested it with the following, simple ``Vagrantfile``:

```ruby
  config.vm.provider 'virtualbox'

  config.vm.define "test" do |main|
    main.vm.box = 'debian/jessie64'
    main.vm.hostname = "test.domain"
  end
end
```

Before (current ``master``), a ``vagrant up`` resulted in the following:

```bash
$ vagrant up && vagrant ssh -- cat /etc/hosts
[...]
127.0.0.1       test.domain     test
127.0.0.1       localhost
127.0.1.1       jessie.vagrantup.com    jessie
[...]
```

With the changes, the result is different:

```bash
$ vagrant up && vagrant ssh -- cat /etc/hosts
[...]
127.0.0.1       localhost                       
127.0.1.1       test.domain     test
[...]
```

I tested a few things:

* *Changing an existing entry*: Created a bogus entry for ``127.0.1.1`` in an existing box, halting it, and then running ``vagrant up`` again. The entry is replaced, as desired.
* *Non-existent entry for 127.0.1.1*: If there is no entry for ``127.0.1.1`` at all, a new one is prepended to the file, as before.
* *Rebooting the machine*: ``/etc/hosts`` is not touched after the first boot, as desired.

@briancain - I changed your ``sed`` command a little bit. It would have matched something like ``127.0.1.100`` also. I had to fight with ruby about escaping the regex ``\s`` escape sequence 😄

briancain:
Thanks @hakoerber ! Looks great. I tested it as well and it looks to behave like I would expect. Thanks for your contribution!

PR #9405, opened 2018-01-26
Debian: Renew DHCP lease on hostname change

### LINKED ISSUE #9402: Debian: Network issues regarding DNS in multi-machine setups
### Vagrant version
* Experienced with all versions ``>1.8.1``, including ``2.0.1``
* Does NOT happen on ``1.8.1``

### Host operating system
Ubuntu 16.04, also experienced on Ubuntu 17.10

### Guest operating system
Debian jessie official image (``debian/jessie64``), same issue with stretch (``debian/stretch64``)

### Expected behavior
The machines in a multi-machine setup can reach each other by their DNS names.

### Actual behavior
Name resolution for other machines in the multi-machine setup fails. External DNS resolution is unaffected.

### Investigation

I traced the issue back to the following commit: https://github.com/hashicorp/vagrant/commit/dbb2d992785c7f8cf23bd04f1dbc1e4c1dbe41f4

The problem lies in the DNS resolution by the host-side dnsmasq. Because the vagrant machines do not renew their DHCP lease after their hostname is set, the machines cannot find each other using that hostname. When looking into the host-side dnsmasq lease file, I can see the machines showing up as the generic "jessie".

Before, vagrant ran ``ifdown -a;ifup -a`` on the machine. The commit changes that to ``invoke-rc.d networking force-reload``. This script explicitly says that "[it] is deprecated because it may not re-enable some interfaces" (see ``/etc/init.d/networking``). If i manually run ``ifdown -a;ifup -a`` in the machine, the name resolution for other machines works afterwards. This means that the new command is not enough to force the machine to renew its DHCP lease.

When I look at the host-side dnsmasq lease file, I see that the machines do not renew their lease after getting a hostname, and stay at "jessie". As soon as they get a new lease, the name is changed to the configured hostname.

### Workaround

To work around the issue, you can force a DHCP renew via a shell provisioner:

```
main.vm.provision "shell", inline: "ifdown -a;ifup -a"
```

### Fix

We are currently working on a fix for vagrant and will report back.

hakoerber:
Closes #9402

### DIFF
--- a/plugins/guests/debian/cap/change_host_name.rb
+++ b/plugins/guests/debian/cap/change_host_name.rb
@@ -32,6 +32,10 @@ module VagrantPlugins
               if test -f /etc/init.d/hostname.sh; then
                 /etc/init.d/hostname.sh start || true
               fi
+              if test -x /sbin/dhclient ; then
+                /sbin/dhclient -r
+                /sbin/dhclient -nw
+              fi
             EOH
           end
         end


hakoerber:
Good idea, I updated the merge request accordingly!

PR #9422, opened 2018-01-31
Use 127.0.0.1 instead of localhost

pvalena:
in test is_port_open_test.rb


pvalena:
This occured in Fedora Rawhide:

```
DEBUG: Failures:
DEBUG:   1) Vagrant::Util::IsPortOpen should report open ports
DEBUG:      Failure/Error: expect(klass.is_port_open?("localhost", open_port)).to be
DEBUG:        expected false to evaluate to true
DEBUG:      # ./test/unit/vagrant/util/is_port_open_test.rb:39:in `block (2 levels) in <top (required)>'
```

Cause of that is that `localhost` was not resolvable, which is fixed by this PR.

PR #9456, opened 2018-02-08
Exit if Hyper-V is enabled and VirtualBox provider is used

### LINKED ISSUE #9258: Hyper-V Prevents Vagrant from Running and Causes BSOD
Using Vagrant with Hyper-V enabled causes a blue screen of death (BSOD).

The only updates on the computer were to Vagrant 2.0.1, VirtualBox 5.2 from 5.1.30, and Windows 10 Pro  from Windows 10 for Hyper-V in order to use `Docker for Windows` since we use both `Vagrant` and `Docker` in development

The only way I was able to get it to work with Vagrant was to drop down to VirtualBox 5.1.30 from 5.2.x and turning off Hyper-V which requires shutting down and restarting, but since we use Docker as well that creates a conflict since in order to use `Docker for Windows` it requires Hyper-V which includes shutting down and restarting.

### Vagrant version
Vagrant 2.0.1

### Host operating system
Windows 10 Pro

### Guest operating system
ubuntu/trusty64

### Vagrantfile

```ruby
end
```

### Debug output
There is no debug statement that gets displayed it just shows the blue screen of death (BSOD) with a stop code of SYSTEM SERVICE EXCEPTION and then it restarts.

### Expected behavior
Running `vagrant up` starts up the box ready for development when Hyper-V is enabled.

### Actual behavior
What actually happened?
Using `vagrant up` begins starting up the box, but after running through a couple lines of setup you get the Windows BSOD.

### Steps to reproduce
1. `vagrant up`

briancain:
This commit adds a function on windows for the VirtualBox provider to
check if Hyper-V is enabled. If so, exit and display a warning to the
user that going forward will result in a system crash.

Fixes #9258

### DIFF
--- a/lib/vagrant/errors.rb
+++ b/lib/vagrant/errors.rb
@@ -372,6 +372,10 @@ module Vagrant
       error_key(:home_dir_unknown_version)
     end
 
+    class HypervVirtualBoxError < VagrantError
+      error_key(:hyperv_virtualbox_error)
+    end
+
     class ForwardPortAdapterNotFound < VagrantError
       error_key(:forward_port_adapter_not_found)
     end
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -97,6 +97,22 @@ module Vagrant
           return @_windows_hyperv_admin
         end
 
+        # Checks if Hyper-V is enabled on the host system and returns true
+        # if enabled.
+        #
+        # @return [Boolean]
+        def windows_hyperv_enabled?
+          return @_windows_hyperv_enabled if defined?(@_windows_hyperv_enabled)
+
+          @_windows_hyperv_enabled = -> {
+            ps_cmd = "$(Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-All -Online).State"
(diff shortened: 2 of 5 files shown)


PatrickLang:
Submitted a PR ^^

Nillth:
This changed unfortunately prevents Vagrant from working in the scenario where there are two entries in the BCD.  (1 with Hyper-V Hyervisor enabled "hypervisorlaunchtype Auto"), 1 with Hyper-V Hypervisor Disabled "hypervisorlaunchtype Off")
By configuring the system this way, with a single reboot you are able to switch between Hyper-V and Virtual Box, as opposed to multiple to uninstall and reinstall.
 
The Current check only looks to see if Hyper-V is installed, not if it is running:
Changing:
`ps_cmd = "$(Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-All -Online).State"`
to
`ps_cmd = "(gcim Win32_ComputerSystem).HypervisorPresent"`
should do the trick
vagranthypervisorcheck

briancain:
Hi @Nillth - please read https://github.com/hashicorp/vagrant/pull/9471 for more information. That change was proposed, however doing so would prevent users from running Vagrant on windows inside a virtual machine as well.

PR #9459, opened 2018-02-08
Update systemd? check for guest inspection

### LINKED ISSUE #9428: Debian guest capability: configure networks
There still seem to be some remaining issues to be resolved identified in #9134 and initially resolved with #9338.

A mix of ubuntu and debian boxes were used for testing the initial PR, so remaining issues need to be identified with specific boxes that display the error states. Conflicting reports on `bento/ubuntu-17.10` (which is a box I tested locally). Error state reported on `ubuntu/artful64` (https://github.com/hashicorp/vagrant/issues/9134#issuecomment-361584017)

briancain:
Prior to this commit, the regex matcher used with grep to determine if a
system was using systemd? was failing on systems that did not exactly
match the old regex. This commit updates that communications test to use
a different method of determining if systemd is in use with the ps
command.

Fixes #9428

### DIFF
--- a/lib/vagrant/util/guest_inspection.rb
+++ b/lib/vagrant/util/guest_inspection.rb
@@ -12,7 +12,7 @@ module Vagrant
         #
         # @return [Boolean]
         def systemd?(comm)
-          comm.test("systemctl | grep '^-\.mount'")
+          comm.test("ps -o comm= 1 | grep systemd")
         end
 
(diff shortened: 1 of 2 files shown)


ljakab:
Is this going into 2.0.3 ?

briancain:
@ljakab yes, that's the plan.

PR #9470, opened 2018-02-14
Pass ruby block to capture stdout when determining PS version

### LINKED ISSUE #9056: "vagrant up" hangs and does not detect outdated PowerShell on Windows 7
### Vagrant version
Vagrant 2.0.0

### Host operating system
64-bit Windows 7 Professional Service Pack 1

### Guest operating system
hashicorp/precise64

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# you're doing.
  # The most common configuration options are documented and commented below.
  # For a complete reference, please see the online documentation at
  # https://docs.vagrantup.com.

  # Every Vagrant development environment requires a box. You can search for
  # boxes at https://vagrantcloud.com/search.
  config.vm.box = "hashicorp/precise64"

  # Disable automatic box update checking. If you disable this, then
  # boxes will only be checked for updates when the user runs
  # `vagrant box outdated`. This is not recommended.
  # config.vm.box_check_update = false

  # Create a forwarded port mapping which allows access to a specific port
  # within the machine from a port on the host machine. In the example below,
  # accessing "localhost:8080" will access port 80 on the guest machine.
  # NOTE: This will enable public access to the opened port
  # config.vm.network "forwarded_port", guest: 80, host: 8080

  # Create a forwarded port mapping which allows access to a specific port
  # within the machine from a port on the host machine and only allow access
  # via 127.0.0.1 to disable public access
  # config.vm.network "forwarded_port", guest: 80, host: 8080, host_ip: "127.0.0.1"

  # Create a private network, which allows host-only access to the machine
  # using a specific IP.
  # config.vm.network "private_network", ip: "192.168.33.10"

  # Create a public network, which generally matched to bridged network.
  # Bridged networks make the machine appear as another physical device on
  # your network.
  # config.vm.network "public_network"

  # Share an additional folder to the guest VM. The first argument is
  # the path on the host to the actual folder. The second argument is
  # the path on the guest to mount the folder. And the optional third
  # argument is a set of non-required options.
  # config.vm.synced_folder "../data", "/vagrant_data"

  # Provider-specific configuration so you can fine-tune various
  # backing providers for Vagrant. These expose provider-specific options.
  # Example for VirtualBox:
  #
  # config.vm.provider "virtualbox" do |vb|
  #   # Display the VirtualBox GUI when booting the machine
  #   vb.gui = true
  #
  #   # Customize the amount of memory on the VM:
  #   vb.memory = "1024"
  # end
  #
  # View the documentation for the provider you are using for more
  # information on available options.

  # Enable provisioning with a shell script. Additional provisioners such as
  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
  # documentation for more information about their specific syntax and use.
  # config.vm.provision "shell", inline: <<-SHELL
  #   apt-get update
  #   apt-get install -y apache2
  # SHELL
end
```

### Debug output
https://gist.github.com/interfrastic/b3e23db685271c6eccf7f8696d9f6be4

### Expected behavior
When I ran the "vagrant up" command, the code in PR #8839 (the fix for #8783) should have determined that my PowerShell version (2.0) is lower than the minimum required version (3.0) and raised an error.

### Actual behavior
The "vagrant up" command hung with no output for 16 hours before I finally interrupted it.

### Steps to reproduce
1. Install VirtualBox.
2. Install Vagrant.
3. Open a Command Prompt window and enter the following commands:

    cd/d %TMP%
    mkdir vagrant-test
    cd vagrant-test
    vagrant init hashicorp/precise64
    vagrant up

### References
- GH-8783

briancain:
Prior to this commit, the function used to determine the version of
Powershell would loop forever inside the Subprocess.execute function
because the process would never exit. This commit fixes that by passing
in a ruby block to capture the version from stdout instead of trying to
capture it from the returned process when it exits.

Fixes #9056

### DIFF
--- a/lib/vagrant/util/powershell.rb
+++ b/lib/vagrant/util/powershell.rb
@@ -11,6 +11,7 @@ module Vagrant
     class PowerShell
       # NOTE: Version checks are only on Major
       MINIMUM_REQUIRED_VERSION = 3
+      LOGGER = Log4r::Logger.new("vagrant::util::powershell")
 
       # @return [Boolean] powershell executable available on PATH
       def self.available?
@@ -85,8 +86,14 @@ module Vagrant
             "Write-Output $PSVersionTable.PSVersion.Major"
           ].flatten
 
-          r = Subprocess.execute(*command)
-          @_powershell_version = r.exit_code != 0 ? nil : r.stdout.chomp
+          version = nil
+          begin
+            r = Subprocess.execute(*command, notify: [:stdout, :stderr], timeout: 10) {|io_name,data| version = data}
+          rescue Vagrant::Util::Subprocess::TimeoutExceeded
+            LOGGER.debug("Timeout exceeded while attempting to determine version of Powershell.")
+          end
+
+          @_powershell_version = version
         end
         @_powershell_version
       end


Occams:
This change is causing issues in our CI environment. It seems like the systems routinely run into the newly defined timeout (10s) while determining the installed version of powershell. This occurs more frequently if the environment experiences heavy load. I try to capture some debug output...

PR #9471, opened 2018-02-15
Updating check for Hyper-V

PatrickLang:
This is an updated check to tell if a hypervisor (including Hyper-V) is running on Windows. Virtualbox uses this check to avoid crashing (See #9258). 

This fix covers a few more corner cases than the previous one in #9456:

* Allow virtualbox to run if Hyper-V is installed, but currently disabled. If you want to switch a machine between using Virtualbox and Hyper-V, you can do it with another boot entry - [Link](https://blogs.msdn.microsoft.com/virtual_pc_guy/2008/04/14/creating-a-no-hypervisor-boot-entry/)
* Works on Windows Server & Windows 10. Windows Server uses `Get-WindowsFeature` to check for install, while `Get-WindowsOptionalFeature` is used on Windows 10 ¯\\_(ツ)_/¯ 
* There are two other features that use the hypervisor, so it's possible to have it loaded but Microsoft-HyperV-All isn't installed: Windows-Defender-ApplicationGuard, and [DeviceGuard](https://docs.microsoft.com/en-us/windows/security/threat-protection/device-guard/deploy-device-guard-enable-virtualization-based-security) which is controlled through group policy

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -105,9 +105,9 @@ module Vagrant
           return @_windows_hyperv_enabled if defined?(@_windows_hyperv_enabled)
 
           @_windows_hyperv_enabled = -> {
-            ps_cmd = "$(Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-All -Online).State"
+            ps_cmd = "$(Get-ComputerInfo).HypervisorPresent"
             output = Vagrant::Util::PowerShell.execute_cmd(ps_cmd)
-            return output == 'Enabled'
+            return output == 'True'
           }.call
 
           return @_windows_hyperv_enabled
--- a/test/unit/vagrant/util/platform_test.rb
+++ b/test/unit/vagrant/util/platform_test.rb
@@ -217,14 +217,14 @@ describe Vagrant::Util::Platform do
 
   describe ".windows_hyperv_enabled?" do
     it "should return true if enabled" do
-      allow(Vagrant::Util::PowerShell).to receive(:execute_cmd).and_return('Enabled')
+      allow(Vagrant::Util::PowerShell).to receive(:execute_cmd).and_return('True')
 
       expect(Vagrant::Util::Platform.windows_hyperv_enabled?).to be_truthy
     end
 
     it "should return false if disabled" do
       Vagrant::Util::Platform.reset!
-      allow(Vagrant::Util::PowerShell).to receive(:execute_cmd).and_return('Disabled')
+      allow(Vagrant::Util::PowerShell).to receive(:execute_cmd).and_return('False')
 
       expect(Vagrant::Util::Platform.windows_hyperv_enabled?).to be_falsey
     end


PatrickLang:
Oops sorry about that. Need to read up on Ruby unit testing. I think the commit I added should fix it.

briancain:
@PatrickLang - Thanks for the update! A couple of questions...


Is there a better check for seeing if Hyper-V is "installed" but disabled? Also do you have an example of how to get to that state in windows 10? Every bit of documentation I see says that it's either enabled or disabled, but I see no language around it being installed but "off". I ask because this command:

```
$(Get-ComputerInfo).HypervisorPresent
```

Won't work for this because it returns true if you're running vagrant on Windows inside of a virtual machine. :(

We will however need some sort of check to ensure that this is only checked against windows 10 (maybe windows 8 works?), because these commands fail and do not exist on older systems like windows 7.

briancain:
Edit on this: I just tried this on windows 7. We will be fine without a OS check, as the `Powershell.execute_cmd` function returns `nil` if a command fails (i.e. doesn't exist).

PatrickLang:
Here's the corner cases I can think of:

Windows 7 - check should always fail. Hyper-V not present, and check for it isn't there
Windows 8 / Windows Server 2012 - check should work. If Hyper-V is present, the check should do the right thing.

One false positive would be:
Windows 8+ or Windows Server 2012+ running under VMWare, KVM or Xen. Those will also return Hypervisor present. If Virtualbox works nested under those hypervisors, then we should switch to a per-hypervisor check which will be much more difficult. You'd have to look at what devices are present to guess what hypervisor you're under and of course hypervisors can lie :/

briancain:
So yesterday I did some tests. We should be good without a "check" for which OS we're on, because if the powershell command fails, [its helper function in Vagrant will just return nil](https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/util/powershell.rb#L69) so we won't have to worry about an uncaught exception for a command that doesn't exist on Windows 7. The original check will just be false, and we should be good to continue on using the virtualbox provider.

Finally that being said, I'm still pretty confident that the check in place is enough to determine if Hyper-V is running...but I might be missing something.

```powershell
$(Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-All -Online).State
```

This returns either `Enabled` or `Disabled`. I'm assuming if it's `Enabled`, Hyper-V is running, and therefore Vagrant needs to bail out before it brings up a virtual machine in virtualbox and blue screens the machine. Are you saying this is not the case? (i.e. the command returns `Disabled`, but some how it's running?)

Thanks!

Quix0r:
Conflicts found, can you please fix them? Is still crashing with version `2.1.1` (docker installed, `vagrant up` gives BSOD).

briancain:
Hi there, this check was improved via a997a27b579d9203e986cfb6d3735e4ad9ae8886 which includes the ability to check on Windows Sever. I'm going to close this pull request now. Thanks!

Quix0r:
Nice work.

PR #9483, opened 2018-02-17
fix defaults for vagrant ssh -c on windows guests

mcandre:
Use more appropriate shell defaults when sending one-off user commands over `vagrant ssh -c ...` to Windows guests.

The current source hardcodes `bash -l`, does not allow users to override this with downstream configurations, and also hardcodes `-c` (POSIX style) flag and single (`'`) quotes, which Windows' Command Prompt does not handle properly.

For the short term, users just want `vagrant ssh -c ...` to work for Windows guests. The current source mangles the command expansion in a way that breaks if a bash.exe is not carefully setup in the Windows box. This patch corrects this behavior, using a more appropriate expansion.

In the long run, we will want all of these things to be more customizable in order to handle exotic environments (e.g. the user actually wants `vagrant ssh [-c]` to use a bash shell, or zsh, or fish, or ion, or whatever. The shell name and flag should ideally be configurable by the normal `config.vm.ssh.shell` variable (it is not!), and an auxilliary `config.vm.ssh.shell.flags` variable (TBD). The wrapping quotes should also be customizable with `config.vm.ssh.quote` (TBD) for maximum flexibility to suit the user's desired environment. For now, let's just get Windows guests working better.

### DIFF
--- a/lib/vagrant/action/builtin/ssh_run.rb
+++ b/lib/vagrant/action/builtin/ssh_run.rb
@@ -35,8 +35,13 @@ module Vagrant
           end
 
           # Get the command and wrap it in a login shell
-          command = ShellQuote.escape(env[:ssh_run_command], "'")
-          command = "#{env[:machine].config.ssh.shell} -c '#{command}'"
+          if env[:machine].guest.name == :windows
+            command = ShellQuote.escape(env[:ssh_run_command], "\"")
+            command = "cmd /c \"#{command}\""
+          else
+            command = ShellQuote.escape(env[:ssh_run_command], "'")
+            command = "#{env[:machine].config.ssh.shell} -c '#{command}'"
+          end
 
           # Execute!
           opts = env[:ssh_opts] || {}


mcandre:
Ah, crud. Looks like the environment variable is not featuring a `guest` accessor. Either the test objects are missing it, or I'm using the wrong accessor. Not really sure how to query the guest type from the environment object, I figured it was this way based on some of the rest of the Vagrant code...

mcandre:
> You can already configure what shell Vagrant should use with the config.ssh.shell option

False. As the documentation states,

"Note that this has no effect on the shell you get when you run vagrant ssh. This configuration option only affects the shell to use when executing commands internally in Vagrant."

https://www.vagrantup.com/docs/vagrantfile/ssh_settings.html

In fact this value is hardcoded to `bash -l` in the Vagrant source code.

chrisroberts:
Hi there. While the shell value is not actually hardcoded the configuration of the generated command within the ssh run action does assume a borne style shell with a `-c` option available for providing the command. This definitely _should_ be updated but the way in which this PR is currently implemented will likely cause existing, working, setups to break. There is a very large variation on how Windows guests are configured and behave and this PR acts too heavy handed when determining what underlying process should be used for running the command. There are multiple different ways to provide the SSH daemon on Windows, and up until recently they have generally included basic posix compatible tooling as well. 

Ideally I think it will be best to add configuration options for how the command is constructed, and to provide sane defaults depending on the communicator that is being used (where the defaults are different when using the ssh communicator rather than the winssh communicator). This will prevent existing setups from encountering breaking changes, and provide better support for users using the winssh communicator.

As a stop gap method when using the winssh communicator, the `vagrant ssh` command can do the "right" thing by using `--`:

```shell
$ vagrant ssh -- dir
```

I'm going to close this PR for the reasons stated above, but will open a new PR shortly to resolve the issue using a mix of new configuration options and defaults based on communicator detection.

Thanks again for the PR and bringing this to our attention.

Cheers!

mcandre:
What are the semantics of `vagrant ssh -- ...` supposed to be?

`vagrant ssh -- dir` executes `dir` in a Command Prompt shell.

`vagrant ssh -- ls` complains of a missing ls command.

`vagrant ssh -- which link` outputs /usr/bin/link, which is strange as both which and /usr/bin/link are cygwin-isms.

`vagrant ssh -- echo %PATH%` outputs a literal "%PATH%", not the contents of the %PATH% Command Prompt environment variable.

`vagrant ssh -- echo \$Env:PATH` outputs a literal "$Env:PATH", not the contents of the PATH powershell environment variable.

`vagrant ssh -- echo \$PATH` outputs a literal "$PATH", not the contents of the $PATH bash environment variable.

`vagrant ssh -- env` output includes some Windows-style environment variables, but also "HOME=/c/Users/vagrant".

I'm confused how users are expected to leverage the -- syntax here. It doesn't seem to work for executing reasonably basic commands, in any shell.

PR #9493, opened 2018-02-20
Update configuration.html.md

carcarlo:
the object v is unrecognized.
I've setted it to 'vb'


briancain:
Hello @carcarlo - Thank you for the pull request! However the change is incorrect. `v` comes from the ruby block on the line above:

```ruby
```

Changing it to `vb` would result in an error, as vb in this case is not defined.

Thank you though!

PR #9502, opened 2018-02-22
Use provider override for box_version if given

### LINKED ISSUE #9452: Existing box reused even when box_version constraint not met
### Vagrant version
Vagrant 2.0.2

### Host operating system
Windows 10 x64 1709

### Guest operating system
bento/ubuntu-16.04

### Vagrantfile
```ruby
  config.vm.box = "bento/ubuntu-16.04"

  config.vm.provider "virtualbox" do |vb, override|
    override.vm.box_version = ">= 201802.02.0"
  end
end
```

### Debug output
https://emorley.pastebin.mozilla.org/9077485

### Expected behavior
* The already-downloaded out of date box version (`201801.02.0`) should not be used, since it doesn't match the `box_version` constraint.
* Instead the new version (`201802.02.0`) should be downloaded.

### Actual behavior
* The already downloaded older version that doesn't match the `box_version` is used.
* This appears to work as expected, if `box` and `box_version` are both specified within the same block. 

ie either of these work fine:
```ruby
  config.vm.box = "bento/ubuntu-16.04"
  config.vm.box_version = ">= 201802.02.0"
end
```

```ruby
  config.vm.provider "virtualbox" do |vb, override|
    override.vm.box = "bento/ubuntu-16.04"
    override.vm.box_version = ">= 201802.02.0"
  end
end
```

For context - the reason we specify the version separately is that the hyper-v box isn't available in as new a version, so we can't use the same `box_version` for both virtualbox and hyper-v.

### Steps to reproduce
1. Install vagrant 2.0.2 and virtualbox 5
2. `vagrant box add bento/ubuntu-16.04 --provider virtualbox --box-version 201801.02.0`
3. Ensure no existing machine exists (eg `vagrant destroy -f`)
3. `vagrant up`

### References

I'm guessing the problem is the lack of check relating to `box_version`, here:
https://github.com/hashicorp/vagrant/blob/9c299a2a357fcf87f356bb9d56e18a037a53d138/lib/vagrant/vagrantfile.rb#L190-L191

briancain:
Prior to this commit if an override was given for a box version, it
would be ignored. This commit ensures that the box_version override is
used when specified in a Vagrantfile.

Fixes #9452

### DIFF
--- a/lib/vagrant/vagrantfile.rb
+++ b/lib/vagrant/vagrantfile.rb
@@ -155,6 +155,7 @@ module Vagrant
       # Track the original box so we know if we changed
       box = nil
       original_box = config.vm.box
+      original_version = config.vm.box_version
 
       # The proc below loads the box and provider overrides. This is
       # in a proc because it may have to recurse if the provider override
@@ -188,10 +189,11 @@ module Vagrant
         end
 
         # If the box changed, then we need to reload
-        if original_box != config.vm.box
+        if original_box != config.vm.box || original_version != config.vm.box_version
           # TODO: infinite loop protection?
 
           original_box = config.vm.box
+          original_version = config.vm.box_version
           load_box_proc.call
         end
       end
--- a/test/unit/vagrant/vagrantfile_test.rb
+++ b/test/unit/vagrant/vagrantfile_test.rb
@@ -288,9 +288,11 @@ describe Vagrant::Vagrantfile do
 
       configure do |config|
         config.vm.box = "base"
+        config.vm.box_version = "1.0"
 
         config.vm.provider "foo" do |_, c|
           c.vm.box = "foobox"
+          c.vm.box_version = "2.0"
         end
       end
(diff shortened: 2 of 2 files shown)


lib/vagrant/vagrantfile.rb
```diff
@@ -194,6 +195,11 @@ def machine_config(name, provider, boxes)
           original_box = config.vm.box
           load_box_proc.call
         end
+
+        if original_version != config.vm.box_version
+          original_version = config.vm.box_version
+          load_box_proc.call
+        end
```
chrisroberts: Lets make this a single conditional instead of broken into two:

```ruby
if original_box != config.vm.box || original_version != config.vm.box_version
  original_box = config.vm.box
  original_version = config.vm.box_version
  load_box_proc.call
end
```

edmorley:
Thank you for fixing this @briancain :-)

PR #9507, opened 2018-02-23
Fix Hyper-v and VirtualBox check

briancain:
Prior to this commit, the hyper-v and virtualbox system crash check
existed within the initialize function of the virtualbox provider. That
caused an issue when running with other providers, because the
virtualbox provider still gets initialized even if not used. This commit
changes that by placing the check inside of one of the virtualbox
provider actions that checks if virtualbox is installed and ready to
use. This action is action is used by the main vbox provider actions,
and should not be called when other providers are being used with
Vagrant.

### DIFF
--- a/plugins/providers/virtualbox/action/check_virtualbox.rb
+++ b/plugins/providers/virtualbox/action/check_virtualbox.rb
@@ -1,3 +1,5 @@
+require 'vagrant/util/platform'
+
 module VagrantPlugins
   module ProviderVirtualBox
     module Action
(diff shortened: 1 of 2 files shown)


briancain:
An example of the problem:

```ruby
  (1..3).each do |i|
    config.vm.define "docker-#{i}"  do |vm|
      vm.vm.provider "docker" do |d|
        d.image = "ubuntu"
      end
    end
  end
end
```

Running `vagrant status` on this would cause the error message to be shown. In reality we just want this check to be displayed if vagrant attempts to do anything with VirtualBox around bringing up virtual machines (which is what causes the crash)

PR #9525, opened 2018-02-28
Support updated WSL implementation

chrisroberts:
This is an extension of great work provided by @BR0kEN- in #9300. Updated to support the old lxrun generated WSL instances as well as the new store installed instances. Included some spec coverage on the behavior as well.

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -1,6 +1,8 @@
 require "rbconfig"
 require "shellwords"
+require "tempfile"
 require "tmpdir"
+require "log4r"
 
 require "vagrant/util/subprocess"
 require "vagrant/util/powershell"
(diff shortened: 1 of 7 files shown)


lib/vagrant/util/platform.rb
```diff
@@ -291,18 +293,105 @@ def wsl_path?(path)
...
+            logger = Log4r::Logger.new("vagrant::util::platform::wsl")
+
+            # Check for lxrun installation first
+            lxrun_path = [wsl_windows_appdata_local, "lxss"].join("\\")
+            paths = [lxrun_path]
+
+            logger.debug("checking registry for WSL installation path")
+            paths += PowerShell.execute_cmd(
+              '(Get-ChildItem HKCU:\Software\Microsoft\Windows\CurrentVersion\Lxss ' \
+                '| ForEach-Object {Get-ItemProperty $_.PSPath}).BasePath').to_s.split("\r\n").map(&:strip)
```
briancain: Should we put a rescue in place here, in case the `execute_cmd` fails and returns nil? Otherwise we'll get an exception that `nil.map` is undefined. We could also run the command first before running map on it and adding it to paths and check the result of the Powershell command.

chrisroberts: The `#to_s` call will force the type so there's no need to rescue here.

PR #9544, opened 2018-03-08
Further attempt to get the IP address of HyperV guests

tuxillo:
For guests that don't have support for Hyper-V Integration Services, attempt to get the IP address from the neighbor cache entries.

Of course this won't always work, it requires that the host and the guest communicate with each other before, but at least gives a chance to some guest OSes to work out of the box. Only for IPv4 addresses for now.

### DIFF
--- a/plugins/providers/hyperv/scripts/get_network_config.ps1
+++ b/plugins/providers/hyperv/scripts/get_network_config.ps1
@@ -22,6 +22,20 @@ foreach ($network in $networks) {
         break
       }
     }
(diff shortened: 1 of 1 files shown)


chrisroberts:
Hi there. Thank you for submitting this PR. Large sections of the Hyper-V provider has been rewritten and is in the 2.1.2 release. It includes a lookup via MAC address very similar to this, so I am going to close this up. Again thank you very much for taking the time to make this PR.

Cheers!

PR #9578, opened 2018-03-15
Fix graceful_halt_timeout issue #8486

### LINKED ISSUE #8486: changing the a windows hostname takes forever when graceful_halt_timeout is set
### Vagrant version
1.9.2

### Host operating system
Windows 10

### Guest operating system
Windows Server 2016

### Vagrantfile
```ruby
    config.vm.box = "windows-2016-amd64"
    config.vm.define "windows-domain-controller"
    config.vm.hostname = "dc"
    config.vm.graceful_halt_timeout = 60*60 # 60 minutes.
    ...
```

### Debug output
It keeps stuck at:
```
INFO subprocess: Command not in installer, restoring original environment...                                                                                            
INFO winrmshell: Attempting to connect to WinRM...                                                                                                                      
INFO winrmshell:   - Host: 127.0.0.1                                                                                                                                    
INFO winrmshell:   - Port: 2200                                                                                                                                         
INFO winrmshell:   - Username: vagrant                                                                                                                                  
INFO winrmshell:   - Transport: plaintext                                                                                                                               
INFO winrm: WinRM is ready!                                                                                                                                             
INFO guest: Execute capability: change_host_name [#<Vagrant::Machine: windows-domain-controller (VagrantPlugins::ProviderVirtualBox::Provider)>, "dc"] (windows)        
```

### Expected behavior
The hostname change should be quick. In fact, when I login into the machine, the hostname is already changed; so, it seems that only an explicit reboot is needed.

### Actual behavior
It takes forever to change.

ajpaul25:
The current code waits for the entire value of machine.config.vm.graceful_halt_timeout when rebooting due to a hostname change.  This is not the intended use of this configuration parameter.  Graceful_halt_timeout is the amount of time to wait for a machine to respond to a graceful halt request before forcefully halting.  This change removes machine.config.vm.graceful_halt_timeout for the module in favor of using the existing wait_for_reboot code, thereby correcting issue #8486.

Fixes #8486

### DIFF
--- a/plugins/guests/windows/cap/change_host_name.rb
+++ b/plugins/guests/windows/cap/change_host_name.rb
@@ -4,10 +4,10 @@ module VagrantPlugins
       module ChangeHostName
 
         def self.change_host_name(machine, name)
-          change_host_name_and_wait(machine, name)
+          change_host_name_and_wait(machine, name, machine.config.vm.graceful_halt_timeout)
         end
 
-        def self.change_host_name_and_wait(machine, name)
+        def self.change_host_name_and_wait(machine, name, sleep_timeout)
           # If the configured name matches the current name, then bail
           # We cannot use %ComputerName% because it truncates at 15 chars
           return if machine.communicate.test("if ([System.Net.Dns]::GetHostName() -eq '#{name}') { exit 0 } exit 1")
@@ -30,6 +30,9 @@ module VagrantPlugins
           # Don't continue until the machine has shutdown and rebooted
 	  if machine.guest.capability?(:wait_for_reboot)
             machine.guest.capability(:wait_for_reboot)
+          else
+            # use graceful_halt_timeout only if guest cannot wait for reboot
+            sleep(sleep_timeout)
 	  end
         end
       end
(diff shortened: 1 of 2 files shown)


plugins/guests/windows/cap/change_host_name.rb
```diff
@@ -28,7 +28,9 @@ def self.change_host_name_and_wait(machine, name, sleep_timeout)
             error_key: :rename_computer_failed)
 
           # Don't continue until the machine has shutdown and rebooted
-          sleep(sleep_timeout)
+	  if machine.guest.capability?(:wait_for_reboot)
+            machine.guest.capability(:wait_for_reboot)
+	  end
```
briancain: We should have an else clause here and just sleep. Perhaps leave the `sleep_timeout` param passed in and use that to fall back on?

plugins/guests/windows/cap/change_host_name.rb
```diff
@@ -28,7 +28,9 @@ def self.change_host_name_and_wait(machine, name, sleep_timeout)
             error_key: :rename_computer_failed)
 
           # Don't continue until the machine has shutdown and rebooted
-          sleep(sleep_timeout)
+	  if machine.guest.capability?(:wait_for_reboot)
+            machine.guest.capability(:wait_for_reboot)
```
briancain: These two lines are the reasons for your rspec failures. You need to `allow` the machine double to be able to receive both `guest`, and the two functions `capability?` and `capability`.

ajpaul25:
Thanks!  I've made the code changes, tested with rake and pushed.  Hopefully travis will be happy too.

PR #9605, opened 2018-03-22
Remove name from command option

amwill04:
`vagrant up` does not accept the name of a box as an argument. It only accepts `id` as per the error message when running with `name`.

### DIFF
--- a/website/source/docs/cli/up.html.md
+++ b/website/source/docs/cli/up.html.md
@@ -9,7 +9,7 @@ description: |-
 
 # Up
 
-**Command: `vagrant up [name|id]`**
+**Command: `vagrant up [id]`**
 
 This command creates and configures guest machines according to your
 [Vagrantfile](/docs/vagrantfile/).


briancain:
Hi @amwill04 - Thanks for your PR, however `vagrant up` does accept name (i.e. the name of the guest). `name` in this case is the name of the defined guest in your Vagrantfile.

```ruby
  config.vm.define "mymachine" do |m|
    m.vm.box = "box"
  end
end
```

```
vagrant up mymachine
```

Thanks!

amwill04:
Thanks, I understand. I think for me when I was reading the documentation to use the command outside the root directory there is a difference - when using `id` you can run `vagrant up boxid` from any directory.

Running `vagrant up mymachine` from anywhere other than the directory hosting the `vagrantfile` brings the following error:

```
 Vagrant environment or target machine is required to run this
command. Run `vagrant init` to create a new Vagrant environment. Or,
get an ID of a target machine from `vagrant global-status` to run
this command on. A final option is to change to a directory with a
Vagrantfile and to try again.
```

I feel the documentation could clear this up as using `id` brings the ability to run the command from any directory. Or is that out of scope?

briancain:
Hi @amwill04 - yes I agree. The documentation could be more clear about what `name` and `id` is. If you wanted to write up a new pull request with two bullet points for `name` and `id` above the options section I'd be happy to review it! Use the format that the options section uses...i.e.

```
* `name` - short description here
* `id` - short description here
```

PR #9686, opened 2018-04-12
work around find syntax limitations in busybox-based distributions

mcandre:
Evidently, busybox-based Linux environments including OpenWrt lack newer features in `find` such as being able to union parenthetical expressions with `-or`.

This breaks rsync chowning for OpenWrt boxes, unfortunately. Fortunately, this is easy to workaround by separating the expressions into distinct `find` calls.

### DIFF
--- a/plugins/synced_folders/rsync/default_unix_cap.rb
+++ b/plugins/synced_folders/rsync/default_unix_cap.rb
@@ -37,8 +37,12 @@ module VagrantPlugins
         end
         "find #{guest_path} #{exclusions}" \
           "'!' -type l -a " \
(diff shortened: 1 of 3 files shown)


chrisroberts:
Hi there,

Thanks for this PR but I do not feel comfortable merging in this changeset. The modifications here are to the default unix capability which means the impact of this change will likely be quite far reaching (outside of the openwrt guest you are currently targeting). I think having this isolated within an openwrt guest implementation, or a generic busybox guest implementation, would be much more ideal and prevent unintended side effects that this change may cause. Thanks again!

Cheers!

PR #9692, opened 2018-04-14
propagate unset guestpath

### LINKED ISSUE #9691: missing shared folder guestpath does not work correctly
Please note that the Vagrant issue tracker is reserved for bug reports and
enhancements. For general usage questions, please use the Vagrant mailing list:
https://groups.google.com/forum/#!forum/vagrant-up. Thank you!

### Vagrant version
Run `vagrant -v` to show the version. If you are not running the latest version
of Vagrant, please upgrade before submitting an issue.

```bash
$ vagrant --version
Vagrant 2.0.4.dev
```

### Host operating system
This is the operating system that you run locally.

```bash
$ uname --all
CYGWIN_NT-10.0 DESKTOP-3DM06TO 2.10.0(0.325/5/3) 2018-02-02 15:16 x86_64 Cygwin
$ cmd /C "ver"
Microsoft Windows [Version 10.0.17133.73]
```

### Guest operating system
This is the operating system you run in the virtual machine.

Using box `generic/alpine36 1.4.12`

```bash
$ uname --all
Linux bazinga 4.9.73-0-hardened #1-Alpine SMP Tue Jan 2 16:33:46 UTC 2018 x86_64 Linux
$ cat /etc/issue
Welcome to Alpine Linux 3.6
Kernel \r on an \m (\l)
```

### Vagrantfile
```ruby
  config.vm.box = "generic/alpine36"
  config.vm.box_version = "1.4.12"
  config.vm.synced_folder ".", name: "vagrant", type: :virtualbox
    vb.functional_vboxsf = true
  end
end
```

Please note, if you are using Homestead or a different Vagrantfile format, we
may be unable to assist with your issue. Try to reproduce the issue using a
vanilla Vagrantfile first.

### Debug output
https://www.vagrantup.com/docs/other/debugging.html. The debug output should
be very long. Do NOT paste the debug output in the issue, just paste the
link to the Gist.

https://gist.github.com/cyungmann/100debe739c74108a2f4d078541f2557

### Expected behavior
What should have happened?

VM is provisioned without an error. Vagrant should not attempt to mount the shared/synced folder inside the VM because guestpath is not specified in the Vagrantfile (see d528902edc43d41cda037222d2b717e70ea4621c).

### Actual behavior
What actually happened?

Vagrant *does* attempt to mount the shared/synced folder inside the VM even though the guestpath is not specified in the Vagrantfile due to this line, which causes `options[:guestpath]` to always be set: https://github.com/hashicorp/vagrant/blob/a9c7b3efe23a4631f28c3716d2f20105a8793144/plugins/kernel_v2/config/vm.rb#L221

```
$ vagrant up
You appear to be running Vagrant outside of the official installers.
Note that the installers are what ensure that Vagrant has all required
dependencies, and Vagrant assumes that these dependencies exist. By
running outside of the installer environment, Vagrant may not function
properly. To remove this warning, install Vagrant using one of the
official packages from vagrantup.com.
==> default: Importing base box 'generic/alpine36'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'generic/alpine36' is up to date...
==> default: Setting the name of the VM: playlistr_default_1523685453108_26960
    default: Adapter 1: nat
    default: 22 (guest) => 2222 (host) (adapter 1)
==> default: Running 'pre-boot' VM customizations...
==> default: Booting VM...
    default:
    default: Vagrant insecure key detected. Vagrant will automatically replace
    default: this with a newly generated keypair for better security.
    default:
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
    default:  => L:/programming-projects/playlistr

mkdir -p ''

Stdout from the command:



Stderr from the command:

mkdir: can't create directory '': No such file or directory
```

cyungmann:
Make sure `options[:guestpath]` is not inadvertently set. Resolves #9691

### DIFF
--- a/plugins/kernel_v2/config/vm.rb
+++ b/plugins/kernel_v2/config/vm.rb
@@ -218,7 +218,7 @@ module VagrantPlugins
           synced_folder_name = guestpath
         end
 
-        options[:guestpath] = guestpath.to_s.gsub(/\/$/, '')
+        options[:guestpath] = guestpath.to_s.gsub(/\/$/, '') if guestpath
         options[:hostpath]  = hostpath
         options[:disabled]  = false if !options.key?(:disabled)
         options = (@__synced_folders[options[:guestpath]] || {}).
@@ -641,7 +641,9 @@ module VagrantPlugins
           guestpath = Pathname.new(options[:guestpath]) if options[:guestpath]
           hostpath  = Pathname.new(options[:hostpath]).expand_path(machine.env.root_path)
 
-          if guestpath.to_s != ""
+          if guestpath.to_s == "" && id.to_s == ""
+            errors << I18n.t("vagrant.config.vm.shared_folder_requires_guestpath_or_name")
+          elsif guestpath.to_s != ""
             if guestpath.relative? && guestpath.to_s !~ /^\w+:/
               errors << I18n.t("vagrant.config.vm.shared_folder_guestpath_relative",
                                path: options[:guestpath])
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -1789,6 +1789,8 @@ en:
         shared_folder_mount_options_array: |-
           Shared folder mount options specified by 'mount_options' must
           be an array of options.
+        shared_folder_requires_guestpath_or_name: |-
+          Shared folder options must include a guestpath and/or name.
 
 #-------------------------------------------------------------------------------
 # Translations for commands. e.g. `vagrant x`
--- a/test/unit/plugins/kernel_v2/config/vm_test.rb
+++ b/test/unit/plugins/kernel_v2/config/vm_test.rb
@@ -557,8 +557,8 @@ describe VagrantPlugins::Kernel_V2::VMConfig do
       expect(sf["/vagrant"][:foo]).to eq(:bar)
     end
 
-    it "is not an error if guest path is empty" do
-      subject.synced_folder(".", "")
+    it "is not an error if guest path is empty but name is not" do
+      subject.synced_folder(".", "", name: "my-vagrant-folder")
       subject.finalize!
       assert_valid
     end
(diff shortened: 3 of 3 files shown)


plugins/kernel_v2/config/vm.rb
```diff
@@ -218,7 +218,7 @@ def synced_folder(hostpath, guestpath, options=nil)
           synced_folder_name = guestpath
         end
 
-        options[:guestpath] = guestpath.to_s.gsub(/\/$/, '')
+        options[:guestpath] = guestpath.to_s.gsub(/\/$/, '') if guestpath
```
briancain: This should be a config validation error, rather than the option being silently ignored if the guestpath isn't specified (or is empty) with a synced_folders config option.

cyungmann: It should not be a config validation error. It is supposed to be an optional parameter but is not currently treated as such. Please see https://github.com/hashicorp/vagrant/pull/7980 as well as my issue write-up. Thanks!

briancain: Hi @cyungmann - yes, sorry for taking a while to get back to you. This is a bit tricky, because the `name` option isn't documented at all so it isn't clear how this should really work, and what might break with this change. I plan on adding some documentation around the `name` option soon (likely with this change).

I think you're change is good, so that we don't attempt to create empty string folders...but I still believe there should be a validation error thrown for a potential invalid configuration state to warn users before vagrant fails out and shows ugly stack traces. I understand `guesthpath` is supposed to be optional, but if it isn't provided then it seems like `name` is required? Otherwise the option breaks in fun ways when it assumes it has a `guestpath` or `name` to mount a share with. It would also be good to get some tests in this class with this change to ensure that if name or guestpath is not provided it is an invalid configuration as well as test that if `name` is provided, the guestpath option is `nil` and not empty string.

If you need any help with that let me know, thanks!

cyungmann: Hello, @briancain. Thank you for the feedback! It all made sense to me, and you are right that things blow up if neither `name` nor `guestpath` are provided. I updated the PR with your suggested changes. Please let me know if I should change anything (I'm a total Ruby newbie!).

PR #9696, opened 2018-04-17
fix rootpath if the vagrantfile is located under the root directory (WSL)

D4rkSh1t:
If you use wsl and the vagrantfile is located under the home of root than i get a wrong root_path:
wrong:
"\\?\C:\Users\%User%\AppData\Local\lxss\rootfs\root\...\..."
correct:
"\\?\C:\Users\%User%\AppData\Local\lxss\root\...\..."


chrisroberts:
@D4rkSh1t What version of Windows are you using?

D4rkSh1t:
@chrisroberts 
Version: 1703
OS Build: 15063.0

PR #9728, opened 2018-04-25
Keep old functionality before PR 7233 by Default

DDSloan96:
PR #9670  breaks existing puppet provisioners, this sets the new structured_facts feature to false by default to avoid breaking existing functioning setups


briancain:
Hello @DDSloan96 - thanks for the pull request, however it's already been fixed via https://github.com/hashicorp/vagrant/pull/9720 and will be updated in the next release.

PR #9729, opened 2018-04-25
Don't raise error if response is HTTP 416

briancain:
This pull request will continue when encountering an HTTP 416 error code in the downloader. It is a rebase/improvement of the PR #7885 to simplify the code a bit.

Fixes #7886

### DIFF
--- a/lib/vagrant/util/downloader.rb
+++ b/lib/vagrant/util/downloader.rb
@@ -292,14 +292,20 @@ module Vagrant
         # show an error message.
         if result.exit_code != 0
           @logger.warn("Downloader exit code: #{result.exit_code}")
-          parts    = result.stderr.split(/\n*curl:\s+\(\d+\)\s*/, 2)
-          parts[1] ||= ""
-          if parts[1].include? "416"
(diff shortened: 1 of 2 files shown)


lib/vagrant/util/downloader.rb
```diff
@@ -292,11 +292,14 @@ def execute_curl(options, subprocess_options, &data_proc)
         # show an error message.
         if result.exit_code != 0
           @logger.warn("Downloader exit code: #{result.exit_code}")
-          parts    = result.stderr.split(/\n*curl:\s+\(\d+\)\s*/, 2)
-          parts[1] ||= ""
-          raise Errors::DownloaderError,
-            code: result.exit_code,
-            message: parts[1].chomp
+          parts    = result.stderr.match(/\n*curl:\s+\(\d+\)\s*/)
```
chrisroberts: Can we make this matching a bit more explicit and include the message returned so it can be provided in the error? Something like:

```ruby
check = result.stderr.match(/\n*curl:\s+\((?<code>\d+)\)\s*(?<error>.*)$/)
if check && check[:code] == "416"
  ...
```

PR #9736, opened 2018-04-26
Add warning if vagrant-triggers plugin is installed

briancain:
To allow users to continue to use the community plugin, this commit adds
a switch that will not load the core trigger feature if it detects the
community plugin is installed. It also allows the user to disable the
warning so they can continue to use the plugin without extra output.

### DIFF
--- a/plugins/kernel_v2/plugin.rb
+++ b/plugins/kernel_v2/plugin.rb
@@ -40,9 +40,28 @@ module VagrantPlugins
         VMConfig
       end
 
-      config("trigger") do
-        require File.expand_path("../config/trigger", __FILE__)
-        TriggerConfig
+      plugins = Vagrant::Plugin::Manager.instance.installed_plugins
+      if !plugins.keys.include?("vagrant-triggers")
+        config("trigger") do
+          require File.expand_path("../config/trigger", __FILE__)
+          TriggerConfig
+        end
+      else
+        if !ENV["VAGRANT_USE_VAGRANT_TRIGGERS"]
+        $stderr.puts <<-EOF
+WARNING: Vagrant has detected the `vagrant-triggers` plugin. This plugin conflicts
+with the internal triggers implementation. Please uninstall the `vagrant-triggers`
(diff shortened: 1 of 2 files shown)


plugins/kernel_v2/plugin.rb
```diff
@@ -40,9 +40,27 @@ class Plugin < Vagrant.plugin("2")
         VMConfig
       end
 
-      config("trigger") do
-        require File.expand_path("../config/trigger", __FILE__)
-        TriggerConfig
+      plugins = Vagrant::Plugin::Manager.instance.installed_plugins
+      if !plugins.keys.include? "vagrant-triggers"
```
chrisroberts: Can you put the param in parentheses: `.include?("vagrant-triggers")`

plugins/kernel_v2/plugin.rb
```diff
@@ -40,9 +40,27 @@ class Plugin < Vagrant.plugin("2")
...
+        config("trigger") do
+          require File.expand_path("../config/trigger", __FILE__)
+          TriggerConfig
+        end
+      else
+        if !ENV["VAGRANT_USE_VAGRANT_TRIGGERS"]
+        $stderr.puts <<-EOF
+ERROR: Vagrant has detected an incompatible plugin installed. Please uninstall
```
chrisroberts: Update `ERROR` to `WARNING`

plugins/kernel_v2/plugin.rb
```diff
@@ -40,9 +40,27 @@ class Plugin < Vagrant.plugin("2")
...
+the `vagrant-triggers` plugin and run the command again if you wish to use the core
+trigger feature. To uninstall the plugin, run the command shown below:
+
+  vagrant plugin uninstall vagrant-triggers
+
+Note that the community plugin `vagrant-triggers` and the core trigger feature
+in Vagrant do not have compeltely compatible syntax.
```
chrisroberts: > in Vagrant do not have compatible syntax.

PR #9746, opened 2018-04-27
Update Windows Hyper-V check

chrisroberts:
Checks only if the Hyper-V hypervisor is enabled instead of all
Hyper-V features. This will allow detection where Hyper-V is in
use for things like ApplicationGuard.

Uses both Get-WindowsOptionalFeature and Get-WindowsFeature to
support check on server and non-server version.

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -107,9 +107,12 @@ module Vagrant
           return @_windows_hyperv_enabled if defined?(@_windows_hyperv_enabled)
 
           @_windows_hyperv_enabled = -> {
-            ps_cmd = "$(Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-All -Online).State"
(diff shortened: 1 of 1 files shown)


j0j00:
This commit breaks the original check, stopping hyper-v from working on Windows 10.

The original command was:
```
"$(Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-All -Online).State"
```
but it now runs
```
"$(Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-Hypervisor).State"
```` 
when it should be 
```
"$(Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-Hypervisor -Online).State"
```` 

It needs to append the `-Online` parameter if it's running `Get-WindowsOptionalFeature`.

Reference to a sample issue: https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1616

-----------------

This is stopping active development from happening, I was wondering if it was possible to easily patch it locally before it's fixed upstream. I'd rather not recompile it since that'll likely complicate things.

PR #9754, opened 2018-05-01
Update docker provisioner for windows, fix command filter behavior for `rm`

### LINKED ISSUE #9280: Windows docker capability missing

### LINKED ISSUE #9339: usage of "rm -f" is not portable

briancain:
This commit rebases #9353. It introduces a windows capability for the docker provisioner to check of the docker daemon is running. It also fixes the behavior of the `rm` filter for winrm. `rm -f dir` on linux does not result in an error, thus the winrm filter was updated since `rm`'ing a directory that doesn't exist on windows is an error.

Fixes #9339
Fixes #9280

### DIFF
--- a/plugins/communicators/winrm/command_filters/rm.rb
+++ b/plugins/communicators/winrm/command_filters/rm.rb
@@ -31,9 +31,9 @@ module VagrantPlugins
 
           ret_cmd = ''
           if recurse
-            ret_cmd = "rm \"#{dir}\" -recurse -force"
+            ret_cmd = "if (Test-Path \"#{dir}\") {Remove-Item \"#{dir}\" -force -recurse}"
           else
-            ret_cmd = "if (Test-Path #{dir}) {Remove-Item #{dir} -force}"
+            ret_cmd = "if (Test-Path \"#{dir}\") {Remove-Item \"#{dir}\" -force}"
           end
           return ret_cmd
         end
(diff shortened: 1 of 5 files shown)


plugins/communicators/winrm/command_filters/rm.rb
```diff
@@ -31,9 +31,9 @@ def filter(command)
 
           ret_cmd = ''
           if recurse
-            ret_cmd = "rm \"#{dir}\" -recurse -force"
+            ret_cmd = "if (Test-Path #{dir}) {Remove-Item #{dir} -force -recurse}"
```
chrisroberts: The `#{dir}`s need to be quoted

plugins/provisioners/docker/client.rb
```diff
@@ -127,12 +127,9 @@ def container_args_changed?(config)
       def create_container(config)
         args = container_run_args(config)
 
-        @machine.communicate.sudo %[
-          rm -f #{config[:cidfile]}
-          docker run #{args}
-        ]
+        @machine.communicate.sudo %[rm -f #{config[:cidfile]}]
```
chrisroberts: cidfile needs to be quoted

Issue #9628, opened 2018-03-28
VMware Desktop Plugin in WSL (maybe some useful info)


jchappell82:
### Vagrant version
`vagrant --version` = 2.0.3
vmware-desktop plugin (latest as of 3/28/2018- I already uninstalled it so I can't easily check now)

### Host operating system
Windows 10 - WSL - version 1709 build 16299.309
Linux flavor - Ubuntu 16.04
VMware Workstation Windows - 12.5

### Guest operating system
Ubuntu 16.04

### Expected behavior
The VM boots

### Actual behavior
The VMware plugin tries to run a hardcoded _Windows_ path to `vmrun.exe`, rather than allowing either a lookup via `$PATH` or converting to a WSL supported `/mnt/c/path/to/vmware/vmrun.exe` path

### Steps to reproduce
1. Install vagrant-vmware-desktop plugin in WSL
2. Install the Vagrant VMware Utility in _Windows_ because WSL does not support background services
3. Copy certificates from the Windows utility's path to the appropriate location in the WSL filesystem
4. Attempt a `vagrant up --provider=vmware_desktop`

It does seem to be able to successfully contact the Utility service, but fails upon attempting to execute `vmrun.exe` due to the hardcoded Windows-style path.

Some general notes - I totally understand that this isn't officially supported yet, but I thought I'd open this issue in case the information is useful to you during your adding the feature for the next version.

I did also attempt to run the Linux version of the VMware utility in a foreground process from another tab, but that fails with an "unable to locate executables" error (which is almost certainly because it doesn't know how to look at `/mnt/c` yet)

All that said, this seems to be _really_ close to working.  Looking forward to the next release, and if you need any help testing it out, feel free to reach out.  Feel free to close this issue as well, as it's just intended to be some hopefully useful info.

chrisroberts:
Hi there,

Thanks for opening this issue, the info, and the offer to help test out support. I have been spending some time working on this and am close to having it working. Working within the WSL can be challenging when due to a POSIX like environment that's really Windows, and proper path conversions in the correct locations to achieve the expected results. As soon as I have things working as expected I will push a new release.

Again, thanks for all the information, and I hope to have support out very soon.

Cheers!

StefanScherer:
Hi there,

Today a colleage asked me how to use Vagrant VMware in WSL. He has a brand new business notebook with Windows 10, so he wants to go with WSL. He started to install it, but had troubles.
[...]

PR #9758, opened 2018-05-02
Call mstsc.exe Windows RDP client in WSL

StefanScherer:
Running `vagrant rdp` in WSL only tries to run the Linux RDP clients `xfreerdp` or `rdesktop`. 

I've added the code from the Windows RDP capability into the Linux `rdp.rb`.

It recognizes that Vagrant is running in WSL and then uses `mstsc.exe` as fallback if the user hasn't installed the other Linux RDP clients. So WSL users have the chance to use the Linux RDP client, but it also just works with the Windows RDP client.

Improves #9628

### DIFF
--- a/plugins/hosts/linux/cap/rdp.rb
+++ b/plugins/hosts/linux/cap/rdp.rb
@@ -13,7 +13,11 @@ module VagrantPlugins
             elsif Vagrant::Util::Which.which("rdesktop")
               "rdesktop"
             else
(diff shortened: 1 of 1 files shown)


StefanScherer:
I tried it manually with
```
sudo vi /opt/vagrant/embedded/gems/2.0.4/gems/vagrant-2.0.4/plugins/hosts/linux/cap/rdp.rb
```

PR #9762, opened 2018-05-03
workaround spurious exit status (Haiku nightly)

mcandre:
Prepare for integration with Haiku nightly, which features an incorrect exit status for `/bin/shutdown`.

In the past with Haiku alpha 4, users could work around this by shimming a shell script at `/bin/shutdown` that ran `/bin/shutdown-actual`, then exiting zero. Unfortunately, Haiku nightly has removed writeability for `/bin` paths, making this harder to patch within the guest. While Haiku is getting its stuff fixed, it would be nice if the Vagrant guest plugin recognized this behavior and wrapped it, ignoring the status for now.

### DIFF
--- a/plugins/guests/haiku/cap/halt.rb
+++ b/plugins/guests/haiku/cap/halt.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
       class Halt
         def self.halt(machine)
           begin
-            machine.communicate.execute("/bin/shutdown")
+            machine.communicate.execute("/bin/shutdown; exit 0")
           rescue IOError, Vagrant::Errors::SSHDisconnected
             # Ignore, this probably means connection closed because it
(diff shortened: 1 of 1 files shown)


chrisroberts:
Same as with #9770, the correct solution should be to resolve the issue within the box.

Cheers!

mcandre:
Unfortunately, working around this issue is quite difficult. Specifically due to the difficulty of shadowing, wrapping, or otherwise capturing the spurious exit code before it passes from Haiku to Vagrant. I don't expect Haiku to fix this exit code anytime soon. I proposed dropping the absolute path to the `shutdown` binary.

Other than waiting indefinitely, or having the Haiku Vagrant plugin override this behavior, I'm not sure how else to solve this.

PR #9770, opened 2018-05-03
allow haiku binary shadowing

mcandre:
Allow shutdown and other binaries to be shadowed with binaries from `/boot/system/non-packaged/bin`, by dropping the absolute path.

This is an alternative fix to help the Vagrant Haiku guest plugin integrate better with Haiku nightly, whose `/bin/shutdown` unfortunately exits with a spurious value.

### DIFF
--- a/plugins/guests/haiku/cap/halt.rb
+++ b/plugins/guests/haiku/cap/halt.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
       class Halt
         def self.halt(machine)
           begin
-            machine.communicate.execute("/bin/shutdown")
+            machine.communicate.execute("shutdown")
           rescue IOError, Vagrant::Errors::SSHDisconnected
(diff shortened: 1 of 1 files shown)


chrisroberts:
Hi there,

I don't think this is a good change to make as the root cause of the problem exists within the guest, not withing Vagrant. Fixing the issue within the box would be the correct resolution.

Cheers!

mcandre:
Fixing this issue within the guest is problematic, as the design of Haiku is to prohibit `/bin/shutdown` from being altered or shadowed, in Haiku nightly. I tried a few different shell tricks to work around this, getting as far as having a working `shutdown` shim script; however Vagrant currently hardcodes a path to `/bin/shutdown`, so the shim is ignored. Unless that is changed, we'll have to wait for the Haiku developers to fix the exit value.

PR #9790, opened 2018-05-05
Leverage POSIX-compliant `find` -o rather than -or 

cdituri:
Leverage POSIX-compliant `find` -o rather than -or when building default rsync capabilities.

In the course of working on a guest plugin for OpenWrt--which I'd also like to contribute to the community once I'm finished with its implementation--I found that OpenWrt's busybox `find` does not support (non-POSIX compliant) `-or` expressions; but rather only POSIX-compliant `-o` expressions.

GNU find explicitly points out the difference between -o and -or in their man page; BSD's man page states their implementation is a super-set of the POSIX standard; and interestingly, SmartOS' man page doesn't even mention `-or` (but one of our SmartOS rsync unit tests uses said -or form!).  

 A couple links for convience:

- https://linux.die.net/man/1/find
- https://www.freebsd.org/cgi/man.cgi?find(1)
- https://smartos.org/man/1/find

### DIFF
--- a/plugins/synced_folders/rsync/default_unix_cap.rb
+++ b/plugins/synced_folders/rsync/default_unix_cap.rb
@@ -37,7 +37,7 @@ module VagrantPlugins
         end
         "find #{guest_path} #{exclusions}" \
           "'!' -type l -a " \
-          "'(' ! -user #{opts[:owner]} -or ! -group #{opts[:group]} ')' -exec " \
+          "'(' ! -user #{opts[:owner]} -o ! -group #{opts[:group]} ')' -exec " \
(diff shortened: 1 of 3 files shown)


cdituri:
Just seeing https://github.com/hashicorp/vagrant/pull/9686#issuecomment-382508176 now - will see what I can do to place this functionality directly in my openwrt guest plugin implementation. For now, closing per @chrisroberts previous comment on the subject.   

> I think having this isolated within an openwrt guest implementation, or a generic busybox guest implementation, would be much more ideal and prevent unintended side effects that this change may cause. Thanks again!

PR #9813, opened 2018-05-09
Reference Hyper-V switch by ID instead of name

0x022b:
Keep track of selected Hyper-V switch using the ID instead of name
to prevent any encoding issues that may occur switching between
PowerShell and Ruby also when importing VM from XML file. With the IDs
staying consistent, the switch name can be fetched from the provided ID.


chrisroberts:
Hi there. Thanks for this PR! However, we are going to close it in favor of #9872 which includes the fix this PR provides.

Cheers!

PR #9839, opened 2018-05-15
Fix unused library

rubensfig:
When creating a fedora 28 guest, dnf failed while installing the nfs-utils-lib package, so I propose this change.


rahulbajaj0509:
@chrisroberts hey, so `nfs-utils-lib` is deprecated while `libnfs-utils` is present. `libnfs-utils` has been there since f24 so there wont be any issues having that instead of `nfs-utils-lib`. 

This patch needs to be modified to have `libnfs-utils` instead of just deleting the `nfs-utils-lib`

rubensfig:
As per @rahulbajaj0509 suggestion, I've updated the pull request.

PR #9861, opened 2018-05-23
Add Rubocop with initial configuration

tibbon:
Adds Rubocop with initial base configuration generated via `rubocop --auto-gen-config`. No changes were made to enforce styles. Lines added to Travis config as well so it will run rubocop. 

If merged, suggested steps forward are to evaluate each style on a case-by-case basis, often removing one from `.rubocop_todo.yml` and either then manually fixing violations or using `rubocop --auto-correct` to fix it, then make a single commit/PR for consideration. 

If there's a style that is intentionally ignored, that should go in the `.rubocop.yml` instead and intentionally disabled with a `description` field to say why it is removed. Then remove it from the `.rubocop_todo.yml`. 

When Rubocop is upgraded to new versions `rubocop --auto-gen-config` can be re-run and generate a new todo file that will deal with any new cops added. 

By doing this, a consistent styleguide can be maintained. Rubocop's defaults generally follows the [bbatsov Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide).


tibbon:
For some reason the todo file isn't being picked up by Travis. Fixing this and repushing soon.

tibbon:
Fixed. There was one Cop that doesn't configure properly with the `--auto-gen-config` and a problem where Rubocop was trying to parse some of the `website/*.builder.xml` files and falling over on the key/value pairs at the top. Proper exclusions have been put in for those.

chrisroberts:
Hi there,

Thanks for the PR but at this point we aren't ready to introduce rubocop. We have discussed it internally, and agree that it is something we would like to use. However, at this point it will just be introducing more overhead and an extremely large todo list of things that would need to be rectified with no attention being paid to it for some time. We are looking at adding this down the road, just not quite yet.

Again thanks for the PR. Cheers!

Issue #9863, opened 2018-05-24
Add an ansible.requirements option to ansible_local provisionner to install roles before running playbook


rgarrigue:
### Vagrant version

2.1.1

### Host operating system

WIndows 7

### Guest operating system

centos/7

### Vagrantfile

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :
Vagrant.configure(2) do |config|
  config.vm.box = "centos/7"
  config.vm.provision "ansible_local" do |ansible|
    ansible.install_mode = "pip"
    ansible.playbook = "/vagrant/playbook.yml"
  end
end
```

### Expected behavior

I wish there was an option like

    ansible.requirements = "/vagrant/requirements.yml"

That would call `ansible-galaxy install -r /vagrant/requirements.txt` after ansible installation via pip and before running ansible.

### Actual behavior

ERROR! the role 'geerlinguy.nginx was not found in /vagrant/roles:/home/vagrant/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:/vagrant

### Steps to reproduce

Write down a simple playbook.yml with this content

```yaml
---
- hosts: all
  become: true
  roles:
    - geerlinguy.nginx
```

and a requirements.yml with

```
- src: geerlingguy.nginx
```

gildegoma:
@rgarrigue I think you're looking for the `galaxy_role_file` [option](https://www.vagrantup.com/docs/provisioning/ansible_common.html#galaxy_role_file) (and related settings).

Please try something like:
```ruby
Vagrant.configure(2) do |config|
  config.vm.box = "centos/7"
  config.vm.provision "ansible_local" do |ansible|
    ansible.install_mode = "pip"
    ansible.playbook = "/vagrant/playbook.yml"
    ansible.galaxy_role_file = "/vagrant/requirements.yml"
  end
end
```

rgarrigue:
Oh... hell yes, this is it. I read it too quickly, and ctrl-f "requirement" returned nothing. 

May I suggest a doc update to include a quick example targeting /vagrant/requirements.yml for the sake of lazy reader like me ? :smile:

gildegoma:
No worries! The "galaxy" keyword could have done the trick though :wink: 

> May I suggest a doc update to include a quick example targeting /vagrant/requirements.yml 

I agree that putting more examples in the documentation could be handy for readers, and reduce support efforts :smiley:. Pull requests with such improvement in mind are always welcome :+1:

PR #9864, opened 2018-05-24
ansible_local provisioner tuning example

rgarrigue:
According to @gildegoma in #9863, here's a concrete example about tuning ansible(_local) provisioner.


rgarrigue:
@gildegoma FYI I just added another bit of doc as a second commit

gildegoma:
Thank you for all this. I took your proposal and use cases, but reorganize them a bit to better fit the way the rest of the Ansible docs are written.
Closing in favor of #9869. Please have a look there and tell me if you agree.

Issue #9592, opened 2018-03-19
vagrant network config file should not have a priority of 99


DDSGooch:
The automatically generated vagrant.network config file shouldn't be installed as 99-vagrant.network because it means this file will nearly always be the lowest priority file.

For example I recently had an issue with a box which already had a 99-dhcp.network. The only way round this was to rename it to 99-zdhcp.network, as it wasn't possibly to change it to a lower priority.

I think giving the vagrant.network file a priority of 50 would probably be the best solution, as this would put it in the middle of the priority list, giving flexibility

https://github.com/hashicorp/vagrant/blob/5bf75dcfb3ab66f6fa0675d051b6c160be2ff1b1/plugins/guests/debian/cap/configure_networks.rb#L100

Many thanks

DDSGooch:
Thanks for that - much appreciated :+1:

jnahorny:
This issue hasn't been fixed by the PR #9646 That code changes the priority of the netplan YAML file. But in this doesn't resolve the issue for a configuration with `systemd-networkd`, which was the suggestion of the person who opened this issue, because the priority of network file is still 99: `/etc/systemd/network/99-vagrant.network`.
I see this in the HEAD of the master, and I've also confirmed this by testing. I believe this issue should be reopened and a proper patch should be applied.
Like this:
```
diff --git a/plugins/guests/debian/cap/configure_networks.rb b/plugins/guests/debian/cap/configure_networks.rb
index 6ba4c2ca9..fcc907a25 100644
--- a/plugins/guests/debian/cap/configure_networks.rb
+++ b/plugins/guests/debian/cap/configure_networks.rb
@@ -109,7 +109,7 @@ module VagrantPlugins
           end
 
           remote_path = upload_tmp_file(comm, net_conf.join("\n"))
-          dest_path = "#{NETWORKD_DIRECTORY}/99-vagrant.network"
+          dest_path = "#{NETWORKD_DIRECTORY}/50-vagrant.network"
           comm.sudo(["mkdir -p #{NETWORKD_DIRECTORY}",
             "mv -f '#{remote_path}' '#{dest_path}'",
             "chown root:root '#{dest_path}'",
```

I've created a pull request for that: https://github.com/hashicorp/vagrant/pull/9867

PR #9867, opened 2018-05-24
Fix issue 9592 when systemd-networkd is used by Debian guest

jnahorny:
The PR #9646 did not fix the isue. It changed the priority of the "netplan" YAML file, not the priority of the  network file in /etc/systemd/network directory used by systemd-networkd. My PR fixes the issue for real.


jnahorny:
Also, I've linked do this PR in the issue (https://github.com/hashicorp/vagrant/issues/9592), but it is CLOSED, so not sure if that one could be re-opened, or I should create a new issue. Please advise.

PR #9919, opened 2018-06-12
Added note on other virtualizers before install.

src-r-r:
I found some other support questions similar across the web. Figured I'd write a note in case anyone else ran into this issue.


briancain:
Thanks for the pull request! I've updated the docs to include a message about windows as well. That's covered in this pull request here https://github.com/hashicorp/vagrant/pull/10065. I'll go ahead and close this one since it also has your changes. Thanks again! 🎉

Issue #9878, opened 2018-05-28
nfs_client.rb needs updating


mjudeikis:
nfs-utils-lib was rolled into nfs-utils as of RHEL 7.2, CentOS 7 now has made the upstream change as well. 
So if you try anything like RHEL 7.2+ Centos, Fedora28+ you get error:
```
The following SSH command responded with a non-zero exit status.                                                                                                                                             
Vagrant assumes that this means the command failed!                                                                                                                                                          
                                                                                                                                                                                                             
if command -v dnf; then                                                                                                                                                                                      
  dnf -y install nfs-utils nfs-utils-lib portmap                                                                                                                                                             
else                                                                                                                                                                                                         
  yum -y install nfs-utils nfs-utils-lib portmap                                                                                                                                                             
fi                                                                                                                                                                                                           
                                                                                                                                                                                                             
if test $(ps -o comm= 1) == 'systemd'; then                                                                                                                                                                  
  /bin/systemctl restart rpcbind nfs                                                                                                                                                                         
else                                                                                                                                                                                                         
  /etc/init.d/rpcbind restart                                                                                                                                                                                
  /etc/init.d/nfs restart
fi

Stdout from the command:

/usr/bin/dnf
```

https://github.com/hashicorp/vagrant/blob/9c299a2a357fcf87f356bb9d56e18a037a53d138/plugins/guests/redhat/cap/nfs_client.rb#L8
^^ code needs to have logic to detects version of the OS and act accordingly:
```
//fedora
OS=$( awk '/ID_LIKE=/' /etc/*-release | sed 's/ID_LIKE=//' | sed 's/[.]0/./')
VERSION=$(awk '/VERSION_ID=/' /etc/*-release | sed 's/VERSION_ID=//' | sed 's/[.]0/./')
//TODO RHEL, centos

echo $OS
echo $VERSION
```

so script in this would know which one to install. 

OR just add `nfs-utils-lib` install as `/dev/null` operation for this usecase:
```
yum -y -q install nfs-utils-lib >/dev/null 2>&1
```

mjudeikis:
Before proceeding with fix wanted to know what people think about this.

chrisroberts:
@mjudeikis Hi and thanks for opening this issue. I'm curious if it might be easier (less required checking) to only install the package if the package is available for installation. Basically just doing a quick search before installation and using that as a conditional. Any thoughts?

mjudeikis:
[...]

PR #9927, opened 2018-06-13
Fix the name for nfs-utils-lib package in fedora28

rahulbajaj0509:
In fedora28, the nfs package `nfs-utils-lib` has been removed and it no longer exists. Rather, now `libnfs-utils` is used. `libnfs-utils` shall work on all distributions of fedora :)


chrisroberts:
Hi there. Thanks for this PR! This is pretty much a duplicate of #9839 and has the same issues that need to be resolved before it would be acceptable to merge. You can find more conversation about this issue in #9878. I'm going to close this PR now (as it's a duplicate) but if you want to tackle what was discussed in #9878 please feel free to open a new PR.

Cheers!

PR #9935, opened 2018-06-14
Fixes the change in packaging for nfs in f28 (#9878)

whitel:
Fixes the change in packaging for nfs in f28 (#9878). Takes in to account comments in PR #9927 and PR #9839 However, removed part of the unit test which will be very difficult to fix (I expect). The unit test is still doing a good test that nfs and rpcbind work though.


plugins/guests/redhat/cap/nfs_client.rb
```diff
@@ -5,7 +5,11 @@ class NFSClient
         def self.nfs_client_install(machine)
           machine.communicate.sudo <<-EOH.gsub(/^ {12}/, '')
             if command -v dnf; then
-              dnf -y install nfs-utils nfs-utils-lib portmap
+              if $(output=$( dnf search nfs-utils-lib 2>&1) && grep -q 'No matches found' <<< "$output") ; then
```
marjune163: Can we use `dnf info -q nfs-utils-lib > /dev/null` to test if the package exists in the repo? That would be better. BTW, if the guest OS's locale is not 'en' like, the output language of the `dnf` may not in English.

whitel:
I figured if I posted it, someone would tell me a better way to do the test :)

plugins/guests/redhat/cap/nfs_client.rb
whitel: I figured if I posted it, someone would tell me a better way to do the test :)

PR #9989, opened 2018-07-04
Fix the net.exe smb share list for non-admins

bastei:
In case Get-SmbShare is not available, using smb shares on Windows is not possible for non-admins at the moment as "net share <share-name>" returns an error. By evaluating the "net share" table, this is solved here.
Additionally, the script was not multilanguage-save, which is solved here, too.

### DIFF
--- a/plugins/hosts/windows/cap/smb.rb
+++ b/plugins/hosts/windows/cap/smb.rb
@@ -155,24 +155,47 @@ module VagrantPlugins
           if result.nil?
             return nil
           end
-          result.sub!(/^.+?\-+/m, "")
-          share_names = result.strip.split("\n").map do |line|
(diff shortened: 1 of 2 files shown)


chrisroberts:
Closing this in favor of #10043 which retains description information. Thanks again for this PR!

Cheers!

bastei:
Fixing the localization problem was only a side effect of the changes in this PR. The main purpose was to enable Windows 7 users to use smb shares without admin rights.
> The share summary will generally not display the entire value which is required for checks.

This is not completely true as the machine id is saved in the share name, too. Thus it would be enough to compare against the share name (where the entire machine id is displayed) instead of the share description (where it is truncated). Would you reopen the PR if I change the code accordingly, @chrisroberts?

PR #9996, opened 2018-07-06
Revert Vagrant plugin loading from #9808

mbevc1:
Need to revert #9808 and restore v2.1.1 behaviour in order to prevent plug-in loading regression introduced with v2.1.2. Emit a warning and proceed if plug-in doesn't load.

### DIFF
--- a/lib/vagrant.rb
+++ b/lib/vagrant.rb
@@ -337,7 +337,7 @@ if Vagrant.plugins_enabled?
             global_logger.info("Loading plugin `#{plugin_name}` with slash require: `#{plugin_slash}`")
             require plugin_slash
           else
-            raise
+            global_logger.warn("Failed to load plugin `#{plugin_name}`. Assuming library and moving on.")
           end
         end
       else


chrisroberts:
Hi there,

Can you please provide some context on the reasoning for reverting the raise on loading failure exception? Without this it masks issues with plugin loading that should be surfaced.

Thanks!

mbevc1:
Hi.

We have a use case where we have a gem(ZCollective) which is used in Cucumber tests, but fails upon loading via Vagrant `require`. Previous behaviour was more tolerant and wasn't trying to load plug-ins in Vagrant internal commands. If gem is installed via `vagrant plugin install zcollective` v2.1.2 fails with error `file not found zcollective`. 

Thanks

chrisroberts:
Thanks for the context! I still believe the behavior for plugin loading is correct. In this case, where a standard file path is not available for requiring, the `--entry-point` option should be used when installing the plugin. I can see where this may be a support library that does not need to be loaded by Vagrant, and in that case the entry point can easily be stubbed with `vagrant` so the plugin require becomes a no-op. In that case, the install command would look like:

```
$ vagrant plugin install zcollective --entry-point vagrant
```

Cheers!

mbevc1:
Thanks for that hint, just a quick question; is it possible to programatically set `entry-point` via pug-in manager? We're loading those via our own lib.

Cheers!

mbevc1:
Okay, found it - a bit confusing, but I've seen CLI command is using `require` to set an `entry_point`.

PR #10004, opened 2018-07-09
Update mount_shared_folder.rb

johnrizzo1:
This change allows special characters in the password such as ) which will cause cmdkey to fail without the quotes.

We should probably implement this in the following files as well.

* plugins\hosts\linux\cap\rdp.rb line 42
* plugins\hosts\windows\cap\rdp.rb line 15


briancain:
Closing this pull request in favor of this one which has your changes plus a unit test https://github.com/hashicorp/vagrant/pull/10219
Thanks for your contribution!

Issue #10011, opened 2018-07-11
Using NFS shared folders doesn't work properly when using a VoidLinux host


Aloz1:
### Vagrant version
vagrant 2.1.2

### Host operating system
Void Linux

### Guest operating system
Alpine (but guest is irrelevant)

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
  config.vm.define :pxeserver do |server|
    server.vm.provider :libvirt
    server.vm.box = "generic/alpine38"
    server.vm.synced_folder "server/tftp",
      "/srv/tftp",
      type: "nfs"
  end
end
```
Note, plugins vagrant-libvirt and vagrant-alpine are installed, but are not relevant as this is a host issue.

### Debug output
https://gist.github.com/Aloz1/66ae28001736d255ea895d06a667f63e
Relevant sections are lines 921 to 927

### Expected behavior
vagrant should have detected what the current status of NFS is.

### Actual behavior
vagrant failed because the command used to check the service is not correct. Void uses an init system called runit, additionally the service is under a different name. The correct command should be '/usr/bin/sv status nfs-server'.

### Steps to reproduce
1. Install voidlinux.
2. Install vagrant on top of voidlinux
3. Try running any Vagrantfile with nfs

Aloz1:
PR is done. Could we please tag it as feedback/review?

GoingFast:
hello?

briancain:
Closing this, as the PR has been merged.

Issue #8279, opened 2017-02-13
Can't get NFS to work on Ubuntu 16.04


octoxan:
### Vagrant version
1.9.1

### Host operating system
Ubuntu 16.04

### Guest operating system
Ubuntu 14.04 (Scotchbox) https://github.com/scotch-io/scotch-box

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|

    config.vm.box = "scotch/box"
    config.vm.network "private_network", ip: "192.168.33.10"
    config.vm.hostname = "scotchbox"
    #config.vm.synced_folder ".", "/var/www", :mount_options => ["dmode=777", "fmode=666"]
    
    # Optional NFS. Make sure to remove other synced_folder line too
    config.vm.synced_folder ".", "/var/www", :nfs => { :mount_options => ["dmode=777","fmode=666"] }

end
```
### Debug output

Here is the entire debug output:
https://gist.github.com/octoxan/436aad239d306ad4cae39f34632a546c

My final error is

```
mount.nfs: access denied by server while mounting 192.168.33.1:/home/xander/Websites/scotchbox
```

I do have nfs-common installed on the vm and nfs-kernel-server installed locally.

Any ideas on how to fix this?

Clayment:
Same problem with vagrant 1.9.2 on Arch linux

kivoli:
I may at least offer some insight as to why and maybe a work-around acceptable for most people. When using NFSv4, which I unfortunately require due to Ruby bundler using file system locks, Ubuntu somehow enforces the prerequisite of a “root” export for NFS.

So this would work:
```exports
/home/testuser 192.168.121.0/24(rw,fsid=0,no_subtree_check,sync)
/home/testuser/project 192.168.121.0/24(rw,nohide,no_subtree_check,sync)
```
```sh
mount 192.168.121.1:/project /vagrant
```
The idea is that all folders to be exported via NFS should be bind-mounted to a single folder structure (usually `/exports`). [Ubuntu NFS Guide](https://help.ubuntu.com/community/SettingUpNFSHowTo)

Funnily enough the way Vagrant does it works perfectly fine with Debian Stretch so it’s not strictly enforced by NFSv4.

Work around for people who do not need NFSv4: [Use NFSv3](https://www.vagrantup.com/docs/synced-folders/nfs.html)
```Vagrantfile
[...]

PR #10012, opened 2018-07-11
Add void linux host support

Aloz1:
Currently, NFS does not work correctly on void linux. This PR adds a new Void Linux host to fix this. An issue has been opened here: #10011

### DIFF
--- a/plugins/hosts/void/cap/nfs.rb
+++ b/plugins/hosts/void/cap/nfs.rb
@@ -3,7 +3,7 @@ module VagrantPlugins
     module Cap
       class NFS
         def self.nfs_check_command(env)
-          "/usr/bin/sv status nfs-server"
+          "sudo /usr/bin/sv status nfs-server"
         end
 
         def self.nfs_start_command(env)
--- a/plugins/hosts/void/cap/nfs.rb
+++ b/plugins/hosts/void/cap/nfs.rb
@@ -1,0 +1,19 @@
+module VagrantPlugins
+  module HostVoid
+    module Cap
+      class NFS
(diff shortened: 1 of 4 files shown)


Aloz1:
Hm, nfs_start_command is wrong. Will need to fix it tonight when I get home. (Should probably be something along the lines of "ln -s /etc/sv/nfs-server /var/services/."

Aloz1:
Hm, need to investigate some more. As it stands, I'm unable to get NFSv3 working, and NFSv4 only works if the export has has no_root_squash. This is probably a void issue, but suggestions would be appreciated.

Aloz1:
NFSv3 works fine now, void requires root access to check service status, hence vagrant was attempting to start up nfs-server rather than 'export -ar'.

Additionally, it appears that NFSv4 works fine. Seems I had the same problem as #8279.

Aloz1:
This PR is ready for review...

GoingFast:
@briancain @chrisroberts going to take another 3 months? the fix is up so whats the problem? nfs is unusable for void users

Aloz1:
Looks like it's been tagged as needing tests (I unfortunately didn't get a notification when the label was added). I'm not very experienced with ruby, let alone unit testing in ruby, so might take me a little bit to figure things out.

I'll investigate and see if I can come up with something over the next couple of days.

chrisroberts:
@Aloz1 I've pushed an update to this PR. If you sign off on the CLA I'll get this merged and included with the next release.

Aloz1:
@chrisroberts thanks for helping out with getting this over the line, its much appreciated!!!

PR #10020, opened 2018-07-12
website: Document Vagrant Cloud search API

justincampbell:


website/source/docs/vagrant-cloud/api.html.md.erb
```diff
@@ -530,6 +530,101 @@ Sends a 2FA code to the requested delivery method.
...
+
+`GET /api/v1/search`
+
+#### Arguments
+
+* `q` - (Optional) The search query. Results will match the `username`, `name`, or `short_description` fields for a box. If omitted, the top boxes based on `sort` and `order` will be returned (defaults to "downloads desc").
+* `provider` - (Optional) Filter results to boxes supporting for a specific provider.
+* `sort` - (Optional, default: `"downloads"`) The field to sort results on. Can be one of `"downloads"`, `"created"`, or `"updated"`.
+* `order` - (Optional, default: `"desc"`) The order to return the sorted field in. Can be `"desc"` os `"asc"`.
+* `limit` - (Optional, default: `10`) The number of results to return (max of TODO).
```
briancain: `max of TODO` should be fixed to whatever the max is

website/source/docs/vagrant-cloud/api.html.md.erb
```diff
@@ -530,6 +530,101 @@ Sends a 2FA code to the requested delivery method.
...
+
+### Search for boxes
+
+`GET /api/v1/search`
+
+#### Arguments
+
+* `q` - (Optional) The search query. Results will match the `username`, `name`, or `short_description` fields for a box. If omitted, the top boxes based on `sort` and `order` will be returned (defaults to "downloads desc").
+* `provider` - (Optional) Filter results to boxes supporting for a specific provider.
+* `sort` - (Optional, default: `"downloads"`) The field to sort results on. Can be one of `"downloads"`, `"created"`, or `"updated"`.
```
briancain: Maybe this is a better question for the corresponding website PR, but what gets returned when you provide a sort key that isn't supported? Is there a new HTTP response code that needs to be documented here?

website/source/docs/vagrant-cloud/api.html.md.erb
```diff
@@ -530,6 +530,101 @@ Sends a 2FA code to the requested delivery method.
...
+    end
+    ```
+  </div>
+</div>
+
+#### Example Response
+
+```json
+{
+  "boxes": [
```
briancain: Does this response body also need to include `downloads`?

website/source/docs/vagrant-cloud/api.html.md.erb
justincampbell: I think right now it just ignores it, but I will check.

justincampbell: Validated now! Will return 400.

briancain:
Looking good, I think we're just missing the new HTTP 400 status code documented with the rest of the response codes.

Issue #9972, opened 2018-07-01
Exec error: fork/exec /usr/bin/ruby: argument list too long


karmux:
### Vagrant version
Vagrant 2.1.2

### Host operating system
Manjaro Linux 17.1.11

### Guest operating system
any

### Vagrantfile
https://github.com/coreos/coreos-vagrant/blob/master/Vagrantfile

### Debug output
Installing plugins: vagrant-ignition
...
Installing plugins: vagrant-ignition
Exec error: fork/exec /usr/bin/ruby: argument list too long
Installation of one or more plugins has failed. Aborting.
...
Installation of one or more plugins has failed. Aborting.

### Expected behavior
`vagrant up` installs plugin and completes.

### Actual behavior
Above error happens when setting up any Vagrant project that requires some plugins.

### Steps to reproduce
Follow steps from here:
https://github.com/coreos/coreos-vagrant

### Workaround
https://github.com/hashicorp/vagrant/issues/8055#issuecomment-264201024

websitesca:
I confirm the above issue as well running Vagrant 2.1.2 on mac osx 10.13.2. the workaround from @zymzxq solved the problem for me

adolfoabegg:
Thank you @zymzxq 👍

jifbrodeur:
I can confirm that the error does not happen on 2.1.1 I downgraded vagrant and the error message: fork/exec /opt/vagrant/embedded/bin/ruby: argument list too long no longer occure.
Thanks

promethe42:
I can confirm that the error does not happen with the exact same `Vagrantfile` after downgrading to 2.1.1.

chrisroberts:
Closing this up as these issues were fixed in 2.1.5

Cheers!

PR #10030, opened 2018-07-16
Disable vagrantfile with plugin commands

### LINKED ISSUE #10013: Vagrantfile loading is not disabled when running plugin commands
### Vagrant version
2.1.2

### Host operating system
macOS 10.12.6

### Guest operating system
Ubuntu 17.10

### Vagrantfile
```ruby
system "vagrant plugin install tty-prompt"

...

require "tty-prompt"
```

OR running the command from the terminal: `vagrant plugin install tty-prompt` from within you vagrant project directory.

### Debug output
Terminal output: https://gist.github.com/FlipABit/f5e335a51ced25c47bfffa0c6598e2b8

### Expected behavior
When running plugin commands other than "list", loading of the Vagrantfile should be disabled.

### Actual behavior
Loading of the Vagrantfile is no longer disabled when running any plugin command. As a result, plugins cannot be installed while working in the directory of the Vagrantfile which requires the plugin. Running the plugin install command will throw an error as a result of the plugin not being installed when the Vagrantfile is loaded.

### Steps to reproduce
1. Require a 3rd party plugin in your Vagrantfile. Make sure that plugin is not already installed on your system.
2. Run `vagrant plugin install <plugin name>` from the Vagrantfile directory.  
3. You should be presented with an error "There was an error loading a Vagrantfile" due to the plugin not being already installed.

### References
This was introduced in v2.1.2. The offending changeset is here: https://github.com/hashicorp/vagrant/commit/1bba1f86bbec387a9297ffd0b705783805b959df. Loading of the vagrant file used to be disabled via `opts[:vagrantfile_name] = ""` (line 41) for `plugin`, `help`, or `box list` command. It is now only disabled for the `box list` command.

Setting the environment variable `VAGRANT_VAGRANTFILE=""` and then running `vagrant plugin install <plugin name>` and then unsetting VAGRANT_VAGRANTFILE temporarily fixes the issue.

chrisroberts:
Fixes #10013 #9972

### DIFF
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -44,6 +44,7 @@ argv.each_index do |i|
 
     # Do not load plugins when performing plugin operations
     if arg == "plugin"
+      opts[:vagrantfile_name] = ""
       ENV['VAGRANT_NO_PLUGINS'] = "1"
       # Only initialize plugins when listing installed plugins
       if argv[i+1] != "list"
--- a/test/unit/bin/vagrant_test.rb
+++ b/test/unit/bin/vagrant_test.rb
@@ -1,0 +1,147 @@
+require File.expand_path("../../base", __FILE__)
+
+describe "vagrant bin" do
+  include_context "unit"
+
+  let(:env) { isolated_environment }
+  let(:ui) { double(:ui) }
+  let(:argv) { [] }
+  let(:exit_code) { 0 }
+
+  let(:run_vagrant) {
+    lambda {
+      begin
+        instance_eval(
+          File.read(
+            File.expand_path(
+              "../../../../bin/vagrant", __FILE__)))
+      rescue SystemExit => e
+        e.status
+      end
+    }.call
+  }
+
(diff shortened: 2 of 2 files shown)


Tanoro:
I stubbed the `vagrantfile_name` option in `bin/vagrant` on line 47 as described above and navigated out of the project directory to manually install the plugin I needed: `vagrant-vbguest`. I navigated back to my project directory and the `vagrant up` command worked and began installing the desired box. Thanks for the reply!

herpaderpaldent:
Getting the same errors. 

Just did a fresh install of vagrant && my whole PC today and running into the issues. The workaround with installing the plugin outside of the working dir didn't work for me either.

https://github.com/eveseat/scripts/blob/master/vagrant/Vagrantfile 
This is the vagrant file i use ... tried to comment all the `vagrant-docker-compose` part out and after installing the plugin creating the vagrant machiene but i failed.

PR #10037, opened 2018-07-18
Define project specific plugins

### LINKED ISSUE #4347: Allow project-specific plugins
Overall, I'm getting to #1789 and #3468 (automatic plugin management), but those issues have been closed without any solutions, so here is a shot:

**Step 1:** Add a `--local` flag to `vagrant plugin install`. This will install the plugin into the project's `.vagrant` folder. When you run vagrant up, it should load both global and local plugins.

**Step 2:** Add a Vagrant config option for plugins, so in the Vagrantfile we can say `config.plugins = { ... }`, and then it will internally invoke the above-said local plugin install.

With this, the whole automatic plugin management thing can be taken care using Vagrant's built-in extensibility mechanisms, without having to do any monkey patching to core. How does that sound?

chrisroberts:
This PR adds support for project specific plugins in Vagrant. Project specific plugins are local to a Vagrant project (identified by a Vagrantfile) and are not available outside the local project. Project plugins can be installed by the user via a `--local` flag:

```
$ vagrant plugin install vagrant-cool-plugin --local
```

or plugins can be defined by the Vagrantfile:

```ruby
  config.vagrant.plugins = "vagrant-cool-plugin"
end
```

Plugins can be provided as a single entry (string), list of entries (array), or as a definition of entries (hash) which can set options like version or entry point. When the Vagrantfile is loaded and defined plugins are not found, the user will be prompted to install the configured list of plugins.

By default, Vagrant will unset the Vagrantfile when using `vagrant plugin` commands. This means that for things like `vagrant plugin list`, only global plugins will be shown. To see local project plugins as well, the `--local` flag must be included. You can remove this requirement by setting the `VAGRANT_LOCAL_PLUGINS_LOAD` environment variable. The reason this behavior is not default is to prevent breakages from things like this in a Vagrantfile:

```ruby
if Vagrant.has_plugin?("vagrant-plugin")
  system "vagrant plugin install vagrant-plugin"
end
```

Fixes #4347

### DIFF
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -44,7 +44,9 @@ argv.each_index do |i|
 
     # Do not load plugins when performing plugin operations
     if arg == "plugin"
-      opts[:vagrantfile_name] = ""
+      if argv.none?{|a| a == "--local" }
+        opts[:vagrantfile_name] = ""
+      end
       ENV['VAGRANT_NO_PLUGINS'] = "1"
       # Only initialize plugins when listing installed plugins
       if argv[i+1] != "list"
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -44,7 +44,7 @@ argv.each_index do |i|
 
     # Do not load plugins when performing plugin operations
     if arg == "plugin"
-      if argv.none?{|a| a == "--local" }
+      if argv.none?{|a| a == "--local" } && !ENV["VAGRANT_LOCAL_PLUGINS_LOAD"]
         opts[:vagrantfile_name] = ""
       end
       ENV['VAGRANT_NO_PLUGINS'] = "1"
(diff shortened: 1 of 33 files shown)


mcandre:
Awesome feature! Are local plugins carried over when exporting Vagrant boxes, or should downstream Vagrantfile's make sure to re-add local plugins?

chrisroberts:
If a custom Vagrantfile nested with the box defines them, then yeah, they'll be there.

goodwillcoding:
Hi folks,

this is a semi-tangential question. But is there a way to include a plugin whose source is included with the project without running an install externally? If there is a small example would be appreciated since I am not a major ruby person

Issue #8648, opened 2017-06-06
Feature request: attach secondary disk on Hyper-V


sandersaares:
Reading the available options, I notice a lack of any capability to add a secondary disk. This would be very useful in order to create a separation between persistent data (stored on the secondary disk) and app code (stored on the main image, which is re-created on demand).

While synced folders are a possible road toward a solution here, not all scenarios can cope with the "shared folder" mechanism employed there (e.g. SQL Server wants a real disk to write on).

Please correct me if I somehow just missed this feature in the documentation.

JohnEricson:
There is a pendingt request on this here https://github.com/hashicorp/vagrant/pull/6321
I think it's not merged yet because the format of configuring it in the Vagrantfile needs to change.

PR #10044, opened 2018-07-20
Feature request: attach secondary disk on Hyper-V

JohnEricson:
This is a pull request to solve reported feature request issue at https://github.com/hashicorp/vagrant/issues/8648.

The original code for this was created in issue https://github.com/hashicorp/vagrant/pull/6321. Part of it has already been merged but this part is left. I've implemented what @mwrock wanted:

* Improved configuration format. No more nested arrays. More future proof.
* Now size of disks are specified in MB instead of GB.

I couldn't implement the configuration format exactly. Instead I came up with this format which I believe is very similar:

```ruby
hv.controller type: :scsi, disks: [ "d2", 10 * 1024 ]
```

This creates a SCSI controller and a new disk that is 10 GB big.

IDE and attaching already existing disks are not implemented but configuration format for this is already implemented like this:

```ruby
# Throws TODO-error because IDE-support is not implemented yet.
hv.controller type: :ide, disks: [ "d1", 20 * 2048 ]

# Throws TODO-error when file exists because this isn't implemented yet.
hv.controller type: :scsi, disks: [ "c:/disk.vhdx" ]
```

I've also added documentation for the new feature. Looking forward to seeing this get merged into Master.

### DIFF
--- a/plugins/providers/hyperv/action/configure.rb
+++ b/plugins/providers/hyperv/action/configure.rb
@@ -35,57 +35,6 @@ module VagrantPlugins
             end
           end
           
-          
-          # TODO: Should be controllers as we want to be able to have several lines.
-          #if !env[:machine].provider_config.controllers.empty?
-          #  puts "hej: controller not empty"
-          #  #env[:ui].detail("Controllers set")
-          #  #controller.each do [key, value]
-          #  #  puts "key: #{id} value: #{value}"
-          #  #  if [type].include?(key)
-          #  #    puts "key matches type"
-          #  #  end
-          #  #end
-          #  
-          #  env[:machine].provider_config.controller.each do |key, value|
-          #    env[:ui].output("#{key} is #{value}")
-          #  end
-          #
-          #  #env[:machine].provider.driver.set_vm_integration_services(
-          #  #  env[:machine].provider_config.vm_integration_services)
-          #end
-
-          disks_to_create = []
-          env[:machine].provider_config.controllers.each { |controller|
-            #puts "configure.rb: controller: #{controller}"
-            
-            next_is_size = false
-            disk_name = ''
-            controller[:disks].each { |i|
-              #puts "configure.rb: i: #{i} disk_name: #{disk_name}"
-              if !next_is_size
-                if File.file?(i)
-                  create_disk = false 
-                  filename_for_disk = i
-                  next_is_size = false
-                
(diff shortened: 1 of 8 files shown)


almightyfoon:
Whats the status on this PR? It would be very helpful to have.

JohnEricson:
Hi @almightyfoon! I'm waiting on a developer to make an approved review of it. I'm very dedicated to get this included as well as I also rely on this feature, so I would really appreciate it if @chrisroberts or anyone else can make it happen.

JohnEricson:
I hope everything should be done for merging now. Please, let me know if you need me to done anything more.

mloskot:
Can we have this merged?

Yivan:
@JohnEricson Very very nice new feature !!
We managed it through custom scripts until now. Main purpose was having a dedicated disk for storing the code, so recreating the VM could be simply done without to lost the code.

Just, when doing a `vagrant destroy` is this secondary disk keeped or deleted (like the main one).
Main purpose of having this secondary disk is to preserve it through recreation of the VM.
If not, may a parameter be added ?
I think it is very valuable for usefullness od this new feature!

Thansk a lot.

JohnEricson:
@Yivan 
> Just, when doing a `vagrant destroy` is this secondary disk keeped or deleted (like the main one) ?

I hope this feature can help you too. Always better if it can be solved by Vagrant out of the box. Now the secondary disk is deleted just like the main one.

But I've implemented support in the config to specify a current disk like this:

```ruby
# Throws TODO-error when file exists because this isn't implemented yet.
hv.controller type: :scsi, disks: [ "c:/disk.vhdx" ]
```

Although the actual logic to attach the disk isn't implemented yet. But perhaps the default behavior when an already existing disk is attached like that could be to preserve it on mount, or specify that behavior using an extra option.

Yivan:
@JohnEricson Thanks for your return on this. Great feature yes!

Having a parameter to prevent destorying disk during `vagrant destroy` would be really great. If this parameter is activated, so the disk should be unmounted before destroying the VM. This is the way we handle it manually actually (with triggers).

Thanks!

briancain:
Hey there everyone! I'm going to go ahead and close this issue, as I have just made a pull request that takes advantage of the recent generic disk config options in Vagrant. You can find that PR here: #11541 It offers a lot of extra goodies such as resizing disks (similar to the VirtualBox implementation). Thanks!

PR #10165, opened 2018-08-31
Update Vagrant.has_plugin? helper to function prior to plugin loading

### LINKED ISSUE #10161: Vagrant.has_plugin?("vagrant-vbguest") always returns false after upgrade to 2.1.4
I upgraded from 2.1.2 to 2.1.4 this morning, and now the plugin test for the VB Guest plugin always returns false.

The section of my Vagrantfile that is no longer working looks like this

unless Vagrant.has_plugin?("vagrant-vbguest")
    raise 'The Vagrant VBGuest plugin must be install prior to building this VM (vagrant plugin install vagrant-vbguest)'
end

This worked in 2.1.2 and fails in 2.1.4 (I didn't try 2.1.3).

I'm running OS X High Sierra, and the Vagrantfile is trying to create a guest for CentOS - but it doesn't get very far.

### Vagrantfile
```ruby
  
  unless Vagrant.has_plugin?("vagrant-vbguest")
    raise 'The Vagrant VBGuest plugin must be install prior to building this VM (vagrant plugin install vagrant-vbguest)'
  end

  config.vm.define 'xxxx-centos-base'
  config.vm.box = 'centos/7'


  config.vm.provider 'virtualbox' do |virtualbox|
    virtualbox.name = 'XXXX Development Base'
    virtualbox.cpus = 2
    virtualbox.memory = 4096
  end

  config.vm.provision 'ansible_local' do |ansible|
    ansible.compatibility_mode = '2.0'
    ansible.playbook = 'xxxx-centos-base.yml'
    ansible.become = true
  end

end
```

chrisroberts:
Due to the Vagrantfile being loaded prior to plugin loading to determine
project local plugin information the Vagrant.has_plugin? helper will always
return false when the Vagrantfile is first loaded. To prevent this behavior
we can check for plugins in the plugin data files prior to the plugins
being loaded, and after they have been loaded we can fallback to the
original specification based check.

Fixes #10161

### DIFF
--- a/lib/vagrant.rb
+++ b/lib/vagrant.rb
@@ -151,16 +151,9 @@ module Vagrant
       return true if plugin("2").manager.registered.any? { |p| p.name == name }
     end
 
-    # Make the requirement object
-    version = Gem::Requirement.new([version]) if version
-
     # Now check the plugin gem names
     require "vagrant/plugin/manager"
-    Plugin::Manager.instance.installed_specs.any? do |s|
-      match = s.name == name
-      next match if !version
-      next match && version.satisfied_by?(s.version)
-    end
+    Plugin::Manager.instance.plugin_installed?(name, version)
   end
 
   # Returns a superclass to use when creating a plugin for Vagrant.
--- a/lib/vagrant/plugin/manager.rb
+++ b/lib/vagrant/plugin/manager.rb
@@ -41,12 +41,14 @@ module Vagrant
         @system_file = StateFile.new(system_path) if system_path && system_path.file?
 
         @local_file = nil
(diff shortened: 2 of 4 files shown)


plowman:
Thanks so much for fixing this, @chrisroberts + @briancain! This has been haunting me for days now.

fnordfish:
Since lot's of projects use constructs like `exit unless Vagrant.has_plugin?("foo")` and both, Vagrant 2.1.3 and 2.1.4 will return a false negative. You'd need to downgrade to Vagrant 2.1.2 as a workaround until this fix is released.

PR #10190, opened 2018-09-07
Misleading warning on Windows (due to improper String comparison)

slackfan:
In Windows a misleading warning may appear, e.g.
```
==> default: This machine used to live in C:/ws/test_vagrant but
 it's now at c:/ws/test_vagrant. 
```

The implemented String comparison is misleading on Windows as it does not consider that paths on Windows are case insensitive. The pull request is to address that by using File.identical?() instead. 

Signed-off-by: Alexander Fischer <github@ordnungsdienst.org>

### DIFF
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -4,7 +4,6 @@ require "digest/md5"
 require "thread"
 
 require "log4r"
-require "vagrant/util/platform"
 
 module Vagrant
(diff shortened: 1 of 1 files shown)


briancain:
Thanks for the contribution @slackfan ! Could you please add a simple test for this? Thanks!

briancain:
Closing this pull request for https://github.com/hashicorp/vagrant/pull/10220. Thanks for your contribution!

PR #10211, opened 2018-09-15
amended ansible local documentation

halfbaked:
included info on what the become flag does and an alternative I found useful when needed to run just some roles as root and others without.

### DIFF
--- a/website/source/docs/provisioning/ansible_local.html.md
+++ b/website/source/docs/provisioning/ansible_local.html.md
@@ -141,6 +141,12 @@ This section lists the _specific_ options for the Ansible Local provisioner. In 
 
     The default value is `/tmp/vagrant-ansible`
 
+- `become` (boolean) - When set to true runs ansible as root. 
+
+    The default value is `false`.
+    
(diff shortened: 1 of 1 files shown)


gildegoma:
@halfbaked thank you for your proposal, but I'm sorry to say that I'm not going to accept it because 

* the (`ansible` and `ansible_local`) common `become` option is already documented: https://www.vagrantup.com/docs/provisioning/ansible_common.html#become
* The fact that `become` can also be set a the ansible **role** or **task** level is relevant for the ansible tool documentation, but beyond the scope of Vagrant's documentation.

I hope you agree with these arguments and thank you again for your interest in improving Vagrant/Ansible ease of use!

halfbaked:
Hi Gilles,

No problem at all. 

I just started into vagrant and ansible this week. In my experience as a novice, getting to grips with it, and getting through that fog of overwhelming information, it would have been useful to have that info there. 

Both your points are 100% valid of course. I appreciate your polite and courteous reply, and loving vagrant so far! Keep up the good work.

PR #10214, opened 2018-09-17
Add in logic to restart NetworkManager.service if it is enabled

jdoss:
This PR adds in logic to check to see if `/usr/bin/systemctl` is present and if `NetworkManager.service` is active. If it is not, it will test for the sysvinit `/etc/init.d/network` initscript. From there it will restart the respective service to set the hostname correctly. 

As of initscripts-9.83-1.fc29.x86_64 in Fedora 29 sysvinit `/etc/init.d/network` initscript [is no longer being included](https://bugzilla.redhat.com/show_bug.cgi?id=1596062#c1) and this PR will ensure that Vagrant can correctly set the hostname on old and new RHEL, CentOS and Fedora versions. 

Related URLs for more context:
https://bugzilla.redhat.com/show_bug.cgi?id=1624068
https://pagure.io/cloud-sig/issue/291#comment-528948

### DIFF
--- a/plugins/guests/redhat/cap/change_host_name.rb
+++ b/plugins/guests/redhat/cap/change_host_name.rb
@@ -29,9 +29,9 @@ module VagrantPlugins
               }
 
               # Restart network
-              if (test -f /etc/init.d/network && /etc/init.d/network status &> /dev/null ); then
+              if test -f /etc/init.d/network; then
                 service network restart
-              elif (test -f /usr/lib/systemd/system/NetworkManager.service && systemctl is-enabled NetworkManager.service &> /dev/null ); then
+              elif systemctl -q is-enabled NetworkManager.service; then
                 systemctl restart NetworkManager.service
               else
                 printf "Could not restart the network to set the new hostname!\n"
--- a/plugins/guests/redhat/cap/change_host_name.rb
+++ b/plugins/guests/redhat/cap/change_host_name.rb
@@ -29,10 +29,10 @@ module VagrantPlugins
               }
 
               # Restart network
-              if test -f /etc/init.d/network; then
-                service network restart
-              elif systemctl -q is-enabled NetworkManager.service; then
+              if (test -f /usr/bin/systemctl && systemctl -q is-active NetworkManager.service); then
                 systemctl restart NetworkManager.service
+              elif test -f /etc/init.d/network; then
+                service network restart
               else
                 printf "Could not restart the network to set the new hostname!\n"
               fi
--- a/plugins/guests/redhat/cap/change_host_name.rb
+++ b/plugins/guests/redhat/cap/change_host_name.rb
@@ -29,7 +29,13 @@ module VagrantPlugins
               }
 
               # Restart network
-              service network restart
+              if (test -f /etc/init.d/network && /etc/init.d/network status &> /dev/null ); then
+                service network restart
+              elif (test -f /usr/lib/systemd/system/NetworkManager.service && systemctl is-enabled NetworkManager.service &> /dev/null ); then
+                systemctl restart NetworkManager.service
+              else
+                printf "Could not restart the network to set the new hostname!\n"
+              fi
             EOH
           end
         end
--- a/plugins/guests/redhat/cap/change_host_name.rb
+++ b/plugins/guests/redhat/cap/change_host_name.rb
@@ -35,6 +35,7 @@ module VagrantPlugins
                 service network restart
               else
                 printf "Could not restart the network to set the new hostname!\n"
+                exit 1
               fi
             EOH
           end
--- a/test/unit/plugins/guests/redhat/cap/change_host_name_test.rb
+++ b/test/unit/plugins/guests/redhat/cap/change_host_name_test.rb
@@ -31,7 +31,7 @@ describe "VagrantPlugins::GuestRedHat::Cap::ChangeHostName" do
       expect(comm.received_commands[1]).to match(/\/etc\/sysconfig\/network-scripts\/ifcfg/)
       expect(comm.received_commands[1]).to match(/hostnamectl set-hostname --static '#{name}'/)
       expect(comm.received_commands[1]).to match(/hostnamectl set-hostname --transient '#{name}'/)
-      expect(comm.received_commands[1]).to match(/service network restart/)
+      expect(comm.received_commands[1]).to match(/service network restart|systemctl restart NetworkManager.service/)
     end
 
     it "does not change the hostname if already set" do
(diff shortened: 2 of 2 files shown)


plugins/guests/redhat/cap/change_host_name.rb
```diff
@@ -29,7 +29,13 @@ def self.change_host_name(machine, name)
...
 
               # Restart network
-              service network restart
+              if (test -f /etc/init.d/network && /etc/init.d/network status &> /dev/null ); then
+                service network restart
+              elif (test -f /usr/lib/systemd/system/NetworkManager.service && systemctl is-enabled NetworkManager.service &> /dev/null ); then
+                systemctl restart NetworkManager.service
+              else
+                printf "Could not restart the network to set the new hostname!\n"
+              fi
```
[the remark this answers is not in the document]
dustymabe: another thing to consider.. if networkmanager is active do we even need to do a `systemctl restart NetworkManager.service` at all?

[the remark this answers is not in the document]
dustymabe: > if networkmanager is active do we even need to do a `systemctl restart NetworkManager.service` at all?

@tyll do you know?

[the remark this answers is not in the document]
jdoss: I am down for simplifying things, but I wanted each check to cover any possible edge case to prevent issues. That most likely was a bit overkill.

plugins/guests/redhat/cap/change_host_name.rb
```diff
@@ -29,7 +29,13 @@ def self.change_host_name(machine, name)
               }
 
               # Restart network
-              service network restart
+              if test -f /etc/init.d/network; then
+                service network restart
+              elif systemctl -q is-enabled NetworkManager.service; then
```
dustymabe: WDYT about `is-active` as opposed to `is-enabled` ?

plugins/guests/redhat/cap/change_host_name.rb
```diff
@@ -29,7 +29,13 @@ def self.change_host_name(machine, name)
               }
 
               # Restart network
-              service network restart
+              if test -f /etc/init.d/network; then
```
dustymabe: i still think we need to check NM first here. It's possible a system has both legacy network plus NetworkManager installed. Checking `systemctl -q is-active NetworkManager.service` first would be best. If it's not active and network scripts exist then try that. If network scripts don't exist then error out.

jdoss: I am fine with that but older vagrant images like CentOS will bomb out. I will add in a check to ensure that systemctl is present when doing that check first.

```
# systemctl
-bash: systemctl: command not found
# cat /etc/redhat-release 
CentOS release 6.10 (Final)```

File: plugins/guests/redhat/cap/change_host_name.rb (hunk 2 above)
Comment (jdoss) [reply plugins/guests/redhat/cap/change_host_name.rb]: :+1:

File: plugins/guests/redhat/cap/change_host_name.rb (hunk 3 above)
Comment (dustymabe) [reply plugins/guests/redhat/cap/change_host_name.rb]: > I am fine with that but older vagrant images like CentOS will bomb out.

good point. so we can say `if -e /usr/bin/systemctl && systemctl -q is-active NetworkManager.service` or something along those lines

File: plugins/guests/redhat/cap/change_host_name.rb [hunk 4]
```diff
@@ -29,7 +29,13 @@ def self.change_host_name(machine, name)
...
 
               # Restart network
-              service network restart
+              if (test -f /usr/bin/systemctl && systemctl -q is-active NetworkManager.service); then
+                systemctl restart NetworkManager.service
+              elif test -f /etc/init.d/network; then
+                service network restart
+              else
+                printf "Could not restart the network to set the new hostname!\n"
+              fi
```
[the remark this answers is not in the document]
jdoss: What about leaving the `printf` with an `exit 1` to exit with a failure?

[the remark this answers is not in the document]
briancain: Yes, that should be ok.

I was trying to think about how we could show a proper Vagrant error, but that will be a larger change than what this PR needs since all the other hostname capabilities just fail if it can't set the hostname. So for now, this is ok.

jdoss:
@briancain OK we should be good now with your requested changes. Let me know if I need to do anything else to get this merged. :rocket:

dustymabe:
thanks for reviewing @briancain - thanks for the fixups @jdoss

briancain:
Hey @jdoss - thanks for the update! Sorry to ask this again, but I was expecting that the tests would cover all three cases:

When:
- `(test -f /usr/bin/systemctl && systemctl -q is-active NetworkManager.service)`
  + systemctl and NetworkManager.service is used
- `test -f /etc/init.d/network`
  +  uses the `service` command
- When case 1 and 2 are false
  + a `printf` and exit code 1 happens

You should be able to copy most of the `it "sets the hostname" do` context for each case, use `stub_command` to stub out those if-conditional results, and then update the `expect` `received_command` statements respectively for each context. Let me know if that makes sense, thank you! ❤️

test/unit/plugins/guests/redhat/cap/change_host_name_test.rb
```diff
@@ -31,13 +31,40 @@
...
+      expect(comm.received_commands[1]).to match(/service network restart|systemctl restart NetworkManager.service/)
     end
 
     it "does not change the hostname if already set" do
       comm.stub_command("hostname -f | grep '^#{name}$'", exit_code: 0)
       cap.change_host_name(machine, name)
       expect(comm.received_commands.size).to eq(1)
     end
+
+    context "restarts the network" do
```
shotop: Hey @briancain - @jdoss asked if I minded jumping into this branch to assist with some tests. I'm a coworker of his who works in testing. I was hoping to get your input on a flaw that I know exists with my tests here.  As you can see below, I'm stubbing the commands to comm that @jdoss introduced and I'm returning 0 or 1 exit_codes depending on the logic I want to test.  The problem I have is that all three assertions will pass, regardless of which of the three commands get stubbed and their return codes.  The reason for this is that `comm.received_commands[1]` is always the entire bash string that gets built up in the .change_host_name method.  I'm not sure I see at the moment how I can assert that comm behaved differently when it received `"test -f /etc/init.d/network"` vs the other commands.  Is there a way that I can verify that comm sent off the `systemctl restart NetworkManager.service` message for example?  I searched in the other plugin test files for precedent, but I'm not seeing anywhere where we test embedded bash IF logic in this way.  I could totally be wrong!  Wondering if you have any insight?

shotop: Awesome - happy to help!

dustymabe:
hey @briancain - do you know when @chrisroberts will be able to pick up this review? Do you anticipate it to take a long time? The reason I ask is we have a Fedora 29 release coming out soon and we'd like to backport the fix that gets merged to the version of vagrant in fedora 29 today.

briancain:
@dustymabe - I've been told the plan is to work on this today. So I would expect the patch to land in the next couple of days!

briancain:
Closing this PR now that the fixup has been merged. Thanks again everyone! 🎉

PR #10222, opened 2018-09-20
Update Docker containerization docs

zwass:
The docs indicate that most systems must use a VM for containerization. This is no longer the case as Docker For Mac now uses HyperKit.

See https://docs.docker.com/v17.12/docker-for-mac/docker-toolbox/

### DIFF
--- a/website/source/intro/vs/docker.html.md
+++ b/website/source/intro/vs/docker.html.md
@@ -15,8 +15,7 @@ that can consistently run software as long as a containerization system exists.
 
 Containers are generally more lightweight than virtual machines, so starting
 and stopping containers is extremely fast. Docker uses the native
-containerization functionality on macOS and Linux, while it must start a Linux
-VM for compatibility with Windows.
+containerization functionality on macOS, Linux, and Windows.
 
 
 Currently, Docker lacks support for certain operating systems (such as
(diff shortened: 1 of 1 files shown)


website/source/intro/vs/docker.html.md
```diff
@@ -14,9 +14,10 @@ workflow across multiple operating systems. Docker is a container management
 that can consistently run software as long as a containerization system exists.
 
 Containers are generally more lightweight than virtual machines, so starting
-and stopping containers is extremely fast. Most common development machines
-don't have a containerization system built-in, and Docker uses a virtual machine
-with Linux installed to provide that.
+and stopping containers is extremely fast. Docker uses the native
+containerization functionality on macOS and Linux, while it must start a Linux
+VM for compatibility with Windows.
```
briancain: Hmm, I think this line is technically false as well? You can have windows containers now https://docs.docker.com/docker-for-windows/install/#what-to-know-before-you-install
https://docs.microsoft.com/en-us/virtualization/windowscontainers/about/

I would maybe just change that last line to mention both for Windows containers and a Linux VM.

zwass: Great point. Looks like the line should say "Docker uses the native containerization functionality on macOS, Linux, and Windows."

briancain: Perfect! 👍Yeah that makes sense to me. Get that updated and it'll be good to merge ❤️

Issue #10095, opened 2018-08-06
Native Triggers Missing `instead_of`?


jamiejackson:
### Vagrant version

Vagrant 2.1.2

### Host operating system

ProductName:	Mac OS X
ProductVersion:	10.13.6
BuildVersion:	17G65

### Guest operating system

ubuntu/trusty64 (but not getting that far)

### Vagrantfile

```ruby
  # prevent networking issues: https://github.com/hashicorp/vagrant/issues/9546
  # why not trigger on reload? https://github.com/emyl/vagrant-triggers/issues/87
  config.trigger.instead_of :reload do
    warn "Because of a vagrant issue (https://github.com/hashicorp/vagrant/issues/9546)"
    warn "  use `vagrant halt && vagrant up` instead of `vagrant reload`"
    exit()
  end
```

### Debug output

```
==> default: Running trigger...
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The 'trigger' provisioner could not be found.

trigger:
* The following settings shouldn't exist: instead_of
```
### Expected behavior

`instead_of` works as it did in `vagrant-triggers` plugin

### Actual behavior

`instead_of` is unimplemented?

briancain:
Hi @jamiejackson - at the moment there is no supported way of overriding a Vagrant command with triggers. If you simply want to block someone from running a command like `reload`, you can do this:


```ruby
config.trigger.before :reload do |trigger|
    trigger.warn = "Because of a vagrant issue (https://github.com/hashicorp/vagrant/issues/9546)\nuse `vagrant halt && vagrant up` instead of `vagrant reload`"
    trigger.run = {inline: "false"}
end
```

```
brian@localghost:vagrant-sandbox % be vagrant reload guest                                                ±[●][master]
==> guest: Running triggers before reload ...
==> guest: Running trigger...
==> guest: Because of a vagrant issue (https://github.com/hashicorp/vagrant/issues/9546) use `vagrant halt && vagrant up` instead of `vagrant reload`
    guest: Running local: Inline script
    guest: false
==> guest: Trigger run failed
==> guest: A script exited with an unacceptable exit code 1.
A script exited with an unacceptable exit code 1.

```

PR #10232, opened 2018-09-24
Add abort option to core triggers

briancain:
This commit adds a new option `abort`, which when configured, will exit
the Vagrant process completely. If set to `true`, it will exit cleanly
with exit code 0. Otherwise, the exit code can be configured.

Fixes: #10095

### DIFF
--- a/plugins/kernel_v2/config/vm_trigger.rb
+++ b/plugins/kernel_v2/config/vm_trigger.rb
@@ -72,6 +72,12 @@ module VagrantPlugins
       # @return [Integer, Array]
       attr_accessor :exit_codes
 
+      # If set to true, trigger will halt Vagrant immediately and exit 0
+      # Can also be configured to have a custom exit code
+      #
+      # @return [Integer]
+      attr_accessor :abort
+
       def initialize(command)
         @logger = Log4r::Logger.new("vagrant::config::vm::trigger::config")
 
@@ -84,6 +90,7 @@ module VagrantPlugins
         @run = UNSET_VALUE
         @run_remote = UNSET_VALUE
         @exit_codes = UNSET_VALUE
+        @abort = UNSET_VALUE
 
         # Internal options
         @id = SecureRandom.uuid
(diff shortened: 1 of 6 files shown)


plugins/kernel_v2/config/vm_trigger.rb
```diff
@@ -149,6 +157,9 @@ def finalize!
           @run_remote = new_run
         end
 
+        if @abort == true
+          @abort = 0
```
chrisroberts: I think that the default value should be a non-zero value here. Since we are aborting the current run, that implies that the run failed and a successful exit code should not be returned. Perhaps defaulting this value to `1` instead.

templates/locales/en.yml
```diff
@@ -1768,6 +1770,12 @@ en:
         on_error_bad_type: |-
           Invalid type set for `on_error` on trigger for command '%{cmd}'. `on_error` can
           only be `:halt` (default) or `:continue`.
+        abort_bad_type: |-
+          Trigger option `abort` for command '%{cmd}' must be either set to
+          `true` (defaults to exit code 0) or an integer to use as an exit code.
```
chrisroberts: Update default exit code value

test/unit/plugins/kernel_v2/config/vm_trigger_test.rb
```diff
@@ -93,6 +94,12 @@ def assert_valid
         expect(a).to be_a(Symbol)
       end
     end
+
+    it "converts aborts true to exit code 0" do
+      cfg.finalize!
+
+      expect(cfg.abort).to eq(0)
```
chrisroberts: update default test exit code

website/source/docs/triggers/configuration.html.md
```diff
@@ -44,3 +44,5 @@ The trigger class takes various options.
 * `warn` (string) - A warning message that will be printed at the beginning of a trigger.
 
 * `exit_codes` (integer, array) - A set of acceptable exit codes to continue on. Defaults to `0` if option is absent. For now only valid with the `run` option.
+
+* `abort` (integer,boolean) - An option that will exit the running Vagrant process once the trigger fires. If set to `true`, Vagrant will use exit code 0. Otherwise, an integer can be provided and Vagrant will it as its exit code when aborting.
```
chrisroberts: Update default exit code

PR #10245, opened 2018-09-27
Add support for passing ssh config file in via config

fatmcgav:
This is a in order to allow using `ProxyCommand` SSH config to allow Vagrant to access hosts behind a bastion or jump host.

Couldn't get the tests to run locally, due to local plugins conflicting with the test env :(


fatmcgav:
This is ready for review...

chrisroberts:
@fatmcgav Thanks for this PR. There were a handful of changes that needed to be applied so I built off your branch and added the adjustments which are now contained withing #10496. I'm going to close this up in favor of the new PR, but your changes will be getting merged in. Thanks again for this PR.

Cheers!

Issue #10282, opened 2018-10-10
Sync folders broken in Docker containers after host VM restart on Windows


MartinMystikJonas:
### Vagrant version
Tested on 2.0.2 and 2.0.15

### Host operating system
Windows 10

### Guest operating system
Linux (Debian)

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|

  config.ssh.insert_key = false
  config.vm.synced_folder '.', '/vagrant', disabled: true

  config.vm.provider "docker" do |d|
        d.force_host_vm = true
        d.vagrant_machine = "vagrant-dockerhost"
        d.vagrant_vagrantfile = "Vagrantfile.winhost"
    end

  config.vm.define "shopmanager", primary: true do |a|
    a.vm.provider "docker" do |d|
      d.image = "***-php"
      d.ports = ["8100:80", "8107:8007"]
      d.name = "shopmanager"
      d.has_ssh = true
      d.create_args = ["--privileged", "--hostname=shopmanager"]
    end
    a.vm.synced_folder ".", "/app"
  end

end
```

### Debug output

First vagrant up:
```
...
==> shopmanager: Docker host is required. One will be created if necessary...
    shopmanager: Docker host VM is already ready.
==> shopmanager: Syncing folders to the host VM...
    vagrant-dockerhost: Mounting shared folders...
    vagrant-dockerhost: /var/lib/docker/docker_153917231_57929 => C:/Users/mpjjo/PhpstormProjects/shopmanager
...
==> shopmanager: Creating the container...
    shopmanager:   Name: shopmanager
    shopmanager:  Image: git.lwg.cz:5005/dev/gj-dev-php
    shopmanager: Volume: /var/lib/docker/docker_153917231_57929:/app
...
```

After restart of vagrant-dockerhost VM and running vagrant up
```
...
==> shopmanager: Docker host is required. One will be created if necessary...
    shopmanager: Docker host VM is already ready.
==> shopmanager: Syncing folders to the host VM...
    vagrant-dockerhost: Mounting shared folders...
    vagrant-dockerhost: /var/lib/docker/docker_1539172319_11189 => C:/Users/mpjjo/PhpstormProjects/shopmanager
==> shopmanager: Vagrant has noticed that the synced folder definitions have changed.
==> shopmanager: With Docker, these synced folder changes won't take effect until you
==> shopmanager: destroy the container and recreate it.
==> shopmanager: Starting container...
...
```

### Expected behavior
[...]

PR #10286, opened 2018-10-10
Deterministic host VM synced folder location for Docker provider 

MartinMystikJonas:
To prevent broken sync after host VM shutdown

Fix for #10282

### DIFF
--- a/plugins/providers/docker/action/host_machine_sync_folders.rb
+++ b/plugins/providers/docker/action/host_machine_sync_folders.rb
@@ -109,8 +109,7 @@ module VagrantPlugins
 
               # If we specify exact then we know what we're doing
               if !data[:docker__exact]
-                data[:guestpath] =
-                  "/var/lib/docker/docker_#{Time.now.to_i}_#{rand(100000)}"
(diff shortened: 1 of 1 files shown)


MartinMystikJonas:
@briancain I am not sure how to write test for thing like this. Basic idea is 1) create docker container using host VM on windows 2) shutdown host VM 3) vagrant up. Overy other behaviour is already covered by tests for Docker synced fodlers. Unfortunatelly writing test of this kind is probably beyond my ruby skills. Could you give me some hint of point to similar testing scenario somewhere? I could at least try :-)

briancain:
@MartinMystikJonas - Welp, apparently this class has no tests :man_facepalming: Ok! In that case, I'll probably take that over. I was hoping there would already be a class of unit tests for this, but I guess not :) That would explain why no tests failed.

Anyway I'll probably get your commit on a branch and write up all the tests for this class myself in a new PR. Thanks though!

briancain:
Going to close this now that I've opened https://github.com/hashicorp/vagrant/pull/10311. It has your changes in this PR plus some unit tests. Thanks again!

Issue #10300, opened 2018-10-16
Setting hostname in Ubuntu 18 with Vagrant conflicts with systemd-networkd dhcp


emusgrave:
I found that setting the hostname in Ubuntu 18 with Vagrant conflicts with systemd-networkd dhcp. In particular, when I set the hostname via my Vagrantfile, I end up getting 2 IP addresses on one network adapter because both systemd-networkd and dhclient are running and getting DHCP leases.

I see the offending code in the [change_host_name.rb](vagrant/plugins/guests/debian/cap/change_host_name.rb) file, but I'm not sure what a proper fix is because I'm not exactly sure why the author @hakoerber put dhclient in there to being with.

### Workaround
I removed the `config.vm.hostname` from my Vagrantfile and put it into an Ansible script.

### Vagrant version
2.1.5

### Host operating system
Windows 10

### Guest operating system
Ubuntu 18.04.1 LTS

### Vagrantfile
```ruby
config.vm.hostname = "myhost"
```

### Expected behavior
Just set the hostname.

### Actual behavior
Setting the hostname on Debian causes the `dhclient` process to be run. On Ubuntu 18 the DHCP is already being set with systemd-networkd configurations, so running dhclient causes the network adapter to fetch a second IP address. 

### Steps to reproduce
1. Configure network adapter for systemd
2. Set hostname with Vagrantfile
3. `vagrant up`
4. SSH into machine and run `ip addr`, see a primary and secondary IP for your network adapter

### References
See file: [change_host_name.rb](vagrant/plugins/guests/debian/cap/change_host_name.rb)

hakoerber:
Hey @emusgrave!

> I see the offending code in the change_host_name.rb file, but I'm not sure what a proper fix is because I'm not exactly sure why the author @hakoerber put dhclient in there to being with.

The change was initially only for debian operating systems. I did not test it with Ubuntu, because I didn't know it was used for Ubuntu in the first place.

Too bad we now have this inconsistency. Does you change work for both Ubuntu and Debian?

briancain:
Hey @hakoerber - I've actually made a PR to fix this behavior. The original workaround you introduced shouldn't be required anymore with my change. Essentially after changing the hostname, Vagrant will restart networking on the guest, which should renew the DHCP lease. So in this case we won't need to use `dhclient` which had a but on older versions of debian.

hakoerber:
Ok, that sounds nice. Thanks for the update!

Issue #9763, opened 2018-05-03
vagrant debian guest stuck after setting hostname


asoki:
### Vagrant version
Vagrant 2.0.4 (2.0.3, 2.1.0)
vagrant-vmware-fusion (5.0.4)

### Host operating system
macOS 10.13.4 (17E199)
running VMWare Fusion Professional Version 10.1.1 (7520154)

### Guest operating system
debian wheezy

### Vagrantfile
```ruby
Vagrant.configure(2) do |config|
  config.vm.define 'foodev' do |foo_config|
    foo_config.vm.box = 'foo'
    foo_config.vm.box_url = 'file:///redacted-path/debian-7.7-64bit.box'
    foo_config.vm.host_name = 'foodev'
    foo_config.vm.network "private_network", ip: "192.168.116.80"
    foo_config.vm.synced_folder "./","/foo-server/foo-code/foo-trunk", id:"foodev-code-sync", type:"nfs", mount_options: ["actimeo=2"]
    foo_config.vm.provision "shell", path: "vagrant_config/foodev_provisioner.sh"
    #foo_config.ssh.insert_key = false
    foo_config.ssh.private_key_path="~/.vagrant_pk/vagrant.foo.redacted.tld_id_rsa"
    foo_config.vm.provider "vmware_fusion" do |v|
      v.vmx['displayname'] = 'foo-Trunk foo (Linux)'
      v.vmx["memsize"] = "8192"
      v.vmx["numvcpus"] = "4"
      v.gui = true
    end
  end
end
```

### Debug output
[https://gist.github.com/asoki/c2d95a3fb243a4bb3d6cb149c62758cc#file-vagrant-debug-log](https://gist.github.com/asoki/c2d95a3fb243a4bb3d6cb149c62758cc#file-vagrant-debug-log)

### Expected behavior
the machine should boot

### Actual behavior
Vagrant get stuck after setting the hostname. (btw, this machine did booted yesterday without any problems and was shut down with vagrant halt)
Inspecting the booted machine reveals, that none of the both network interfaces has an ip.
The last log entry in the syslog is:
```
May  3 08:55:57 foodev acpid: waiting for events: event logging is off
May  3 08:55:57 foodev dhclient: DHCPREQUEST on eth0 to 255.255.255.255 port 67
May  3 08:55:57 foodev dhclient: DHCPOFFER from 192.168.116.254
May  3 08:55:57 foodev dhclient: DHCPACK from 192.168.116.254
[...]

PR #10330, opened 2018-10-24
Reliably restart Debian and Ubuntu guests networking on setting hostname

briancain:
This pull request adds some additional logic that falls back to using the
ifdown/ifup tools to restart networking. On Ubuntu 14.04, the init
script was designed to always fail to restart newtorking, so it needs
to use the ifdown/up tools instead.

This pull request also uses a workaround due to how older debian
and ubuntu systems fail to properly restart networking. Instead of
relying on the init scripts or ifup/down tools to restart each interface,
this commit instead restarts each interface individually.

Related PR #10301
Related issues:  #9763 #10300
Tested against this Vagrantfile: https://github.com/briancain/congenial-octo-palm-tree/blob/master/debian/Vagrantfile against these OS's:

- debian 7
- debian 8
- debian 9
- ubuntu 12
- ubuntu 14
- ubuntu 16
- ubuntu 18

### DIFF
--- a/plugins/guests/debian/cap/change_host_name.rb
+++ b/plugins/guests/debian/cap/change_host_name.rb
@@ -92,9 +92,9 @@ module VagrantPlugins
           interfaces.each do |iface|
             logger.debug("Restarting interface #{iface} on guest #{machine.name}")
             if nettools
-             restart_command = "ifdown #{iface} && ifup #{iface}"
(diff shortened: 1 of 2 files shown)


SaschaHowe:
👍 works for me. thank you very much for looking into it and fixing it

briancain:
@sruffell - thanks for testing it! I don't mind, especially since this was caught before release :sweat_smile:

Issue #9356, opened 2018-01-10
Script: get_network_config.ps1 error "Failed to determine IP address" occurs often at random points of provisioning (suspected regression)


sandersaares:
Please note that the Vagrant issue tracker is reserved for bug reports and
enhancements. For general usage questions, please use the Vagrant mailing list:
https://groups.google.com/forum/#!forum/vagrant-up. Thank you!

### Vagrant version
Vagrant 2.0.1

### Host operating system
Windows 10 v1709 and Windows 2016

### Guest operating system
Windows 2016 on Hyper-V

### Vagrantfile
https://gist.github.com/sandersaares/c2946dbd47732c7e09bf1ef4d117a30d

### Debug output
[vagrant_stdout.log](https://github.com/hashicorp/vagrant/files/1617797/vagrant_stdout.log)
[vagrant_stderr.log](https://github.com/hashicorp/vagrant/files/1617798/vagrant_stderr.log)

### Expected behavior

VM is provisioned without problems.

### Actual behavior

At random steps in provisioning, I get

```
An error occurred while executing a PowerShell script. This error
is shown below. Please read the error message and see if this is
a configuration error with your system. If it is not, then please
report a bug.

Script: get_network_config.ps1
Error:

Failed to determine IP address
```

Sometimes this comes immediately after "Waiting for machine to boot", sometimes at any other point after that. It does not appear to come before that.

I have tried to also monitor the data returned by `Get-VmNetworkAdapter` in parallel and all my queries have properly returned an IP address (though perhaps it only disappears for an instant, breaking it for Vagrant but not for my separate queries?).

Back when I was on Vagrant 1.x.x I was using the same Vagrantfile and did not experience this issue, so it seems a regression.

### Steps to reproduce
1. `vagrant up`
2. If no error, `vagrant destroy -f` and go to step 1

sandersaares:
I have updated the gist to reduce the Vagrantfile to a smaller scope that still reproduces the issue. The attached log is from the first revision that had a bit more steps included.

PierreR:
Same problem here.

m-ritchie:
I have been seeing this issue too. This is on the snag list of issues for my project so if anyone has any thoughts I'm all ears! I'm running on Windows 10 v1709 and have been trying to narrow down any potential environmental issues. 

[...]

PR #10347, opened 2018-10-29
Fixes #10229: Add timeout for getting ip addr in ssh_info

### LINKED ISSUE #10229: Race condition after setting host name causes "up" to fail.
A race condition in the rename host procedure causes the provisioning to fail.
This is a specific case of the error described in issue #9356. I've created a separate case because the cause of #9356 is somewhat ambiguous and is likely caused by more than one scenario. This case only deals with errors caused while waiting for a reboot after renaming the machine during a provisioning of a Windows guest.

### Vagrant version
Vagrant 2.1.5

### Host operating system
Windows 10 Pro Version 1803 Build 17134.286

### Guest operating system
Windows 7 Enterprise with SP1

###Provider
Hyper-V

### Vagrantfile
```
# -*- mode: ruby -*-
# vi: set ft=ruby :

require 'io/console'
require 'openssl'
require 'base64'

if not defined?(INTERNAL_VAGRANT_SHARE)
	INTERNAL_VAGRANT_SHARE = "[REDACTED]"
end

#used to encrypt .auth
if not defined?(VM_PWKEY)
	VM_PWKEY = "horse-bucket-danger-bush"
end

  config.vm.box = "win7-sc-base"
  config.vm.box_url = "#{INTERNAL_VAGRANT_SHARE}/Asteres/win7-sc-base.json"
  config.vm.box_check_update = true
  
  config.vm.define "kiosk1" do |kiosk1|
    kiosk1.vm.hostname = "VM-Kiosk1"
  end #kiosk1

   config.vm.define "kiosk2" do |kiosk2|
     kiosk2.vm.hostname = "VM-Kiosk2"
   end #kiosk2

  config.vm.network "public_network"
  config.vm.synced_folder "share", "/share", create: true
 
  config.vm.provider :hyperv do |v, override|
    v.memory = 2048
    v.cpus = 2
    v.ip_address_timeout = 130
	# Use differencing disk instead of cloning whole VHD
    if Vagrant::VERSION < '2.1.2'
      v.differencing_disk = true
    else
      v.linked_clone = true
    end
	
	v.enable_virtualization_extensions = true
    v.vm_integration_services = {
      guest_service_interface: true,
      heartbeat: true,
      key_value_pair_exchange: true,
      shutdown: true,
      time_synchronization: true,
      vss: true
    }
  end #hyperv

  # timeout of waiting for image to stop running - may be a deprecated setting
  config.windows.halt_timeout = 20
 
  config.vm.guest = :windows
  config.vm.communicator = "winrm"

  config.winrm.username = "REDACTED"
  config.winrm.password = ENV['BoxPassword']
  config.winrm.transport = :negotiate

end

```

### Debug output
[Log 1: PowerShellError](https://gist.github.com/General-Fault/4579642973c9722d37601744f8203a41)
[Log 2: ExecutionError](https://gist.github.com/General-Fault/5a77aac61f8dffddf4dc0ce9ecabe963).

### Expected behavior
Provisioning a VM with a <vm.hostname> parameter should not fail. 

### Actual behavior
Occasionally the reboot occurs more quickly that Vagrant can start the reboot listener. As a result the provisioning script fails.

### Steps to reproduce
1. Create a VagrantFile for a Windows guest with a Hyper-V provider.
2. Add config.vm.hostname = "new-hostname" to the VagrantFile
3. On a Windows 10 host, "up" the guest.

4. As part of the provision process, the rename operation schedules a reboot with a 5 second delay.
```
[10:28:48.628] DEBUG winrmshell: [WinRM] Command created for $computer = Get-WmiObject -Class Win32_ComputerSystem
$retval = $computer.rename("new-hostname").returnvalue
if ($retval -eq 0) {
shutdown /r /t 5 /f /d p:4:1 /c "Vagrant Rename Computer"
}
exit $retval
```

Immediately after, it attempts to wait for the reboot
```
            if machine.guest.capability?(:wait_for_reboot)
                machine.guest.capability(:wait_for_reboot)
```

If the wait_for_reboot step process takes more than 5 seconds to initialize, as it often does, this call can fail.

5. If the guest is successfully provisioned, repeat the test until the provision fails.

### Details
The error is usually a <VagrantPlugins::HyperV::Errors::PowerShellError> with the message "Failed to determine IP address". [detail](https://gist.github.com/General-Fault/4579642973c9722d37601744f8203a41) 
However the race can fail at later steps and generate other errors. At least one other observed instance resulted in a <VagrantPlugins::CommunicatorWinRM::Errors::ExecutionError> [detail](https://gist.github.com/General-Fault/5a77aac61f8dffddf4dc0ce9ecabe963).

In instances where the provisioning is successful, the next get-ipaddress occurs within that 5 seconds. In instances where the provisioning fails, that time is longer (or I suspect Windows is taking longer to shutdown). The IP address check seems to be a part of asking windows for a shutdown event. So it's just late.

### Workaround
A workaround is to increase the delay to 15 seconds (in Vagrant\embedded\gems\2.1.2\gems\vagrant-2.1.2\plugins\guests\windows\cap\change_host_name.rb:20 - see below). But I think that a better solution here would be to assume that a failed IP address meant that the system had already started the shutdown, and continue normally.

          script = <<-EOH
            $computer = Get-WmiObject -Class Win32_ComputerSystem
            $retval = $computer.rename("#{name}").returnvalue
            if ($retval -eq 0) {
              shutdown /r /t 15 /f /d p:4:1 /c "Vagrant Rename Computer"
            }
            exit $retval
          EOH

###References
See also issue #9356.

briancain:
Prior to this commit, the Hyper-V provider would attempt to invoke a
script to get the guests IP address without any sort of timeout in
place. This caused a race condition when setting the hostname for a
windows guest if the machine took too long to reboot after setting the
hostname. This commit updates that behavior by putting a timeout in
place to ensure the machine is running before giving up trying to
determine the guests IP.


briancain:
I'm less confident that my updates will fix issue #9356 now, but this should fix the original issue #10229.

plugins/guests/windows/cap/change_host_name.rb
```diff
@@ -27,13 +30,26 @@ def self.change_host_name_and_wait(machine, name, sleep_timeout)
...
 
-          # Don't continue until the machine has shutdown and rebooted
-	  if machine.guest.capability?(:wait_for_reboot)
-            machine.guest.capability(:wait_for_reboot)
-          else
-            # use graceful_halt_timeout only if guest cannot wait for reboot
-            sleep(sleep_timeout)
-	  end
+          timeout = 0
+          while true
```
chrisroberts: Remove the `while true` loop, it's not required

plugins/guests/windows/cap/change_host_name.rb
```diff
@@ -27,13 +30,26 @@ def self.change_host_name_and_wait(machine, name, sleep_timeout)
...
+                @logger.debug("No wait_for_reboot capability, sleeping for #{sleep_timeout} instead...")
+                # use graceful_halt_timeout only if guest cannot wait for reboot
+                sleep(sleep_timeout)
+                break
+              end
+            rescue Vagrant::Errors::MachineGuestNotReady => e
+              raise if timeout == 24
+              @logger.warn("Machine not ready, cannot wait for reboot yet. Trying again")
+              timeout = timeout + 1
+              sleep(5)
```
chrisroberts: If we are not bailing out yet then we should `retry` after the sleep

plugins/guests/windows/cap/change_host_name.rb
```diff
@@ -27,13 +30,26 @@ def self.change_host_name_and_wait(machine, name, sleep_timeout)
...
+                machine.guest.capability(:wait_for_reboot)
+                break
+              else
+                @logger.debug("No wait_for_reboot capability, sleeping for #{sleep_timeout} instead...")
+                # use graceful_halt_timeout only if guest cannot wait for reboot
+                sleep(sleep_timeout)
+                break
+              end
+            rescue Vagrant::Errors::MachineGuestNotReady => e
+              raise if timeout == 24
```
chrisroberts: Instead of having a `timeout` variable that is incremented and checked it might be better to have a known amount of time we will attempt retries and store that in a constant (like `MAX_REBOOT_DURATION`). Then we can use that value and decrement it on each retry.

```ruby
MAX_REBOOT_DURATION = 120

wait_remaining = MAX_REBOOT_DURATION
begin
  ...
rescue Vagrant::Errors::MachineGuestNotReady
  raise if wait_remaining < 0
  sleep(5)
  wait_remaining -= 5
  retry
end
```

Having the time defined in a constant will help to make it clearer in the future how much time is allowed if it needs to be adjusted for some reason.

plugins/guests/windows/cap/reboot.rb
```diff
@@ -1,21 +1,51 @@
...
+          else
+            @logger.debug("A reboot is already in progress")
+          end
+
+          @logger.debug("Waiting for machine to finish rebooting")
 
+          wait_remaining = MAX_REBOOT_RETRY_DURATION
+          begin
+            wait_for_reboot(machine)
+          rescue Vagrant::Errors::MachineGuestNotReady, WinRM::WinRMHTTPTransportError => e
```
chrisroberts: Only the `Vagrant::Errors::MachineGuestNotReady` exception should be needed here. Also the `=> e` can be dropped since it is unused.

plugins/communicators/winrm/communicator.rb
```diff
@@ -109,7 +109,7 @@ def ready?
 
         @logger.info("WinRM is ready!")
         return true
-      rescue Errors::TransientError => e
+      rescue Errors::TransientError, VagrantPlugins::CommunicatorWinRM::Errors::WinRMNotReady => e
```
chrisroberts: Add a quick test case for properly rescuing out this exception type.

Issue #10345, opened 2018-10-28
Allow configuration of synced_folder options for provisioners chef and puppet.


calbrecht:
I encountered the lack of possibility to specify finer grained synced_folder options to the puppet provisioner (but i think this could be relevant to the chef provisioner as well).

Currently i am able to specify ```synced_folder_type``` which does not let me specify neither the ```nfs_version``` in case i want to use nfsv4 nor the ```nfs_export``` option, which i need to set to false because i'm running on NixOS and i'm handling exports through NixOS configuration that is leading to /etc/exports not beeing writeable.

I propose to replace the configuration setting ```synced_folder_type``` with a setting ```synced_folder_opts``` which could hold a hash of synced_folder configuration options that will be merged into the ```folder_opts``` of the relevant provisioner plugins.

P.S. i already have a patched solution for my puppet provisioner needs and am willing to create a PR for chef and puppet provisioner plugins.

briancain:
Hi @calbrecht - you can already set the NFS version with `nfs_version`: https://www.vagrantup.com/docs/synced-folders/nfs.html#nfs_version
Any other specific options for NFS can be set with `mount_options`. Otherwise the update you proposed is already how `synced_folders` works. It takes a hash that Vagrant uses to set various synced folder options.

The synced_folder option should be fairly agnostic to any plugin, so I'm not sure what specifically you are needing with the puppet or chef provisioner. I believe these plugins might do a check that the correct folders are in place on the guest, but I don't think it should care _how_ the folders got there.

I'll go ahead and close this, but if you have any more feedback feel free to comment here. Otherwise if you have more questions I recommend directing them to the mailing list. Thanks!

calbrecht:
@briancain The puppet and probably the chef provisioner are creating synced_folders where i am not able to set the before mentioned options.
See ```plugins/provisioners/puppet/provisioner/puppet.rb```
line 31 through 35
```
folder_opts = {}
          folder_opts[:type] = @config.synced_folder_type if @config.synced_folder_type
          folder_opts[:owner] = "root" if !@config.synced_folder_type
          folder_opts[:args] = @config.synced_folder_args if @config.synced_folder_args
          folder_opts[:nfs__quiet] = true
```
and line 44 through 46
```
root_config.vm.synced_folder(
                File.expand_path(@config.environment_path[1], root_path),
                environments_guest_path, folder_opts)
```

calbrecht:
[...]

PR #10370, opened 2018-11-03
WIP: Add synced_folder_opts to puppet provisioner #10345

calbrecht:
Provide the possibility to configure the synced folders
created by the puppet and chef provisioner with options 
supported by any of the existing synced folder plugins.

***Deprecate*** ```synced_folder_type``` and ```synced_folder_args```.

See #10345 

@TODO test provisioner/puppet.rb

### DIFF
--- a/plugins/provisioners/puppet/config/puppet.rb
+++ b/plugins/provisioners/puppet/config/puppet.rb
@@ -17,6 +17,8 @@ module VagrantPlugins
         attr_accessor :environment_variables
         attr_accessor :module_path
         attr_accessor :options
+        attr_accessor :synced_folder_type
+        attr_accessor :synced_folder_args
         attr_accessor :synced_folder_opts
         attr_accessor :temp_dir
         attr_accessor :working_directory
@@ -34,29 +36,16 @@ module VagrantPlugins
           @module_path           = UNSET_VALUE
           @options               = []
           @facter                = {}
+          @synced_folder_type    = UNSET_VALUE
+          @synced_folder_args    = UNSET_VALUE
           @synced_folder_opts    = {}
           @temp_dir              = UNSET_VALUE
           @working_directory     = UNSET_VALUE
           @structured_facts   = UNSET_VALUE
         end
 
-        def synced_folder_type=(value)
-          puts "DEPRECATION: The 'synced_folder_type' setting for the Puppet provisioner is"
-          puts "deprecated. Please use the 'synced_folder_opts' setting instead."
-          puts "The 'synced_folder_type' setting will be removed in the next version of Vagrant."
-
-          @synced_folder_type = value
-        end
-
-        def synced_folder_args=(value)
-          puts "DEPRECATION: The 'synced_folder_args' setting for the Puppet provisioner is"
-          puts "deprecated. Please use the 'synced_folder_opts' setting instead."
-          puts "The 'synced_folder_args' setting will be removed in the next version of Vagrant."
-
-          @synced_folder_args = value
-        end
-
         def synced_folder_opts(**opts)
+          [:nfs, :rsync, :smb, :virtualbox].each {|sym| opts[:type] = sym.to_s if opts.key?(sym)}
           @synced_folder_opts = @synced_folder_opts.merge!(opts)
         end
 
@@ -103,10 +92,17 @@ module VagrantPlugins
           end
 
           if @synced_folder_opts.eql?({})
-            @synced_folder_opts[:type] = @synced_folder_type if @synced_folder_type
-            @synced_folder_opts[:owner] = "root" if !@synced_folder_type
-            @synced_folder_opts[:args] = @synced_folder_args if @synced_folder_args
+            @synced_folder_opts[:type] = @synced_folder_type unless @synced_folder_type == UNSET_VALUE
+            @synced_folder_opts[:owner] = "root" if @synced_folder_type == UNSET_VALUE
+            @synced_folder_opts[:args] = @synced_folder_args unless @synced_folder_args == UNSET_VALUE
             @synced_folder_opts[:nfs__quiet] = true
+          else
+            unless @synced_folder_type == UNSET_VALUE or @synced_folder_opts.key?(:type)
+              @synced_folder_opts[:type] = @synced_folder_type
+            end
+            unless @synced_folder_args == UNSET_VALUE or not @synced_folder_opts[:type] == "rsync"
+              @synced_folder_opts[:args] = @synced_folder_args
+            end
           end
 
           @binary_path        = nil     if @binary_path == UNSET_VALUE
@@ -132,6 +128,22 @@ module VagrantPlugins
 
         def validate(machine)
           errors = _detected_errors
+
+          unless @synced_folder_type == UNSET_VALUE
+            machine.ui.warn(I18n.t("vagrant.provisioners.puppet.synced_folder_type_deprecation"))
+            if @synced_folder_opts.key?(:type)
+              errors << I18n.t("vagrant.provisioners.puppet.synced_folder_type_ignored",
+                               synced_folder_type: @synced_folder_type,
+                               type: @synced_folder_opts[:type])
+            end
+          end
+
+          unless @synced_folder_args == UNSET_VALUE
+            machine.ui.warn(I18n.t("vagrant.provisioners.puppet.synced_folder_args_deprecation"))
+            unless @synced_folder_opts[:type] == "rsync"
+              errors << I18n.t("vagrant.provisioners.puppet.synced_folder_args_ignored", args: @synced_folder_args)
+            end
+          end
 
           # Calculate the manifests and module paths based on env
           this_expanded_module_paths = expanded_module_paths(machine.env.root_path)
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -2533,6 +2533,22 @@ en:
           machine. The fix is to run a `vagrant reload` so that the proper shared
           folders will be prepared and mounted on the VM.
         module_path_missing: "The configured module path doesn't exist: %{path}"
+        synced_folder_type_deprecation: |-
+          DEPRECATION: The 'synced_folder_type' setting for the Puppet provisioner is
+          deprecated. Please use the 'synced_folder_opts' setting instead.
+          The 'synced_folder_type' setting will be removed in the next version of Vagrant.
+        synced_folder_args_deprecation: |-
+          DEPRECATION: The 'synced_folder_args' setting for the Puppet provisioner is
+          deprecated. Please use the 'synced_folder_opts' setting instead.
+          The 'synced_folder_args' setting will be removed in the next version of Vagrant.
+        synced_folder_type_ignored: |-
+          The option 'synced_folder_type' with value '%{synced_folder_type}' was ignored
+          in favour of the option 'synced_folder_opts[:type]' with value '%{type}'.
+          Please remove the option 'synced_folder_type' and set valid 'synced_folder_opts' instead'.
+        synced_folder_args_ignored: |-
+          The option 'synced_folder_args' with value '%{args}' was ignored
+          because it is only applicable to the 'rsync' synced_folder type.
+          Please remove the option 'synced_folder_args' and set valid 'synced_folder_opts' instead'.
 
       puppet_server:
         cert_requires_node: |-
--- a/test/unit/plugins/provisioners/puppet/config/puppet_test.rb
+++ b/test/unit/plugins/provisioners/puppet/config/puppet_test.rb
@@ -7,12 +7,36 @@ describe VagrantPlugins::Puppet::Config::Puppet do
 
   subject { described_class.new }
 
-  let(:machine) { double("machine") }
+  let(:machine) { double("machine", ui: double("ui"), env: double("env", root_path: "")) }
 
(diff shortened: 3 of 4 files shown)


plugins/provisioners/puppet/config/puppet.rb
```diff
@@ -35,22 +34,30 @@ def initialize
...
         end
 
-        def nfs=(value)
-          puts "DEPRECATION: The 'nfs' setting for the Puppet provisioner is"
-          puts "deprecated. Please use the 'synced_folder_type' setting instead."
-          puts "The 'nfs' setting will be removed in the next version of Vagrant."
+        def synced_folder_type=(value)
+          puts "DEPRECATION: The 'synced_folder_type' setting for the Puppet provisioner is"
+          puts "deprecated. Please use the 'synced_folder_opts' setting instead."
+          puts "The 'synced_folder_type' setting will be removed in the next version of Vagrant."
```
briancain: This text should be abstracted out into our string translations file with `I18n`. If you're not familiar with that, essentially you copy this and add a I18n "key" to be referenced later in the `en.yml` file (I have an example below). Also the warning should be printed in the `validation` method so that we can used the passed in `machine` object to properly print the warning like so:

```ruby
def validate(machine)
...
  if @synced_folder_type
    machine.ui.warn(I18n.t("vagrant.provisioners.puppet.synced_folder_type_deprecation"))
  end
...
end
```

And same goes for the `synced_folder_args` method below!

Also then because it's just setting a value to this option, we don't need a special method for it so they can go back to how they were prior to this PR. And you'll need to add @synced_folder_args as a value being initialized with `UNSET_VALUE` too (It looks like it didn't exist before...).

plugins/provisioners/puppet/config/puppet.rb
```diff
@@ -95,10 +102,15 @@ def finalize!
...
+            @synced_folder_opts[:type] = @synced_folder_type if @synced_folder_type
+            @synced_folder_opts[:owner] = "root" if !@synced_folder_type
+            @synced_folder_opts[:args] = @synced_folder_args if @synced_folder_args
+            @synced_folder_opts[:nfs__quiet] = true
+          end
+
           @binary_path        = nil     if @binary_path == UNSET_VALUE
           @module_path        = nil     if @module_path == UNSET_VALUE
-          @synced_folder_type = nil     if @synced_folder_type == UNSET_VALUE
-          @synced_folder_args = nil if @synced_folder_args == UNSET_VALUE
```
briancain: This should still be here I think. Otherwise these objects will be left as `UNSET_VALUE` internally and could lead to some exceptions down the road.

plugins/provisioners/puppet/config/puppet.rb
```diff
@@ -95,10 +102,15 @@ def finalize!
             @environment_variables = {}
           end
 
+          if @synced_folder_opts.eql?({})
+            @synced_folder_opts[:type] = @synced_folder_type if @synced_folder_type
+            @synced_folder_opts[:owner] = "root" if !@synced_folder_type
+            @synced_folder_opts[:args] = @synced_folder_args if @synced_folder_args
+            @synced_folder_opts[:nfs__quiet] = true
+          end
```
briancain: Woot! This is a general comment not for here but for the `validate` method.

I think we should print a warning that checks if both `:type` and `@synced_folder_type` are set just so it's clear what is getting overridden here in case both options are set. If they're different, we should probably just make that an error?

calbrecht:
@briancain thanks for your feedback, i appreciate it. :) Nevertheless some points i disagree, but will comment on them separtely. :tada:

plugins/provisioners/puppet/config/puppet.rb
calbrecht: Sure, this looks very reasonable to me and i will follow your advice.

plugins/provisioners/puppet/config/puppet.rb
calbrecht: Right now this will not be the case, because i removed the ```@synced_folder_type``` from above at line 38. And the ```@synced_folder_args``` was not present there anyway, and i guess they were present at this place to be able to access them from ```@config.synced_folder_*``` within ```plugins/provisioners/puppet/provisioner/puppet.rb```. 
And since i moved that piece of code into the config we don't need them here any longer. 
Or do i misunderstand the ruby language here?

Shouldn't it be sufficient to re-add lines 20 and 21 above and then access them within between lines 105 - 110 above?

plugins/provisioners/puppet/config/puppet.rb
calbrecht: Hm, unsure, because there is no need to set ```:type``` within the ```synced_folder_opts``` hash. This is intended to be exactly the same hash symbols that are supported by any ```synced_folder``` type, like e.g.
```
synced_folder_opts :nfs => true, :nfs_udp => false, :nfs_version => 4, :nfs_export => false
```
which will be translated to ```:type``` within the ```vm.synced_folder``` method i guess and then if someone explicitly sets ```synced_folder_opts``` the one would probably know that there is no need to mix up with the deprecated and then redundant ```synced_folder_type``` option. But of course if you insist i will implement it like you wish.

And then maybe should i include the example within the website documentation, so to make it clear that there is no need to set both of the options?

plugins/provisioners/puppet/config/puppet.rb
calbrecht: Sure, thanks :) Now i am able to follow your reasoning.

plugins/provisioners/puppet/provisioner/puppet.rb
```diff
@@ -28,18 +28,12 @@ def configure(root_config)
...
-          folder_opts[:args] = @config.synced_folder_args if @config.synced_folder_args
-          folder_opts[:nfs__quiet] = true
-
           if @config.environment_path.is_a?(Array)
             # Share the environments directory with the guest
             if @config.environment_path[0].to_sym == :host
               root_config.vm.synced_folder(
                 File.expand_path(@config.environment_path[1], root_path),
-                environments_guest_path, folder_opts)
+                environments_guest_path, @config.synced_folder_opts)
```
calbrecht: These calls i really would like to test with some sort of integration test, at least against a mocked ```root_config.vm.synced_folder``` method, but did not find any good copy-paste template for it so to understand what i have to do here. Maybe you know of some place i can peak a solution to test this?

plugins/provisioners/puppet/config/puppet.rb
briancain: So these options get directly passed into the `synced_folder` config method: https://github.com/hashicorp/vagrant/blob/master/plugins/kernel_v2/config/vm.rb#L204

Which means you should be able to give it any of the options that are valid here: https://www.vagrantup.com/docs/synced-folders/basic_usage.html#options

So there's a possibility that someone could set `{type: "nfs"}` as a key like

```ruby
config.vm.provider "puppet" do |p|
  p.synced_folder_type = "nfs"
  p.synced_folder_opts = {type: "smb"}
end
```

Which is silly of course, but I think it would be good to warn users if both happen to be set is all.

But yeah in the docs it should specifically call out the doc page I link to as all the valid options you can provide to `synced_folder_opts`.

plugins/provisioners/puppet/provisioner/puppet.rb
briancain: You should be able to mock this up with Rspec. Unfortunately the puppet provisioner is very sparse on tests :( You'll want to create a new describe `context` for this classes rspec tests around the `#configure` method. Ideally here I would want to mock up that a given _double_ called `vm` received the method `:synced_folder` with the specific config options. That will be good enough locally for unit tests.

Something like

```ruby
describe "#configure" do
  let(:vm) { double("vm") }
  let(:root_config) { double("root_config" } # This should actually be real when writing this test, not just pseudo code
...
  it "configures the provisioner" do
    # other stuff probably goes here!!
    expect(vm).to receive(:synced_folder).with("/path/to/the/manifest", {type=>"nfs", other=>"keys"})

    subject.configure(root_config)
  end
end
```

This is super basic and missing a lot of what would probably be required :sweat_smile: But that is basically how I would test that the _right_ thing is being passed to the `#synced_folder` function to be stored as a config.

briancain: You will also probably need to ensure that the double `vm` is the right class and not just a double? Since it's apart of `root_config`, you'll likely need to do a bit of mocking to ensure that it is a `VagrantPlugins::Kernel_V2::VMConfig` type. But anyway, hopefully that gives you a good idea :eyes:

plugins/provisioners/puppet/config/puppet.rb
calbrecht: Ok, i was too lazy to code for silly folks ;) but i will follow your advice here as well.
I did already link to the ```basic_usage.html``` page though in the documentation but will add the ```#options``` to it. 
Do you think i should add an extra section within the example section of the provisioner documentation page?

plugins/provisioners/puppet/provisioner/puppet.rb
calbrecht: Yes, exactly :) Thanks very much, because it would have been tedious to search this together without knowing what to do in a new language :) Very much appreciated :+1:

calbrecht:
> I know it's a WIP pull request but I left some feedback for you.
No need to apologize, in fact i was hoping to get some feedback soon that's why i created this PR early. Maybe next time i'll have to pull feedback more actively :)

plugins/provisioners/puppet/config/puppet.rb
briancain: AH you're right I'm not sure if that's what we want then. I could see users expecting `synced_folder_type` to work as well as `synced_folder_opts` with other various options. Like maybe it's an old Vagrantfile with `synced_folder_type` set, and then they wanted to add new options like `mount_options` later and realize their `synced_folder_type` param will be ignored.

briancain: I also do think it's probably worth while to give a simple example of using the new `synced_folder_opts` option for the puppet provisioner.

calbrecht: Ok, will add an example to the documentation then.

As for the merging of ```synced_folder_type``` i'm still not fully convinced, if someone is able to find the new ```synced_folder_opts``` option and set the values, they will probably have been reading that the older ```synced_folder_type``` is deprecated. 
If we really check against collisions we'll have to check against all the synced_folder specific symbols as well like ```:nfs => true``` or the like, that is a lot of code to support an option that is deprecated in favor of a more capable option (and currently i don't understand from the code what will happen in the case when both ```:nfs => true, :type => "rsync"``` are given).

I agree that a warning should be given when both options are present though. What do you think?

briancain: So the `:nfs` key for the puppet provisioner config has been deprecated for a while now, and it's no longer even on the puppet provisioner docs. And `:nfs=>true` is not a valid `synced_folder` option. So we should be ok to just check for `:type` in `synced_folder_opts` and `synced_folder_type` existing instead. Anyone still using `:nfs=>true` on the puppet provisioner has seen that warning for a while now and shouldn't be too surprised if it is gone on an upgrade.

But yeah, ultimately we should display a warning if both `type` and `synced_folder_type` is present, and it should tell the user that the `synced_folder_opts` `:type` key will actually be used and that `synced_folder_type` will be ignored.

> (and currently i don't understand from the code what will happen in the case when both :nfs => true, :type => "rsync" are given)

Yep...unfortunately if both are set, it is dependent on what  _order_  they are defined in the Vagrantfile. Because `#nfs=` also sets `synced_folder_type`, which ever option is defined _last_ in the provider config block will be the one that wins out setting that variable. Not a great user experience here, but I think your PR here will make this better!

calbrecht: Sorry for the confusion, should have made clear which keys/options i mean. I was always talking about the keys given to ```synced_folder_opts``` here if not referenced with ```synced_folder_type```. 

So i guess i'm going to respect and merge the ```synced_folder_type``` as ```:type => @synced_folder_type``` into ```synced_folder_opts``` if none of the keys ```:nfs | :rsync | :smb | :virtualbox | :type``` are present within ```synced_folder_opts``` and discard ```synced_folder_type``` if the opposite is true.

Same goes for the ```synced_folder_args``` but only if ```:type``` will result in ```rsync```, because i think this option is only applicable to rsync folders (https://www.vagrantup.com/docs/provisioning/puppet_apply.html#synced_folder_args)

I will give a warning if collisions occur, any of the two deprecated options are set, or when no ```:type``` is given at all.

Will need some time right now to setup my rubymine environment properly now because that is not playing nice with NixOs as it is, but will for sure finish this PR properly.

briancain: Sounds great, thank you! Let me know if you need any more help finishing it :tada:

briancain:
@calbrecht - Hey there! I was just curious if you needed anymore help finishing this up, or if you needed to hand it off. Thanks! :)

calbrecht:
@briancain thank you very much for getting back on me. 

To be honest, i got distracted by the fact that, for me, the isolated_environment is not really isolated and is pulling the plugins from my actual installation home folder (which makes running the "integration" tests that are left to implement so tedious) and eventually lost interest in writing the remaining tests after trying and failing to fix the isolation for good.

PR #10374, opened 2018-11-05
Move login command warning

briancain:
This commit moves where the warning is for the login command so that it
only displays when the command is invoked, and not when the plugin is
loaded.


briancain:
Ah hrmm...I remember why it's in the plugin load now. Need to figure out how to only show this when the command is invoked, not just loaded.

Issue #10362, opened 2018-11-02
VirtualBox NAT interface DHCP failures


balapuram:
I am trying to run vagrant up command , i am getting below error.

Vagrant Host Machine: CentOS7
Vagrant Guest Machine: CentOS7

The vagrant file as below

========================
Vagrant.configure("2") do |config|
  config.vm.box = "centos/7"
  config.vm.network "forwarded_port", guest: 80, host: 8080, host_ip: "10.10.10.20"
 end

========================

Looks like it is failing when it attempt for ssh authentication. Can you anyone please let me know what is missing here.

Thanks in Advance.

[root@jenkins2 ansible_provisioner]# vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Checking if box 'centos/7' is up to date...
==> default: Clearing any previously set forwarded ports...
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
==> default: Forwarding ports...
    default: 80 (guest) => 8080 (host) (adapter 1)
    default: 22 (guest) => 2222 (host) (adapter 1)
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
Timed out while waiting for the machine to boot. This means that
Vagrant was unable to communicate with the guest machine within
the configured ("config.vm.boot_timeout" value) time period.

If you look above, you should be able to see the error(s) that
Vagrant had when attempting to connect to the machine. These errors
are usually good hints as to what may be wrong.

If you're using a custom box, make sure that networking is properly
working and you're able to connect to the machine. It is a common
problem that networking isn't setup properly in these boxes.
Verify that authentication configurations are also setup properly,
as well.

If the box appears to be booting properly, you may want to increase
the timeout ("config.vm.boot_timeout") value.
[root@jenkins2 ansible_provisioner]#

Regards,
Bhaskar.

briancain:
@balapuram - Can you bring up the machine using the virtualbox gui instead of Vagrant?

robotpony:
I'm seeing the same logs here (with Ubuntu 14.04) on previously tested/running images. When starting up via VirtualBox, networking devices are failing to start up:

[...]

PR #10395, opened 2018-11-11
Added new config parameter "booting_wait" for delay before communicat…

arion85:
(Sorry for my english) I had a problem with getting IP v4 for a NAT adapter. The delay before the launch of the communicator solves this problem.

### DIFF
--- a/plugins/kernel_v2/config/vm.rb
+++ b/plugins/kernel_v2/config/vm.rb
@@ -21,6 +21,7 @@ module VagrantPlugins
       attr_accessor :allowed_synced_folder_types
       attr_accessor :base_mac
       attr_accessor :boot_timeout
+      attr_accessor :booting_wait
       attr_accessor :box
       attr_accessor :ignore_box_vagrantfile
       attr_accessor :box_check_update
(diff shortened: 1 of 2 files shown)


chrisroberts:
@arion85 Can you provide an example of the error you encountered without this pull request? Did you try modifying the boot_timeout option? https://www.vagrantup.com/docs/vagrantfile/machine_settings.html#config-vm-boot_timeout

arion85:
Hello! I had the same problem as here https://github.com/hashicorp/vagrant/issues/10362. When I studied the problem in more detail, I realized that my NAT adapter was not assigned an IP address v4 when the VM was started via Vagrant. For some reason, the responses from the DHCP server could not go inside the VM. "booting_wait" makes a delay after starting a VM before starting the communicator and SSH. This is not associated with boot_timeout.

chrisroberts:
Thanks! Moved conversation about this into the related issue: https://github.com/hashicorp/vagrant/issues/10362#issuecomment-438303618

chrisroberts:
With the ongoing conversation in #10362, I'm going to close this PR. The addition of this setting only seems to help in specific situations and I'd like to find a root cause before adding anything like this.

Thanks again for the PR! Cheers!

PR #10397, opened 2018-11-12
Fix "srv" to "src"

mingrammer:
"/srv/website" seems like a typo, but it is not a critical issue.


briancain:
Hey there @mingrammer thanks, but this is actually intended! `/srv` is for website specific data served by a system. More info can be found here https://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/srv.html

Thanks though! :heart:

mingrammer:
Ah! I didn't know that! Thank you for that let me know the existence of `/srv` directory :)

PR #10399, opened 2018-11-12
Add communicator reset

### LINKED ISSUE #10388: Cannot use Docker as unprivileged user on first "up"
### Vagrant version
Vagrant 2.2.0

### Host operating system

```
$ uname -a
Linux yui 4.18.0-2-amd64 #1 SMP Debian 4.18.10-2 (2018-11-02) x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:  Debian GNU/Linux unstable (sid)
Release:  unstable
Codename: sid
```

### Guest operating system

```
$ uname -a
Linux vagrant-ubuntu-trusty-64 3.13.0-92-generic #139-Ubuntu SMP Tue Jun 28 20:42:26 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:  Ubuntu 14.04.5 LTS
Release:  14.04
Codename: trusty
```

### Vagrantfile
```ruby
  config.vm.provision :docker

  config.vm.provision :shell,
    keep_color: true,
    privileged: false,
    run: "always",
    inline: <<-SCRIPT
      echo "Groups of $(whoami)"
      groups
      docker run hello-world
    SCRIPT
end
```

### Debug output
https://gist.github.com/mbrodala/49813f8feceb77c6a4cfa747f72e6697

### Expected behavior
Vagrant should execute the `docker run` command and display its output without errors.

### Actual behavior
The `docker run` fails even though Docker has been set up properly by the `docker` provisioner:

```
...
    default: Running: inline script
    default: Groups of vagrant
    default: vagrant
    default: docker: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.38/containers/create: dial unix /var/run/docker.sock: connect: permission denied.
    default: See 'docker run --help'.
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

This seems to be caused by Vagrant [re-using the existing SSH connection](https://github.com/hashicorp/vagrant/blob/b0fc33e9b728f8575890a5a9e358394983b5298c/plugins/communicators/ssh/communicator.rb#L333-L339) and the way Linux behaves in general when adding groups to a user, in this case the `docker` group for the `vagrant` user. Newly added groups only affect new shells, existing shells will keep their list of user groups unchanged. Since Vagrant does not terminate the SSH connection after installing Docker and [adding `vagrant` to the `docker` group](https://github.com/hashicorp/vagrant/blob/e8cee899a8034ca89130575e2f540b003ae006e1/plugins/provisioners/docker/cap/linux/docker_configure_vagrant_user.rb#L11), any subsequent SSH command won't have the `docker` group and won't have the permissions to access the Docker socket. On subsequent `up` a new SSH connection is set up which then has the updated list of user groups.

### Steps to reproduce
1. Put the `Vagrantfile` somewhere
2. Run `vagrant up`
3. See the error and `vagrant` missing the `docker` group
4. Run `vagrant up` once more
5. See the error gone and now `vagrant` having the `docker` group

### References

Initially discovered in https://github.com/leighmcculloch/vagrant-docker-compose/issues/53

chrisroberts:
Adds a `Communicator#reset!` method to allow resetting the communicator to force establishing a new connection. Fixes linux docker install to reset communicator after group updates so `vagrant` user has `docker` group available. Includes new option `:reset` to the shell provisioner to allow resetting the connection after a provisioner has completed. Also allows a provisioner to run without a `path` or `inline` defined simply to reset the connection.

Fixes #10388

### DIFF
--- a/lib/vagrant/plugin/v2/communicator.rb
+++ b/lib/vagrant/plugin/v2/communicator.rb
@@ -117,6 +117,13 @@ module Vagrant
         # @see #execute
         def test(command, opts=nil)
         end
+
+        # Reset the communicator. For communicators which establish
+        # a persistent connection to the remote machine, this connection
+        # should be terminated and re-established. The communicator
+        # instance should be in a "fresh" state after calling this method.
+        def reset!
+        end
       end
     end
   end
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -69,11 +69,14 @@ module VagrantPlugins
           end
 
           # Got it! Let the user know what we're connecting to.
-          @machine.ui.detail("SSH address: #{ssh_info[:host]}:#{ssh_info[:port]}")
-          @machine.ui.detail("SSH username: #{ssh_info[:username]}")
-          ssh_auth_type = "private key"
-          ssh_auth_type = "password" if ssh_info[:password]
-          @machine.ui.detail("SSH auth method: #{ssh_auth_type}")
+          if !@ssh_info_notification
+            @machine.ui.detail("SSH address: #{ssh_info[:host]}:#{ssh_info[:port]}")
+            @machine.ui.detail("SSH username: #{ssh_info[:username]}")
+            ssh_auth_type = "private key"
+            ssh_auth_type = "password" if ssh_info[:password]
+            @machine.ui.detail("SSH auth method: #{ssh_auth_type}")
+            @ssh_info_notification = true
+          end
 
           previous_messages = {}
           while true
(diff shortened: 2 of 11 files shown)


mbrodala:
Awesome, thanks a lot!

leighmcculloch:
Thanks @chrisroberts! That was a fast turn around! 🎉

PR #10414, opened 2018-11-14
fixed vdisk management -- not required -- on VBox 6

scoter-oracle:
Starting from VirtualBox 6 vdisk management has changed while importing ovf/ova and so also Vagrant Boxes.
If we still use "version_5_0.rb" (as we already do for 5_1 / 5_2) while importing a box, the vdisks will be created within the Vagrant Project; to overcome this possible issue and get all the vdisks correctly created under VirtualBox "Default Machine Folder", I would propose a dedicated "version_6_0.rb" file containing the same thing we've for 5_0 but without vdisk management.

### DIFF
--- a/plugins/providers/virtualbox/driver/version_6_0.rb
+++ b/plugins/providers/virtualbox/driver/version_6_0.rb
@@ -1,14 +1,768 @@
-require File.expand_path("../version_5_0", __FILE__)
+require 'log4r'
+
+require "vagrant/util/platform"
+
+require File.expand_path("../base", __FILE__)
 
 module VagrantPlugins
   module ProviderVirtualBox
     module Driver
-      # Driver for VirtualBox 6.0.x
-      class Version_6_0 < Version_5_0
+      # Driver for VirtualBox 5.0.x
+      class Version_6_0 < Base
         def initialize(uuid)
-          super
+          super()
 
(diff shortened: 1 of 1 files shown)


scoter-oracle:
I think that, maybe, the best option could be to remove vdisk management for all VBox releases > 5.0; am I wrong or by Vagrant CLI we do not have any option on where to put vdisks coming from a Box ? If this is true I think that the vdisk management is useless for VBox 5.x too.

chrisroberts:
Hi there. Thank you for taking the time to make this pull request. Unfortunately we will not be able to accept this pull request. It introduces a large amount of unnecessary code duplication and the code which has been removed will result in file collisions when multiple guests are created with the same source box. I'm going to close this up in favor of #10417.

Cheers!

scoter-oracle:
Hey Chris, I've opened this thread just because if you do not apply those changes for VirtualBox 6.0, you'll encounter the problem described (file collisions when multiple guests are created from the same source box).
VirtualBox changed how the vdisk are treated in 6.0; you should give a try, we've just released VirtualBox 6.0 BETA2, available at http://download.virtualbox.org/virtualbox/6.0.0_BETA2 .
That said, I'm perfectly fine if you want to close but you'll get issues opened once VirtualBox 6.0 will be GA.

Cheers!

PR #10467, opened 2018-11-30
Fixes #10224: Clear our registered providers when validating configs …

### LINKED ISSUE #10224: vagrant validate without any provider?
Is possible to run vagrant validate without any provider?
I am using GitLab CI and I would like to check Vagrantfile for syntax errors after commit.
But its impossible because vagrant expecting at least one provider.
Can I install some dummy provider?
Thanks

briancain:
…with no provider

Prior to this commit, if you went to validate your Vagrantfile and
wanted to ignore the provider, Vagrant would still fail as it checks if
there are any registered providers that are installed and usable. This
commit mocks out all registered providers to bypass that for the
validate command so that Vagrant can just validate the config and ignore
any provider config blocks.

```
vagrant@vagrant:~/test$ vagrant validate
No usable default provider could be found for your system.

Vagrant relies on interactions with 3rd party systems, known as
"providers", to provide Vagrant with resources to run development
environments. Examples are VirtualBox, VMware, Hyper-V.

The easiest solution to this message is to install VirtualBox, which
is available for free on all major platforms.

If you believe you already have a provider available, make sure it
is properly installed and configured. You can see more details about
why a particular provider isn't working by forcing usage with
`vagrant up --provider=PROVIDER`, which should give you a more specific
error message for that particular provider.
vagrant@vagrant:~/test$ vagrant validate --ignore-provider
==> ubuntu: Ignoring provider config for validation...
==> hashicorp: Ignoring provider config for validation...
==> windows: Ignoring provider config for validation...
Vagrantfile validated successfully.
```

### DIFF
--- a/plugins/commands/validate/command.rb
+++ b/plugins/commands/validate/command.rb
@@ -72,7 +72,6 @@ module VagrantPlugins
         Vagrant.plugin("2").manager.providers.each do |key, data|
           data[0].class_eval do
             def initialize(machine)
-              @machine = machine
             end
 
             def machine_id_changed
--- a/plugins/commands/validate/command.rb
+++ b/plugins/commands/validate/command.rb
@@ -30,7 +30,7 @@ module VagrantPlugins
         action_env = {}
         if options[:ignore_provider]
           action_env[:ignore_provider] = true
-          mockup_providers!
+          tmp_data_dir = mockup_providers!
         end
 
         # Validate the configuration of all machines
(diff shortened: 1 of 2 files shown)


plugins/commands/validate/command.rb
```diff
@@ -41,6 +43,57 @@ def execute
...
+      # any provider blocks.
+      def mockup_providers!
+        require 'log4r'
+        logger = Log4r::Logger.new("vagrant::validate")
+        logger.debug("Overriding all registered provider classes for validate")
+
+        # Without setting up a tmp Environment, Vagrant will completely
+        # erase the local data dotfile and you can lose state after the
+        # validate command completes.
+        tmp_data_dir = Dir.mktmpdir("vagrant-validate-")
```
chrisroberts: This path should be stored and cleaned up before the command exits.

Issue #10352, opened 2018-10-31
Salt provisioner ignores version parameter


marek-obuchowicz:
### Vagrant version
2.2.0

### Host operating system
Linux (Debian 9.5, 64-bit)

### Guest operating system
Linux (Debian 9.5, 64-bit)

### Vagrantfile
```ruby
(...)
    config.vm.provision :salt do |salt|
      salt.minion_config = "salt_minion"
      salt.run_highstate = true
      salt.bootstrap_options = "-F -P -c /tmp"
      salt.version = "2017.7.4"
    end
```

### Expected behavior
Salt version 2017.7.4 should be installed

### Actual behavior
Parameter is ignored and latest salt is installed: `2018.3.3`
Additionally, it is also not consistent with documentation, which suggests that old version should be used if parameter is not specified:

```
version (string, default: "2017.7.1") - Version of minion to be installed.
```

briancain:
Hey there @marek-obuchowicz - you didn't provide a debug gist, but I assume this is what's going on:

```
0 upgraded, 0 newly installed, 0 to remove and 57 not upgraded.
 *  INFO: Git revision matches a Salt version tag, shallow cloning enabled.
 *  INFO: Attempting to shallow clone 2017.7.4 from Salt's repository https://github.com/saltstack/salt.git
Cloning into 'salt'...
w
arning:
C
ould not find remote branch
2
017.7.4

to clone.

f
atal:
R
emote branch
2
017.7.4

not found in upstream
o
rigin

 *  WARN: Failed to shallow clone.
 *  INFO: Resuming regular git clone and remote SaltStack repository addition procedure
```

The salt provisioner uses git to clone and install salt using the [salt bootstrap script](https://github.com/saltstack/salt-bootstrap), and it looks like the branch you want doesn't exist anymore. It then fails over to the latest, but other things break later on in the setup because it still assumes 2017.7.4.

I've made a PR here to fix up the docs so that's clearer: https://github.com/hashicorp/vagrant/pull/10355

briancain:
Looks like you need to actually do this:

```ruby
 salt.version = "v2017.7.4"
```
Doing this matches the branch they have on Salts repo.

```
==> salt: Running provisioner: shell...
    salt: Running: inline script
    salt: salt-call 2017.7.4 (Nitrogen)
```

marek-obuchowicz:
@briancain thanks a lot for feedback!
I've tried your suggestion, but it caused the following error:
```
[...]

PR #10474, opened 2018-12-03
Fixes #10358: Add validation error in salt provider for certain options

### LINKED ISSUE #10358: Add validation error for Salt provisioner if `version` option is specified with no `install_type`
The following Vagrant config block can fail when provisioning a guest with Salt:

```ruby
salt.vm.provision :salt do |s|
  s.version = "v2017.7.4"
end
```

The Salt bootstrap is expecting an installation type to use. Vagrant should at least validate that an `install_type` was provided before attempting to run the bootstrap script.

briancain:
Prior to this commit, if you specified a `version` for the salt provider
but no `install_type` Vagrant would fail to pass the proper parameters
to the bootstrap install script. This commit fixes that by adding some
validation to the salt provider if `version` is specified but not
`install_type`. It also adds some extra context for certain config
validation error messages so that the user knows what option was
incorrect, rather than the message just referring to the option as
*this*.

### DIFF
--- a/plugins/provisioners/salt/config.rb
+++ b/plugins/provisioners/salt/config.rb
@@ -177,6 +177,10 @@ module VagrantPlugins
           errors << I18n.t("vagrant.provisioners.salt.python_version")
         end
 
+        if @version && !@install_type
+          errors << I18n.t("vagrant.provisioners.salt.version_type_missing")
+        end
+
         return {"salt provisioner" => errors}
       end
     end
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -2701,9 +2701,11 @@ en:
         must_accept_keys: |-
           You must accept keys when running highstate with master!
         args_array: |-
-          You must set this value as an array.
+          You must set `args_array` value as an array.
         python_version: |-
-          You must set this as an integer or string that represents an integer.
+          You must set `python_version` as an integer or string that represents an integer.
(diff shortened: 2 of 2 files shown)


marek-obuchowicz:
@briancain this happened to be BC-breaking change, so if semantic versioning is used, it should bump vagrant release number to 3.x.x i guess...

Consider the scenario, user has before this release:
```
salt.version = "git v2017.7.4"
```

Salt won't run after this patch is added, and vagrantfiles needs to be modified to:
```
salt.install_type = "git"
salt.version = "v2017.7.4"
```

In order to keep backwards compatibility, it could be improved so that old syntax is automatically handled (split salt.version into install_type and version). As it was already released, i'm not sure if such adjusting makes sense now. Alternatively it could be mentioned in the documentation of salt provider, that configuration file migration is needed on vagrant upgrade to `v2.2.3`.

Despite the fact that this PR is already merged and released, please consider either approaches above. In our environments it caused a lot of confusion and problems, because of broken backward compatibility.

marek-obuchowicz:
@briancain reading up your comments on https://github.com/hashicorp/vagrant/issues/10352, i can confirm that it worked for us only because `Salt provisioner just uses Vagrants options to build up a list of arguments to the salt bootstrap bash script`, so we were exploiting configuration options in a non-desired way and now this PR adds protection against that. I think the comment above does not apply for this case.

PR #10477, opened 2018-12-04
Use |> redirect operator

c960657:
When booting Vagrant, I get the following error message when trying to set the hostname:

```
==> dev: Checking for guest additions in VM...
==> dev: Setting hostname...

# Set the hostname
echo 'dev' > /etc/hostname

# Prepend ourselves to /etc/hosts
grep -w 'dev' /etc/hosts || {
  if grep -w '^127\.0\.1\.1' /etc/hosts ; then
    sed -i'' 's/^127\.0\.1\.1\s.*$/127.0.1.1\tdev\tdev/' /etc/hosts
  else
    sed -i'' '1i 127.0.1.1\tdev\tdev' /etc/hosts
  fi
}

# Update mailname
echo 'dev' > /etc/mailname



Stdout from the command:

192.168.1.101  switch-dev


Stderr from the command:

bash: line 6: /etc/hostname: cannot overwrite existing file
bash: line 18: /etc/mailname: cannot overwrite existing file
```

The reason is that I have configured the root account to use noclobber, `set -o noclobber`. This prevents overwriting files using the `>` redirection operator.

While this configuration is somewhat unusual, it is easy to support by just using the `>|` operator instead.

I am not sure whether this operator is supported everywhere, but it is old feature, and it is supported by at least bash, zsh, ksh and tcsh.

An alternative (more conservative) approach is to simply remove the file before writing to it using `>`.

### DIFF
--- a/plugins/guests/darwin/cap/change_host_name.rb
+++ b/plugins/guests/darwin/cap/change_host_name.rb
@@ -24,7 +24,7 @@ module VagrantPlugins
 
               # Prepend ourselves to /etc/hosts - sed on bsd is sad
               grep -w '#{name}' /etc/hosts || {
-                echo -e '127.0.0.1\\t#{name}\\t#{basename}' | cat - /etc/hosts > /tmp/tmp-hosts &&
+                echo -e '127.0.0.1\\t#{name}\\t#{basename}' | cat - /etc/hosts >| /tmp/tmp-hosts &&
                   mv /tmp/tmp-hosts /etc/hosts
               }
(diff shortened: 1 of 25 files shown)


chrisroberts:
Hi there,

Thanks for this PR! After reviewing it I do not believe that I am comfortable to introduce such a large changeset to known working scripts for all of these guests. It seems like it might be better to simply disable the noclobber option when generating the box and enable it with a provisioner once it has been setup.

Cheers!

PR #10482, opened 2018-12-04
Fixes #10463: Display version update on stderr instead of stdout

### LINKED ISSUE #10463: New version message shouldn't appear in machine-readable output
**Vagrant version:** 2.2.1 (before upgrading to 2.2.2)
**Host operating system:** Ubuntu Xenial

### Expected behavior
Commands like `vagrant port` should only include the requested information on stdout. If any additional messages are printed, they should either appear on stderr or only in interactive/human-only command output.

### Actual behavior
`vagrant port --guest 22` returned
```
vagrant: A new version of Vagrant is available: 2.2.2 (installed version: 2.2.1)! ==> vagrant: To upgrade visit: https://www.vagrantup.com/downloads.html 2222
```
which resulted in CI failing since this mess is not a valid port number.

briancain:
This commit updates the behavior of printing the checkpoint version
check for Vagrant. To allow users to filter out the message, it updates
the version check to go to stderr. It updates the UI class for printing
the version check to be a "basic" class, so that the message continues
to be the same color instead of a red error message.

### DIFF
--- a/lib/vagrant/util/checkpoint_client.rb
+++ b/lib/vagrant/util/checkpoint_client.rb
@@ -159,11 +159,11 @@ module Vagrant
       def version_check
         latest_version = Gem::Version.new(result["current_version"])
         installed_version = Gem::Version.new(VERSION)
-        ui = Vagrant::UI::Prefixed.new(Vagrant::UI::Basic.new, "vagrant")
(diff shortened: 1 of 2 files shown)


briancain:
So the only downside to this change is, using the `Basic` UI class means we can't (I think) keep the new version message bolded. If we go back to the regular UI class with colors, the `error` message will be in red. :thinking:

PR #10485, opened 2018-12-06
Add experimental flag to guard development features

briancain:
This commit introduces a special flag for enabling features that are not
ready for release. It can either be enabled by setting the
`VAGRANT_EXPERIMENTAL` flag to "1", or by setting it to a string of one
or more comma seperated values for specific features. It also adds a
couple of Vagrant developer focused methods for making it easier to
determine if the flag has been enabled, and if so, what features.

### DIFF
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -159,6 +159,31 @@ begin
     env.ui.warn(I18n.t("vagrant.general.not_in_installer") + "\n", prefix: false)
   end
 
+  # Acceptable experimental flag values include:
+  #
+  # Unset  - Disables experimental features
+  # 0      - Disables experimental features
+  # 1      - Enables all features
+  # String - Enables one or more features, separated by commas
+  if ENV["VAGRANT_EXPERIMENTAL"]
+    experimental = ENV["VAGRANT_EXPERIMENTAL"].to_s
+    if experimental == "0"
+      logger.debug("Experimental flag is set but disabled")
+    else
+      ui = Vagrant::UI::Prefixed.new(env.ui, "vagrant")
+      if experimental != "1"
+        logger.debug("Experimental flag is enabled")
+        features = experimental.split(',')
+        ui.warn(I18n.t("vagrant.general.experimental.features", features: features.join(", ")), bold: true, prefix: true, channel: :error)
+      elsif experimental == "1"
+        logger.debug("Experimental flag is enabled")
+        ui.warn(I18n.t("vagrant.general.experimental.all"), bold: true, prefix: true, channel: :error)
+      else
+        logger.warn("Experimental flag is set to an unsupported value")
+      end
+    end
+  end
+
   begin
     # Execute the CLI interface, and exit with the proper error code
     exit_status = env.cli(argv)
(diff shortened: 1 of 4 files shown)


bin/vagrant
```diff
@@ -159,6 +159,31 @@ begin
     env.ui.warn(I18n.t("vagrant.general.not_in_installer") + "\n", prefix: false)
   end
 
+  # Acceptable experimental flag values include:
+  #
+  # Unset  - Disables experimental features
+  # 0      - Disables experimental features
+  # 1      - Enables all features
+  # String - Enables one or more features, separated by commas
+  if ENV["VAGRANT_EXPERIMENTAL"]
```
chrisroberts: Should just use the `Util::Experimental.enabled?`. If explicitly disabled with `"0"` value I don't think we need it logged. For the rest of the output I think this can be cleaned up a bit by using helpers in the util instead of direct inspection of the environment variable here.

lib/vagrant/util/experimental.rb
```diff
@@ -0,0 +1,48 @@
+module Vagrant
+  module Util
+    class Experimental
+      VALID_FEATURES = [].freeze
+      class << self
+
+        # A method for determining if the experimental flag has been enabled
+        #
+        # @return [Boolean]
+        def enabled?
```
chrisroberts: Update to: `global_enabled? || !features.empty?`

lib/vagrant/util/experimental.rb
```diff
@@ -0,0 +1,48 @@
...
+          end
+          @_experimental
+        end
+
+        # A method for Vagrant internals to determine if a given feature
+        # has been abled and can be used.
+        #
+        # @param [String] - An array of strings of features to check against
+        # @return [Boolean] - A hash containing the original array and if it is valid
+        def feature_enabled?(feature)
```
chrisroberts: Could be reduced down by adding a `features` method and then doing `features.include?(feature.to_s)`. Should also cast to string to allow symbols to be provided.

lib/vagrant/util/experimental.rb
```diff
@@ -0,0 +1,48 @@
...
+          experimental = ENV["VAGRANT_EXPERIMENTAL"].to_s.downcase
+          if experimental == "1"
+            return true
+          elsif VALID_FEATURES.include?(feature) &&
+                experimental.split(',').include?(feature)
```
chrisroberts: Move the `experimental.split` into an isolated method and store result so it's only done once.

lib/vagrant/util/experimental.rb
```diff
@@ -0,0 +1,48 @@
...
+            return true
+          else
+            return false
+          end
+        end
+
```
chrisroberts: Add a `guard` or `guard_with` method that accepts feature names which can unlock it, and yield if allowed:

```ruby
def guard(*names)
  if block_given? && names.any?{|n| feature_enabled?(n) }
    yield
  end
end
```

something like that.

lib/vagrant/util/experimental.rb
```diff
@@ -0,0 +1,48 @@
+module Vagrant
+  module Util
+    class Experimental
+      VALID_FEATURES = [].freeze
```
chrisroberts: I don't think we want to have a hard restriction on allowed names so I would vote to remove this.

PR #10500, opened 2018-12-13
Update project_setup.html.md

Brandon9598:


briancain:
@Brandon9598 - Update is now live! Thanks again. :heart:

PR #10555, opened 2019-01-03
Fix test `should not display upgrade information`

pvalena:
Convert to `should display upgrade information` when latest is newer than current.


chrisroberts:
Hi there. Thanks for opening this PR. I'm going to close this in favor of #10557 which fixes other tests within this file and adds some new tests to validate expected behavior. Cheers!

PR #10562, opened 2019-01-04
Remove tmpdir after rsync completes

aeikum:
I noticed that rsync doesn't clean up its directories in /tmp. Using rsync-auto, by the end of the working day, I'd have dozens of /tmp/vagrant-rsync dirs. This patch removes the tmpdir after the rsync command completes.

This is my first Ruby code, so I'm not sure I used appropriate library function, or if any library requires are needed for FileUtils. Anyway, it seems to work for me.

### DIFF
--- a/plugins/synced_folders/rsync/helper.rb
+++ b/plugins/synced_folders/rsync/helper.rb
@@ -1,3 +1,4 @@
+require "fileutils"
 require "ipaddr"
 require "shellwords"
 require "tmpdir"
@@ -232,6 +233,8 @@ module VagrantPlugins
               message: err.to_s
           end
         end
+      ensure
+        FileUtils.remove_entry_secure(controlpath) if controlpath
       end
 
       # Check if rsync versions support using chown option
--- a/test/unit/plugins/synced_folders/rsync/helper_test.rb
+++ b/test/unit/plugins/synced_folders/rsync/helper_test.rb
@@ -300,6 +300,10 @@ describe VagrantPlugins::SyncedFolderRSync::RsyncHelper do
         end
 
         subject.rsync_single(machine, ssh_info, opts)
+
+        unless Vagrant::Util::Platform.windows?
+          expect(File.exist?("/tmp/vagrant-rsync-12345")).to be_falsey
+        end
       end
     end
   end
--- a/test/unit/plugins/synced_folders/rsync/helper_test.rb
+++ b/test/unit/plugins/synced_folders/rsync/helper_test.rb
@@ -284,13 +284,20 @@ describe VagrantPlugins::SyncedFolderRSync::RsyncHelper do
 
     context "control sockets" do
       it "creates a tmp dir" do
-        allow(Vagrant::Util::Platform).to receive(:windows?).and_return(false)
         allow(Dir).to receive(:mktmpdir).with("vagrant-rsync-").
           and_return("/tmp/vagrant-rsync-12345")
 
-        expect(Vagrant::Util::Subprocess).to receive(:execute).with(any_args) { |*args|
-          expect(args[9]).to include("ControlPath=/tmp/vagrant-rsync-12345")
-        }.and_return(result)
+        if Vagrant::Util::Platform.windows?
+          #Windows does not use a ControlPath
+          expect(Vagrant::Util::Subprocess).to receive(:execute).with(any_args) { |*args|
+            expect(args[9]).not_to include("ControlPath=/tmp/vagrant-rsync-12345")
+          }.and_return(result)
+        else
+          FileUtils.mkdir("/tmp/vagrant-rsync-12345")
+          expect(Vagrant::Util::Subprocess).to receive(:execute).with(any_args) { |*args|
+            expect(args[9]).to include("ControlPath=/tmp/vagrant-rsync-12345")
+          }.and_return(result)
+        end
 
         subject.rsync_single(machine, ssh_info, opts)
       end


plugins/synced_folders/rsync/helper.rb
```diff
@@ -213,6 +213,10 @@ def self.rsync_single(machine, ssh_info, opts)
           r = Vagrant::Util::Subprocess.execute(*(command + [command_opts]))
         end
 
+        unless Vagrant::Util::Platform.windows?
+          FileUtils.remove_entry_secure(controlpath)
+        end
```
briancain: Can we instead move this to the end of the function with an `ensure`? Then we can be sure this directory is removed just in case an exception is thrown before we reach here. You'll also want to make sure `controlpath` exists in the ensure before removing it just in case the function returns before `controlpath` is created.

briancain: It's probably also a good thing to have `require 'fileutils'` at the top of this file just in case.

aeikum:
Thanks for the review, @briancain. I've pushed a new version with your changes. I noticed it fails the tests, but I don't understand how the test worked before. AFAICT the ControlPath directory in the test doesn't exist, so it isn't really being usefully tested during the `rsync_single` call. It fails after my change because we try to remove the non-existent directory.

One fix may be to change the test to run only on Unix, actually create the temp dir, and then verify it is removed after the `rsync_single` call.

aeikum:
OK, one more version, this time with tests. Again, total Ruby noob here, please review closely.

test/unit/plugins/synced_folders/rsync/helper_test.rb
```diff
@@ -284,15 +284,26 @@
...
       it "creates a tmp dir" do
-        allow(Vagrant::Util::Platform).to receive(:windows?).and_return(false)
         allow(Dir).to receive(:mktmpdir).with("vagrant-rsync-").
           and_return("/tmp/vagrant-rsync-12345")
 
-        expect(Vagrant::Util::Subprocess).to receive(:execute).with(any_args) { |*args|
-          expect(args[9]).to include("ControlPath=/tmp/vagrant-rsync-12345")
-        }.and_return(result)
+        if Vagrant::Util::Platform.windows?
+          #Windows does not use a ControlPath
```
briancain: This is true, which is why the above `allow` was there. I'd put all that back, and remove the if statements here. The `#Platform.windows?` check is not intended for unit tests to call. We can set the result of invoking that however with the `allow` in rspec like we had before.

I would revert back to the original changes. And now we could add an expect test below here but _before_ the `subject.rsync_single(..)` call that looks like:

```ruby
expect(FileUtils).to receive(:remove_entry_secure).with("/tmp/vagrant-rsync-12345")
```

And it would probably be good to have a _Windows_ specific test that checks that FileUtils does not receive `:remove_entry_secure` at all.

test/unit/plugins/synced_folders/rsync/helper_test.rb
```diff
@@ -284,15 +284,26 @@
...
+          expect(Vagrant::Util::Subprocess).to receive(:execute).with(any_args) { |*args|
+            expect(args[9]).to include("ControlPath=/tmp/vagrant-rsync-12345")
+          }.and_return(result)
+        end
 
         subject.rsync_single(machine, ssh_info, opts)
+
+        unless Vagrant::Util::Platform.windows?
+          expect(File.exist?("/tmp/vagrant-rsync-12345")).to be_falsey
+        end
```
briancain: This `unless` and `expect` should be removed completely. We mock out in rspec the creation of tmp files for the method we're testing so this will always be false anyway.

test/unit/plugins/synced_folders/rsync/helper_test.rb
aeikum: My understanding of `allow` is it basically overrides a function's implementation for the sake of the test. If we override `Platform.windows` to return false on a Windows platform, we'll be testing the non-Windows codepath on Windows, which strikes me as not terribly useful. Similarly for returning true when running on a non-Windows platform.

As a concrete example, if we run the non-Windows codepath on Windows, then `rsync_single` will pass `ControlPath=/tmp/vagrant-rsync-12345` to ssh, which is wrong, as that path is invalid on Windows. `rsync_single` should fail (though maybe ssh happens to silently ignore an invalid ControlPath, I don't know).

In other words, the behavior of `rsync_single` is and should be different on each platform, so it makes sense to me to test for different behavior on each platform.

Another problem of using the stubbed `mktmpdir` implementation on Unix is that the temp dir is not actually created, so the ControlPath given to ssh doesn't exist, and again it should fail. That's why I added the `FileUtils.mkdir` call before invoking `rsync_single`.

I'm also confused about your request to test that `remove_entry_secure` is called. That seems like an implementation detail. The functional goal of the patch is to ensure the controlpath temp dir doesn't exist after `rsync_single` exits, so that's what my test verifies. How the implementation accomplishes that isn't important. If we change the implementation to use `FileUtils.rm_rf` for some reason, I don't think the tests should start failing.

briancain: > My understanding of allow is it basically overrides a function's implementation for the sake of the test. If we override Platform.windows to return false on a Windows platform, we'll be testing the non-Windows codepath on Windows, which strikes me as not terribly useful. Similarly for returning true when running on a non-Windows platform.
> As a concrete example, if we run the non-Windows codepath on Windows, then rsync_single will pass ControlPath=/tmp/vagrant-rsync-12345 to ssh, which is wrong, as that path is invalid on Windows. rsync_single should fail (though maybe ssh happens to silently ignore an invalid ControlPath, I don't know).
> In other words, the behavior of rsync_single is and should be different on each platform, so it makes sense to me to test for different behavior on each platform.

Yes that's right. Linux based hosts should be receiving ControlPath, and Windows should not. That's why I am asking for a second test below this that tests for Windows. Which can be done by allowing `Platform.windows?` to return true. Then we can test those code paths there, as well as test that it does not receive `ControlPath`.

> Another problem of using the stubbed mktmpdir implementation on Unix is that the temp dir is not actually created, so the ControlPath given to ssh doesn't exist, and again it should fail. That's why I added the FileUtils.mkdir call before invoking rsync_single.

Unit tests shouldn't be creating tmp files, and we don't actually need it to exist to test the behavior of this method. We can spec out how these methods _should_ work, and test its behavior without creating files on the system. If you take a look at the other tests in this file, we are mostly testing that this method passes the correct rsync arguments to Vagrants Subprocess class:

```ruby
expect(Vagrant::Util::Subprocess).to receive(:execute).with(any_args) { |*args|
  expect(args[9]).to include("ControlPath=/tmp/vagrant-rsync-12345")
}.and_return(result)
```

That path doesn't need to exist, because it's not being used to rsync in these tests. We just want to make sure that, under certain conditions, the correct rsync arguments are passed through.

> I'm also confused about your request to test that remove_entry_secure is called. That seems like an implementation detail. The functional goal of the patch is to ensure the controlpath temp dir doesn't exist after rsync_single exits, so that's what my test verifies. How the implementation accomplishes that isn't important. If we change the implementation to use FileUtils.rm_rf for some reason, I don't think the tests should start failing.

If we change the implementation of how the method removes the tmp dir, then we also change the tests.

Let's take a step back here and see what the original test is doing. I've put notes as ruby comments:

```ruby
it "creates a tmp dir" do
  # These `allow`s are for mocking out _how_ these libraries, external to this method, should work
  allow(Vagrant::Util::Platform).to receive(:windows?).and_return(false)
  allow(Dir).to receive(:mktmpdir).with("vagrant-rsync-").
    and_return("/tmp/vagrant-rsync-12345")

  # this is the main test. We are simply testing that we are passing the
  # correct arguments to the Subprocess class. We are not testing that rsync is actually executing these arguments on a guest.
  # Because of this, we don't actually need tmp files to exist on the system
  expect(Vagrant::Util::Subprocess).to receive(:execute).with(any_args) { |*args|
    expect(args[9]).to include("ControlPath=/tmp/vagrant-rsync-12345")
  }.and_return(result)

  # This expect is now required based on the change you made. We need to test that this is called when this method is invoked.
  expect(FileUtils).to receive(:remove_entry_secure).with("/tmp/vagrant-rsync-12345")
  subject.rsync_single(machine, ssh_info, opts)
end
```

Essentially we are testing that the `rsync_single` class builds the correct set of argumetns that are passed to the Subprocess class. And with your change, we want to test that the temp folder is properly cleaned up, since it was not previously. We want to care about the implementation detail here, because that's what unit tests are for.

So now, we need to test the Windows path. The test should be similar to this, but instead we want to make sure it does _not_ get ControlPath nor does it need to call `:remove_secure_entry`, since the tmp folder never existed in the first place on Windows.

```ruby
it "does not create a tmp dir or send ControlPath on Windows" do
  allow(Vagrant::Util::Platform).to receive(:windows?).and_return(true)

  # I'm not sure exactly how this should look, but essentially we need to test that
  # the variable `args` does not contain "ControlPath" at all in any of the strings
  expect(Vagrant::Util::Subprocess).to receive(:execute).with(any_args) { |*args|
    expect(args).not_to include("ControlPath=/tmp/vagrant-rsync-12345")
  }.and_return(result)

  expect(FileUtils).not_to receive(:remove_entry_secure)
  subject.rsync_single(machine, ssh_info, opts)
end
```

I didn't test the code above yet, but I'm expecting the Windows test will look very similar to that. Hopefully that makes sense.

briancain:
Hi there. Thanks for your contribution! I've opened a new pull request to get this fixed up and merged into Vagrant here: https://github.com/hashicorp/vagrant/pull/10690

If you could, please sign the contributors license agreement! Thanks.

PR #10566, opened 2019-01-05
Do not assume root group on FreeBSD during rsync_post

### LINKED ISSUE #10443: Synced Folders plugin wrongly assumes root group when using rsync on freebsd
### Vagrant version

Vagrant:

    % vagrant -v
    Vagrant 2.2.1

Plugins;

    % vagrant plugin list
    vagrant-berkshelf (5.1.2, global)
      - Version Constraint: > 0
    vagrant-cloudstack (1.5.0, global)
      - Version Constraint: > 0
    vagrant-omnibus (1.5.0, global)
      - Version Constraint: > 0

### Host operating system

MacOS Mojave

    Darwin xxxx 18.2.0 Darwin Kernel Version 18.2.0: Fri Oct  5 19:41:49 PDT 2018; root:xnu-4903.221.2~2/RELEASE_X86_64 x86_64 i386 MacBookPro12,1 Darwin

### Guest operating system

FreeBSD 10.3

### Vagrantfile

```ruby
  config.ssh.shell = 'sh'
  config.ssh.forward_agent = true
  config.ssh.username = "root"
  config.ssh.private_key_path = "~/.ssh/id_rsa"
  config.vm.allowed_synced_folder_types = [:rsync]

  config.vm.define 'node01' do |box|
    box.vm.hostname = 'node01'

    config.vm.provider :cloudstack do |cloudstack, _|
      # Cloudstack api key and other configuration redacted
      cloudstack.display_name  = box.vm.hostname
      cloudstack.template_name = 'freebsd-10.3-20181124'
    end
  end
end
```

### Debug output

https://gist.github.com/nrocco/93f7f7943a59a244d03bcacbf83f52f0

### Expected behavior

I expect that - given the above configuration - when running

    vagrant up

Vagrant successfully finishes with an exit code of `0` and the current working directory on my workstation is rsync'ed to the guest.

### Actual behavior

When running

    vagrant up

Vagrant terminates with the following error:

     INFO interface: Machine: error-exit ["Vagrant::Errors::VagrantError", "The following SSH command responded with a non-zero exit status.\nVagrant assumes that this means the command failed!\n\nfind /vagrant '!' -type l -a '(' ! -user root -or ! -group root ')' -exec chown root:root '{}' +\n\nStdout from the command:\n\n\n\nStderr from the command:\n\nfind: -group: root: no such group\n"]

The problem seems to originate from `plugins/synced_folders/rsync/helper.rb` on line 66

    64         # Folder options
    65         opts[:owner] ||= ssh_info[:username]
    66         opts[:group] ||= ssh_info[:username]

The `helper.rb` script assumes that the name of the group is the same as the name of the user. But in FreeBSD a group with name `root` does not exist. Instead the `root` user belongs to the `wheel` group.

### Steps to reproduce

1. Take the above `Vagrantfile`
2. Use a FreeBSD based box with root user access
3. Run `vagrant up`

nrocco:
This is an attempt to fix the issue #10443

It "fixes" the issue described above but hardcoding `:freebsd` in the `helper.rb` file does not seem like the 'correct' approach.

I am not expecting the code to be merged as-is. I have not tested this PR for regressions.
Any help in pointing me into the right direction is much appreciated.

```
nrocco@macbook vagrant-testing % vagrant provision
==> vm0: Rsyncing folder: /Users/nrocco/Develop/vagrant-testing/ => /vagrant
==> vm0: Running provisioner: shell...
    vm0: Running: /var/folders/bh/z8cxtzk971d45xd7gz50sb5r0000gn/T/vagrant-shell20190105-3221-1448omt.sh
    vm0: Hello World!
    vm0: FreeBSD e4f8c67a 10.3-RELEASE FreeBSD 10.3-RELEASE #0 r297264: Fri Mar 25 02:10:02 UTC 2016     root@releng1.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC  amd64
```

Fixes #10443

### DIFF
--- a/plugins/synced_folders/rsync/helper.rb
+++ b/plugins/synced_folders/rsync/helper.rb
@@ -67,6 +67,11 @@ module VagrantPlugins
         # Folder options
         opts[:owner] ||= ssh_info[:username]
         opts[:group] ||= ssh_info[:username]
+        
+        # On freebsd the root user belongs to the wheel group
+        if machine.guest.name == :freebsd && ssh_info[:username] == 'root'
+          opts[:group] = 'wheel'
+        end
 
         # set log level
         log_level = ssh_info[:log_level] || "FATAL"


plugins/synced_folders/rsync/helper.rb
```diff
@@ -67,6 +67,11 @@ def self.rsync_single(machine, ssh_info, opts)
         # Folder options
         opts[:owner] ||= ssh_info[:username]
         opts[:group] ||= ssh_info[:username]
+        
+        # On freebsd the root user belongs to the wheel group
+        if machine.guest.name == :freebsd && ssh_info[:username] == 'root'
+          opts[:group] = 'wheel'
```
briancain: We should log to the user that we changed the group here so that it's clear what is going on in debug mode:

You'll need to make sure our logging library is loaded at the top....

```ruby
require 'log4r'
```

And we need to initialize a logger object:

```ruby
def self.rsync_single(.... )
  logger = Log4r::Logger.new("vagrant::synced_folders::rsync::helper::rsync_single")
  ...
  ...
```

Then now back here in this if-statement after updating `opts[:group]` we can log to debug that we're updating the group:

```ruby
logger.debug("FreeBSD guest detected, updating guests destination shared folder group to 'wheel' instead of 'root'")
```

briancain:
Hello @nrocco - so after some thinking, I think we are going to end up closing this. Our reasoning behind this is using the `root` user is a non-standard configuration for Vagrant boxes, where Vagrant typically expects a `vagrant` user to be configured. Since using `root` is considered non-standard, it's typically on the users shoulders to properly configure the settings to be used. So for example, you can work around this by telling Vagrant what group the rsynced folders group should be:

```ruby
  config.vm.synced_folder ".", "/vagrant", type: "rsync", group: "wheel"
```

Or in the original Vagrantfile:

```ruby
  config.ssh.shell = 'sh'
  config.ssh.forward_agent = true
  config.ssh.username = "root"
  config.ssh.private_key_path = "~/.ssh/id_rsa"
  config.vm.allowed_synced_folder_types = [:rsync]

  config.vm.synced_folder ".", "/vagrant", type: "rsync", group: "wheel"

  config.vm.define 'node01' do |box|
    box.vm.hostname = 'node01'

    config.vm.provider :cloudstack do |cloudstack, _|
      # Cloudstack api key and other configuration redacted
      cloudstack.display_name  = box.vm.hostname
      cloudstack.template_name = 'freebsd-10.3-20181124'
    end
  end
end
```

Thanks again for making this pull request. We do appreciate it a lot!

nrocco:
Thanks for the feedback @briancain .

I can confirm that indeed the following works

```
config.vm.synced_folder ".", "/vagrant", type: "rsync", group: "wheel"
```

PR #10570, opened 2019-01-07
Show source and destination locations with file provisioner

briancain:
Prior to this commit, Vagrants output would only show that a file
provisioner was running, but had no detail as to what file was being
copied to where. This is especially confusing if a Vagrantfile has
multiple file provisioners. This commit updates that by showing the
source and destination of the file so that it's clearer what is being
copied and to where.


briancain:
Before:

screenshot from 2019-01-07 11-33-00

After!

screenshot from 2019-01-07 11-32-21

PR #10595, opened 2019-01-17
- FIX for #10594

### LINKED ISSUE #10594: [Bug/Master] Network configuration on Solaris 11 boxes is only working when the interface is already present
### Vagrant version
2.2.3 (bug present in code of master branch)

### Host operating system
Windows 7 64 Bit

### Guest operating system
Solaris 11.4 / Solaris 11.3

### Expected behavior
On Solaris 11 boxes, Vagrant should be able to create the appropriate IP addresses to use.

### Actual behavior
Vagrant tries to create an IP adress for an interface, but fails because the interface has not been created.

The corresponding code can be seen in the file: https://github.com/hashicorp/vagrant/blob/master/plugins/guests/solaris11/cap/configure_networks.rb

On Solaris 11 (maybe also on other Solaris versions), your physical network interfaces are "data links".
Vagrant adds the physical data link via VirtualBox configuration, But before one can add an IP address using ipadm create-addr, one has to check if the underlying network interface has been created.

Look what datalinks are there using dladm. Then look for the corresponding interfaces using  ipadm and grep. If the interface has not been created, create the interface using
```
# (e.g. net1)
"#{su_cmd} ipadm create-if #{device}"
```
Only then can the address be created using 
```
"#{su_cmd} ipadm create-addr -T static -a #{network[:ip]}/#{cidr} #{device}/v4"
```

### Steps to reproduce
1. Have a Vagrant Solaris 11 box with only the first network adapter (the NAT one)
2. vagrant up
3. Error is shown

### References
https://docs.oracle.com/cd/E26502_01/html/E28987/gmajn.html (Oracle docs for dladm and ipadm)

danowar2k:
This is the fix in Pull Request form.

Fixes #10594

### DIFF
--- a/plugins/guests/solaris11/cap/configure_networks.rb
+++ b/plugins/guests/solaris11/cap/configure_networks.rb
@@ -17,6 +17,9 @@ module VagrantPlugins
               if machine.communicate.test("ipadm | grep #{device}/v4")
                 machine.communicate.execute("#{su_cmd} ipadm delete-addr #{device}/v4")
               end
+              unless machine.communicate.test("ipadm show-if #{device}")
+                machine.communicate.execute("#{su_cmd} ipadm create-ip #{device}")
+              end
               machine.communicate.execute("#{su_cmd} ipadm create-addr -T static -a #{network[:ip]}/#{cidr} #{device}/v4")
             elsif network[:type].to_sym == :dhcp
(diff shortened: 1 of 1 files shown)


danowar2k:
Huh, I have one Github account and two email addresses (private and work-related). I committed using my work address. I have added the work address to Github afterwards.

I signed the CLA using my private address. Is that why this is blocked?

briancain:
@danowar2k - No worries, the bot just needed to recheck the CLA. I reran the check and it seems to be ok now. Thanks!

PR #10625, opened 2019-01-28
Fix issue 9584

### LINKED ISSUE #9584: ansible_local's pip installer has a hardcoded curl get_pip.py piped to python
When the `ansible_local` provisioner configures `ansible.install_mode = pip` this command is invoked: https://github.com/hashicorp/vagrant/blob/cb61497ad61873653a8042da4293dea06dffc545/plugins/provisioners/ansible/cap/guest/pip/pip.rb#L23

Piping curl into python with a static command prevents me from setting [get_pip.py args](https://pip.pypa.io/en/stable/installing/#get-pip-py-options). In my case I need to set a proxy, but I don't want to set a proxy for the entire provisioner as that causes issues when running certain playbooks. As far as I can tell there's no way to set a proxy for only the ansible install phase of the provisioner.

As a workaround, I use a shell provisioner to install pip manually with my custom get_pip.py args before `ansible_local` runs, but I still have to wait for `ansible_local`'s static curl | python command to timeout on every install which is fairly annoying. I would use the package manger ansible install_mode, but my ansible playbooks require specific ansible versions and you can only target an ansible version with the pip install_mode.

### Vagrant version
Vagrant 2.0.2

vdebroy:
In keeping with the suggested fix discussed in the issue: an additional option for pip_install_cmd has been included. Default behavior is maintained and it is only if the additional option is provided in the config that it is used. Unit tests have been included. 

Fixes #9584

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/pip/pip.rb
+++ b/plugins/provisioners/ansible/cap/guest/pip/pip.rb
@@ -28,7 +28,7 @@ module VagrantPlugins
             # Thus, we stick with the default on an empty string.
             # Typecast added in the check for safety.
 
-            if pip_install_cmd == pip_install_cmd.to_s.empty?
+            if pip_install_cmd.to_s.empty?
               pip_install_cmd=DEFAULT_PIP_INSTALL_CMD
             end
 
--- a/plugins/provisioners/ansible/cap/guest/pip/pip.rb
+++ b/plugins/provisioners/ansible/cap/guest/pip/pip.rb
@@ -18,9 +18,9 @@ module VagrantPlugins
             machine.communicate.sudo "pip install #{args_array.join(' ')}"
           end
 
-          def self.get_pip(machine)
+          def self.get_pip(machine, pip_install_cmd="curl https://bootstrap.pypa.io/get-pip.py | sudo python")
             machine.ui.detail I18n.t("vagrant.provisioners.ansible.installing_pip")
-            machine.communicate.execute "curl https://bootstrap.pypa.io/get-pip.py | sudo python"
+            machine.communicate.execute pip_install_cmd
           end
 
         end
--- a/plugins/provisioners/ansible/cap/guest/pip/pip.rb
+++ b/plugins/provisioners/ansible/cap/guest/pip/pip.rb
@@ -5,6 +5,8 @@ module VagrantPlugins
       module Guest
         module Pip
 
+          DEFAULT_PIP_INSTALL_CMD = "curl https://bootstrap.pypa.io/get-pip.py | sudo python"
+
           def self.pip_install(machine, package = "", version = "", pip_args = "", upgrade = true)
             upgrade_arg = "--upgrade" if upgrade
             version_arg = ""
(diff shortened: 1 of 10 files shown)


test/unit/plugins/provisioners/ansible/cap/guest/pip/pip_test.rb
```diff
@@ -0,0 +1,52 @@
...
+            subject.get_pip(machine,pip_install_command)
+        end
+
+        it "installs pip using the default command if the argument is empty" do
+            pip_install_command = ""
+            expect(communicator).to receive(:execute).with("curl https://bootstrap.pypa.io/get-pip.py | sudo python")
+            subject.get_pip(machine,pip_install_command)
+        end 
+        
+        it "installs pip using the default command if the argument is UNSET_VALUE" do
```
chrisroberts: This test would be better to pass in a `nil` value.

test/unit/plugins/provisioners/ansible/cap/guest/pip/pip_test.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  let(:machine) { iso_env.machine(iso_env.machine_names[0], :dummy) }
+  let(:communicator) { double("comm") }
+
+  before do
+    allow(machine).to receive(:communicate).and_return(communicator)
+    allow(communicator).to receive(:execute).and_return(true)
+  end
+
```
chrisroberts: Please fix the indentation within the `describe` blocks below. They appear to be indented with 4 spaces while 2 are used elsewhere.

vdebroy:
Thanks so much for reviewing this and for the helpful feedback @chrisroberts! I should be able to make the updates pretty soon. Thanks again!

plugins/provisioners/ansible/cap/guest/pip/pip.rb
```diff
@@ -18,9 +20,20 @@ def self.pip_install(machine, package = "", version = "", pip_args = "", upgrade
...
-          def self.get_pip(machine)
+          def self.get_pip(machine, pip_install_cmd=DEFAULT_PIP_INSTALL_CMD)
+
+            # The objective here is to get pip either by default
+            # or by the argument passed in. The objective is not 
+            # to circumvent the pip setup by passing in nothing.
+            # Thus, we stick with the default on an empty string.
+            # Typecast added in the check for safety.
+
+            if pip_install_cmd == pip_install_cmd.to_s.empty?
```
chrisroberts: This conditional should just be:

```ruby
if pip_install_cmd.to_s.empty?
```

vdebroy: Thanks so much for catching that Sir! That was silly of me. You're absolutely right. I just made the update locally and re-ran the tests and they passed. I will commit. 
tests-passing-locally

plugins/provisioners/ansible/cap/guest/pip/pip.rb
```diff
@@ -18,9 +20,20 @@ def self.pip_install(machine, package = "", version = "", pip_args = "", upgrade
...
-          def self.get_pip(machine)
+          def self.get_pip(machine, pip_install_cmd=DEFAULT_PIP_INSTALL_CMD)
+
+            # The objective here is to get pip either by default
+            # or by the argument passed in. The objective is not 
+            # to circumvent the pip setup by passing in nothing.
+            # Thus, we stick with the default on an empty string
+            # or if it is an UNSET_VALUE.
+
+            if pip_install_cmd == Vagrant.plugin("2", :config)::UNSET_VALUE || pip_install_cmd.empty?
```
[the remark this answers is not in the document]
vdebroy: Should be fixed in most recent commit. Thanks for your tip, I think I did it correctly this time.

plugins/provisioners/ansible/cap/guest/pip/pip.rb
```diff
@@ -5,6 +5,8 @@ module Cap
       module Guest
         module Pip
 
+          DEFAULT_PIP_INSTALL_CMD = "curl https://bootstrap.pypa.io/get-pip.py | sudo python"
```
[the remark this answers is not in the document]
vdebroy: Should be fixed in most recent commit.

test/unit/plugins/provisioners/ansible/cap/guest/pip/pip_test.rb
vdebroy: Should be fixed in most recent commit.

test/unit/plugins/provisioners/ansible/cap/guest/pip/pip_test.rb
vdebroy: Should be fixed in most recent commit.

vdebroy:
Thanks for all your help @chrisroberts! The PR should be ready for review again.

PR #10659, opened 2019-02-10
implement nfs shared folders for ipv6

uwej711:
I tried to use nfs shared folders with ipv6 (virtualbox and linux guest), but it did not work. The changes below should make it work. Maybe not the cleanest solution, so any feedback is welcome.


uwej711:
Dear approvers, 

can I do anything to help / getting feedback? Currently nfs shared folders do not work, when you set up your host-only network with an ipv6 address. Due to some issues with unrelated software we have to use, using ipv6 would solve some issues for us.

Kind regards
Uwe

PR #10690, opened 2019-02-22
Cleanup rsync tmp folders

briancain:
This pull request ensures that the tmp folder created by rsync gets properly cleaned up on the host.

### DIFF
--- a/plugins/synced_folders/rsync/helper.rb
+++ b/plugins/synced_folders/rsync/helper.rb
@@ -234,7 +234,7 @@ module VagrantPlugins
           end
         end
       ensure
-        FileUtils.remove_entry_secure(controlpath) if controlpath
+        FileUtils.remove_entry_secure(controlpath, true) if controlpath
       end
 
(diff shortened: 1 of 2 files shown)


aeikum:
Pretty sure we don't have any patents on cleaning up temp directories, but lawyers gotta eat, I guess.

Thanks for moving this forward.

plugins/synced_folders/rsync/helper.rb
```diff
@@ -232,6 +233,8 @@ def self.rsync_single(machine, ssh_info, opts)
               message: err.to_s
           end
         end
+      ensure
+        FileUtils.remove_entry_secure(controlpath) if controlpath
```
chrisroberts: Can we update this to be `FileUtils.remove_entry_secure(controlpath, true)` so it becomes a best effort removal and does not force a failure if the removal fails for some reason?

briancain:
@chrisroberts - updated with feedback!

Issue #9873, opened 2018-05-25
vagrant docker: docker network create


StuporHero:
Vagrant 2.1.1
Windows 10
Docker 18.03.1-ce-win65 (17513)

I have a custom Ubuntu 16.04 image that has the ChefDK setup in it to run `kitchen test` on a cookbook. Kitchen is setup to use Docker to execute and test the cookbook. In order for this to work in Vagrant, I need to manually create a docker network and include it my configuration. This works just fine, but for maximum portability of the build configuration, I'd like to be able to create the network as part of the `vagrant up` command. The manual command is simple:

```
Usage:  docker network create [OPTIONS] NETWORK

Create a network

Options:
      --attachable           Enable manual container attachment
      --aux-address map      Auxiliary IPv4 or IPv6 addresses used by Network driver (default map[])
      --config-from string   The network from which copying the configuration
      --config-only          Create a configuration only network
  -d, --driver string        Driver to manage the Network (default "bridge")
      --gateway strings      IPv4 or IPv6 Gateway for the master subnet
      --ingress              Create swarm routing-mesh network
      --internal             Restrict external access to the network
      --ip-range strings     Allocate container ip from a sub-range
      --ipam-driver string   IP Address Management Driver (default "default")
      --ipam-opt map         Set IPAM driver specific options (default map[])
      --ipv6                 Enable IPv6 networking
      --label list           Set metadata on a network
  -o, --opt map              Set driver specific options (default map[])
      --scope string         Control the network's scope
      --subnet strings       Subnet in CIDR format that represents a network segment
```

This should probably get its own subsection in the `config.vm.provider 'docker'` block with the implied addition of the `--network` flag on the `docker run` command. This would also be useful for multi-container setups where multiple containers are created together by Vagrant since the `--link` flag is now deprecated. Thanks for your consideration!

StuporHero:
A couple of additional notes:
- If the network already exists, the `vagrant up` command should not fail.
- There should be some kind of "remove on destroy" option to automatically run `docker network rm <network>` on `vagrant destroy`.

briancain:
I have created a pull request https://github.com/hashicorp/vagrant/pull/10702 which addresses this issue. Thanks!

Issue #10656, opened 2019-02-09
Struggling with static IP + Docker provider


jesusgn90:
Hi team,

I'm struggling with static IP assignment for Docker provider.

```rb
  config.vm.define "dockeragent" do |dockeragent|
    dockeragent.vm.provider "docker" do |d|
      d.build_dir = "../installers/docker/"
    end
    dockeragent.vm.network :private_network, ip: "172.16.1.3"
  end
```

I want my container named `dockeragent` uses `172.16.1.3` but it's using always `172.16.1.1`, my problem is that I want multiple containers, and each container must have its own IP.

Help!

Thanks in advance!

briancain:
Hey there @jesusgn90 - It looks like this isn't currently supported with the docker provider :thinking: https://www.vagrantup.com/docs/docker/basics.html#synced-folders-and-networking

> Private and public networks are not currently supported.

I _think_ this should be covered in this issue here https://github.com/hashicorp/vagrant/issues/9873

jesusgn90:
This use case is critical for us, is there an ETA for this feature?

Regards!

briancain:
@jesusgn90 - I'm going to try to get started on it soon, but it won't be a small feature by any means so I can't give you a definite time frame. It does look possible that you could use triggers as a work around for setting up networking with docker in the meantime:

```ruby
config.trigger.before :up do |trigger|
  trigger.name = 'Create network vagrant'
  trigger.run = { inline: 'docker network create vagrant' }
end

config.trigger.after :destroy do |trigger|
  trigger.name = 'Remove network vagrant'
  trigger.run = { inline: 'docker network rm vagrant' }
end
```

But replace the inline scripts with the networks you _actually_ wish to create and delete, of course!

PR #10702, opened 2019-02-27
Docker Provider Network Support

briancain:
This is a pull request for adding networking support for the Docker provider.

A simple Docker guest for testing:

```ruby
config.vm.define "docker"  do |docker|
  docker.vm.network :private_network, ip: "172.20.128.2", subnet: "172.20.0.0/16"
  docker.vm.network :private_network, type: "dhcp"
  docker.vm.provider "docker" do |d|
    d.git_repo = "https://github.com/briancain/nginx-docker-test.git"
  end
end
```

Fixes https://github.com/hashicorp/vagrant/issues/9873 https://github.com/hashicorp/vagrant/issues/10656


plugins/providers/docker/action/network.rb
```diff
@@ -0,0 +1,96 @@
...
+
+          # Splits the networking options to generate the proper CLI flags for docker
+          if options[:type] != "dhcp"
+            options.each do |opt, value|
+              vals = opt.to_s.split("__")
+              if vals[0] == "docker" && vals.size == 3
+                cli_opts[vals[1].to_sym].concat(["--#{vals[2]}", value])
+              end
+            end
+          end
```
briancain: Ok, this is generally the idea I was thinking about in the comments above. Essentially a user would give us:

```ruby
 docker.vm.network :private_network,
        docker__connect__ip: "172.20.128.2", docker__create__subnet: "172.20.0.0/16",
        docker__create__driver: "bridge"
```

And this function would generate the following flags internally to use for the right `docker` commands:

```ruby
{create: ["--subnet", "172.20.0.0/16", "--driver", "bridge"], connect: ["--ip", "172.20.128.2"]}
```

Then later on we can use this hashes keys to invoke the proper `docker network` command with the configured options. Maybe this is a too auto-magical but it seems to work at the moment.

briancain: The pro to this approach at least is that it decouples Vagrant from having to know every single `docker network` subcommand and all of its flags. The con however is that it shifts the responsibility of the user to know which options have to go with each subcommand, which on the surface might not be super intuitive. However if they're going to use networks in docker, it is likely true that they will have to understand each command and what options it needs to configure a network anyway. Also a pro for this subcommand "namespaced" approach is that it makes sure we don't have any config option collisions between `docker network` subcommand options if we had only gone with the `docker__option` approach.

jesusgn90:
Hi @briancain , this is music to my ears.

Is there any ETA for a beta version? I meant not a production stage but enough to test it.

I also may give you some feedback.

Regards,
Jesús

briancain:
Hey @jesusgn90 - it's technically working now. I'm finishing up the tests and I need to write documentation for the website for the feature. I'd imagine it should be out in the next release of Vagrant. If you're running Vagrant from source feel free to test. But since this is still in pull request form I'd recommend waiting unless you are just curious and want to play around with the feature/give feedback!

plugins/providers/docker/action/network.rb
```diff
@@ -0,0 +1,92 @@
...
+        def initialize(app, env)
+          @app = app
+          @logger = Log4r::Logger.new('vagrant::plugins::docker::network')
+        end
+
+        # @param[Hash] options - options from the network config
+        # @returns[Array] cli_opts - an array of strings used for the network commnad
+        def generate_create_cli_arguments(options)
+          cli_opts = []
+          ignored_options = ["ip", "protocol", "id", "alias"]
```
chrisroberts: These would be better as a constant

jesusgn90:
It's almost done guys! Go go go!

website/source/docs/docker/networking.html.md
```diff
@@ -0,0 +1,222 @@
...
+
+```ruby
+docker_connect__option: "value"
+```
+
+It should be noted that if you want the container to have a specific IP instead
+of using DHCP, you also will have to specify a subnet due to how docker networks behave.
+
+It should also be noted that if you want a specific IPv6 address, your network
+option should use `ip6` rather than `ip`. If you just want to use DHCP, you can
```
chrisroberts: The `ip6` and `ipv6` references should be removed here as they are no longer used (ipv4 and ipv6 are detected automatically during setup).

website/source/docs/docker/networking.html.md
```diff
@@ -0,0 +1,222 @@
...
+simply say `type: "dhcp"` insetad. More examples are shared below which demonstrate
```
chrisroberts: `insetad` -> `instead`

website/source/docs/docker/networking.html.md
```diff
@@ -0,0 +1,222 @@
...
+1. A IPv4 IP address assigned by DHCP
+2. A IPv4 IP address 172.20.128.2 on a network with subnet 172.20.0.0/16
+3. A IPv6 IP address assigned by DHCP on subnet 2a02:6b8:b010:9020:1::/80
+
+```ruby
+Vagrant.configure("2") do |config|
+  config.vm.define "docker"  do |docker|
+    docker.vm.network :private_network, type: "dhcp", docker_network__internal: true
+    docker.vm.network :private_network,
+        ip: "172.20.128.2", subnet: "172.20.0.0/16"
```
chrisroberts: Use `netmask` instead of `subnet`: `netmask: 16`

website/source/docs/docker/networking.html.md
```diff
@@ -0,0 +1,222 @@
...
+    docker.vm.network :private_network, type: "dhcp", ipv6: "true", subnet: "2a02:6b8:b010:9020:1::/80"
```
chrisroberts: Remove `ipv6` option

website/source/docs/docker/networking.html.md
```diff
@@ -0,0 +1,222 @@
...
+```
+A Docker command executed by Vagrant didn't complete successfully!
+The command run along with the output from the command is shown
+below.
+
+Command: ["docker", "network", "create", "vagrant_network_172.20.0.0/16", "--subnet=172.20.0.0/16", "--unsupported=option", {:notify=>[:stdout, :stderr]}]
+
+Stderr: unknown flag: --unsupported
+See 'docker network create --help'.
+
```
chrisroberts: This section is no longer try as the options are not passed to the CLI directly anymore (scoped options would allow this behavior: `docker_network__unsupported: "option"`)

website/source/docs/docker/networking.html.md
```diff
@@ -0,0 +1,222 @@
...
+Where `option` corresponds to the given flag that will be provided to the `docker network create`
+command. Similarly, if there is a value you wish to enable when connecting a container
+to a given network, you can use the following value in your network config:
+
+```ruby
+docker_connect__option: "value"
+```
+
+It should be noted that if you want the container to have a specific IP instead
+of using DHCP, you also will have to specify a subnet due to how docker networks behave.
```
chrisroberts: Defining the netmask/subnet is not required. If undefined, it will default to `/24` for ipv4 and `/64` for ipv6. If a network already exists and the provided IP address is within that network's subnet, that network will be used.

website/source/docs/docker/networking.html.md
```diff
@@ -0,0 +1,222 @@
...
+## Useful Debugging Tips
+
+If you provide Vagrant with a faulty config option when setting up a network, Vagrant
+will pass that option along to the `docker network` commands it uses. That command
+line tool should give you some insight if there is something wrong with the option
+you configured:
+
+```ruby
+docker.vm.network :private_network,
+  ip: "172.20.128.2", subnet: "172.20.0.0/16",
```
chrisroberts: Use `netmask: 16` instead of `subnet`

website/source/docs/docker/networking.html.md
```diff
@@ -0,0 +1,222 @@
...
+33: eth3@if34: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
+    link/ether 02:42:ac:15:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
+    inet 172.21.0.2/16 brd 172.21.255.255 scope global eth3
+       valid_lft forever preferred_lft forever
+    inet6 2a02:6b8:b010:9020:1::2/80 scope global nodad
+       valid_lft forever preferred_lft forever
+    inet6 fe80::42:acff:fe15:2/64 scope link
+       valid_lft forever preferred_lft forever
+```
+
```
chrisroberts: Might also want to add an example for using an existing non-vagrant managed network via naming. Like, if `my-network` is already defined, using the following for attaching to it:

```
docker.vm.network :private_network, type: "dhcp", name: "my-network"
```

Issue #10705, opened 2019-02-28
bsdtar errors when running as root


steve-simpson:
### Vagrant version
Vagrant 2.2.4

### Host operating system
Operating System: CentOS Linux 7 (Core)
Kernel: Linux 3.10.0-957.5.1.el7.x86_64
Architecture: x86-64

### Guest operating system
CentOS 7

### Debug output
https://gist.github.com/steve-simpson/1852909f112770ba0c282abb50dc3d2a

### Expected behavior
mybasebox1 should have been added normally

### Actual behavior
Box is not added and errors with: `bsdtar: Error exit delayed from previous errors.`

### Steps to reproduce
1. Create a simple base box using virtual box as a non privileged user. In my case this was made on Debian 10 using VirtualBox 5.2.2 (Vagrant 2.2.3)
2. Attempt to add the created base box to CentOS 7 server as root user. I've attempted to "pull" the box from Vagrant Cloud and also by simply copying the `.box` to the remote server and adding manually
3. We should see `bsdtar: Error exit delayed from previous errors.`

### References
Similar issue but not an exact copy. The resolution appears to be the same
https://github.com/hashicorp/vagrant/issues/3025

I am confident that this issue can be resolved by changing `box_collection.rb` line 119 to:
```ruby
"bsdtar", "--no-same-owner", "--no-same-permissions", "-v", "-x", "-m", "-s", "|\\\\\|/|", "-C", temp_dir.to_s, "-f", path.to_s)
```
I will submit a pull request if this is suitable?

steve-simpson:
I believe that this issue is due to the default behaviour of bsdtar, specifically:

>       --no-same-owner
> 	     (x	mode only) Do not extract owner	and group IDs.	This is	the
> 	     reverse of	--same-owner and the default behavior if tar is	run as
> 	     non-root.
> 
>      --no-same-permissions
> 	     (x	mode only) Do not extract full permissions (SGID, SUID,	sticky
> 	     bit, ACLs,	extended attributes or extended	file flags).  This is
> 	     the reverse of -p and the default behavior	if tar is run as non-
> 	     root.

chrisroberts:
@steve-simpson Reading through this, I would agree that including those options will likely resolve the issue (though generally I would argue that running as non-root would be the better solution). Since there are times where running Vagrant as root is desirable, it makes sense to include these options so the behavior is consistent when running as root or non-root user. 

A pull request would be great!

steve-simpson:
[...]

PR #10707, opened 2019-03-01
Added --no-same-owner & --no-same-permissions flags to bsdtar in box_…

steve-simpson:
…collection.rb

As discussed in #10705 this pull request fixes some edge case behaviour when trying to unpack a vagrant box as a root user.


JuPlutonic:
--no-same-owner = -o on gnu-tar/Linux and on bsdtar/Mac
So use -o

PR #10717, opened 2019-03-07
Add proper VirtualBox share folders support for FreeBSD guests

### LINKED ISSUE #8884: VBox Shared Folders on FreeBSD Guest Error Out
*Edit* Since I closed this, VirtualBox has added support for VBoxSF on FreeBSD hosts as well as guests.

Hi guys! When I boot a FreeBSD guest with my Vagrantfile, I receive the following message.
`Vagrant is not able to mount VirtualBox shared folders on BSD-based guests. BSD-based guests do not support the VirtualBox filesystem at this time.`
The error goes on to state that this is not a bug in Vagrant, however this is untrue.
With the port [virtualbox-ose-additions](http://www.freshports.org/emulators/virtualbox-ose-additions), 
one can simply run `mount -t vboxsf Vagrant /vagrant` to mount the VirtualBox Shared Folder inside a FreeBSD guest. They function the exact same as they would in Linux, at this point.

### Vagrant version
Vagrant 1.9.5

### Host operating system
FreeBSD 11.0 x64

### Guest operating system
FreeBSD 11.0 x64

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

  config.vm.box = "freebsd/FreeBSD-11.0-RELEASE-p1"
  config.ssh.shell = "sh -c 'SH_ENV=/etc/profile exec sh'"

  config.vm.define 'dev-genericname', autostart: false do |genericname|

    genericname.vm.box = 'freebsd/FreeBSD-11.0-RELEASE-p1'
    genericname.vm.network "private_network", type: "dhcp"
    genericname.vm.base_mac = "macaddress"

    genericname.vm.provider "virtualbox" do |vb|
      vb.name   = "genericname"
      vb.memory = 1024
      vb.cpus   = 1
    end

# Provisioning stuff
  end
end

```

Please note, if you are using Homestead or a different Vagrantfile format, we
may be unable to assist with your issue. Try to reproduce the issue using a
vanilla Vagrantfile first.

### Expected behavior
Guest should mount shared folder

### Actual behavior
Vagrant claims freebsd has no support for shared folders

### Steps to reproduce
1. Download FreeBSD vagrantbox from atlas
2. Write vagrantfile
3. `vagrant up freebsdvm`

briancain:
This commit adds proper support for FreeBSD guests when mounting share folders, assuming that the proper guest additions are in place on the guest.

Tested using the `bento/freebsd-11`  and the `generic/hardenedbsd11` guest.

Fixes #8884


thompsa:
This change modifies all BSD guest clients. Here is a version which adds it for FreeBSD and also ensures the kernel module exists and is loaded before attempting the mount.

https://github.com/thompsa/vagrant/commit/ee4bb779c2c4775f314d01de6fdbf512db53cef3

briancain:
Hi @thompsa - from what I understand and through my testing, invoking `vboxvfs` with the mount command will load the module the module for you, so checking/loading it with Vagrant is not required. It is also not something we do with other guests when using the virtualbox guest additions:

```
$ kldstat
Id Refs Address            Size     Name
 1   22 0xffffffff80200000 20647c8  kernel
 2    1 0xffffffff82266000 381130   zfs.ko
 3    2 0xffffffff825e8000 a380     opensolaris.ko
 4    3 0xffffffff825f3000 8b3d8    vboxdrv.ko
 5    1 0xffffffff82819000 2e790    vboxguest.ko
 6    2 0xffffffff82848000 2af0     vboxnetflt.ko
 7    2 0xffffffff8284b000 9a20     netgraph.ko
 8    1 0xffffffff82855000 14b8     ng_ether.ko
 9    1 0xffffffff82857000 3f70     vboxnetadp.ko
$ sudo mount -t vboxvfs vagrant /vagrant
$ kldstat
Id Refs Address            Size     Name
 1   24 0xffffffff80200000 20647c8  kernel
 2    1 0xffffffff82266000 381130   zfs.ko
 3    2 0xffffffff825e8000 a380     opensolaris.ko
 4    3 0xffffffff825f3000 8b3d8    vboxdrv.ko
 5    2 0xffffffff82819000 2e790    vboxguest.ko
 6    2 0xffffffff82848000 2af0     vboxnetflt.ko
 7    2 0xffffffff8284b000 9a20     netgraph.ko
 8    1 0xffffffff82855000 14b8     ng_ether.ko
 9    1 0xffffffff82857000 3f70     vboxnetadp.ko
10    1 0xffffffff8285b000 200b0    vboxvfs.ko
```

This change should also be just fine for other BSD guests, unless they use a different `vbox` option for the `mount` command (like what the original linux cap uses, `vboxsf`)? The intention is to do away with hard coded error message completely.

briancain:
Also just tested this using the `generic/hardenedbsd11` box and it mounts folders no problem :+1:

briancain:
It does look like mostly only flavors of freeBSD will work with this patch, as the other ones I've tried like NetBSD and openBSD still don't have proper virtualbox guest tools. Happy to move this over to the freeBSD guest since that seems to be the case. The other thing is we could keep this on the BSD guest, and then the support would still be there for whenever those OS's get proper virtualbox guest support. :thinking:

thompsa:
Good to know that the module loading isnt needed. The only difference now is the generic bsd vs freebsd layout. If the others don't support it it would be good to retain the existing error message rather than trying the mount and not working.

simon0191:
does this work for MacOS guests?

briancain:
@simon0191 - No, but macOS guests should already be supported!

PR #10763, opened 2019-03-29
FreeBSD: Simplify the listing of interfaces

jlduran:
According to [ifconfig(8)], to list only Ethernet interfaces, excluding all other interface types, including the loopback interface, the command to use should be:

    ifconfig -l ether

Related to: #8760

[ifconfig(8)]: https://www.freebsd.org/cgi/man.cgi?query=ifconfig&sektion=8

### DIFF
--- a/plugins/guests/freebsd/cap/configure_networks.rb
+++ b/plugins/guests/freebsd/cap/configure_networks.rb
@@ -18,8 +18,8 @@ module VagrantPlugins
           # Remove any previous network additions to the configuration file.
           commands << "sed -i '' -e '/^#VAGRANT-BEGIN/,/^#VAGRANT-END/ d' /etc/rc.conf"
 
-          comm.sudo("ifconfig -a | grep -o '^[0-9a-z]*' | grep -v '^lo'", options) do |_, stdout|
-            interfaces = stdout.split("\n")
+          comm.sudo("ifconfig -l ether", options) do |_, stdout|
+            interfaces = stdout.split
           end
 
           networks.each.with_index do |network, i|
--- a/test/unit/plugins/guests/freebsd/cap/configure_networks_test.rb
+++ b/test/unit/plugins/guests/freebsd/cap/configure_networks_test.rb
@@ -13,8 +13,7 @@ describe "VagrantPlugins::GuestFreeBSD::Cap::ConfigureNetworks" do
 
   before do
     allow(machine).to receive(:communicate).and_return(comm)
-    comm.stub_command("ifconfig -a | grep -o '^[0-9a-z]*' | grep -v '^lo'",
-      stdout: "em1\nem2")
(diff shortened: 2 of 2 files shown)


jlduran:
By using a regex there will always be a corner case that we miss. For example, if we rename the interfaces to contain uppercase letters...

1. Create a `Vagrantfile` with the following:

   ```ruby
     config.vm.box = 'freebsd/FreeBSD-12.0-RELEASE'
     config.vm.hostname = 'host.example.com'
     config.vm.network 'private_network', type: 'dhcp'
     config.vm.synced_folder '.', '/vagrant', type: 'nfs'
     config.ssh.shell = 'sh'

     config.vm.provider :virtualbox do |vb|
       vb.customize ['modifyvm', :id, '--nictype1', 'virtio']
       vb.customize ['modifyvm', :id, '--nictype2', 'virtio']
     end

     config.vm.provision 'shell', inline: <<-ROOT
       # Rename vtnet0 to Vagrant0
       #    and vtnet1 to NFS
       #
       # NB: The capital letters intentionally miss
       #     the current regex
       sysrc ifconfig_vtnet0_name=Vagrant0
       sysrc ifconfig_vtnet1_name=NFS
     ROOT
   end
   ```

2. Reload:

       vagrant reload

3. Output:

   ```
   ...

   ifconfig -a | grep -o '^[0-9a-z]*' | grep -v '^lo'

   Stdout from the command:



   Stderr from the command:

   ```

jlduran:
You are welcome!

For a really old (and unsupported) version of FreeBSD, or Darwin, which grabbed most userland from FreeBSD years ago, something like this could be used:

```ruby
comm.sudo("ifconfig -l | sed 's/lo[0-9]* //g'", options) do |_, stdout|
  interfaces = stdout.split
end
```

This should catch interface names that start with `lo`, i.e. `ifconfig <iface> name logical`. But, again, there is a chance of a corner case I might be missing.

briancain:
@jlduran - ah I see. The address family filter probably doesn't exist for older versions of `ifconfig`? I think that's probably alright. I looked around on Vagrant Cloud for some older freebsd boxes....I tried the bento boxes for freebsd-10.2 and 9.2. Both seemed to handle the update alright. But most of the boxes on VC seem to be either version 10 or 11.

jlduran:
The `ether` _addres_family_ was added in freebsd/freebsd@16f88d4b690af7f1a6b7b9cfd5721f7a60c53fa7 (2000).

For this particular case, which lists all Ethernet interfaces excluding loopbacks (non-ether), the code was added in freebsd/freebsd@c7017d0550079fead9d0fe4bf431577d879ca9a8 (2009). `ether` is an alias to `link`, the commit just treats it different when used with the `-l`flag.

So yeah, I think we should be safe! 😅

PR #10769, opened 2019-04-01
Add a new --args parameter to rsync command

aeikum:
This command lets the user override the default arguments to rsync on the command line. I also added the --verbose switch, and it now prints the full rsync command when in verbose mode.

### DIFF
--- a/plugins/synced_folders/rsync/command/rsync.rb
+++ b/plugins/synced_folders/rsync/command/rsync.rb
@@ -27,6 +27,9 @@ module VagrantPlugins
             o.on("--[no-]rsync-chown", "Use rsync to modify ownership") do |chown|
               options[:rsync_chown] = chown
             end
+            o.on("--verbose", "Enable verbose output") do |v|
+              options[:verbose] = v
+            end
           end
 
           # Parse the options and return if we don't have any target.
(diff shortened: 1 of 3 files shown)


briancain:
Hey there @aeikum - is there a reason why you would want to one-off change the rsync arguments used for a guest instead of using the existing `rsync__args` option for your synced folder?

I think adding this would add some confusion as to where your source of truth is coming from when rsyncing folders, and it would also mean your arguments would be lost for others when using your Vagrantfile. It would also mean that your rsync arguments would be lost when bringing up the guest with `vagrant up`.

aeikum:
Sure. I have a project which has a symlink to an external directory that isn't synced to the VM. I wanted to use `rsync -L` to copy the symlink targets into the VM, and this seemed like an elegant solution to that. Like `vagrant ssh`, this is more intended for active development on a particular VM than for setting up a reproducible build system.

briancain:
@aeikum - hmm ok. I think this makes more sense to use something like the `vagrant upload` command: https://www.vagrantup.com/docs/cli/upload.html

The `vagrant rsync` command just re-syncs your shared folders defined in your Vagrantfile that are defined to use `rsync`. `vagrant upload` doesn't have any notion I think of uploading symlinks yet however, so there might be some improvements with that command instead.

aeikum:
Ah, I didn't know about upload. I agree I should be able to make that work.

Feel free to merge the first 2 commits if you like them, otherwise we can close this. Thanks for your time.

briancain:
Thanks! I'll go ahead and close this.

PR #10770, opened 2019-04-01
Advanced folder sync config for puppet shared folders.

vStone:
Allows configuring the shared folder config by supplying
a hash in stead of an array. Only few (rsync?) shared-folder
types use the :args key used here.


vStone:
Looks like this might be a dupe #10345 #10370

PR #10790, opened 2019-04-12
VAGRANT_DISABLE_VBOXSYMLINKCREATE is not true by default

dentarg:
This text was introduced in https://github.com/hashicorp/vagrant/pull/9354 ([commit](https://github.com/hashicorp/vagrant/commit/b16ca2e384a4d548d5b907e00662693ce05cca4b#diff-aa3fa2a3966ebd8cc74d249fa15dfc46)):

> `VAGRANT_DISABLE_VBOXSYMLINKCREATE`
> If set, this will disable the ability to create symlinks with all virtualbox shared folders. Defaults to true if the option is not set. This can be overridden on a per-folder basis within your Vagrantfile config by settings the `SharedFoldersEnableSymlinksCreate` option to true.

I read "Defaults to true if the option is not set." that `VAGRANT_DISABLE_VBOXSYMLINKCREATE` defaults to true if not set, and that is not correct. If not set, I see this warning on `vagrant up`:

```
Vagrant is currently configured to create VirtualBox synced folders with
the `SharedFoldersEnableSymlinksCreate` option enabled. If the Vagrant
guest is not trusted, you may want to disable this option. For more
information on this option, please refer to the VirtualBox manual:

  https://www.virtualbox.org/manual/ch04.html#sharedfolders

This option can be disabled globally with an environment variable:

  VAGRANT_DISABLE_VBOXSYMLINKCREATE=1

or on a per folder basis within the Vagrantfile:

  config.vm.synced_folder '/host/path', '/guest/path', SharedFoldersEnableSymlinksCreate: false
```

### DIFF
--- a/website/source/docs/other/environmental-variables.html.md
+++ b/website/source/docs/other/environmental-variables.html.md
@@ -135,7 +135,7 @@ This can be disabled setting this environment variable.
 ## `VAGRANT_DISABLE_VBOXSYMLINKCREATE`
 
 If set, this will disable the ability to create symlinks with all virtualbox
-shared folders. Defaults to true if the option is not set. This can be overridden
+shared folders. Defaults to false if the option is not set. This can be overridden
 on a per-folder basis within your Vagrantfile config by settings the
 `SharedFoldersEnableSymlinksCreate` option to true.
 


briancain:
Hey there @dentarg - Thanks for the pull request! Perhaps our wording around this flag is confusing, but it really is `true` by default if not set by the user:

https://github.com/hashicorp/vagrant/blob/cd3909d7cede3a58a2b3439add3f8cb53a770919/plugins/providers/virtualbox/synced_folder.rb#L109-L113

That warning is essentially saying if you don't set that flag, its feature will be enabled by default. Enabling it or setting it to false will disable that feature, but without it, that feature is on.

I'll close this for now, but thanks for contributing!! :heart:

dentarg:
@briancain Yeah, your wording is very confusing.

If you look at, for example https://www.vagrantup.com/docs/other/environmental-variables.html#vagrant_alias_file, it says "By default, this is set to ~/.vagrant.d/aliases." and I'm pretty sure that means `VAGRANT_ALIAS_FILE=~/.vagrant.d/aliases`

But for `VAGRANT_DISABLE_VBOXSYMLINKCREATE` you write "Defaults to true if the option is not set. " but it doesn't mean that default is `VAGRANT_DISABLE_VBOXSYMLINKCREATE=true`. The default in that section is instead referring to `SharedFoldersEnableSymlinksCreate` that, yes, is indeed `true` by default.

Do you see the problem? Can I help you clarify this?

briancain:
@dentarg - maybe instead, when we talk about default behavior with this feature, we should instead be talking about the actual virtualbox option, rather than the public facing vagrant flag that adjusts the behavior. It seems like that's the confusion here? Because actually by default that flag is not set at all, it's an internal option that is set to true.

briancain:
I'm happy to fix that wording if that makes sense to you. I agree that it reads a little strange!

dentarg:
> when we talk about default behavior with this feature, we should instead be talking about the actual virtualbox option, rather than the public facing vagrant flag that adjusts the behavior. It seems like that's the confusion here?

Yes, exactly

dentarg:
I'm happy to review any change in wording around this if you want @briancain

PR #10808, opened 2019-04-24
Recommend adding lines to end of sudoers, ensuring %admin and root exist

mrgnw:
The placement of example entries may affect if NOPASSWD works properly.

For example, on macOS the entries would need to be added after this:
```
# root and users in group wheel can run anything on any machine as any user
root            ALL = (ALL) ALL
%admin          ALL = (ALL) ALL
```
Results for questions like "Adding NOPASSWD in /etc/sudoers doesn't work" can lead to complex or unrelated solutions.

Placing entries at the end of sudoers prevents this issue and requires less explanation.


mrgnw:
Fixing github account & email

PR #10812, opened 2019-04-25
Prevent nopasswd issues, add vagrant entries at end of sudoers

mrgnw:
Vagrant sudoer entries need to be after `root` and `%admin` are set up. Adding the entries at the end by default ensures the order is correct.

If the order isn't correct, users will still have to type their password every time and might not understand why.

Example from macOS 10.14 Mojave sudoers:
```
# root and users in group wheel can run anything on any machine as any user
root            ALL = (ALL) ALL
%admin          ALL = (ALL) ALL
# entries need to be added after this point.
```

### DIFF
--- a/website/source/docs/synced-folders/nfs.html.md
+++ b/website/source/docs/synced-folders/nfs.html.md
@@ -139,6 +139,8 @@ have to modify them _slightly_ on certain hosts because the way Vagrant
 modifies `/etc/exports` changes a bit from OS to OS. If the commands below
 are located in non-standard paths, modify them as appropriate.
 
+Also note that in the sudoer file format, entries are applied in order. If you've added the appropriate entries but still have to type in your password, make sure the entries aren't inserted too early. From the sudoers man page: "When multiple entries match for a user, they are applied in order. Where there are multiple matches, the last match is used (which is not necessarily the most specific match)."
+
 For \*nix users, make sure to edit your `/etc/sudoers` file with `visudo`. It protects you against syntax errors which could leave you without the ability to gain elevated privileges.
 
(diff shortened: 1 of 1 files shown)


website/source/docs/synced-folders/nfs.html.md
```diff
@@ -134,10 +134,9 @@ If you do not want to type your password on every `vagrant up`, Vagrant
 uses thoughtfully crafted commands to make fine-grained sudoers modifications
 possible to avoid entering your password.
 
-Below, we have a couple example sudoers entries. Note that you may
-have to modify them _slightly_ on certain hosts because the way Vagrant
-modifies `/etc/exports` changes a bit from OS to OS. If the commands below
-are located in non-standard paths, modify them as appropriate.
+Below are example entries to add at the end of your sudoers file.
+
+Note that you may have to modify them _slightly_ on certain hosts because the way Vagrant modifies `/etc/exports` changes a bit from OS to OS. If the commands below are located in non-standard paths, modify them as appropriate.
```
briancain: I think instead of splitting up this paragraph, we should simply add an additional one below it talking about how the suoders file format applies these rules in order of entry. Meaning that the order in which you add these rules to this file matters, so pay attention. We could even quote the sudoers man page here:

>   When multiple entries match for a user, they are applied in order.  Where there are multiple matches, the last
     match is used (which is not necessarily the most specific match).

And tell users to refer to the man page for how to use this file.

mrgnw: Thanks Brian! I put everything in a single paragraph where you recommended.

PR #10819, opened 2019-04-29
Clarify that the method of default folder syncing for the project directory is RSync

xuhdev:


briancain:
Hey there @xuhdev - This is not actually true. Typically by default Vagrant will try to use the providers built in shared folder mechanism (like virtualbox or vmwares protocol for sharing folders). Sometimes it will fall back to rsync, but it's not the default. Thanks for taking the time to make a pull request though!

xuhdev:
@briancain Are you sure about this :smiley: ? I always got rsync in the default regardless whether vboxfs is available or not. I also couldn't find code that prioritizes vboxfs over rsync.

briancain:
Yes I am sure:

https://github.com/hashicorp/vagrant/blob/c22a145c59790c098f95d50141d9afb48e1ef55f/lib/vagrant/action/builtin/mixin_synced_folders.rb#L37-L40

```ruby
[[10, :docker, VagrantPlugins::DockerProvider::SyncedFolder],
 [10, :virtualbox, VagrantPlugins::ProviderVirtualBox::SyncedFolder],
 [10, :vmware, HashiCorp::VagrantVMwareDesktop::SyncedFolder],
 [7, :smb, VagrantPlugins::SyncedFolderSMB::SyncedFolder],
 [5, :nfs, VagrantPlugins::SyncedFolderNFS::SyncedFolder],
 [5, :rsync, VagrantPlugins::SyncedFolderRSync::SyncedFolder]]
```

It actually looks like rsync is the last option Vagrant uses if no other synced folder implementations are usable.

briancain:
@xuhdev - It could be that the box you are using is explicitly configured to use rsync for synced folders with its packaged Vagrantfile. That would potentially explain why Vagrant was picking rsync over vboxfs. This is often the case for when box maintainers don't want to install virtualbox guest additions, but want the box to work without any configuration on the user side.

xuhdev:
Hmm, interesting. Thanks for the info!

PR #10820, opened 2019-04-29
Fixes #10798: Enhance how docker compose driver path expands

### LINKED ISSUE #10798: Unable to mount nfs volume in docker compose (vagrant)
Vagrant 2.2.0
### Host operating system
MacOs Mojave 10.14.2

### Vagrantfile
```
config.vm.define "iceland" do |iceland|
    iceland.vm.provider "docker" do |docker|
      docker.build_dir = "/iceland/server/vagrant"
      docker.create_args = ["--add-host=iceland:127.0.0.1"]
      docker.name = "iceland"
      docker.compose=true
      docker.compose_configuration={
          "volumes" => {
              'iceland_data_volume' => {
                  "driver" => "local",
                  "driver_opts" => {
                      "type" => "nfs",
                      "o" => "addr=host.docker.internal,nolock",
                      "device" => ":/iceland/data"
                  }
              }
          },
      }
      docker.volumes =['iceland_data_volume:/temp_directory/data']
      docker.ports =
          [
              "1000:1000",
          ]
    end
  end
  end
```
### Dockerfile in the build_dir
```FROM ubuntu:18.04```

### Expected behavior
Running vagrant up should have generated a docker-compose.yml that looks like this 
```---
version: '2'
volumes:
  iceland_data_volume:
    driver: local
    driver_opts:
      type: nfs
      o: addr=host.docker.internal,nolock
      device: ":/iceland/data"
services:
  iceland:
    build:
      context: "/iceland/server/vagrant/"
    environment: {}
    expose: []
    ports:
    - 1000:1000
    volumes:
    - "iceland_data_volume:/temp_directory/data"
    - "/iceland/server/vagrant:/vagrant"
    links: []
    command: []
```

### Actual behavior
What actually happened?
```
---
version: '2'
volumes:
  iceland_data_volume:
    driver: local
    driver_opts:
      type: nfs
      o: addr=host.docker.internal,nolock
      device: ":/iceland/data"
services:
  iceland:
    build:
      context: "/iceland/server/vagrant"
    environment: {}
    expose: []
    ports:
    - 1000:1000
    volumes:
    - "/iceland/server/vagrant/iceland_data_volume:/temp_directory/data"
    - "/iceland/server/vagrant:/vagrant"
    links: []
    command: []
```
### Steps to reproduce
Use the above vagrant and docker (vagrant up) file to generate the following docker-compose.yml inside of `.vagrant/docker-compose` directory

### The bug
volumes explicitly created in docker compose are appended with the current path 
`"/iceland/server/vagrant/iceland_data_volume:/temp_directory/data" ` where as in my case it should have been only `"iceland_data_volume:/temp_directory/data"` where `iceland_data_volume` is a nfs volume.
This behaviour is happening because of https://github.com/hashicorp/vagrant/blob/85acf0cac724ef4bfda593a66e0c17f7e4110da0/plugins/providers/docker/driver/compose.rb#L105 line of code I think.

briancain:
Prior to this commit, the docker compose driver would _always_ path
expand a host volume no matter what. This is not always the correct
option, for example if that host volume is actually a reference to a key
inside a `volumes` hash instead of a path on disk. This commit changes
that by looking to see if the requested host volume is actually a
defined key inside the compose config, and if not, it will path expand
it like before. Otherwise it will leave the key "as is".

### DIFF
--- a/plugins/providers/docker/driver/compose.rb
+++ b/plugins/providers/docker/driver/compose.rb
@@ -102,7 +102,19 @@ module VagrantPlugins
               # will be fixed someday and the gsub below can be removed.
               host.gsub!(/^[^A-Za-z]+/, "")
             end
-            host = @machine.env.cwd.join(host).to_s
+            # if host path is a volume key, don't expand it.
+            # if both exist (a path and a key) show warning and move on
+            # otherwise assume it's a realative path and expand the host path
+            compose_config = get_composition
+            if compose_config["volumes"] && compose_config["volumes"].keys.include?(host)
+              if File.directory?(@machine.env.cwd.join(host).to_s)
+                @machine.env.ui.warn(I18n.t("docker_provider.volume_path_not_expanded",
+                                           host: host))
+              end
+            else
+              @logger.debug("Path expanding #{host} to current Vagrant working dir instead of docker-compose config file directory")
+              host = @machine.env.cwd.join(host).to_s
+            end
             "#{host}:#{guest}"
           end
           cmd     = Array(params.fetch(:cmd))
--- a/templates/locales/providers_docker.yml
+++ b/templates/locales/providers_docker.yml
@@ -104,6 +104,12 @@ en:
       destroy the container and recreate it.
     waiting_for_running: |-
       Waiting for container to enter "running" state...
+    volume_path_not_expanded: |-
+      Host path `%{host}` exists as a `volumes` key and is a folder on disk. Vagrant
+      will not expand this key like it used to and instead leave it as is defined.
+      If this folder is intended to be used, make sure its full path is defined
+      in your `volumes` config. More information can be found on volumes in the
+      docker compose documentation.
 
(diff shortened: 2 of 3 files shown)


briancain:
Tested this PR with the configuration below:

```ruby
config.vm.define "docker-guest"  do |docker|
  docker.vm.provider "docker" do |d|
    d.compose = true
    d.compose_configuration = {
      "volumes" => {
        "vagrant_home_mount" => {
          "driver": "local",
          "driver_opts" => {
            "type": "none",
            "device": "/home/vagrant/datadir",
            "o": "bind"
          }
        }
      },
    }
    d.volumes = ["vagrant_home_mount:/temp_dir/datadir"]
    d.git_repo = "https://github.com/briancain/nginx-docker-test.git"
  end
end
```

Couldn't actually get NFS to work (which is what the original poster was using), but this setup produces the same results/issue.

PR #10829, opened 2019-05-06
Continue updating environment boxes if metadata not found

briancain:
Prior to this commit, if a user ran a `vagrant box update` on their
entire environment and one of the boxes did not have a metadata file,
the rest of the boxes in the update would be skipped. This commit fixes
that by providing an option to ingore those boxes, so that the rest of
the boxes could check for updates.

### DIFF
--- a/plugins/commands/box/command/update.rb
+++ b/plugins/commands/box/command/update.rb
@@ -128,7 +128,13 @@ module VagrantPlugins
             if download_options[:insecure].nil?
               download_options[:insecure] = machine.config.vm.box_download_insecure
             end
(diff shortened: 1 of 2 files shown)


plugins/commands/box/command/update.rb
```diff
@@ -37,6 +37,10 @@ def execute
               options[:force] = f
             end
 
+            o.on("-i", "--ignore-failures", "Ignores box and attempts to continue boxes in environment") do |i|
+              options[:ignore] = i
```
briancain: I'm flexible on this name. Right now it only ignores boxes with no metadata, (i.e. boxes added manually, likely from packer builds).

PR #10840, opened 2019-05-09
Ubuntu 12 is now EOL'd, so use Ubuntu 18 instead

karlbecker:
While reading the docs, I noticed Ubuntu 12 was referenced. Since [Ubuntu 12.04 was officially EOL'd in April 2019](https://www.ubuntu.com/about/release-cycle), I recommend switching to a newer version as the default image. 

I'm mostly thinking about this to help newcomers avoid unexpected dependency installation troubles right after they pull up a new VM, since I imagine some packages will stop being supported soon on 12.

Perhaps the version should be specified more precisely than "Ubuntu 18" ?

### DIFF
--- a/website/source/intro/getting-started/index.html.md
+++ b/website/source/intro/getting-started/index.html.md
@@ -36,13 +36,13 @@ provider for the getting started guide, please install that as well.
 ## Up and Running
 
 ```
-$ vagrant init hashicorp/precise64
+$ vagrant init hashicorp/bionic64
 $ vagrant up
 ```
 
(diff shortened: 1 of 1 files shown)


karlbecker:
Bummer. I would have loved to have a little contribution here, but I'm not currently interested in signing a multi-page legal agreement for literally a half sentence tweak. Good luck with your CLA's, I'm sure plenty of people have signed it, but I hope you might be able to let that requirement go in the future @hashicorp-cla !

PR #10841, opened 2019-05-09
Search for running_vm_list only in `machines` folder

wzur-r3:
otherwise it might catch a non-machine id, in my case, for hostmanager plugin


briancain:
Hi @wzur-r3 - What did you see inside `.vagrant` that was picked up during tab complete?

wzur-r3:
Hi @briancain  - as I wrote in the initial report, it was `hostmanager` plugin.

It creates a folder with the same name and it has a file name `id` inside, too.

wzur-r3:
for clarification I use Vagrant version 2.2.4 and vagrant-hostmanager version 1.8.9

PR #10862, opened 2019-05-17
Fixes #10861: Trigger abort fixups for multi-threaded context

### LINKED ISSUE #10861: Abort option for triggers in a multi-machine environment should only stop the processing of the machine matched by the only_on option
Related to #10823

After commit 35ee3e2342a3af12e23c8f7813687f94e71a95ab the behaviour is now as follow:
In a multi-machine environment the `abort` option for triggers exits the Vagrant process and all machines are affected.
If `only_on` is defined I believe that the `abort` option should only apply to the machine matched by the `only_on` option and therefore only the thread for that guest should be stopped. Processing for other guests should continue.

Before commit 35ee3e2342a3af12e23c8f7813687f94e71a95ab this was the observed behaviour.

briancain:
Prior to this commit, Vagrant would simply exit completely even if there
were other guests still being configured. This commit fixes that by
changing the abort trigger to exit the thread if there are other threads
running. Otherwise if it's the final thread configuring a guest, the
trigger will exit Vagrant.

### DIFF
--- a/lib/vagrant/plugin/v2/trigger.rb
+++ b/lib/vagrant/plugin/v2/trigger.rb
@@ -296,12 +296,22 @@ module Vagrant
           end
         end
 
-        # Exits Vagrant immediately
+        # Exits Vagrant immediately, unless in a parallel context. If other
+        # threads are detected, just *this* thread exists. If it is the final
+        # thread, it will exit with the configured exit code.
         #
         # @param [Integer] code Code to exit Vagrant on
         def trigger_abort(exit_code)
           @ui.warn(I18n.t("vagrant.trigger.abort"))
-          Process.exit!(exit_code)
+
+          # We assume 2 is the smallest since thread Main spawns at least 1 thread
+          # to configure a given guest.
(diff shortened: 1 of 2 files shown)


lib/vagrant/plugin/v2/trigger.rb
```diff
@@ -296,12 +296,22 @@ def run_remote(config, on_error, exit_codes)
...
+        # thread, it will exit with the configured exit code.
         #
         # @param [Integer] code Code to exit Vagrant on
         def trigger_abort(exit_code)
           @ui.warn(I18n.t("vagrant.trigger.abort"))
-          Process.exit!(exit_code)
+
+          # We assume 2 is the smallest since thread Main spawns at least 1 thread
+          # to configure a given guest.
+          if Thread.list.size == 2
```
chrisroberts: We can't use this condition to determine exit type to perform since there can be multiple threads in use even when not running with parallel enabled.

chrisroberts:
Closing this in favor of #10891

PR #10879, opened 2019-05-28
providers/docker: Add usability test

andersk:
This allows Vagrant to fallback from the Docker provider to an alternate provider if `docker` isn’t installed, like the documentation [claims](https://www.vagrantup.com/docs/providers/basic_usage.html#default-provider) it should.

### DIFF
--- a/plugins/providers/docker/provider.rb
+++ b/plugins/providers/docker/provider.rb
@@ -11,6 +11,14 @@ module VagrantPlugins
     class Provider < Vagrant.plugin("2", :provider)
       @@host_vm_mutex = Mutex.new
 
+      def self.usable?(raise_error=false)
+        Driver.new.execute("docker", "version")
+        true
+      rescue Vagrant::Errors::CommandUnavailable, Errors::ExecuteError
+        raise if raise_error
+        return false
+      end
+
       def initialize(machine)
         @logger  = Log4r::Logger.new("vagrant::provider::docker")
         @machine = machine
--- a/test/unit/support/shared/virtualbox_context.rb
+++ b/test/unit/support/shared/virtualbox_context.rb
@@ -28,6 +28,11 @@ shared_context "virtualbox" do
     allow(subprocess).to receive(:execute).
       with("VBoxManage", "showvminfo", kind_of(String), kind_of(Hash)).
       and_return(subprocess_result(exit_code: 0))
+
+    # apparently this is also used in docker tests
+    allow(subprocess).to receive(:execute).
+      with("docker", "version", an_instance_of(Hash)).
+      and_return(subprocess_result(exit_code: 0))
   end
 
   around do |example|


plugins/providers/docker/provider.rb
```diff
@@ -11,6 +11,14 @@ module DockerProvider
     class Provider < Vagrant.plugin("2", :provider)
       @@host_vm_mutex = Mutex.new
 
+      def self.usable?(raise_error=false)
+        Driver.new.execute("docker", "version")
+        true
+      rescue Vagrant::Errors::CommandUnavailable, Errors::ExecuteError
```
briancain: I _think_ we will also need to rescue on `Vagrant::Errors::CommandUnavailableWindows` here

andersk: Not necessary, because `CommandUnavailableWindows` [inherits](https://github.com/hashicorp/vagrant/blob/v2.2.4/lib/vagrant/errors.rb#L315) from `CommandUnavailable`. I have tested this on Linux and Windows.

test/unit/support/shared/virtualbox_context.rb
```diff
@@ -28,6 +28,11 @@ def subprocess_result(options={})
     allow(subprocess).to receive(:execute).
       with("VBoxManage", "showvminfo", kind_of(String), kind_of(Hash)).
       and_return(subprocess_result(exit_code: 0))
+
+    # apparently this is also used in docker tests
```
briancain: Interesting. I removed this and ran the tests myself and didn't receive any error at all requiring this. Do you have an example of the error you were receiving without this?

andersk: Yeah, see https://travis-ci.org/hashicorp/vagrant/builds/538053443.

andersk:
TBH, I have essentially no experience with Ruby and I don’t feel confident in my ability to bring up a test framework for this. (I already spent a couple hours failing to figure out how to do such a basic thing as running a single test instead of the entire test suite. It would be nice if that were documented, by the way.) If you’re willing to take care of it I think that’s the way to go. Thanks!

briancain:
@andersk - no worries, I understand! Starting a new class for tests is a lot of work. I will do that myself this week and bring in your commit so that you get credit. Thanks!

test/unit/support/shared/virtualbox_context.rb
andersk: It’s possible that the discrepancy is because my commit is based on v2.2.4 and Travis ran on a merge of my commit with master.

briancain: Hmm, strange. Well I'll see if it happens to me when I open a new PR with the rest of the tests. I've tried a couple of times locally and it was totally fine without it :thinking: Thanks for the info!

briancain:
Thanks for your PR @andersk ! I've included that in my pull request here! https://github.com/hashicorp/vagrant/pull/10890 I'll go ahead and close this then.

test/unit/support/shared/virtualbox_context.rb
andersk: The discrepancy turns out to be because the `usable?` method was no longer called after you changed it from a class method to an instance method.

Issue #10506, opened 2018-12-15
Feature Request: vagrant global-status format


jchook:
### Vagrant version
Vagrant 2.2.0

Hey so, `vagrant global-status` is fantastic and incredibly useful to other scripts.

Currently my scripts require `grep '^[0-9a-f]'` to access the VM lines.

Personally I think the column order, etc appears obvious and requires zero additional lines of output, such as the header and long explanation.

If we can't clean-up the default output, could we at least have some kind of argument like `--interop` or something that allows other programs or scripts to easily use/interpret the results of this command?

For example, say that `vagrant global-status` yields:

```
id       name    provider   state    directory                           
-------------------------------------------------------------------------
6f499e8  dev-vm  virtualbox running  /home/jchook/projects/by      
a5ff8a8  virtu   virtualbox poweroff /home/jchook/projects/virtu         
4bfd46f  default virtualbox poweroff /home/jchook/projects/rps           
456a9d0  default virtualbox poweroff /home/jchook/dev/wireguard          
 
The above shows information about all known Vagrant environments
on this machine. This data is cached and may not be completely
up-to-date (use "vagrant global-status --prune" to prune invalid
entries). To interact with any of the machines, you can go to that
directory and run Vagrant, or you can use the ID directly with
Vagrant commands from any directory. For example:
"vagrant destroy 1a2b3c4d"
```

Perhaps `vagrant global-status --interop` would return:

```
6f499e8  dev-vm  virtualbox running  /home/jchook/projects/by       
a5ff8a8  virtu   virtualbox poweroff /home/jchook/projects/virtu         
4bfd46f  default virtualbox poweroff /home/jchook/projects/rps           
456a9d0  default virtualbox poweroff /home/jchook/dev/wireguard          
 ```

For example, here is a script I use to quickly ssh into vagrant VMs without `cd`:

```
#!/usr/bin/env bash
VAGPATHS=$(vagrant global-status | grep '^[0-9a-f]' | awk '{ print $5 }')
for dir in $VAGPATHS; do
  base="$(basename $dir)"
  if [ "$1" == "$base" ]; then
    cd "$dir"
    shift
    vagrant "$@"
    exit 0
  fi
done
echo "Unable to find VM called $1 :("
exit 1
```

**Comment by paco3346** [comment]
You want the `--machine-readable` flag.

```
vagrant global-status
1546285649,,ui,info,id       
1546285649,,ui,info,name    
1546285649,,ui,info,provider   
[...]

PR #10884, opened 2019-05-31
#10506 - Machine Readable Global-Status

GregJPreece:
As noted in #10506, the `vagrant global-status --machine-readable` command produces a rather unhelpful format for machine processing. It does not specify any `@ui.machine` outputs, only `@ui.info`, meaning that the formatted table outputted by this command is fudged into a series of machine-readable lines that are difficult to parse and not particularly useful. This PR adds `@ui.machine` calls alongside them, to make parsing the output of this command much easier, and consistent with similar commands such as `vagrant status --machine-readable`.

Before:

```
1559265195,,ui,info,id       
1559265195,,ui,info,name    
1559265195,,ui,info,provider   
1559265195,,ui,info,state    
1559265195,,ui,info,directory                                                        
1559265195,,ui,info,
1559265195,,ui,info,------------------------------------------------------------------------------------------------------
1559265195,,ui,info,2e5afa9  
1559265195,,ui,info,default 
1559265195,,ui,info,virtualbox 
1559265195,,ui,info,poweroff 
1559265195,,ui,info,/path/to/vagrant/file 
1559265195,,ui,info,
1559265195,,ui,info, \nThe above shows information about all known Vagrant environments\non this machine. This data is cached and may not be completely\nup-to-date (use "vagrant global-status --prune" to prune invalid\nentries). To interact with any of the machines%!(VAGRANT_COMMA) you can go to that\ndirectory and run Vagrant%!(VAGRANT_COMMA) or you can use the ID directly with\nVagrant commands from any directory. For example:\n"vagrant destroy 1a2b3c4d"
```

After:
```
1559265166,,metadata,machine-count,1
1559265166,,machine-id,62cfe71
1559265166,,provider-name,virtualbox
1559265166,,machine-home,/path/to/vagrant/file
1559265166,,state,poweroff
```

Please see the comments for a request for info re: suppressing UI output

I am currently writing a build system wrapper for Vagrant, so I will be using the machine-readable option on most of the available Vagrant commands. If Hashicorp are receptive to patches along similar lines to this one, I would like to help clean up the machine-readable mode and make it more consistent. It's a very useful mode to have, and makes integrating Vagrant with automation/build tools much easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,15 @@ Once your Vagrant bundle is installed from Git repository, you can run the test 
 
 This will run the unit test suite, which should come back all green!
 
+If you are developing Vagrant on a machine that already has a Vagrant package installation present, both will attempt to use the same folder for their configuration (location of this folder depends on system). In this case, override the
+`VAGRANT_HOME` environment variable for your development version of Vagrant before running any commands. For example, in Bash:
+
+    export VAGRANT_HOME=/path/to/project/.vagrant
+
+This folder will be ignored by Git. You can now run Vagrant commands against the development version:
+
+    bundle exec vagrant
+
 Please take time to read the [HashiCorp Community Guidelines](https://www.hashicorp.com/community-guidelines) and the [Vagrant Contributing Guide](https://github.com/hashicorp/vagrant/blob/master/.github/CONTRIBUTING.md).
 
 Then you're good to go!
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ between Windows, Mac OS X, and Linux.
 
 ## Quick Start
 
+Package dependencies: Vagrant requires `bsdtar` to be available on your system PATH to run successfully.
+
 For the quick-start, we'll bring up a development machine on
 [VirtualBox](https://www.virtualbox.org/) because it is free and works
 on all major platforms. Vagrant can, however, work with almost any
@@ -49,8 +51,6 @@ If you want the bleeding edge version of Vagrant, we try to keep master pretty s
 and you're welcome to give it a shot. Please review the installation page [here](https://www.vagrantup.com/docs/installation/source.html).
 
 ## Contributing to Vagrant
-
-Package dependencies: Vagrant requires `bsdtar` to be available on your system PATH to run tests successfully.
 
 Once your Vagrant bundle is installed from Git repository, you can run the test suite with:
 
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ and you're welcome to give it a shot. Please review the installation page [here]
 
 ## Contributing to Vagrant
 
+Package dependencies: Vagrant requires `bsdtar` to be available on your system PATH to run tests successfully.
+
 Once your Vagrant bundle is installed from Git repository, you can run the test suite with:
 
     bundle exec rake
--- a/README.md
+++ b/README.md
@@ -58,12 +58,11 @@ Once your Vagrant bundle is installed from Git repository, you can run the test 
 
 This will run the unit test suite, which should come back all green!
 
-If you are developing Vagrant on a machine that already has a Vagrant package installation present, both will attempt to use the same folder for their configuration (location of this folder depends on system). In this case, override the
-`VAGRANT_HOME` environment variable for your development version of Vagrant before running any commands. For example, in Bash:
+If you are developing Vagrant on a machine that already has a Vagrant package installation present, both will attempt to use the same folder for their configuration (location of this folder depends on system). This can cause errors when Vagrant attempts to load plugins. In this case, override the `VAGRANT_HOME` environment variable for your development version of Vagrant before running any commands, to be some new folder within the project or elsewhere on your machine. For example, in Bash:
 
-    export VAGRANT_HOME=/path/to/project/.vagrant
+    export VAGRANT_HOME=~/.vagrant-dev
 
-This folder will be ignored by Git. You can now run Vagrant commands against the development version:
+You can now run Vagrant commands against the development version:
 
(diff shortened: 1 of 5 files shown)


lib/vagrant/ui.rb
```diff
@@ -103,12 +103,6 @@ def ask(*args)
         raise Errors::UIExpectsTTY
       end
 
-      [:detail, :warn, :error, :info, :output, :success].each do |method|
```
GregJPreece: Removing this is premature, I know. Several other commands still output `@ui.info` only, which would leave them outputting nothing in machine-readable mode. Hopefully once those commands have been updated we can remove this code block.

In the meantime, I was hoping to add a `--suppress-ui` flag, which would output only `@ui.machine` calls in that mode, to clean up and declutter the output. However, I am unsure of how to read global arguments in the `@ui` library. I can see how to do it in plugins, but not the library.

This is the very first Ruby I've ever written, so please bear with me if I'm a bit slow :-)

briancain:
Hey there @GregJPreece - Thanks for taking the time to make a pull request with Vagrant!

While generally I do agree that the `--machine-readable` flag needs some love and attention, I don't think we should be removing the ui output as a default behavior. I can only imagine how many users at this point have built up automation around being able to read in the UI level (like for errors) and parse what vagrant prints. Removing this context completely will break those users tools, so I'm :-1: here on removing it.

One enhancement that could be worth making is having a difference between "messages" and general output, but I'm not entirely sure how to do that without adding more columns and keeping it consistent with how it works now :thinking:

GregJPreece:
Indeed, as mentioned in [my comment](https://github.com/hashicorp/vagrant/pull/10884#discussion_r289229293) I'm not myself in favour of breaking current default behaviour, but I'm not familiar enough with Ruby or Vagrant to know how to get a flag check into the UI lib function. Do you have any suggestions?

README.md
```diff
@@ -50,12 +50,23 @@ and you're welcome to give it a shot. Please review the installation page [here]
...
 
+If you are developing Vagrant on a machine that already has a Vagrant package installation present, both will attempt to use the same folder for their configuration (location of this folder depends on system). In this case, override the
+`VAGRANT_HOME` environment variable for your development version of Vagrant before running any commands. For example, in Bash:
+
+    export VAGRANT_HOME=/path/to/project/.vagrant
+
+This folder will be ignored by Git. You can now run Vagrant commands against the development version:
+
+    bundle exec vagrant
+
```
chrisroberts: I'm okay with including this within the README as an optional step. This will generally only be useful when you have plugins installed (as the new version may require the plugins to be re-installed when being loaded). However, the `VAGRANT_HOME` directory should not be set to the project's local data directory (`.vagrant`). While it may work okay there is a good chance of unexpected behavior based on what data is being written to Vagrant's home directory and the project's local data directory. This would be better as a project local home (`.vagrant-home`) or something located in `/tmp`, etc.

lib/vagrant/ui.rb
chrisroberts: :-1: on removing these methods. This is mainly due to the existing behavior of the UI class and is not backwards compatible. I'd also argue that it is trivial to filter out `ui` type messages if they are considered extraneous for the use case.

plugins/commands/global-status/command.rb
```diff
@@ -65,6 +65,17 @@ def execute
...
 
+        # Machine-readable (non-formatted) output
+        @env.ui.machine("metadata", "machine-count", entries.length.to_s);
+        entries.each do |entry|
+          opts = { "target" => entry.send(:name).to_s }
+          @env.ui.machine("machine-id", entry.send(:id).to_s[0...7], opts)
+          @env.ui.machine("provider-name", entry.send(:provider).to_s, opts)
+          @env.ui.machine("machine-home", entry.send(:vagrantfile_path).to_s, opts)
+          @env.ui.machine("state", entry.send(:state).to_s, opts)
+        end
```
chrisroberts: The `#send` method isn't required here. It is useful for accessing instance methods which may not be publicly accessible (are marked as `protected` or `private`) or dynamically calling a method (like when the method name is defined within a variable). Instead the methods can just be called directly: `entry.name`.

test/unit/vagrant/ui_test.rb
```diff
@@ -231,16 +231,8 @@
...
-          parts = message.split(",")
-          expect(parts.length).to eq(5)
-          expect(parts[1]).to eq("")
-          expect(parts[2]).to eq("ui")
-          expect(parts[3]).to eq(method.to_s)
-          expect(parts[4]).to eq("foo")
-          true
-        }
+      it "does not output UI type to the machine-readable output" do
+        expect(subject).to receive(:safe_puts).never
```
chrisroberts: We would expect the UI output to still be included within the machine readable output. A new test should also be added which validates the new machine readable content is output as expected.

README.md
GregJPreece: Yes, the plugin scenario is how I ran into it. I figured it would be a common enough scenario for new devs that a spot in the README was warranted. I'll update as requested.

lib/vagrant/ui.rb
GregJPreece: Methods have been restored.

plugins/commands/global-status/command.rb
GregJPreece: Removed the unnecessary send() calls.

GregJPreece:
> Thanks for the updates. Everything looks great. I did toss in some test coverage on the new behavior, just FYI.
> 
> Cheers!

You beat me to it! I was going to figure those out this evening. Thanks for taking care of that for me.

PR #10888, opened 2019-06-04
Update Example Ubuntu init

JonathanAndrews:
Ubuntu 12.04 is no longer supported and it can be confusing for new users to install such an old OS.


briancain:
Closing in favor of https://github.com/hashicorp/vagrant/pull/11029. Thanks for making a PR!

PR #10890, opened 2019-06-04
Add usability check for docker plus tests

briancain:
A cleanup pull request built off of https://github.com/hashicorp/vagrant/pull/10879

Adds proper unit tests for the docker and virtualbox provider class, and a little bit of cleanup for the `usable?` method for the docker provider.

### DIFF
--- a/plugins/providers/docker/provider.rb
+++ b/plugins/providers/docker/provider.rb
@@ -11,6 +11,14 @@ module VagrantPlugins
     class Provider < Vagrant.plugin("2", :provider)
       @@host_vm_mutex = Mutex.new
 
+      def self.usable?(raise_error=false)
+        Driver.new.execute("docker", "version")
+        true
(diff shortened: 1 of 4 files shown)


briancain:
Going to wait and see if travis fails, and if not, I'll remove those commented out tests. They do not fail for me locally.

plugins/providers/docker/provider.rb
```diff
@@ -53,6 +45,14 @@ def driver
         @driver
       end
 
+      def usable?(raise_error=false)
```
andersk: By changing my class method to an instance method, you’ve broken the usability check, because it’s called on the class:

https://github.com/hashicorp/vagrant/blob/v2.2.5/lib/vagrant/environment.rb#L361

https://github.com/hashicorp/vagrant/blob/v2.2.5/lib/vagrant/vagrantfile.rb#L155

Issue #10801, opened 2019-04-20
Vagrant not working on Japanese Windows OS - Encoding problem


JordanJapan:
### Vagrant version
2.2.4

### Host operating system
Windows 8.1 (Japanese)

### Debug output 1
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'ubuntu/trusty64'...
C:/HashiCorp/Vagrant/embedded/gems/2.2.4/gems/childprocess-0.6.3/lib/childprocess/windows/process_builder.rb:43:in `join': incompatible character encodings: Windows-31J and UTF-8 (Encoding::CompatibilityError)

### Debug output 2
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'ubuntu/trusty64'...
C:/HashiCorp/Vagrant/embedded/gems/2.2.4/gems/vagrant-2.2.4/plugins/providers/virtualbox/driver/base.rb:408:in `block in execute': invalid byte sequence in UTF-8 (ArgumentError)

### Description
I'm new to using this and I was following a video online.  However, when I got the step of using vagrant up, I got an encoding error (debug 1).  After looking around, I tried changing command prompt's code page to 65001 from 932.  Unfortunately, a new error occurred (debug 2)

Is there a clean and simple solution to just make it work or even better automatically set the encoding as needed?  Thanks.

briancain:
Hey there @JordanJapan - could you please provide the full output of running `vagrant up --debug` in a [gist](https://gist.github.com/)?  Thanks!

JordanJapan:
Hello.  Thanks for the reply.

When I tried that command a lot of output appeared and I couldn't copy it all.  I was looking for a way to try outputting the entire log to a file, but with command prompt it refuses to do so even with the redirect command.

vagrant up --debug > c:\file.txt

It only copies the first 2 lines of the debug output.  Is there another way to output to a log file in Windows?  Thanks in advance.

briancain:
@JordanJapan - you should be able to do something like....

```
vagrant up --debug 2>&1 | Tee-Object -FilePath ".\vagrant.log"
```

Hopefully that works!

JordanJapan:
Alright.  The above command wasn't working in command prompt, but I installed Power Shell after reading about it and I got a file this time.  Same errors still occur, however.

I attached the .txt file with the complete output.

Thanks for taking the time to help.

[...]

PR #10898, opened 2019-06-07
Relax childprocess dependency restriction.

voxik:
Testing against childprocess 1.0.1, the test suite is passing just fine.

The newer childprocess might fix issues such as #10801

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |s|
   s.rubyforge_project         = "vagrant"
 
   s.add_dependency "bcrypt_pbkdf", "~> 1.0.0"
-  s.add_dependency "childprocess", "~> 0.6.0"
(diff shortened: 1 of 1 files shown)


voxik:
The test suite passed.

I am not going to sing CLA, so do whatever you need to do ...

briancain:
Hi @voxik - thanks for making a PR. HashiCorp requires all contributors sign the CLA, so if that's the case I'll close this PR. We would also likely want to be more explicit about what version of childprocess we are using rather than that wide range of versions if the newer 1.0.x version does solve the encoding issue on windows.

voxik:
I would prefer if you are not that explicit. If you were not that explicit, windows users would have the bug fixed a while ago.

Also the strict versions are troubling us on Fedora, because we have to relax the dependencies, while the resolver does not respect local gems and is using upstream dependency info for resolution instead, making the result incompatible.

Issue #9156, opened 2017-11-08
Enhancement: Copy-VMFile -> Synced Folders for Hyper-V


anlx-sw:
### Vagrant version
Vagrant 2.0.1

### Host operating system
Windows 2016 with Hyper-V

### Guest operating system
Works with Windows and Linux Guests if integration services are running on the guest.

### Hyper-V Copy-VMFile

As i had numerous problems using vagrant on hyper-v with the current possibilities to sync files from a hyper-v host to guests:

- SMB: Insecure (credentials safed in cleartext, smbv1), network connectivity between client and host is necessairy, ...
- NFS: Not working for windows host
- RSYNC: Works for linux guests if you install the tools on the host. Doesn't really work for windows guests.

As an enhancement i now use the copy-vmfile commandlet to sync files from the host to the client using a simple powershell script:

```
$paramHash = @{
 Name = 'vagrant-guest'
 SourcePath = ''
 DestinationPath = ''
 CreateFullPath = $True
 FileSource = 'Host'
 Force = $True
 Verbose = $True
}
dir C:\sourcedirectory\vagrant\ | foreach {$destination = Join-path -Path "C:\vagrant" -ChildPath $_.name
      $paramHash.SourcePath = $_.fullname
      $paramHash.DestinationPath = $destination
      Copy-VMFile @paramHash 
      }
```

I run this with the vagrant host shell plugin currently as the first provisioning step:
`vagrant plugin install vagrant-host-shell`

```
    config.vm.provision :host_shell do |host_shell|
      host_shell.inline = 'Powershell.exe -executionpolicy bypass -File ./copy-filestovm.ps1'
    end
```

I thought this would be a nice addition for everyone stuck with hyper-v as the host. As I lack skill and time to do a proper PR for this enhancement i opened this issue.

anlx-sw:
I adopted my powershell script above a little bit to be able to transfer files in subfolders:
```

$paramHash = @{
 Name = 'test.example.com'
 SourcePath = ''
 DestinationPath = ''
 CreateFullPath = $True
 FileSource = 'Host'
 Force = $True
 Verbose = $false
}

Push-Location
Set-Location .\provision
get-childitem -Recurse -File '.\' | foreach {
      $filename = $_.FullName | Resolve-Path -Relative 
      $destination = Join-path -Path "C:\vagrant" -ChildPath $filename 
      $paramHash.SourcePath = $_.fullname
      $paramHash.DestinationPath = $destination
[...]

PR #10920, opened 2019-06-19
Adds Hyper-V sync folder plugin

laozc:
This PR adds Hyper-V sync folder plugin which leverages Guest Integration Service to sync files from Hyper-V host to guests.

Guest Integration Service is available on variety of guest OSes including Windows and Linux.
For Linux guest currently only systemd based distros are supported.

Verified on both Windows 10 and Ubuntu 18.04 guest.

### DIFF
--- a/lib/vagrant/errors.rb
+++ b/lib/vagrant/errors.rb
@@ -388,6 +388,10 @@ module Vagrant
       error_key(:hyperv_virtualbox_error)
     end
 
+    class HypervNotSupported < VagrantError
+      error_key(:hyperv_not_supported)
+    end
+
     class ForwardPortAdapterNotFound < VagrantError
       error_key(:forward_port_adapter_not_found)
     end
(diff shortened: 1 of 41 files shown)


laozc:
Close #9156

jrmcdona:
@laozc When do you see this PR going through?  What is a workaround on the meantime?

laozc:
I'm not sure.
It's quite a large PR and I think it would need some time.

In the meantime you may leverage a few alternatives:

1. SMB or rsync shared folder plugin
* If you don't have SMB disabled in your local net, you may use SMB plugin.
* Leveraging rsync shared folder if possible. However, rsync shared folder plugin may have some issue on some symbol links. (Works for *nix guest only)

2. You may install the dev branch
* If you're under Windows, you may install the Ruby with Devkit (which can be found on Ruby site) and it should be installed without any issue.
* If you do not want to use Windows installer, you may use WSL. I've tested on WSL environment and it should function well under WSL.

Hopefully it will get merged soon to solve one pain point to deal with on Windows.

PR #10938, opened 2019-06-28
virtualbox: Fix usability test to reject bad installs without crashing

andersk:
If VirtualBox is installed but the kernel module is missing or the service is stopped, the usability test should fail without crashing so we can fall back to other providers.

### DIFF
--- a/plugins/providers/virtualbox/provider.rb
+++ b/plugins/providers/virtualbox/provider.rb
@@ -8,9 +8,10 @@ module VagrantPlugins
       def self.installed?
         Driver::Meta.new
         true
-      rescue Vagrant::Errors::VirtualBoxInvalidVersion
-        return false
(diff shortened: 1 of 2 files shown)


timabbott:
Can this be reviewed?  It breaks the use case of having a Vagrantfile designed to support both Docker and Virtualbox on macOS (with the the developer deciding by passing `--provider`); one ends up unable to access the Docker host unless Virtualbox is installed.

andersk:
Sure. Added tests for `VirtualBoxInvalidVersion` and `VirtualBoxKernelModuleNotLoaded` equivalent to the existing tests for `VirtualBoxNotDetected` and `VirtualBoxInvalidVersion`. Or do you also want tests for the `usable?(false)` case?

briancain:
@andersk - I think for now it's fine without the `usable?(false)` case. Thanks for adding the tests!

PR #10960, opened 2019-07-10
don't pass build_args as extra_args

### LINKED ISSUE #10959: building with compose does not pass build args
### Vagrant version
Vagrant 2.2.5

### Vagrantfile
```ruby
  config.vm.provider "docker" do |d|
    d.build_dir = Dir.pwd
    d.compose = true
    d.build_args = {
      "foo" => "bar"
    }
  end
end
```
### Dockerfile
```
FROM debian:latest
CMD [ '/bin/bash' ]
```

### Expected behavior
Create a compose file containing build args
example:
```yml
---
version: '2'
services:
  default:
    build:
      context: "/home/schoen/tmp/vagrant-bug"
      args:
        foo: bar
    environment: {}
    expose: []
    ports: []
    volumes:
    - "/home/schoen/tmp/vagrant-bug:/vagrant"
    links: []
    command: []
```

### Actual behavior
Compose file does not contain build args
```yml
---
version: '2'
services:
  default:
    build:
      context: "/home/schoen/tmp/vagrant-bug"
    environment: {}
    expose: []
    ports: []
    volumes:
    - "/home/schoen/tmp/vagrant-bug:/vagrant"
    links: []
    command: []
```

### Steps to reproduce
1. copy vagrantfile and dockerfile
2. vagrant up
3. cat .vagrant/docker-compose/docker-compose.yml

florianschoen:
fixes #10959

### DIFF
--- a/plugins/providers/docker/action/build.rb
+++ b/plugins/providers/docker/action/build.rb
@@ -68,7 +68,7 @@ module VagrantPlugins
 
             image = machine.provider.driver.build(
               build_dir || git_repo,
-              extra_args: args) do |type, data|
+              build_args: args) do |type, data|
               data = remove_ansi_escape_codes(data.chomp).chomp
               env[:ui].detail(data) if data != ""
             end
--- a/plugins/providers/docker/action/build.rb
+++ b/plugins/providers/docker/action/build.rb
@@ -68,7 +68,7 @@ module VagrantPlugins
 
             image = machine.provider.driver.build(
               build_dir || git_repo,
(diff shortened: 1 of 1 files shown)


briancain:
Hey there! Thanks for taking the time to make a pull request! I ended up making a new one here: https://github.com/hashicorp/vagrant/pull/11106. Instead of passing an additional key that duplicates args into the build method, we should just update docker composes build method to look for the key that the docker build action is passing in.

PR #10991, opened 2019-07-26
Bump rspec-its dependency

dcermak:
rspec-its version 1.3.0 has been released, this PR updates the gemspec to use the latest version.


dcermak:
FYI: we've had this patch in openSUSE Tumbleweed for a while now and it hasn't caused any issues for the testsuite. I see this as relatively risk free, as it is a test only dependency (and the tests pass ;-) ).

PR #10992, opened 2019-07-26
Unrestrict rspec-its dependency

pvalena:
as It works fine [in Fedora](https://copr-be.cloud.fedoraproject.org/results/pvalena/vagrant/fedora-rawhide-x86_64/00978596-vagrant/) with rspec-its 1.3.0 (test suite execution is in [build.log](https://copr-be.cloud.fedoraproject.org/results/pvalena/vagrant/fedora-rawhide-x86_64/00978596-vagrant/build.log.gz)).

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -46,7 +46,7 @@ Gem::Specification.new do |s|
   # from within rspec
   s.add_development_dependency "rake", "~> 12.0.0"
   s.add_development_dependency "rspec", "~> 3.5.0"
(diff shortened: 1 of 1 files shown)


briancain:
Thanks for making a pull request with Vagrant! We already have a pull request that covers this (https://github.com/hashicorp/vagrant/pull/10991), and I don't think we want to change how vagrant defines its dependency on this gem. I'm going to go ahead and close this.

PR #10993, opened 2019-07-26
Only return interfaces where addr is not nil

dcermak:
tunnel interfaces on Linux have `addr` set to `nil` which makes the function `list_interfaces` fail with a backtrace.


briancain:
Thanks for making a pull request for Vagrant! I've made a new pull request to include some tests for this change here: https://github.com/hashicorp/vagrant/pull/11116

dcermak:
Thank you!

Issue #10584, opened 2019-01-10
`vagrant up` fails if the hostname needs to be changed on Alpine


maxbrunet:
### Vagrant version
`2.2.3`

### Host operating system
Linux (openSUSE Tumbleweed)

### Guest operating system
Alpine Linux 3.8

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "generic/alpine38"
  config.vm.hostname = "myhostname"
end
```

### Debug output
https://gist.github.com/maxbrunet/7872a0c2f7ad87ce2a08b7e0438f83a0

### Expected behavior
Alpine VM is up and requested hostname is set.

### Actual behavior

`vagrant up` fails on `Setting hostname...`.

```shell
==> default: Setting hostname...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

# Save current hostname saved in /etc/hosts
CURRENT_HOSTNAME_FULL="$(hostname -f)"
CURRENT_HOSTNAME_SHORT="$(hostname -s)"

# New hostname to be saved in /etc/hosts
NEW_HOSTNAME_FULL='myhostname'
NEW_HOSTNAME_SHORT="${NEW_HOSTNAME_FULL%%.*}"

# Update sysconfig
sed -i 's/\(HOSTNAME=\).*/\1myhostname/' /etc/sysconfig/network

# Set the hostname - use hostnamectl if available
if command -v hostnamectl; then
  hostnamectl set-hostname --static 'myhostname'
  hostnamectl set-hostname --transient 'myhostname'
else
  hostname 'myhostname'
fi

# Update ourselves in /etc/hosts
if grep -w "$CURRENT_HOSTNAME_FULL" /etc/hosts; then
  sed -i -e "s/( )$CURRENT_HOSTNAME_FULL( )/$NEW_HOSTNAME_FULL/g" -e "s/( )$CURRENT_HOSTNAME_FULL$/$NEW_HOSTNAME_FULL/g" /etc/hosts
fi
if grep -w "$CURRENT_HOSTNAME_SHORT" /etc/hosts; then
  sed -i -e "s/( )$CURRENT_HOSTNAME_SHORT( )/$NEW_HOSTNAME_SHORT/g" -e "s/( )$CURRENT_HOSTNAME_SHORT$/$NEW_HOSTNAME_SHORT/g" /etc/hosts
fi

# Restart network
service network restart


Stdout from the command:

[...]

PR #11000, opened 2019-08-03
Fix the hostname config logic for Alpine.

ladar:
This should fix and close https://github.com/lavabit/robox/issues/70 and https://github.com/hashicorp/vagrant/issues/10584 ... 

Would benefit from a second tester before merging.

### DIFF
--- a/plugins/guests/alt/cap/change_host_name.rb
+++ b/plugins/guests/alt/cap/change_host_name.rb
@@ -17,7 +17,11 @@ module VagrantPlugins
               NEW_HOSTNAME_SHORT="${NEW_HOSTNAME_FULL%%.*}"
 
               # Update sysconfig
-              sed -i 's/\\(HOSTNAME=\\).*/\\1#{name}/' /etc/sysconfig/network
+              if [ -f /etc/sysconfig/network ]; then
+                sed -i 's/\\(HOSTNAME=\\).*/\\1#{name}/' /etc/sysconfig/network
+              elif [ -f /etc/hostname ]; then
+                sed -i 's/.*/#{name}/' /etc/hostname
+              fi
 
               # Set the hostname - use hostnamectl if available
               if command -v hostnamectl; then
@@ -36,7 +40,14 @@ module VagrantPlugins
               fi
 
               # Restart network
-              service network restart
+              if command -v hostnamectl; then
+                if [ -f /etc/init.d/network ]; then
+                  service network restart
+                elif [ -f /etc/init.d/networking ]; then
+                  service networking restart
+                fi
+              fi
+
             EOH
           end
         end


timschumi:
Vagrant will presumably include seperate Alpine support starting with the next release (see #10975), and changing the hostname is working there.

Since the `alt` guest type isn't really intended for anything outside of [ALT Linux](https://en.wikipedia.org/wiki/ALT_Linux) anyways, I guess this change shouldn't be included.

ladar:
Happy to hear Alpine support is being merged. The PR is purely defensive though, it adds checks for the `/etc/sysconfig/network` / `/etc/hostname` / `/etc/init.d/network` / `/etc/init.d/networking` files and only runs the appropriate logic if the file exists. The existing code makes assumptions, and those assumptions lead to the failed exit. So this patch may avoid errors with other operating systems.

ladar:
Updated the script so it prints an error message if it can't persist the hostname change, but will still exit with 0. Which seemed more user friendly to me.

I was also thinking, a) `:alt` might be needed for other distros, like a Linux from Scratch system, or Devuan, or any of the others on my TODO list, so better error detection, and letting the user know what is wrong, when the system doesn't conform to the expected patter, will still be better than the obtuse error which occurs now... and b) this patch could still help people who update their `vagrant` version, but have `robox` and/or Alpine images, with `:alt` still in their system cache.

As always, I won't be offended if the PR gets rejected.

briancain:
> I was also thinking, a) `:alt` might be needed for other distros, like a Linux from Scratch system, or Devuan, or any of the others on my TODO list, so better error detection, and letting the user know what is wrong, when the system doesn't conform to the expected patter, will still be better than the obtuse error which occurs now... and b) this patch could still help people who update their `vagrant` version, but have `robox` and/or Alpine images, with `:alt` still in their system cache.

In response to a) This is what the `linux` capability should be used for, as it is generally the base "linux" class of capabilities that should work in the general case. Since `:alt` is its own distro, I'm :-1: on any more generalization stuff just because it should actually be the linux cap that does this. That being said I think this patch is still fine to inlucde with `:alt` since technically it could include all of those network tools, and we also don't seem to have a base linux change_host_name capability.

Other than some unit tests I think this change is fine to merge :+1:

ladar:
I just noticed your request for unit tests. I need to spend some time figuring out how the test you pointed me at is supposed to work, but I'll give it a shot in the next couple of days.

briancain:
Thanks @ladar - if you need any help getting those tests going please don't hesitate to reach out! They can be tricky to work with sometimes, so I'm happy help or give guidance.

ladar:
@briancain any guidance would be appreciated. Is there documentation you can point at? And/or someplace else were similar test cases like the ones you want already exist?

Ruby isn't one of my primary languages, so without any guidance I'll have to figure out how to write the tests the hard way.

ladar:
@briancain I managed to back into the suggestions you made above. I don't quite understand how the dummy communicator simulates a real box environment (without actually creating one), but I was able to back into a collection of tests by studying the unit tests for the other providers. 

Let me know if this checks all the boxes.

PR #11004, opened 2019-08-07
File provisioner must search files in VAGRANT_CWD

pedroarthur:
Currently, if you try to use `VAGRANT_CWD` to `vagrant up` a machine in a different folder than the current shell `$PWD`, it will fail while provisioning files.

Minimum failing example:

```bash
mkdir -p machine/files/

cat > machine/files/test <<EoF
# placeholder
EoF

cat > machine/test.vagrant <<EoF
# -*- mode: ruby -*-
# vi: set ft=ruby :

  config.vm.box = "debian/buster64"

  config.vm.provision "file",
    source: "files/test",
    destination: "/tmp/files/test"
end
EoF

export VAGRANT_VAGRANTFILE=test.vagrant
export VAGRANT_CWD=machine

vagrant up
```

### DIFF
--- a/plugins/provisioners/file/provisioner.rb
+++ b/plugins/provisioners/file/provisioner.rb
@@ -3,7 +3,7 @@ module VagrantPlugins
     class Provisioner < Vagrant.plugin("2", :provisioner)
       def provision
         @machine.communicate.tap do |comm|
-          source = File.expand_path(config.source)
+          source = "#{@machine.env.cwd}/#{config.source}"
           destination = expand_guest_path(config.destination)
 
           # If the source is a directory determine if any path modifications


plugins/provisioners/file/provisioner.rb
```diff
@@ -3,7 +3,7 @@ module FileUpload
     class Provisioner < Vagrant.plugin("2", :provisioner)
       def provision
         @machine.communicate.tap do |comm|
-          source = File.expand_path(config.source)
+          source = "#{@machine.env.cwd}/#{config.source}"
```
briancain: This should actually be:

```ruby
source = File.join(@machine.env.cwd, config.source)
```

briancain:
Hi @pedroarthur - thanks for opening an pull request with Vagrant! We're going to go ahead and close this in favor of https://github.com/hashicorp/vagrant/pull/11099 since it has the proper changes to fix this bug.

pedroarthur:
> Hi @pedroarthur - thanks for opening an pull request with Vagrant! We're going to go ahead and close this in favor of #11099 since it has the proper changes to fix this bug.

Great!

PR #11011, opened 2019-08-08
Add support of Astra Linux

g00dvin:
Add support of Astra Linux (Debian based distro) in order to allow use of guest capabilities.

### DIFF
--- a/plugins/guests/astra/guest.rb
+++ b/plugins/guests/astra/guest.rb
@@ -1,0 +1,10 @@
+require_relative '../linux/guest'
+
+module VagrantPlugins
+  module GuestAstra
+    class Guest < VagrantPlugins::GuestLinux::Guest
+      # Name used for guest detection
+      GUEST_DETECTION_NAME = "astra".freeze
+    end
+  end
+end
--- a/plugins/guests/astra/plugin.rb
+++ b/plugins/guests/astra/plugin.rb
@@ -1,0 +1,15 @@
+require "vagrant"
+
+module VagrantPlugins
+  module GuestAstra
+    class Plugin < Vagrant.plugin("2")
+      name "Astra Linux guest"
+      description "Astra Linux guest support."
+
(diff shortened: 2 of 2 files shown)


plugins/guests/astra/guest.rb
```diff
@@ -0,0 +1,10 @@
+require_relative '../linux/guest'
+
+module VagrantPlugins
+  module GuestAstra
+    class Guest < VagrantPlugins::GuestLinux::Guest
+      # Name used for guest detection
+      GUEST_DETECTION_NAME = "astra".freeze
```
briancain: Is this all that's required with the Astra guest? Just a different name to check against inside the release file?

g00dvin: Yes, it is debian based distro with custom DE (Fly). 
/etc/os-releses contains:
`  PRETTY_NAME="Astra Linux (Orel 2.12.14)"  
  NAME="Astra Linux (Orel)"  
  ID=astra  
  ID_LIKE=debian  
`

briancain: Hmm ok...I think in this case our linux guest detection for for `/etc/os-release` file is a little off.

https://github.com/hashicorp/vagrant/blob/1dda6cc14c2d221a8e2e7b5091fed2fca9ba1ecf/plugins/guests/linux/guest.rb#L9-L11

It looks at `ID`, but in this case we could also try to match against `ID_LIKE` since it seems that Astra is the same as Debian. We could potentially add a fourth check there at the end as a failsafe if nothing else matched. That would save us adding an additional guest capability here.

briancain:
Hey there! Thanks again for taking the time to make a pull request on Vagrant. Due to the lack of response on this PR, I'm going to go ahead and close it. If you wish to continue working on this later please feel free to open a new one with the requested changes, or open an issue so that we can keep track of proposed changes for the future. Thanks!

Issue #10484, opened 2018-12-06
Unable to install Vagrant development environment using Vagrantfile from the repo


czukowski:
This is probably an issue in rmv install script that is used by the Vagrantfile with an added factor of an old Ubuntu version.

### Vagrantfile

https://github.com/hashicorp/vagrant/blob/master/Vagrantfile

### Output

Gist: https://gist.github.com/czukowski/8fd2dc50c231f7ca3239170ef7d2df42

Specifically this:

```
    default: Downloading https://github.com/rvm/rvm/archive/.tar.gz
    default: curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
    default:
    default: Could not download 'https://github.com/rvm/rvm/archive/.tar.gz'.
    default:   curl returned status '35'.
    default: Downloading https://bitbucket.org/mpapis/rvm/get/.tar.gz
    default: curl: (35) error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure
    default:
    default: Could not download 'https://bitbucket.org/mpapis/rvm/get/.tar.gz'.
    default:   curl returned status '35'.
```

It is coming from the provision command:

```sh
# Install RVM
su -l -c 'curl -sL https://get.rvm.io | bash -s stable' vagrant
```

### Expected behavior

`vagrant up` completed successfully with rvm installed correctly.

### Actual behavior

rvm has not been installed and everything else depending on it as well.

### Steps to reproduce

1. `git clone https://github.com/hashicorp/vagrant.git .`
2. `vagrant up`

### Additional info

When trying the same steps, but with a newer Ubuntu (just replacing the box with `ubuntu/xenial64`), the installation has completed successfully, but the "latest" Ruby version detected by the provision script was `ruby-2.6.0-preview2` which was in conflict with the Gemfile:

```
    default: Resolving dependencies...
    default: Bundler could not find compatible versions for gem "ruby ":
    default:   In Gemfile:
    default:     ruby
    default:
    default:     vagrant was resolved to 2.2.3.dev, which depends on
    default:       ruby  (~> 2.2, < 2.6)
```

After installing Ruby 2.5.1 using rvm manually, all the tests comleted ok (3478 examples, 0 failures, 3 pending).

daryl-sf:
Having the same issue. Did you manage to get it resolved?

czukowski:
[...]

PR #11012, opened 2019-08-09
[#10484] Fixed Vagrantfile for Vagrant development

aielo:
Quick fix to #10484. Logs can be found [here](https://gist.github.com/aielo/d59f144256d80a6477ab071f0acb5ca0).

Summary of changes:
- Updated box to `ubuntu/bionic64`
- Removed `BUNDLER_VER_REQ` and `bundler` install
Its dependency is no longer in `vagrant.gemspec`. Using installation-provided version.
- RVM install via `apt-get` ([ubuntu_rvm](https://github.com/rvm/ubuntu_rvm))
- Next-to-last version of ruby installed
`vagrant.gemspec` sets max version `"< 2.7"`. Therefore, 2.7.0-preview1 was being used.
- Removed `curl` from basic dependencies
It was there just to download the RVM installer script.
- Suppressed `grep` error on `.bash_profile`
File is not expected to exist and will be created anyways.
- Box memory set to 2GB
8 tests were failing due to `Errno::ENOMEM: Cannot allocate memory`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,7 +10,7 @@ Vagrant.configure("2") do |config|
 
   ["vmware_fusion", "vmware_workstation", "virtualbox"].each do |provider|
     config.vm.provider provider do |v, override|
-      v.memory = "1024"
+      v.memory = "2048"
     end
   end
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,11 +4,11 @@
 # Ruby, run unit tests, etc.
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "ubuntu/bionic64"
+  config.vm.box = "hashicorp/bionic64"
   config.vm.hostname = "vagrant"
   config.ssh.shell = "bash -c 'BASH_ENV=/etc/profile exec bash'"
 
-  ["vmware_fusion", "vmware_workstation", "virtualbox"].each do |provider|
+  ["vmware_desktop", "virtualbox", "hyperv"].each do |provider|
     config.vm.provider provider do |v, override|
       v.memory = "2048"
     end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -36,20 +36,21 @@ if [ -f "${MARKER_FILE}" ]; then
   exit 0
 fi
 
+# Add ubuntu_rvm repo
+apt-add-repository -y ppa:rael-gc/rvm
+
 # Update apt
 apt-get update --quiet
 
-# Install basic dependencies
-apt-get install -qy build-essential bsdtar curl
+# Add vagrant user to sudo group:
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -4,13 +4,13 @@
 # Ruby, run unit tests, etc.
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "hashicorp/precise64"
+  config.vm.box = "ubuntu/bionic64"
```
briancain: This should be `hashicorp/bionic64` now. Thanks!

Vagrantfile
```diff
@@ -29,32 +29,32 @@ $shell = <<-'CONTENTS'
...
-
-# Add the vagrant user to the RVM group
-#usermod -a -G rvm vagrant
-
-# Install latest Ruby that complies with Vagrant's version constraint
-RUBY_VER_LATEST=$(su -l -c 'rvm list known' vagrant | tr '[]-' ' ' | awk "/^ ruby  ${RUBY_VER_REQ:0:1}\./ { print \$2 }" | sort | tail -n1)
-su -l -c "rvm install ${RUBY_VER_LATEST}" vagrant
-su -l -c "rvm --default use ${RUBY_VER_LATEST}" vagrant
+# Install next-to-last Ruby that complies with Vagrant's version constraint
+RUBY_VER=$(su -l -c 'rvm list known' vagrant | tr '[]-' ' ' | awk "/^ ruby  ${RUBY_VER_REQ:0:1}\./ { print \$2 }" | sort -r | sed -n '2p')
```
briancain: What's the reasoning behind changing this to next to last, instead of what it was previously?

Vagrantfile
```diff
@@ -4,13 +4,13 @@
 # Ruby, run unit tests, etc.
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "hashicorp/precise64"
+  config.vm.box = "ubuntu/bionic64"
   config.vm.hostname = "vagrant"
   config.ssh.shell = "bash -c 'BASH_ENV=/etc/profile exec bash'"
 
   ["vmware_fusion", "vmware_workstation", "virtualbox"].each do |provider|
```
briancain: This can now just be:

```ruby
  ["vmware_desktop", "virtualbox", "hyperv"].each do |provider|
```

Vagrantfile
briancain: Oh sorry, I re-read the PR message. :+1:

Vagrantfile
aielo: Didn't realize there was a hashicorp box for bionic64. Switched!

Vagrantfile
aielo: Fixed, thanks for the feedback!

Issue #9823, opened 2018-05-10
Hyper-V: no support for EnhancedSessionTransportType 


kopfpilot:
This is more a feature request, than a real bug:
https://blogs.technet.microsoft.com/virtualization/2018/02/28/sneak-peek-taking-a-spin-with-enhanced-linux-vms/

Please add a new configuration parameter in the Vagrantfile that switches to either execute `Set-VM -VMName <your_vm_name>  -EnhancedSessionTransportType HvSocket` or not.

When Xrdp is compiled in ubuntu 16.04, the rendering is faster and allows screen resizing :)

kopfpilot:
It applies to Windows guests as well: https://blogs.technet.microsoft.com/virtualization/2017/01/27/introducing-vmconnect-dynamic-resize/

draggeta:
Same, would like for this feature as well!

zeenlym:
It would be helpful to have a command to open HyperV session (VMConnect) in addition.

Workaround:
````
vmconnect.exe $env:computername $(Get-VM -Id $(Get-Content .\.vagrant\machines\default\hyperv\id)).Name
````

**Comment by mgrottenthaler** [comment]
I am currently using a PowerShell script and an `after` `up` trigger to work around this:

```ruby
  # Trigger to set EnhancedSessionTransportType after the VM is up
  config.trigger.after :up do |trigger|
    trigger.name = "Set EnhancedSessionTransportType"
    trigger.info = "Configuring Enhanced Session"
    trigger.run = {inline: "powershell -ExecutionPolicy Bypass -File \"#{File.join(__dir__, 'set-enhanced-session.ps1')}\" -vmName '#{vm_name}'"}
  end
```

```powershell
param(
    [string]$vmName
)

$vm = Get-VM -Name $vmName
if ($vm -ne $null) {
    # Enable Enhanced Session Mode
    Set-VM -VMName $vmName -EnhancedSessionTransportType HvSocket
    Get-VM -VMName $vmName | Select-Object -Property EnhancedSessionTransportType
} else {
    Write-Output "VM not found: $vmName"
}
```

maybe this can help somebody. Any updates on an implementation in vagrant?

PR #11014, opened 2019-08-10
hyper-v: add support for EnhancedSessionTransportType. 

camjjack:
Addresses #9823

First contribution to this project so keen to find out if I've missed anything else that I need to do.


zeenlym:
Does anyone plan to review this pull request ?

plugins/providers/hyperv/action/configure.rb
```diff
@@ -97,6 +97,11 @@ def call(env)
               env[:machine].provider_config.vm_integration_services)
           end
 
+          if env[:machine].provider_config.enable_enhanced_session_mode
```
soapy1, line 100: This looks great, just one more thing. Users should be able to disable EnhancedSessionTransportType as well as enable it. 
So something like
```
$ cat Vagrantfile
# Vagrantfile
 ...
    h.vm.provider "hyperv" do |b|
      b.enable_enhanced_session_mode = true
    end
 ...

$ vagrant up box
```
will bring up the box with the EnhancedSessionTransportType set to `HvSocket`. Then editing the Vagrantfile setting enable_enhanced_session_mode to false 
```
$ cat Vagrantfile
# Vagrantfile
 ...
    h.vm.provider "hyperv" do |b|
      b.enable_enhanced_session_mode = false
    end
  ...

$ vagrant reload box
```
should reload the box with the EnhancedSessionTransportType set to `VMBus`. However, currently it leaves the EnhancedSessionTransportType set to `HvSocket`

camjjack: Thanks for the feedback. Added this in commit. 1ed4c882dc6adc19b330da146a432265cd152a34
Note as its a bool setting to false is the same as not present, which is I think the best behavior for a new feature.

soapy1: Awesome, this looks great! Thanks 🙏

camjjack:
Is anyone able to action this pull request? Its been approved for a while but its becoming stale because it hasn't been merged.

PR #11032, opened 2019-08-21
Update middleman-hashicorp container and Gemfile.lock

nfagerlund:
Time marches on, and so do security vulnerabilities in Nokogiri. So it's time
for a new container.

As with last time, here's a reminder for the next person who needs to update
this:

- You shouldn't just update the dependency in Gemfile.lock, because your build
  times will go to heck as you compile Nokogiri from source on every run. So you
  need an updated container with all the dependencies.
- To update the container, you need to push a new tag to the middleman-hashicorp
  repo. Teamcity does the rest, and will ship a new container to Docker Hub
  (unless its credentials are out of date, in which case go ask team-eng-serv.)
- Once that's pushed:
    - Update Makefile
    - Update the Gemfile
    - Delete Gemfile.lock
    - `make website` until it comes up, then ctrl-C
    - Commit the changes


nfagerlund:
This supersedes #10771.

nfagerlund:
@chrisroberts Thanks! I don't have write access here, so I'll need someone else to merge.

PR #11034, opened 2019-08-22
Fix documentation typos/punctuation.

ladar:
This was originally supposed to fix the typo at the bottom of the page, but it appears that fix was already made 17 days ago (and my fork was 19 days old). 

I'm still submitting the pull request because I added periods throughout the page, and because I think the "Choosing the Right Box" section should be overhauled. The first item only lists `bento` and `canonical` the latter of which isn't even a user.  A more comprehensive list would be helpful, and possibly broken out between distro maintained boxes like `ubuntu` / `centos`, and projects which maintained collections of boxes for a variety of operating systems, like `bento`. But before I submit, I have a conflict of interest, so I don't know if should I make that change.

I wanted to add the support email address to the blurb on malicious boxes as well, so that if people find evil lurking somewhere, they can report it. The latest release point needs to be rewritten so it makes sense. Namely, look for boxes which are updated periodically, and/or which have been released recently, as they will contain more up-to-date software.

And finally the last item just doesn't make sense. I think it's trying to say the page will say whether a box is `Hosted by Vagrant Cloud ` or `Externally hosted (cloud.centos.org) ` /  ` Externally hosted (cloud-images.ubuntu.com) `, etc, but wanted to check before submitting.


ladar:
@chrisroberts if someone gives the go-ahead, I'll try and give it a quick makeover?

briancain:
Hi! Thanks for making a pull request on Vagrant! I'm going to go ahead and close this pull request, since the commits got brought in and merged in https://github.com/hashicorp/vagrant/pull/11084. Thanks!

PR #11059, opened 2019-09-08
Explain benefits better

valeriemettler:
Added a memorable catchphrase that would help the new users immediately understand one of the main the benefits of using Vagrant.


briancain:
Hello @valeriemettler ! Thanks for making a pull request with Vagrant! Unfortunately we're going to have to close this one. We appreciate you taking the time to make a pull request.

PR #11061, opened 2019-09-08
FreeBSD: Update Ansible's package name

jlduran:
The Ansible package in FreeBSD is now `py36-ansible`.

If `pkg` itself needs updating, passing only one 'yes' is not enough to confirm the installation of the package.  Use the `-y` option, specifically made for this purpose.  Also add `-q` to force a quiet output, inline with other guests.

References:

- [`pkg-install(8)`](https://www.freebsd.org/cgi/man.cgi?query=pkg-install&sektion=8)
- freebsd/freebsd-ports@cbf5576945e1582d1a4a8080c3c440addc73cd4e

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/freebsd/ansible_install.rb
+++ b/plugins/provisioners/ansible/cap/guest/freebsd/ansible_install.rb
@@ -11,7 +11,7 @@ module VagrantPlugins
               if install_mode != :default
                 raise Ansible::Errors::AnsiblePipInstallIsNotSupported
               else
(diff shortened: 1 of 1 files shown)


briancain:
Hello @jlduran - thanks for taking the time to make a pull request on Vagrant! These changes look good to me. I had to fix up some other issues unrelated to this change for the `ansible_install` capability, so I've cherry picked your commit into my PR here: https://github.com/hashicorp/vagrant/pull/11263

Thanks again! :tada:

PR #11072, opened 2019-09-16
Fix three bugs in WinSSH

cliffe:
Fixes:

* a hang when executing powershell via SSH (was stuck on keep alive messages since the powershell instance was waiting on interactive input)
* when using cmd shell (default) renaming computer was attempting to execute powershell commands via a batch script

We can't use WinRM due to the nature of our proxied environment, but ssh works as intended.

### DIFF
--- a/plugins/communicators/winssh/communicator.rb
+++ b/plugins/communicators/winssh/communicator.rb
@@ -1,4 +1,5 @@
 require File.expand_path("../../ssh/communicator", __FILE__)
+require_relative "../winrm/command_filter"
 
 module VagrantPlugins
   module CommunicatorWinSSH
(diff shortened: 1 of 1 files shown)


cliffe:
With these fixes, WinSSH is now in a state where a Windows VM guest can have puppet, ssh, rsync, all working. Previously various commands were failing to execute via WinSSH.

cliffe:
I'd love for this to be merged so that we wouldn't have to have a modified version of vagrant in order for WinSSH to work

jbonhag:
Hi @cliffe,

Thanks for the PR! A bunch of WinSSH issues were fixed in the Vagrant 2.2.8 release. Would you mind checking it out to see if it resolves your issues?

Thanks!

PR #11100, opened 2019-10-03
Use hostnamectl instead of hostname to set the hostname under SUSE

chkpnt:
`hostname` isn't part of the boxes openSUSE currently provides. As openSUSE and SLE are using `systemd` quite a while, setting the hostname using `hostnamectl` should be fine.

Without this fix, the _transient hostname_ will stay on localhost until the VM is restarted.

I've tested this change against openSUSE Leap 15.0, Leap 15.1 and Tumbleweed using this Vagrantfile:

```ruby

  config.vm.define "Leap15-0" do |sut|
    sut.vm.box = "Leap-15.0.x86_64"
    sut.vm.box_url = "https://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/openSUSE-Leap-15.0/images/boxes/Leap-15.0.x86_64.json"
    sut.vm.hostname = "leap15-0.example.com"
  end

  config.vm.define "Leap15-1" do |sut|
    sut.vm.box = "Leap-15.1.x86_64"
    sut.vm.box_url = "https://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/openSUSE-Leap-15.1/images/boxes/Leap-15.1.x86_64.json"
    sut.vm.hostname = "leap15-1.example.com"
  end

  config.vm.define "Tumbleweed" do |sut|
    sut.vm.box = "opensuse/openSUSE-Tumbleweed-Vagrant.x86_64"
    sut.vm.hostname = "tumbleweed.example.com"
  end

end

```

### DIFF
--- a/plugins/guests/suse/cap/change_host_name.rb
+++ b/plugins/guests/suse/cap/change_host_name.rb
@@ -5,11 +5,10 @@ module VagrantPlugins
         def self.change_host_name(machine, name)
           comm = machine.communicate
 
-          if !comm.test("hostname -f | grep '^#{name}$'", sudo: false)
+          if !comm.test("getent hosts '#{name}'", sudo: false)
             basename = name.split(".", 2)[0]
             comm.sudo <<-EOH.gsub(/^ {14}/, '')
-              echo '#{basename}' > /etc/HOSTNAME
-              hostname '#{basename}'
+              hostnamectl set-hostname '#{basename}'
 
               # Prepend ourselves to /etc/hosts
               grep -w '#{name}' /etc/hosts || {
--- a/test/unit/plugins/guests/suse/cap/change_host_name_test.rb
+++ b/test/unit/plugins/guests/suse/cap/change_host_name_test.rb
@@ -25,15 +25,15 @@ describe "VagrantPlugins::GuestSUSE::Cap::ChangeHostName" do
     let(:basename) { "banana-rama" }
 
     it "sets the hostname" do
-      comm.stub_command("hostname -f | grep '^#{name}$'", exit_code: 1)
+      comm.stub_command("getent hosts '#{name}'", exit_code: 1)
 
       cap.change_host_name(machine, name)
-      expect(comm.received_commands[1]).to match(/echo '#{basename}' > \/etc\/HOSTNAME/)
-      expect(comm.received_commands[1]).to match(/hostname '#{basename}'/)
+      expect(comm.received_commands[1]).to match(/hostnamectl set-hostname '#{basename}'/)
     end
 
     it "does not change the hostname if already set" do
-      comm.stub_command("hostname -f | grep '^#{name}$'", exit_code: 0)
+      comm.stub_command("getent hosts '#{name}'", exit_code: 0)
+
       cap.change_host_name(machine, name)
       expect(comm.received_commands.size).to eq(1)
     end


chkpnt:
Nevertheless I'm unsure if this is the right way or if `hostname` should be part of the boxes. Maybe @dcermak, who seems to be the maintainer of [`Virtualization:Appliances:Images:openSUSE-Leap-15.1`](https://build.opensuse.org/package/show/Virtualization:Appliances:Images:openSUSE-Leap-15.1/kiwi-images-vagrant), should proclaim his opinion.

plugins/guests/suse/cap/change_host_name.rb
```diff
@@ -5,11 +5,10 @@ class ChangeHostName
         def self.change_host_name(machine, name)
           comm = machine.communicate
 
-          if !comm.test("hostname -f | grep '^#{name}$'", sudo: false)
+          if !comm.test("getent hosts '#{name}'", sudo: false)
```
dcermak: This does not query the hostname though, this only queries the settings from `/etc/hosts`. `hostnamectl status --static` will give you the current full hostname.

plugins/guests/suse/cap/change_host_name.rb
```diff
@@ -5,11 +5,10 @@ class ChangeHostName
...
             basename = name.split(".", 2)[0]
             comm.sudo <<-EOH.gsub(/^ {14}/, '')
-              echo '#{basename}' > /etc/HOSTNAME
-              hostname '#{basename}'
+              hostnamectl set-hostname '#{basename}'
```
dcermak: Why not `hostnamectl set-hostname '#{name}'`? In the example Vagrantfile that you provide this results in a hostname `leap15-0` instead of `leap15-0.example.com`. Note that if you run `hostnamectl set-hostname leap15-0.example.com` then the bash prompt will still display `vagrant@leap15-0:~>` if that is what you wanted to achieve.
```suggestion
              hostnamectl set-hostname '#{name}'
```

dcermak:
> Nevertheless I'm unsure if this is the right way or if `hostname` should be part of the boxes. Maybe @dcermak, who seems to be the maintainer of [`Virtualization:Appliances:Images:openSUSE-Leap-15.1`](https://build.opensuse.org/package/show/Virtualization:Appliances:Images:openSUSE-Leap-15.1/kiwi-images-vagrant), should proclaim his opinion.

I try to keep the images as bare-bones as possible and not include anything that is not in the default patterns. So if `hostname` got dropped, we should switch to using `hostnamectl`.

plugins/guests/suse/cap/change_host_name.rb
chkpnt: Thanks for your comments!

`hostname status --static` will print the hostname, not the FQDN which depends on the resolver. And this is what `hostname -f` is doing: It uses the hostname, gives it to the resolver and returns the result. 
`getent hosts '#{name}'` is the command I've found which seems to be the closest to `hostname -f`, as on the base box the resolver will use `/etc/hosts`.

Of course one could discuss if this test is really necessary or if comparing the hostname (basename) would be sufficient. I guess the original authors wanted vagrant to be able to change the domain part of the box' fqdn after the initial start.

plugins/guests/suse/cap/change_host_name.rb
chkpnt: But this would write `#{name}` into `/etc/hostname` and `/etc/HOSTNAME`. As far as I know only the host's name, which is the '#{basename}' belongs to these files.

dcermak: That is right, so let's keep it that way.

plugins/guests/suse/cap/change_host_name.rb
dcermak: > Of course one could discuss if this test is really necessary or if comparing the hostname (basename) would be sufficient. I guess the original authors wanted vagrant to be able to change the domain part of the box' fqdn after the initial start.

I think this merely tries to find out if vagrant already changed the hostname of the box, so it makes sense to query the boxes' fqdn instead of the hostname.

Thanks for clarifying this!

PR #11105, opened 2019-10-06
Fix #10961 NFS sharing in macOS 10.15 host

andyfowler:
### Description

On macOS 10.15 (Catalina), `/` is read-only and paths inside of `/Users` (and elsewhere) are mounted via a "firmlink" (which is a new invention in APFS). These must be resolved to their full path to be shareable via NFS.

```
/Users/johnsmith/mycode  =>  /System/Volumes/Data/Users/johnsmith/mycode
```

This code tests each shared folder on the host, to see if it's 1) inside of an APFS filesystem, and 2) in this firmlink situation. Firmlinks are only createable by the OS, so this hardcoded path to check should be fine — at least until Apple adds more.

And this all was supposed to be visible to applications anyway. There's a useful intro to APFS changes in 10.15 here (linked to the right time)
https://developer.apple.com/videos/play/wwdc2019/710/?time=481

### Fixes

This fixes #10961 — and there were some helpful commenters in there who helped me track down where to begin.

### Tests

There weren't tests for the BSD NFS class, so I added one, but it's pretty isolated. This should also be testable by trying to share any folder below your home directory via NFS on 10.15. You'll see the side effects in `/etc/exports` on the host, which will have a `/System/Volumes/Data` prefix.

[Our team at Nutshell](https://github.com/nutshellcrm/) has been using vagrant for 6 years — thanks for making our lives easier, Mitchell & co!

PR #11108, opened 2019-10-08
Fix NFS sharing in macOS 10.15 (based on #11105)

chrisroberts:
This is a continuation of the work started in #11105

The firmlinks defined on the system can be located within
a system file (/usr/share/firmlinks). This can be read to
generate a simple map that can then be used for updating
paths as required.

Added a new host capability (resolve_host_path) which is
used by the bsd host nfs_export capability. The bsd host
implements this capability as a no-op, and the darwin
host implements this capability to handle firmlinks.

### DIFF
--- a/plugins/hosts/bsd/cap/nfs.rb
+++ b/plugins/hosts/bsd/cap/nfs.rb
@@ -214,9 +214,9 @@ module VagrantPlugins
 
         def self.nfs_check_folders_for_apfs(folders)
           folders.each do |_, opts|
-            # check to see if this path is mounted in an APFS filesystem, and if it's in the
-            # firmlink which must be prefixed.
-            is_mounted_apfs_command = "df -t apfs #{opts[:hostpath]}"
+            # check to see if this path is mounted in an APFS filesystem, and if it's under the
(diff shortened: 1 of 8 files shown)


andyfowler:
Thanks, @chrisroberts! Appreciate you running with this. Isolating path resolution from the NFS class definitely feels like a cleaner approach! I saw that `/usr/share/firmlinks` file, and wasn't really sure how authoritative it was, but Apple refer to it, so I assume it's a supported means of decoding firmlinks. 

I tested this branch on the "GM" version of 10.15, and it works great! 🚢 

```
❯ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.15
BuildVersion:   19A582a
```

vanderlee:
Any idea on when this fix will become available to the public?

briancain:
Hi @vanderlee - We are currently working on pushing out the next release of Vagrant (2.2.6) as soon as possible. Sorry I don't have an exact date.

jakedowns:
Just tested Vagrant 2.2.6 after seeing https://github.com/hashicorp/vagrant/commit/ae9c3e28d671f9230c18639f452d573914fabee0

i'm getting this error https://cl.ly/f271e8a02b9e

Issue #9689, opened 2018-04-13
Poor Documentation in Ansible inventory_path option without reference to implicit ansible.limit


Ilhicas:
According to the documentation present at : https://www.vagrantup.com/docs/provisioning/ansible_intro.html

Under section:
Static Inventory

One can read:
"The second option is for situations where you would like to have more control over the inventory management. With the inventory_path option, you can reference a specific inventory resource (e.g. a static inventory file, a dynamic inventory script or even multiple inventories stored in the same directory). Vagrant will then use this inventory information instead of generating it."

However one should read:
"The second option is for situations where you would like to have more control over the inventory management. With the inventory_path option, you can reference a specific inventory resource (e.g. a static inventory file, a dynamic inventory script or even multiple inventories stored in the same directory). Vagrant will then use this inventory information instead of generating it.

**Note: This option doesn't override ansible implicit command defining --limit "TheMachineProvisioned", limiting the playbook defined to this machine name. To allow for the playbook to have access to all machines, please declare in your ansible provision block: ansible.limit = "all"** 
"

This is because of the implicit ansible.limit, that should be either documented under this section, or overriden if using the ansible.inventory_path, if the user has a bad inventory file, let him find out.

This reminds me of the opposite of one of Python's mantras - Explicit is better than implicit.

sk1u:
I bumped into this myself, I could work on this.

PR #11117, opened 2019-10-10
docs: Add clarification about ansible.limit in static inventory usecase

sk1u:
Hi all,

This PR attempts to address https://github.com/hashicorp/vagrant/issues/9689
Please let me know if it can be improved.

Fixes https://github.com/hashicorp/vagrant/issues/9689

### DIFF
--- a/website/source/docs/provisioning/ansible_intro.html.md
+++ b/website/source/docs/provisioning/ansible_intro.html.md
@@ -239,6 +239,7 @@ config.vm.network :private_network, ip: "192.168.111.222"
 **Notes:**
 
   - The machine names in `Vagrantfile` and `ansible.inventory_path` files should correspond, unless you use `ansible.limit` option to reference the correct machines.
+  - The `ansible.inventory_path` option by default is only scoped to apply to a single guest in the inventory file, and does not apply to all defined guests. To allow access to all available machines in the inventory, it is necessary to set `ansible.limit = "all"`.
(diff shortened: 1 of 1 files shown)


website/source/docs/provisioning/ansible_intro.html.md
```diff
@@ -239,6 +239,7 @@ config.vm.network :private_network, ip: "192.168.111.222"
 **Notes:**
 
   - The machine names in `Vagrantfile` and `ansible.inventory_path` files should correspond, unless you use `ansible.limit` option to reference the correct machines.
+  - The `ansible.inventory_path` option doesn't override the `ansible.implicit` command. To allow access to all available machines in the inventory, it is necessary to set `ansible.limit = "all"`.
```
briancain: > The `ansible.inventory_path` option doesn't override the `ansible.implicit` command.

I don't think this part makes sense. There's no `ansible.implicit` command. Instead, I believe this is more about the _implicit_ notion that it works on a single guest by default, rather than all guests. Instead, this should probably read something like:


 > - The `ansible.inventory_path` option by default is only scoped to apply to a single guest in the inventory file, and does not apply to all defined guests. To allow access to all available machines in the inventory, it is necessary to set `ansible.limit = "all"`.

sk1u: Ok, it makes sense.
I pushed a change.

gildegoma:
@sk1u + @briancain ❤️ ❤️ ❤️

Issue #11144, opened 2019-10-22
NFS issue in Catalina with secondary filesystems


Mezzle:
**The workaround in vagrant 2.2.6 for Catalina does not work when the shared folder is not on the boot volume.**

I use a separate, case-sensitive partition for all my projects. It shows up on Catalina at `/Volumes/Development` as well as at `/System/Volumes/Data/Volumes/Development`.

Vagrant dutifully creates an entry in the exports file like this:

```exports
/System/Volumes/Data/Volumes/Development/example -alldirs -mapall=501:80 172.17.0.19
```

However, this gives an error with respect to `nfs check exports`

```
exports:2: exported dir/fs mismatch: /System/Volumes/Data/Volumes/Development/example /Volumes/Development
```

_Modified from a post originally posted by @oddnoc in https://github.com/hashicorp/vagrant/issues/10961#issuecomment-544623375_

paoga87:
I have a similar issue and have tried many workarounds. 
```
NFS is reporting that your exports file is invalid. Vagrant does
this check before making any changes to the file. Please correct
the issues below and execute "vagrant reload":

exports:3: exported dir/fs mismatch: /Users /System/Volumes/Data
```
However, in my case Vagrant is not even creating the entry in my exports file.

nathan-fiscaletti:
Might be related to https://github.com/hashicorp/vagrant/issues/11134

Mezzle:
They're ... slightly related, but not quite.

oddnoc:
Ugh. In Catalina 10.15.1, the workaround of editing /etc/exports and restarting nfsd no longer works.

hoskogmo:
Experiencing the same problem using MacOS Catalina 10.15 with Vagrant 2.2.6 when mounting a APFS volume that is case sensitive. Vagrant does prefix the paths as expected with `/System/Volumes/Data` after upgrading from 2.2.5. I have yet to find a workaround.

jbeker:
So, 10.15.1., `nfsd` is crashing. If you look in the console app into the `system.log` you will see entries like this:

```
Oct 30 09:30:27 hobbiton-wired com.apple.xpc.launchd[1] (com.apple.nfsd[1303]): Service exited due to SIGSEGV | sent by exc handler[1303]
```

I have submitted a ticket with Apple (ID: FB7418234). But if others can duplicate and report the issue that would be helpful.

jbeker:
[...]

PR #11150, opened 2019-10-24
Update to check for mount points before checking firmlinks

Mezzle:
As Catalina lists a firm link for "/Volumes" - but *also* actually mounts underneath it, this check checks for volumes mounted first, to see if the prefix matches that, and if so, uses that instead of modifying for firm link.

Resolves: #11144


chrisroberts:
As noted in #11144 the root cause was within either nfsd or apfs/firmlinks stuff and has been fixed in the first catalina update.

Cheers!

PR #11151, opened 2019-10-24
Catch NetworkNoInterfaces error in docker prepare_networks_test

dcermak:
The test "generates a network name and configuration" calls at the end `process_public_network()`, which can return an empty list if the currently executing machine has no usable network interfaces (this is typically the case for workers that build rpm packages in OBS or Koji). This results in an exception of type `Errors::NetworkNoInterfaces` to be thrown and causing this test to fail.

This commit wraps this call in a rescue block to suppress this failure.

### DIFF
--- a/test/unit/plugins/providers/docker/action/prepare_networks_test.rb
+++ b/test/unit/plugins/providers/docker/action/prepare_networks_test.rb
@@ -324,8 +324,11 @@ describe VagrantPlugins::DockerProvider::Action::PrepareNetworks do
 
   describe "#process_public_network" do
     let(:options) { {:ip=>"172.30.130.2", :subnet=>"172.30.0.0/16", :driver=>"bridge", :id=>"30e017d5-488f-5a2f-a3ke-k8dce8246b60"} }
+    let(:addr) { double("addr", ip: true, ip_address: "192.168.1.139") }
+    let(:netmask) { double("netmask", ip_unpack: ["255.255.255.0"]) }
     let(:ipaddr) { double("ipaddr", prefix: 22, succ: "10.1.10.2", ipv4?: true,
-                          ipv6?: false, to_i: 4294967040) }
+                          ipv6?: false, to_i: 4294967040, name: "ens20u1u2",
+                          addr: addr, netmask: netmask) }
 
     it "raises an error if there are no network interfaces" do
       expect(subject).to receive(:list_interfaces).and_return([])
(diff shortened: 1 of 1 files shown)


dcermak:
@briancain done! It's not too pretty unfortunately, I'm a Ruby beginner.

test/unit/plugins/providers/docker/action/prepare_networks_test.rb
```diff
@@ -343,7 +344,21 @@
...
+      addr = double("ip")
+      netmask = double("netmask")
+      allow(netmask).to receive(:ip_unpack).and_return(["255.255.255.0"])
+      allow(addr).to receive(:ip).and_return(true)
+      allow(addr).to receive(:ip_address).and_return("192.168.1.139")
+      allow(ifaddr).to receive(:addr).and_return(addr)
+      allow(ifaddr).to receive(:name).and_return("ens20u1u2")
+      allow(ifaddr).to receive(:netmask).and_return(netmask)
+      allow(subject).to receive(:list_interfaces).
+        and_return([ifaddr])
```
briancain: So from the changes here, I _think_ we could actually simplify this and just simply have:

```ruby
allow(subject).to receive(:list_interfaces).
        and_return([ipaddr])
```

Notice the change from `ifaddr` to the double `ipaddr` defined above already.

I'm assuming maybe there are some missing fields though that we'd want to add to the double based on what you've `allow`'d above? You might also need to mock up `netmask` like you did....and if so I recommend defining it like we have above with the `let(:netmask)` way of defining doubles. Then instead of having allows everywhere you can do something like:

```ruby
let(:netmask) { double("netmask", ip_unpack: ["255.255.255.0"]) }
```

Similarly, you can probably add some missing fields to the double `ipaddr` above instead of the additional `allow`s.

Let me know if that works for you.

test/unit/plugins/providers/docker/action/prepare_networks_test.rb
```diff
@@ -1,5 +1,6 @@
 require_relative "../../../../base"
 require_relative "../../../../../../plugins/providers/docker/action/prepare_networks"
+require_relative "../../../../../../plugins/providers/docker/errors"
```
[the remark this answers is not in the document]
dcermak: whoopsie…

test/unit/plugins/providers/docker/action/prepare_networks_test.rb
dcermak: Something like this?

Issue #11027, opened 2019-08-16
Hyper-V Snapshot results in 'vmID is empty' error


pauby:
### Vagrant version
2.2.5

### Host operating system
Windows 10 1903 18362.295

### Debug output
https://gist.github.com/pauby/e2011ebcabb7b41eb270c7780c11a4ce

### Expected behavior
No error.

### Actual behavior
An error. 

```text
An error occurred while executing a PowerShell script. This error
is shown below. Please read the error message and see if this is
a configuration error with your system. If it is not, then please
report a bug.

Script: list_snapshots.ps1
Error:

C:\HashiCorp\Vagrant\embedded\gems\2.2.5\gems\vagrant-2.2.5\plugins\providers\hyperv\scripts\list_snapshots.ps1 : Cannot bind argument to parameter
'VmId' because it is an empty string.
At line:1 char:263
+ ... lugins\providers\hyperv\scripts\list_snapshots.ps1') -VmID '' -ErrorA ...
+                                                                ~~
    + CategoryInfo          : InvalidData: (:) [list_snapshots.ps1], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorEmptyStringNotAllowed,list_snapshots.ps1
```

Note the snapshot is still taken.

### Steps to reproduce
1. vagrant snapshot save base --force

PR #11152, opened 2019-10-24
Fixes #11139: Skip machines that haven't been created for snapshot save

### LINKED ISSUE #11139: `snapshot save <name>` tries to use snapshot name as machine name
# Vagrant version
2.2.6 (does not happen in 2.2.5)

### Host operating system
Ubuntu

### Debug output
https://gist.github.com/wolfgang42/976afb5e5db1827bf11a3031b6ff9c15

### Expected behavior
```
$ vagrant snapshot save foo
==> default: Snapshotting the machine as 'foo'...
==> default: Snapshot saved! You can restore the snapshot at any time by
==> default: using `vagrant snapshot restore`. You can delete it using
==> default: `vagrant snapshot delete`.
```

### Actual behavior
```
$ vagrant snapshot save foo
The machine with the name 'foo' was not found configured for
this Vagrant environment.
```

briancain:
This commit fixes the original #11027 fix, which assumed that the
hyper-v provider just wasn't properly getting a VM id when it listed
snapshots. In reality, it was just that if you invoke the
`with_target_vm` method with no arguments, it runs on the entire environment.
This meant that the capability snapshot_list attempted to be invoked on
machines that didn't exist yet, which is the original cause for why the
list_snapshot method did not recieve a vm ID. This commit fixes that by
simply skipping the machine if it does not yet exist.

### DIFF
--- a/plugins/commands/snapshot/command/save.rb
+++ b/plugins/commands/snapshot/command/save.rb
@@ -15,6 +15,9 @@ module VagrantPlugins
             o.separator "can be restored via `vagrant snapshot restore` at any point in the"
             o.separator "future to get back to this exact machine state."
             o.separator ""
+            o.separator "If no vm-name is given, Vagrant will take a snapshot of"
+            o.separator "the entire environment with the same snapshot name."
+            o.separator ""
             o.separator "Snapshots are useful for experimenting in a machine and being able"
             o.separator "to rollback quickly."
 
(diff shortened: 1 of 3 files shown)


briancain:
Alternatively, we could mirror the fix that https://github.com/hashicorp/vagrant/commit/75d4aa42a1e34633928ccf32c04c23340ee5572b used in VirtualBox for the other providers (i.e. returning empty array). This would also fix this issue, but would require fixing it in every known Vagrant provider. :thinking:

PR #11164, opened 2019-10-29
update getting started guide - provision

slavrd:
Some users have had trouble in the getting started guide as the provisioned web server returned a 404.

### DIFF
--- a/website/source/intro/getting-started/provisioning.html.md
+++ b/website/source/intro/getting-started/provisioning.html.md
@@ -32,6 +32,7 @@ and save it as `bootstrap.sh` in the same directory as your Vagrantfile:
 apt-get update
 apt-get install -y apache2
 if ! [ -L /var/www ]; then
+  [ -d /var/www/html ] && cp -r /var/www/html /vagrant/html
   rm -rf /var/www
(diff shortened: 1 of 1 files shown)


website/source/intro/getting-started/provisioning.html.md
```diff
@@ -32,6 +32,7 @@ and save it as `bootstrap.sh` in the same directory as your Vagrantfile:
 apt-get update
 apt-get install -y apache2
 if ! [ -L /var/www ]; then
+  [ -d /var/www/html ] && cp -r /var/www/html /vagrant/html
```
briancain: This line shouldn't be required. We're already symlinking our vagrant directory to `/var/www` below. Instead, I think you're getting a 404 likely because this guide doesn't ever mention that you should have an `html` folder with some html files for apache to serve. We should add an example of that after this script (or maybe after the Vagrantfile example below...).

slavrd: You do have a point. My idea was to make it easier for the user to copy/paste => it's working. Asking them to explicitly create the html themselves should definitely demonstrate better what is happening, though. I'll change it and we'll see. Thanks.

PR #11165, opened 2019-10-29
Introduce disk management base config layer to core Vagrant

briancain:
This pull request adds a basic `disk` config option to Vagrant core that will allow providers to consume a config and manage a guests disk options. Similar to the provisioners, it has a few base options that all disks likely will need, but allows for more provider specific options with a `config` which will be implemented by each provider. Ideally in a Vagrantfile, a guests disk config block should be provider agnostic, unless there are some provider specific options that need to be set when configuring a disk.

By default, a new disk block is a non-primary disk that will be attached to the guest. If a user wishes to modify the primary disk attached to the guest, they can add the option `primary: true` to a disk config block.

```ruby
  config.vm.define "hashicorp" do |b|
    b.vm.box = "hashicorp/bionic64"

    b.vm.disk :disk, size: 1000,
      virtualbox__diskoption: "1234", libvirt__otheroption: true

    b.vm.disk :disk, size: 10000, name: "bigger_disk",
      virtualbox__diskoption: "1234"

  end
end
```

TODOs:

- [x] Define an entry action that all providers can call to configure their disks
- [x] We should provide "shortcuts" for defining disk size. Such as `size: "10GB"`
- [x] Better validation of disk config types in `validate`
- [x] Docs for website for new config option (and document how to use config class for provider authors)

Future TODOs after this PR:
- [x] Implement the disk provider for VirtualBox for "baseline" functionality. Should also reveal what is missing for the basic disk config layer. (Could be done in a different PR...)

### DIFF
--- a/lib/vagrant/action/builtin/disk.rb
+++ b/lib/vagrant/action/builtin/disk.rb
@@ -12,6 +12,8 @@ module Vagrant
           defined_disks = get_disks(machine, env)
 
           # Call into providers machine implementation for disk management
+          #
+          # machine.provider.configure_disks(defined_disks)
 
           # Continue On
           @app.call(env)
--- a/lib/vagrant/action/builtin/disk.rb
+++ b/lib/vagrant/action/builtin/disk.rb
@@ -9,9 +9,24 @@ module Vagrant
 
         def call(env)
           machine = env[:machine]
+          defined_disks = get_disks(machine, env)
+
+          # Check that provider plugin is installed for disk
+          # If not, log warning or error to user that disks won't be managed
 
           # Continue On
           @app.call(env)
+        end
+
+        def get_disks(machine, env)
+          return @_disks if @_disks
+
+          @_disks = []
+          @_disks = machine.config.vm.disks.map do |disk|
+            # initialize the disk provider??
+          end
+
+          @_disks
         end
       end
     end
--- a/lib/vagrant/action/builtin/disk.rb
+++ b/lib/vagrant/action/builtin/disk.rb
@@ -1,0 +1,19 @@
+module Vagrant
+  module Action
+    module Builtin
+      class Disk
+        def initialize(app, env)
+          @app    = app
+          @logger = Log4r::Logger.new("vagrant::action::builtin::disk")
+        end
+
+        def call(env)
+          machine = env[:machine]
+
+          # Continue On
+          @app.call(env)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


lib/vagrant/action/builtin/disk.rb
```diff
@@ -0,0 +1,33 @@
...
+          @logger = Log4r::Logger.new("vagrant::action::builtin::disk")
+        end
+
+        def call(env)
+          machine = env[:machine]
+          defined_disks = get_disks(machine, env)
+
+          # Call into providers machine implementation for disk management
+          #
+          # machine.provider.configure_disks(defined_disks)
```
briancain: I'm imagining it'll probably be something like this? Or maybe this is just a "base" call and providers will have to override the builtin disk action? :thinking:

plugins/kernel_v2/config/disk.rb
```diff
@@ -0,0 +1,151 @@
...
+require "securerandom"
+
+module VagrantPlugins
+  module Kernel_V2
+    class VagrantConfigDisk < Vagrant.plugin("2", :config)
+      #-------------------------------------------------------------------
+      # Config class for a given Disk
+      #-------------------------------------------------------------------
+
+      DEFAULT_DISK_TYPES = [:disk, :dvd, :floppy].freeze
```
briancain: Maybe we don't need to be so rigid right now with these default disk types :man_shrugging:

briancain:
I can possibly guard these all on EXPERIMENTAL, and move out or change the website docs if need be.

plugins/kernel_v2/config/disk.rb
chrisroberts: At this point I think this is acceptable for the first pass.

plugins/kernel_v2/config/vm.rb
```diff
@@ -748,6 +804,21 @@ def validate(machine, ignore_provider=nil)
...
 
+        # Validate disks
+        # Check if there is more than one primrary disk defined and throw an error
+        primary_disks = @disks.select { |d| d.primary && d.type == :disk }
+        if primary_disks.size > 1
+          errors << I18n.t("vagrant.config.vm.multiple_primary_disks_error",
+                           name: machine.name)
+        end
+
+        # TODO: Check for duplicate disk names?
```
chrisroberts: Yeah, even though we have an ID we should still keep the disk names unique for clarity

plugins/kernel_v2/config/vm.rb
```diff
@@ -123,6 +126,28 @@ def merge(other)
             end
           end
 
+          # Merge defined disks
+          other_disks = other.instance_variable_get(:@disks)
+          new_disks   = []
+          @disks.each do |p|
+            other_p = other_disks.find { |o| p.id == o.id }
```
chrisroberts: it would be nice if we could do merging by `id` or by `name` since from the user point of view using `name` to identify disks would be much easier.

PR #11178, opened 2019-11-05
Fixes #10170: Check for modinfo in /sbin

### LINKED ISSUE #10170: "modinfo" not found in $PATH
### Vagrant version
Vagrant 2.1.4

### Host operating system
* lsb_release -a
```
No LSB modules are available.
Distributor ID:	Devuan
Description:	Devuan GNU/Linux 2.0 (ascii)
Release:	2.0
Codename:	ascii
``` 
* cat /etc/debian_version
```
9
```

### Guest operating system
Debian 9

### Vagrantfile
```ruby
  config.vm.box = "debian/stretch64"

  config.vm.network "private_network", ip: "192.168.63.10"

  config.vm.synced_folder "/<REDACTED>/national", "/home/vagrant/national", type: "nfs"

    vb.name = "Debian 9 multiusage"
  end

  config.vm.provision "shell", inline: <<-SHELL
    rm /var/lib/apt/lists/* -rf
    apt-get update
    apt-get upgrade -y
    apt-get install -y vim htop curl wget
  SHELL
end
```

### Debug output
https://www.vagrantup.com/docs/other/debugging.html. The debug output should
be very long. Do NOT paste the debug output in the issue, just paste the
link to the Gist.

### Expected behavior
* NFS should be configured
* VM should be up

### Actual behavior
* vagrant up ends with an error message
* VM is down

### Steps to reproduce
1. Use devuan
2. Install Oracle VirtualBox from Oracle website
3. Install vagrant from vagrantup website
4. Use a minimal Vagrantfile with a NFS shared directory

### References
- GH-6304 : add use of modinfo with nfs capapbility

### How I see this issue

* modinfo lies in /sbin 
* /sbin is in $PATH only for root
* Given the full path (/sbin/modinfo) of modinfo in plugins/hosts/linux/cap/nfs.rb: everythingd works fine
```
Vagrant is currently configured to create VirtualBox synced folders with
the `SharedFoldersEnableSymlinksCreate` option enabled. If the Vagrant
guest is not trusted, you may want to disable this option. For more
information on this option, please refer to the VirtualBox manual:

  https://www.virtualbox.org/manual/ch04.html#sharedfolders

This option can be disabled globally with an environment variable:

  VAGRANT_DISABLE_VBOXSYMLINKCREATE=1

or on a per folder basis within the Vagrantfile:

  config.vm.synced_folder '/host/path', '/guest/path', SharedFoldersEnableSymlinksCreate: false
    default: Adapter 1: nat
    default: Adapter 2: hostonly
    default: 22 (guest) => 2222 (host) (adapter 1)
==> default: Booting VM...
==> default: Exporting NFS shared folders...
==> default: Preparing to edit /etc/exports. Administrator privileges will be required...
==> default: Mounting NFS shared folders...
    default: /vagrant => /<REDACTED>/debian9
==> default: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> default: flag to force provisioning. Provisioners marked to run always will still run.
```
 * The warning about SharedFoldersEnableSymlinksCreate is new and didn't occur with vagrant 2.1.2 on chakralinux ( old and different OS crashed last week-end, but I got safe elsewhere)

Thanks for your awesome work everyone :D

jbonhag:
Check for `modinfo` in /sbin if it doesn't appear on the PATH -- this could be expanded to scan other directories as well.

If it's not found on the PATH or in /sbin, the command will default back to `modinfo` so the user sees the error message about adding it to their PATH.

### DIFF
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -262,16 +262,25 @@ module VagrantPlugins
         end
 
         def self.modinfo_path
-          path = Vagrant::Util::Which.which("modinfo")
-          return path if path
+          if !defined?(@_modinfo_path)
+            @_modinfo_path = Vagrant::Util::Which.which("modinfo")
 
-          folders = ["/sbin"]
-          folders.each do |folder|
-            path = "#{folder}/modinfo"
-            return path if File.file?(path)
+            if @_modinfo_path.to_s.empty?
+              folders = ["/sbin"]
+              folders.each do |folder|
+                path = "#{folder}/modinfo"
+                if File.file?(path)
+                  @_modinfo_path = path
+                  break
+                end
+              end
+            end
+
+            if @_modinfo_path.to_s.empty?
+              @_modinfo_path = "modinfo"
+            end
           end
-
-          "modinfo"
+          @_modinfo_path
         end
 
         # @private
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -270,6 +270,8 @@ module VagrantPlugins
             path = "#{folder}/modinfo"
             return path if File.file?(path)
           end
+
+          "modinfo"
         end
 
         # @private
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -93,7 +93,7 @@ module VagrantPlugins
               "systemctl --no-pager --no-legend --plain list-unit-files --all --type=service " \
                 "| grep #{nfs_service_name_systemd}").exit_code == 0
           else
-            Vagrant::Util::Subprocess.execute("modinfo", "nfsd").exit_code == 0 ||
+            Vagrant::Util::Subprocess.execute(modinfo_path, "nfsd").exit_code == 0 ||
               Vagrant::Util::Subprocess.execute("grep", "nfsd", "/proc/filesystems").exit_code == 0
           end
         end
@@ -259,6 +259,17 @@ module VagrantPlugins
 
         def self.nfs_running?(check_command)
           Vagrant::Util::Subprocess.execute(*Shellwords.split(check_command)).exit_code == 0
+        end
+
+        def self.modinfo_path
+          path = Vagrant::Util::Which.which("modinfo")
+          return path if path
+
+          folders = ["/sbin"]
+          folders.each do |folder|
+            path = "#{folder}/modinfo"
+            return path if File.file?(path)
+          end
         end
 
         # @private
--- a/plugins/hosts/linux/cap/nfs.rb
+++ b/plugins/hosts/linux/cap/nfs.rb
@@ -266,13 +266,9 @@ module VagrantPlugins
             @_modinfo_path = Vagrant::Util::Which.which("modinfo")
 
             if @_modinfo_path.to_s.empty?
-              folders = ["/sbin"]
-              folders.each do |folder|
-                path = "#{folder}/modinfo"
-                if File.file?(path)
-                  @_modinfo_path = path
-                  break
-                end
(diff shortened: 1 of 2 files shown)


plugins/hosts/linux/cap/nfs.rb
```diff
@@ -261,6 +261,19 @@ def self.nfs_running?(check_command)
...
+          return path if path
+
+          folders = ["/sbin"]
+          folders.each do |folder|
+            path = "#{folder}/modinfo"
+            return path if File.file?(path)
+          end
+
+          "modinfo"
+        end
```
briancain: We have a pattern for methods like this. We can cache the value locally to prevent us from looking up this value multiple times during a Vagrant run:

```ruby
def self.modinfo_path
  if !defined?(@_modinfo_path)
    @_modinfo_path = Vagrant::Util::Which.which("modinfo")

    if @_modinfo_path.to_s.empty?
      folders = ["/sbin"]
      folders.each do |folder|
        path = "#{folder}/modinfo"
        if File.file?(path)
          @_modinfo_path = path
          break
        end
      end
    end

    if @_modinfo_path.to_s.empty?
      @_modinfo_path = "modinfo"
    end
  end
  @_modinfo_path
end
```

I haven't tested that, but something like this! :sweat_smile: :pray:

jbonhag: Thanks! 🙏 You were right on.  I updated the method and added a test for when the value is cached.

plugins/hosts/linux/cap/nfs.rb
```diff
@@ -261,6 +261,28 @@ def self.nfs_running?(check_command)
           Vagrant::Util::Subprocess.execute(*Shellwords.split(check_command)).exit_code == 0
         end
 
+        def self.modinfo_path
+          if !defined?(@_modinfo_path)
+            @_modinfo_path = Vagrant::Util::Which.which("modinfo")
+
+            if @_modinfo_path.to_s.empty?
+              folders = ["/sbin"]
+              folders.each do |folder|
```
briancain: I'm re-reading this now...are we expecting more folders here to be searched besides `/sbin`? :thinking: if not, we could remove the block enumerator and just check the path against `/sbin`.

jbonhag: Good point!  I patterned this after plugins/hosts/gentoo/cap/nfs.rb, but in this case we're really only checking one place.

This could be made even simpler if we say "use `/sbin/modinfo` if it exists, otherwise use `modinfo`" -- what do you think?

briancain: Aha, I see :thinking: Yeah I think in this case since we know we explicitly want to just check `/sbin` we can remove the iterator and just check that here. If we need to add more paths in the future we can change it then. I'd say just replace that, but otherwise leave the method as is so that we can first check if `Which` gives us a result, if not check`/sbin` and as a last resort just try `modinfo`

jbonhag:
Tested using nested virtualization, with the hashicorp/precise64 Vagrant box as host and the following Vagrantfile as guest:

```
  config.vm.box = "debian/stretch64"

  config.vm.network "private_network", ip: "192.168.63.10"

  config.vm.synced_folder "./national", "/home/vagrant/national", type: "nfs"

    vb.name = "Debian 9 multiusage"
  end

  config.vm.provision "shell", inline: <<-SHELL
    export DEBIAN_FRONTEND=noninteractive
    rm /var/lib/apt/lists/* -rf
    apt-get update
    apt-get upgrade -y
    apt-get install -y vim htop curl wget
  SHELL
end
```

I verified that the nfs folder synced from guest to host and vice versa.

jbonhag:
btw: had a brief scare where `vagrant up` worked _without_ the fix, until I realized that the precise box has `/sbin` on the PATH for the vagrant user.  Removing it from the PATH broke `vagrant up`, which worked again after the patch was applied.

PR #11181, opened 2019-11-07
Bump rubyzip to v1.3.0 to fix CVE-2019-16892

stefansundin:
https://github.com/advisories/GHSA-5m2v-hc64-56h6

~~Perhaps we should bump it to v2.0.0 instead?~~ https://rubygems.org/gems/rubyzip

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency "net-scp", "~> 1.2.0"
   s.add_dependency "rb-kqueue", "~> 0.2.0"
   s.add_dependency "rest-client", ">= 1.6.0", "< 3.0"
-  s.add_dependency "rubyzip", "~> 1.2.2"
+  s.add_dependency "rubyzip", "~> 1.3"
   s.add_dependency "wdm", "~> 0.1.0"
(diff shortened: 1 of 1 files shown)


chrisroberts:
If you want to bump this constraint up to version 2.0.0 that would be great and we can merge it in. It doesn't look like there were API changes that we care about. The CVE is related to unpacking archives which we don't do with this library, but since having it vendored locks the constraint for plugins it would be good to include a fixed version.

Thanks!

stefansundin:
Thanks for looking @chrisroberts. Unfortunately there seems to be a problem doing that.. :(

```
Bundler could not find compatible versions for gem "rubyzip":
  In Gemfile:
    vagrant was resolved to 2.2.7.dev, which depends on
      rubyzip (~> 2.0.0)

    vagrant was resolved to 2.2.7.dev, which depends on
      winrm-fs (~> 1.0) was resolved to 1.3.3, which depends on
        rubyzip (~> 1.1)
```

chrisroberts:
@stefansundin Ah, okay. setting at `~> 1.3` should be sufficient then and we can circle back to the winrm-fs constraints later. :slightly_smiling_face:

stefansundin:
I updated it to `~> 1.3`. Thanks.

PR #11194, opened 2019-11-14
Fixes #10753: Add reboot support for Linux guests

### LINKED ISSUE #10753: add support for shell :reboot to linux guests
it would really nice to have support for shell :reboot on linux based guests like we can do with windows guests:

```ruby
config.vm.provision :shell, inline: "echo 'Rebooting...'", reboot: true
```

without this, we still have to use the vagrant-reload plugin as:

```ruby
config.vm.provision :reload
```

jbonhag:
This commit adds support for setting `reboot: true` while using the
shell provisioner with Linux guests.

### DIFF
--- a/plugins/guests/linux/cap/reboot.rb
+++ b/plugins/guests/linux/cap/reboot.rb
@@ -1,0 +1,42 @@
+require "log4r"
+
+module VagrantPlugins
+  module GuestLinux
+    module Cap
+      class Reboot
+        MAX_REBOOT_RETRY_DURATION = 120
+
+        def self.reboot(machine)
+          @logger = Log4r::Logger.new("vagrant::linux::reboot")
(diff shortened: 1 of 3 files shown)


plugins/guests/linux/cap/reboot.rb
```diff
@@ -0,0 +1,42 @@
...
+        MAX_REBOOT_RETRY_DURATION = 120
+
+        def self.reboot(machine)
+          @logger = Log4r::Logger.new("vagrant::linux::reboot")
+          reboot_script = "reboot"
+
+          comm = machine.communicate
+
+          @logger.debug("Issuing reboot command for guest")
+          comm.sudo("reboot")
```
briancain: We should use `reboot_script` here since it's defined above.

PR #11197, opened 2019-11-15
Build quietly and capture the image hash from its output

kontza:
Consider the following combo:
- macOS Catalina
- Docker desktop 2.1.0.4
- Docker engine 19.03.4, experimental mode enabled
- Vagrant 2.2.6

With that combo the original line `matches = result.scan(/Successfully built (.+)$/i)` fails to generate a match, thus `up` fails because it doesn't find an image ID. With this change one can at least `vagrant up --provider=docker` successfully.

Downside may be the loss of output. Perhaps in the case where there is no `^sha256...` an exception could be raised and the user could be instructed to run `docker build` manually to see what exactly fails.

### DIFF
--- a/plugins/providers/docker/driver.rb
+++ b/plugins/providers/docker/driver.rb
@@ -18,12 +18,16 @@ module VagrantPlugins
       def build(dir, **opts, &block)
         args   = Array(opts[:extra_args])
         args   << dir
-        result = execute('docker', 'build', '-q', *args, &block)
-        matches = result.scan(/^sha256:([0-9a-f]+)$/i)
+        result = execute('docker', 'build', '--progress', 'plain', *args, &block)
+        matches = result.scan(/Successfully built (.+)$/i)
         if matches.empty?
-          # This will cause a stack trace in Vagrant, but it is a bug
-          # if this happens anyways.
-          raise "UNKNOWN OUTPUT: #{result}"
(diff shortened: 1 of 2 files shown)


kontza:
You're absolutely right. At the time, when I made the pull request, I had banged my head against the virtual build wall for hours, wandering why it doesn't work. Also, I had run separately `docker build` and thus I had the image ready and the quiet-option seemed fine. After a good night's sleep I retried the build on a fresh clone, and now instead I kept wandering "What's happening? Why I'm not seeing what's going on?" :)

Here's a sample log:
[vagrant-up.log](https://github.com/hashicorp/vagrant/files/3854267/vagrant-up.log)

I studied it a bit further and the culprit is the buildkit-based build which prints its output to stderr, not stdout. And, furthermore, the output does not contain "Successfully built [ID]", but "writing image sha256:[ID]".

kontza:
Here's a new take. Changes in `local.rb` check for an empty stdout, and revert to stderr if its empty. Changes in `driver.rb` check first for the 'Successfully built', if match is not found, check for 'writing image sha256:...', and fail only if that's not matched, either.

briancain:
Thank you @kontza  for your pull request! I have pulled in your commits and changes here https://github.com/hashicorp/vagrant/pull/11205 and added some enhancements. Thanks again for bringing up the original issue with docker and buildkit! :tada:

PR #11198, opened 2019-11-15
Fixes #11196: Add more examples for ruby heredoc shell scripts

### LINKED ISSUE #11196: Misleading example in shell provisioner documentation
One suggestion: Please use
```ruby
$script = <<-'SCRIPT'
echo I am provisioning...
date > /etc/vagrant_provisioned_at
SCRIPT
```
instead of
```ruby
$script = <<-SCRIPT
echo I am provisioning...
date > /etc/vagrant_provisioned_at
SCRIPT
```
here: https://www.vagrantup.com/docs/provisioning/shell.html#inline-scripts

When the single quotes are not used, ruby starts to deescape the string, which is really confusing if you try to use vagrant without prior knowledge in ruby:
```bash
echo "\"foo; rm -rf /\""
```
Would output this on a usual Linux:
```
"foo; rm -rf /"
```
But when it runs as a vagrant shell provisioner, it instead runs `echo ""foo; rm -rf /""`, outputs `foo` and deletes the disk:
```ruby
$script = <<-SCRIPT
echo "\"foo; rm -rf /\""
SCRIPT
```

briancain:
This commit adds an additional example for how to preserve quotes in a
script that uses heredocs.

### DIFF
--- a/website/source/docs/provisioning/shell.html.md
+++ b/website/source/docs/provisioning/shell.html.md
@@ -129,6 +129,33 @@ Vagrant.configure("2") do |config|
 end
 ```
 
+In the code block above, the script block starts with `<<-SCRIPT` and ends with `SCRIPT`.
+This is known as a _Here Document_ or a _heredoc_.  Additionally, if your script
+relies on quotes and you do not wish for Ruby to escape your quotes, you may
+want to use this style of _heredoc_ where `SCRIPT` is surrounded in single quotes:
+
+```ruby
+$script = <<-'SCRIPT'
+echo "These are my \"quotes\"! I am provisioning my guest."
+date > /etc/vagrant_provisioned_at
+SCRIPT
+
+Vagrant.configure("2") do |config|
+  config.vm.provision "shell", inline: $script
+end
+```
+
+Now that our _heredoc_ is quoted, our script will preserve the quotes in the string to `echo`:
+
+```
+==> default: Running provisioner: shell...
+    default: Running: inline script
+    default: These are my "quotes"! I am provisioning my guest.
+```
+
+For more examples of how to use _heredoc_ in Ruby, please refer to the
+[Ruby documentation](https://ruby-doc.org/core-2.5.0/doc/syntax/literals_rdoc.html#label-Here+Documents).
+
 It is understandable that if you are not familiar with Ruby, the above may seem very
 advanced or foreign. But do not fear, what it is doing is quite simple:
 the script is assigned to a global variable `$script`. This global variable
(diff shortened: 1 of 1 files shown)


website/source/docs/provisioning/shell.html.md
```diff
@@ -129,6 +129,33 @@ Vagrant.configure("2") do |config|
 end
 ```
 
+In the code block above, the script block starts with `<<-SCRIPT` and ends with `SCRIPT`.
+This is known as a _Here Document_ or a _heredoc_.  Additionally, if your script
+relies on quotes and you do not wish for Ruby to escape your quotes, you may
+want to use this style of _heredoc_ where `SCRIPT` is surrounded in single quotes:
```
jbonhag: I would change _Here Document_ to "here document" and _heredoc_ to "heredoc", and only put the terms in quotes the first time we use them.

jbonhag:
Let's use quotes instead of backticks so the reader doesn't mistake these for commands.  A couple of examples from other docs:

Perl:

> A line-oriented form of quoting is based on the shell "here-document" syntax.

Ruby:

> If you are writing a large block of text you may use a “here document” or “heredoc”

Bash:

> The format of here-documents is:

website/source/docs/provisioning/shell.html.md
```diff
@@ -130,9 +130,9 @@ end
 ```
 
 In the code block above, the script block starts with `<<-SCRIPT` and ends with `SCRIPT`.
-This is known as a _Here Document_ or a _heredoc_.  Additionally, if your script
+This is known as a `Here Document` or a `heredoc`.  Additionally, if your script
```
jbonhag: This is known as a "here document" or a "heredoc".  Additionally, if your script

website/source/docs/provisioning/shell.html.md
```diff
@@ -129,6 +129,33 @@ Vagrant.configure("2") do |config|
...
+echo "These are my \"quotes\"! I am provisioning my guest."
+date > /etc/vagrant_provisioned_at
+SCRIPT
+
+Vagrant.configure("2") do |config|
+  config.vm.provision "shell", inline: $script
+end
+```
+
+Now that our _heredoc_ is quoted, our script will preserve the quotes in the string to `echo`:
```
jbonhag: Now that our heredoc is quoted, our script will preserve the quotes in the string to `echo`:

website/source/docs/provisioning/shell.html.md
```diff
@@ -129,6 +129,33 @@ Vagrant.configure("2") do |config|
...
+
+```
+==> default: Running provisioner: shell...
+    default: Running: inline script
+    default: These are my "quotes"! I am provisioning my guest.
+```
+
+For more examples of how to use _heredoc_ in Ruby, please refer to the
```
jbonhag: For more examples of how to use heredocs in Ruby, please refer to the

briancain: Oops, missed these

PR #11208, opened 2019-11-20
 Open listen dependency up to `< 4`

pvalena:
As there should be no breaking changes.

I've not encountered any issues during testing.


pvalena:
I didn't get the ACK to accept the CLA from our legal team yet, and it can take some time, sorry.
Feel free to make the change in a different PR.

PR #11231, opened 2019-12-03
Fixes #11228: Allow to force check for box updates

### LINKED ISSUE #11228: "Vagrant box outdated" and "vagrant up" both are not showing latest version of box from locally hosted catalog file
# Vagrant version
Vagrant 2.2.6

### Host operating system
Debian 10 and Windows 10

### Guest operating system
Debian 10

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

 config.vm.box = "debian" 
  #config.vm.box_url = "http://localhost/vagrant/debian/debian.json"
  config.vm.box_url = "file:///var/www/html/vagrant/debian/debian.json"
  
    v.gui = true
  end
end
```

### Debug output
https://gist.github.com/eddi0815/cae742ee6e4a962c43c442de755d5609

```
$ vagrant box outdated
Checking if box 'debian' version '0.1.0' is up to date...
``` 

``` 
$ vagrant box outdated --global
* 'generic/debian10' for 'virtualbox' is outdated! Current: 1.9.38. Latest: 2.0.6
* 'debian' for 'virtualbox' is outdated! Current: 0.1.0. Latest: 0.2.0
``` 

### Expected behavior
**vagrant box outdated** should inform about a new version, too.

### Actual behavior
No information about a new version. The same behaviour with **vagrant up**

### Steps to reproduce
1. I'm using catalog for box versioning as you can see in my vagrantfile and I have this file on localhost
2. I have even tried putting this file at same path of vagrantfile but still no success
3.    Box updates are not even being check on vagrant up

### References
I have the same problem like it is described in issue https://github.com/hashicorp/vagrant/issues/10254

briancain:
Prior to this commit, if a user had recently checked for updates, there
was no way to force Vagrant to re-check without manually deleting a
state file in the local `.vagrant` data dir. This commit fixes that by
giving users the ability to force check for updates for a given box with
a flag to the `vagrant box outdated` command.

### DIFF
--- a/lib/vagrant/action/builtin/box_check_outdated.rb
+++ b/lib/vagrant/action/builtin/box_check_outdated.rb
@@ -40,7 +40,7 @@ module Vagrant
           # Have download options specified in the environment override
           # options specified for the machine.
           download_options = {
-            automatic_check: true,
+            automatic_check: !env[:box_outdated_force],
             ca_cert: env[:ca_cert] || machine.config.vm.box_download_ca_cert,
             ca_path: env[:ca_path] || machine.config.vm.box_download_ca_path,
             client_cert: env[:client_cert] ||
--- a/plugins/commands/box/command/outdated.rb
+++ b/plugins/commands/box/command/outdated.rb
@@ -26,6 +26,10 @@ module VagrantPlugins
               options[:global] = g
             end
 
+            o.on("-f", "--force", "Force checks for latest box updates") do |f|
+              options[:force] = f
+            end
+
             build_download_options(o, download_options)
           end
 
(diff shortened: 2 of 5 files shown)


ZiluckMichael:
@briancain  @chrisroberts Shouldn't this also be added as a flag to `vagrant box update`? It seems odd to only include it for `vagrant box outdated` when `update`.

briancain:
@ZiluckMichael - No, because the `box update` command doesn't pass along a `:automatic_check` key, so that command doesn't have the same issue as the outdated command did.

PR #11250, opened 2019-12-11
Fixes #11249: Add VirtualBox provider support for version 6.1.x

### LINKED ISSUE #11249: Support Virtualbox 6.1
Virtualbox 6.1 was released December 10 and it appears Vagrant does not support this version.

briancain:
This commit adds support for VirtualBox version 6.1.x. It simply
inherits from the base 6.0.x provider class.

### DIFF
--- a/plugins/providers/virtualbox/driver/meta.rb
+++ b/plugins/providers/virtualbox/driver/meta.rb
@@ -64,6 +64,7 @@ module VagrantPlugins
             "5.1" => Version_5_1,
             "5.2" => Version_5_2,
             "6.0" => Version_6_0,
+            "6.1" => Version_6_1,
           }
 
           if @@version.start_with?("4.2.14")
--- a/plugins/providers/virtualbox/driver/version_6_1.rb
+++ b/plugins/providers/virtualbox/driver/version_6_1.rb
@@ -1,0 +1,16 @@
+require File.expand_path("../version_6_0", __FILE__)
+
+module VagrantPlugins
+  module ProviderVirtualBox
+    module Driver
+      # Driver for VirtualBox 6.1.x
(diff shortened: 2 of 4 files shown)


awatry:
I have absolutely no access to review/merge this, but I can at least state that this is working for my use case on virtualbox 6.1 on an ubuntu host with a centos 7 guest image (applied on top of the 2.2.6 release).

Redsandro:
@chrisroberts can we get a release for this sooner rather than later? I've already manually patched 3 machines, and now I have to do a bunch more. It would be sooo nice if we could just release this now, rather than wait to accumulate more changes. :+1:

jens1o:
@Redsandro did you received any response outside of this communication method? I'd also be interested in when a new release will be created.

jens1o:
Is there something I could do to help filing a new release faster (e.g. is there some blocker left) or do we wait for people coming back to the office after the New Year Break?

Redsandro:
@jens1o I think it would make sense to do an intermediate release **right now** because this affects a lot of users and the fix is available. However, that requires action by @chrisroberts, and he took some well deserved time off this Christmas.

Otherwise, [these are the tasks that need to be completed](https://github.com/hashicorp/vagrant/milestone/44) before the planned 2.2.7 release.

glaszig:
i'd like sort of an out-of-order maintenance release with 6.1 support as well. cannot use vagrant right now and don't really like the idea of downgrading virtualbox.

AGausmann:
Productivity at work has ground to a halt. Currently living in dependency hell, trying to find old versions of packages that work together so I can get my environment working again. Tried to downgrade VirtualBox and it failed to build kernel modules, so I have to get an older kernel version as well.

PR #11227, opened 2019-12-01
ansible_local: Add `cap` bug fixes and related RSpec coverage

gildegoma:
## Includes

**Bug Fixes:**
- [x] Review of #11148 (ArchLinux bug fix)
- [x] Fix method signature for FreeBSD and SuSE 

**Improvements:**
- [x] Trim the generated 'pip install' command in Ansible::Cap::Guest::Pip::pip_install

**RSpec Coverage:**
- [x] ArchLinux Guest
- [x] Debian Guest
- [x] Ubuntu Guest
- [x] FreeBSD Guest
- [x] SuSE Guest
- [ ]  Fedora Guest (will be implemented in a distinct PR)
- [ ]  RedHat/CentOS Guest (will be implemented in a distinct PR)

**Integration Tests:**
- [x] `archlinux/archlinux` -> caught  909cc20007bae1ead613c5f2c98c1508ba065a19
- [ ] `freebsd/FreeBSD-12.1-RELEASE` (will very likely be tested by @briancain for #11061/#11263)
- [ ] SuSE 

## References

- GH-6633
- GH-11048
- GH-11263

Issue #6633, opened 2015-12-02
Add test coverage for ansible_local provisioner


gildegoma:
**Context:**
For refactoring safety on the `ansible` provisioner, the introduction of `ansible_local` has been made without major changes in the unit test parts. The downside: unit tests should also be refactored to ease the coverage of both ansible provisioners.

See #2103, #5340 and #6500 

**Goal:**

Keep codebase stable after release 1.8.0, and introduce a new set of unit tests covering both provisioners before accepting new changes.

See http://betterspecs.org/

gildegoma:
Unit Test coverage is added for `cap` code via #11227

PR #11263, opened 2019-12-17
Fix ansible_install for several guest caps

briancain:
Prior to this commit, several guests were not updated to support the additional parameter `pip_install_cmd`. This commit includes that param, even if it goes unused in the guests. Without this param, Vagrant would crash expecting 4 params on the guest capability, rather than the passed in 5.

This pull request also brings in the PR https://github.com/hashicorp/vagrant/pull/11061, which updates the package used on FreeBSD for installing ansible.

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/arch/ansible_install.rb
+++ b/plugins/provisioners/ansible/cap/guest/arch/ansible_install.rb
@@ -7,7 +7,7 @@ module VagrantPlugins
         module Arch
           module AnsibleInstall
 
-            def self.ansible_install(machine, install_mode, ansible_version, pip_args)
+            def self.ansible_install(machine, install_mode, ansible_version, pip_args, pip_install_cmd="")
               if install_mode != :default
                 raise Ansible::Errors::AnsiblePipInstallIsNotSupported
               else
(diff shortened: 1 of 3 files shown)


gildegoma:
@briancain could you please maybe consider bringing changes in #11227 (instead of 02b1ed65ec0824e5fbac5b853c63c1d77d7e412e).

briancain:
@gildegoma - hmm, only if they are ready to be merged. It seems like that pull request is still being worked on? While this change doesn't fully support the pip install method, it will at least prevent people from seeing a stacktrace where Vagrant passed in the wrong number of parameters to the ansible_install capability.

gildegoma:
@briancain sure, this is important to get this fixed anyways in the next release. That said, you can review (and merge) #11227 as it is. I'll add more unit tests in a next step (not that urgent, as you can see in #6633 for the overdue records ;-)

briancain:
@gildegoma - oooh ok, sounds good! I'll do that today, and rebase this PR and fix up any merge conflicts. Then maybe if you could add a couple of issue enhancements around the remaining work so it doesn't get lost, that would be great! :smile: Thanks again :tada:

briancain:
Sorry for the run-around everyone. I've moved these changes into a _new_ PR, https://github.com/hashicorp/vagrant/pull/11265. This includes the work described above, along with the original patch to fixup installing ansible on FreeBSD.

PR #11306, opened 2020-01-09
Remove Windows GEMRC fix/workaround

thijsputman:
The RubyGems pull-request (https://github.com/rubygems/rubygems/pull/2735) this temporary fix/workaround was waiting for to land, got merged in April of 2019.

The workaround is causing Vagrant to crash for many of our Windows users when they try to run `vagrant up` through a NPM-script. Removing the workaround code appears to solve their issue.

### DIFF
--- a/lib/vagrant/bundler.rb
+++ b/lib/vagrant/bundler.rb
@@ -40,20 +40,6 @@ module Vagrant
     def initialize
       @plugin_gem_path = Vagrant.user_data_path.join("gems", RUBY_VERSION).freeze
       @logger = Log4r::Logger.new("vagrant::bundler")
-
-      # TODO: Remove fix when https://github.com/rubygems/rubygems/pull/2735
-      # gets merged and released
-      #
-      # Because of a rubygems bug, we need to set the gemrc file path
-      # through this method rather than relying on the environment varible
-      # GEMRC. On windows, that path gets split on `:`: and `;`, which means
-      # the drive letter gets treated as its own path. If that path exists locally,
-      # (like having a random folder called `c` where the library was invoked),
-      # it fails thinking the folder `c` is a gemrc file.
-      gemrc_val = ENV["GEMRC"]
-      ENV["GEMRC"] = ""
-      Gem.configuration = Gem::ConfigFile.new(["--config-file", gemrc_val])
-      ENV["GEMRC"] = gemrc_val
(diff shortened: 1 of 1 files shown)


thijsputman:
The unit-test at [vagrant/bundler_test.rb:39](https://github.com/thijsputman/vagrant/blob/72ee8a2f0ba6f66480e9e307ef94aec83d0f5755/test/unit/vagrant/bundler_test.rb#L39) now (logically) fails. 

Please advise on how to proceed.

briancain:
@thijsputman - That test can be removed if the original code is being removed. If it's being removed though, we'll need to have a replacement test where the path is a windows path (i.e. `C:/Users/vagrant/folders/and/stuff`) that ensures we get the proper behavior from rubygems!

I'm also not sure if this update requires us being on a newer version of ruby, but when we get around to reviewing the PR we can test all of that. Thanks for taking the time to make a pull request on vagrant though! :tada:

thijsputman:
@briancain It appears the unit-test in question was added as part of the workaround: https://github.com/hashicorp/vagrant/pull/10803. Prior to this, no such test was present.

I would thus propose to simply remove the test in question.

I'm not familiar with Ruby at all, so I'm a bit out of my depth where it comes to writing a new test to ensure proper behaviour of RubyGems/`GEMRC` on Windows I'm afraid...

briancain:
Hey there @thijsputman - I just re-read the original text you wrote at the beginning of this PR. For what it's worth, although that fix was merged into rubygems, it is likely only in a new version of ruby. Vagrant is currently a couple of versions behind the latest release, and I don't have an answer for when that specific version of rubygems will be available to Vagrant.

So for that reason, do you think you could open an issue here on Vagrant with the information we ask showing us what is crashing with your scripts? Thanks!

thijsputman:
@briancain sure, here you go: https://github.com/hashicorp/vagrant/issues/11315.

I can reproduce the issue on most, but not all, of the Windows machines we have in the office here.

The root cause appears to be that somehow NPM adds a lot of extra noise to the environment when running scripts, which is causing RubyGems to error out...

briancain:
Hey! FYI, I'm going to go ahead and close this, as it will be covered in this PR: https://github.com/hashicorp/vagrant/pull/11363 thanks again!

PR #11310, opened 2020-01-10
Fix issues with Windows SSH provisioner

### LINKED ISSUE #11232: Cannot pass environment variables to provisioner with winssh communicator
### Vagrant version
Vagrant 2.2.3, but the same error exists in the current master.

### Host operating system
Linux 4.19, Debian 10.2 (Buster)

### Guest operating system
Windows 10 via VirtualBox

### Vagrantfile
```ruby
  config.vm.communicator = "winssh"
  config.vm.box = "win10"
  config.vm.provision "shell", path: "provision.ps1", env: { "foo" => "bar" }
end
```

#### provision.ps1
```powershell
Set-StrictMode -Version Latest
Write-Output "Hello World!"
```

### Debug output
Cannot provide the full debug output, but the following stack trace completely explains the issue:
https://gist.github.com/jojonas/e7a9cc179614869e745daf07703a81cd

Important error message:
```
vagrant-2.2.3/plugins/provisioners/shell/provisioner.rb:116:in `block (3 levels) in provision_winssh': protected method `generate_environment_export' called for #<VagrantPlugins::CommunicatorWinSSH::Communicator:0x000055f467fccbf8> (NoMethodError)
```

### Expected behavior
The Powershell script should have been executed, printing "Hello World!"

### Actual behavior
An error was thrown, see above.

### Steps to reproduce
1. Provision a box via WinSSH.
2. Pass the `env` hash to `config.vm.provision`.
3. Get a stack trace. Provisioner is not executed.

### Explanation
[1] calls protected function `generate_environment_export` [2].

[1] https://github.com/hashicorp/vagrant/blob/v2.2.3/plugins/provisioners/shell/provisioner.rb#L116
[2] https://github.com/hashicorp/vagrant/blob/v2.2.3/plugins/communicators/ssh/communicator.rb#L729

jbonhag:
Windows commands that run over SSH are wrapped in a script that writes a
special marker to the two output streams (stdout and stderr).  This
allows Vagrant to consume the output streams.

Unfortunately, this leads to a sort of chicken-and-egg problem where no
commands can be run before a wrapper script exists. For example, you
can't make a destination directory to upload the wrapper script without
first creating a wrapper script to make the directory. :)

This PR adds the ability to create "raw" commands with the WinSSH 
communicator (i.e. not wrapped in a shell script). This way we can 
create the `config.winssh.upload_directory` for our wrapper scripts 
without needing to make any assumptions 
about the existing directory structure on the guest.

It also moves the default `upload_path` from the shell provisioner
config so we can have OS-specific defaults.

What remains:
- [x] When `config.winssh.shell = "cmd"`, `.ps1` files should be run 
through PowerShell
- [x] Uploading files with spaces in them -- invoke the `scp.exe` command as an encoded powershell command.
- [ ] Currently the [docs](https://www.vagrantup.com/docs/provisioning/shell.html#inline-scripts) say "For Windows guest machines, the inline 
script must be PowerShell. Batch scripts are not allowed as inline 
scripts." -- for the WinSSH communicator we could assume inline scripts 
to be written in the language of `config.winssh.shell`. Not sure if 
would be more valuable to standardize this across communicators.

Fixes #11232.

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -213,11 +213,18 @@ module Vagrant
           return path if !cygwin?
 
           # Replace all "\" with "/", otherwise cygpath doesn't work.
-          path = path.gsub("\\", "/")
+          path = unix_windows_path(path)
 
           # Call out to cygpath and gather the result
           process = Subprocess.execute("cygpath", "-w", "-l", "-a", path.to_s)
           return process.stdout.chomp
+        end
+
+        # This takes any path and converts Windows-style path separators
+        # to Unix-like path separators.
+        # @return [String]
+        def unix_windows_path(path)
+          path.gsub("\\", "/")
         end
 
         # This checks if the filesystem is case sensitive. This is not a
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -65,7 +65,7 @@ module VagrantPlugins
           while true
             ssh_info = @machine.ssh_info
             break if ssh_info
-            sleep 0.5
+            sleep(0.5)
           end
 
           # Got it! Let the user know what we're connecting to.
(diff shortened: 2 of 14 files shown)


chrisroberts:
Closing this with continued work on #11430. Changes here are included in the new PR, I just didn't want to force push to the branch on this PR.

PR #11349, opened 2020-01-28
[FEATURE] Disk management with the VirtualBox provider

briancain:
This pull request introduces a new experimental feature for managing virtual disks with the VirtualBox provider.

It allows uses to be able to resize the primary disk of a guest, as well as attach additional disks to guests.

### DIFF
--- a/plugins/kernel_v2/config/disk.rb
+++ b/plugins/kernel_v2/config/disk.rb
@@ -140,9 +140,14 @@ module VagrantPlugins
 
           if machine.provider.capability?(:validate_disk_ext)
             if !machine.provider.capability(:validate_disk_ext, @disk_ext)
+              if machine.provider.capability?(:get_default_disk_ext)
+                disk_exts = machine.provider.capability(:get_default_disk_ext).join(', ')
+              else
+                disk_exts = "not found"
+              end
               errors << I18n.t("vagrant.config.disk.invalid_ext", ext: @disk_ext,
                                name: @name,
-                               exts: machine.provider.capability(:get_default_disk_ext).join(', '))
+                               exts: disk_exts)
             end
           else
             @logger.warn("No provider capability defined to validate 'disk_ext' type")
--- a/plugins/kernel_v2/config/disk.rb
+++ b/plugins/kernel_v2/config/disk.rb
@@ -11,7 +11,6 @@ module VagrantPlugins
       #-------------------------------------------------------------------
 
       DEFAULT_DISK_TYPES = [:disk, :dvd, :floppy].freeze
-      DEFAULT_DISK_EXT = ["vdi", "vmdk", "vhd", "vhdx"].freeze
 
       # Note: This value is for internal use only
       #
@@ -136,13 +135,18 @@ module VagrantPlugins
                            types: DEFAULT_DISK_TYPES.join(', '))
         end
 
-        if !DEFAULT_DISK_EXT.include?(@disk_ext)
-          errors << I18n.t("vagrant.config.disk.invalid_ext", ext: @disk_ext,
-                           name: @name, exts: DEFAULT_DISK_EXT.join(', ') )
-        end
-
         if @disk_ext
           @disk_ext = @disk_ext.downcase
+
+          if machine.provider.capability?(:validate_disk_ext)
+            if !machine.provider.capability(:validate_disk_ext, @disk_ext)
+              errors << I18n.t("vagrant.config.disk.invalid_ext", ext: @disk_ext,
+                               name: @name,
+                               exts: machine.provider.capability(:get_default_disk_ext).join(', '))
+            end
+          else
+            @logger.warn("No provider capability defined to validate 'disk_ext' type")
+          end
         end
 
         if @size && !@size.is_a?(Integer)
--- a/plugins/providers/virtualbox/cap/configure_disks.rb
+++ b/plugins/providers/virtualbox/cap/configure_disks.rb
@@ -186,6 +186,10 @@ module VagrantPlugins
           dsk_info
         end
 
+        # @param [Vagrant::Machine] machine
+        # @param [Config::Disk] disk_config - the current disk to configure
+        # @param [Hash] defined_disk - current disk as represented by VirtualBox
+        # @return [Hash] - disk_metadata
         def self.resize_disk(machine, disk_config, defined_disk)
           machine.ui.detail(I18n.t("vagrant.cap.configure_disks.resize_disk", name: disk_config.name), prefix: true)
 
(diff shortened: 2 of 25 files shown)


plugins/kernel_v2/config/disk.rb
```diff
@@ -127,6 +136,15 @@ def validate(machine)
                            types: DEFAULT_DISK_TYPES.join(', '))
         end
 
+        if !DEFAULT_DISK_EXT.include?(@disk_ext)
+          errors << I18n.t("vagrant.config.disk.invalid_ext", ext: @disk_ext,
+                           name: @name, exts: DEFAULT_DISK_EXT.join(', ') )
```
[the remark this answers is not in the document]
briancain: Mostly just so there's some kind of validation that if an extension was picked, it's something expected so they aren't presented with a potentially cryptic provider error. But yeah, since this is the generic config, it might be a chore to keep these accurate across providers :man_shrugging: :thinking:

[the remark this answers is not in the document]
chrisroberts: Would it be possible to push the validation check off to the provider in use?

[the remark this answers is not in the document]
briancain: Definitely possible to have the provider to this check, it just won't be in the config validate step. So basically they won't know until Vagrant goes to create the disk. Then they'll have to destroy/re-up once they fix the extension in their Vagrantfile. I don't think there's a good way otherwise without having different separate provider config.

plugins/kernel_v2/config/disk.rb
```diff
@@ -127,6 +135,20 @@ def validate(machine)
...
         end
 
+        if @disk_ext
+          @disk_ext = @disk_ext.downcase
+
+          if machine.provider.capability?(:validate_disk_ext)
+            if !machine.provider.capability(:validate_disk_ext, @disk_ext)
+              errors << I18n.t("vagrant.config.disk.invalid_ext", ext: @disk_ext,
+                               name: @name,
+                               exts: machine.provider.capability(:get_default_disk_ext).join(', '))
```
chrisroberts: Need a check on the `get_default_disk_ext` capability before using it.

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -0,0 +1,262 @@
...
+              machine.provider.driver.close_medium(defined_disk["UUID"])
+
+              # clone back to original vmdk format and attach resized disk
+              vmdk_disk_file = machine.provider.driver.vdi_to_vmdk(vdi_disk_file)
+              machine.provider.driver.attach_disk(disk_info[:port], disk_info[:device], vmdk_disk_file, "hdd")
+            rescue Exception => e
+              LOGGER.warn("Vagrant encountered an error while trying to resize a disk. Vagrant will now attempt to reattach and preserve the original disk...")
+              machine.ui.error("Vagrant has encountered an exception while trying to resize a disk. It will now attempt to reattach the original disk, as to prevent any data loss.")
+              machine.ui.error("The original disk is located at: #{original_disk["Location"]}")
+              machine.ui.error("If Vagrant fails to reattach the original disk, it is recommended that you open the VirtualBox GUI and navigate to the current guests settings for '#{machine.name}' and look at the 'storage' section. Here is where you can reattach a missing disk if Vagrant fails to do so...")
```
chrisroberts: These UI outputs should be defined in the locales file

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -0,0 +1,262 @@
...
+              # Attach disk
+              machine.provider.driver.
+                attach_disk(disk_info[:port], disk_info[:device], original_disk["Location"], "hdd")
+
+              # Remove cloned disk if still hanging around
+              if vdi_disk_file
+                machine.provider.driver.close_medium(vdi_disk_file)
+              end
+
+              machine.ui.warn("Disk has been reattached. Vagrant will now continue on an raise the exception receieved")
```
chrisroberts: Locales file

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -0,0 +1,262 @@
...
+              FileUtils.mv(defined_disk["Location"], backup_disk_location)
+
+              # we have to close here, otherwise we can't re-clone after
+              # resizing the vdi disk
+              machine.provider.driver.close_medium(defined_disk["UUID"])
+
+              # clone back to original vmdk format and attach resized disk
+              vmdk_disk_file = machine.provider.driver.vdi_to_vmdk(vdi_disk_file)
+              machine.provider.driver.attach_disk(disk_info[:port], disk_info[:device], vmdk_disk_file, "hdd")
+            rescue Exception => e
```
chrisroberts: `e` doesn't need to be captured

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -0,0 +1,262 @@
...
+              machine.provider.driver.
+                attach_disk(disk_info[:port], disk_info[:device], original_disk["Location"], "hdd")
+
+              # Remove cloned disk if still hanging around
+              if vdi_disk_file
+                machine.provider.driver.close_medium(vdi_disk_file)
+              end
+
+              machine.ui.warn("Disk has been reattached. Vagrant will now continue on an raise the exception receieved")
+              raise e
```
chrisroberts: just calling `raise` will suffice here

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -0,0 +1,262 @@
...
+
+              # we have to close here, otherwise we can't re-clone after
+              # resizing the vdi disk
+              machine.provider.driver.close_medium(defined_disk["UUID"])
+
+              # clone back to original vmdk format and attach resized disk
+              vmdk_disk_file = machine.provider.driver.vdi_to_vmdk(vdi_disk_file)
+              machine.provider.driver.attach_disk(disk_info[:port], disk_info[:device], vmdk_disk_file, "hdd")
+            rescue Exception => e
+              LOGGER.warn("Vagrant encountered an error while trying to resize a disk. Vagrant will now attempt to reattach and preserve the original disk...")
```
chrisroberts: This rescue block should be wrapped within a `begin/rescue` with any new exceptions being logged as errors and allowing the original exception to be raised. It might make it a bit cleaner to put the cleanup logic in an isolated method you can call from within here.

plugins/providers/virtualbox/cap/validate_disk_ext.rb
```diff
@@ -0,0 +1,27 @@
+require "log4r"
+
+module VagrantPlugins
+  module ProviderVirtualBox
+    module Cap
+      module ValidateDiskExt
+        LOGGER = Log4r::Logger.new("vagrant::plugins::virtualbox::validate_disk_ext")
+
+        # The default set of disk formats that VirtualBox supports
+        DEFAULT_DISK_EXT = ["vdi", "vmdk", "vhd"].freeze
```
chrisroberts: `.map(&:freeze).freeze` to freeze everything. Otherwise the contents of the array can be modified (`DEFAULT_DISK_EXT.first << 'extra'`)

plugins/providers/virtualbox/cap/configure_disks.rb
chrisroberts: Instead of grabbing `Exception`, lets limit this down to just `ScriptError, SignalException, StandardError`

PR #11355, opened 2020-01-30
docker provider: Don't explode when removing an image if it is in use

### LINKED ISSUE #7245: vagrant destroy docker error because it cannot delete image
### Vagrant version

[root@centos-play ~]# vagrant -v
Vagrant 1.8.1
### Host operating system

[root@centos-play ~]# cat /etc/redhat-release
CentOS Linux release 7.2.1511 (Core)
### Debug output

[root@centos-play icinga_monitoring]# vagrant destroy piper
    piper: Are you sure you want to destroy the 'piper' VM? [y/N] y
==> piper: Stopping container...
==> piper: Deleting the container...
==> piper: Removing built image...
A Docker command executed by Vagrant didn't complete successfully!
The command run along with the output from the command is shown
below.

Command: ["docker", "rmi", "55a589d98865", {:notify=>[:stdout, :stderr]}]

Stderr: Failed to remove image (55a589d98865): Error response from daemon: conflict: unable to delete 55a589d98865 (cannot be forced) - image is being used by running container 9e12171e549d

Stdout:
### Expected behavior

Delete the machine without error.
If the image is used by another container return success if the current container was deleted, and throw an warning about not being able to also delete the image . The image will be deleted anyway when you delete the last container that uses this image.
### Actual behavior

An error is thrown about not being able to delete the image. This is misleading for less experienced users

terryburton:
Output from recent Docker has changed:

    Stderr: Error response from daemon: conflict: unable to delete 0ba49dd235e5 (cannot be forced) - image is being used by running container 250bbe980448

Fixes #7245

### DIFF
--- a/plugins/providers/docker/driver.rb
+++ b/plugins/providers/docker/driver.rb
@@ -166,6 +166,7 @@ module VagrantPlugins
         return true
       rescue => e
         return false if e.to_s.include?("is using it")
+        return false if e.to_s.include?("is being used")
         raise if !e.to_s.include?("No such image")
       end
 
--- a/test/unit/plugins/providers/docker/driver_test.rb
+++ b/test/unit/plugins/providers/docker/driver_test.rb
@@ -389,8 +389,17 @@ describe VagrantPlugins::DockerProvider::Driver do
       end
     end
 
-    context 'image is being used' do
+    context 'image is being used by running container' do
       before { allow(subject).to receive(:execute).and_raise("image is being used by running container") }
+
+      it 'does not remove the image' do
+        expect(subject.rmi(id)).to eq(false)
+        subject.rmi(id)
+      end
+    end
+
+    context 'image is being used by stopped container' do
+      before { allow(subject).to receive(:execute).and_raise("image is being used by stopped container") }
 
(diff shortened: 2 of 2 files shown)


soapy1:
@terryburton, thanks for the contribution! These changes look good to me. Do you think you could add a test for this?

terryburton:
What would the form a such a test look like in order to actually be useful?

For example would the test be against Docker changing the output in future:

up VM1, which creates a new image.
up VM2 based on same image
Remove VM1, which will not be able to remove the image but not result in an error.
Remove VM2, which should remove the image.

I can add this to my backlog but it will not be high priority as it seems like a lot of supporting infrastructure for a basic strings update.

If a staffer wants to make the update then I'm happy to review.

soapy1:
I think unit tests can be effective enough here. Sent you a PR for this https://github.com/terryburton/vagrant/pull/1

terryburton:
> I think unit tests can be effective enough here. Sent you a PR for this [terryburton#1](https://github.com/terryburton/vagrant/pull/1)

Thanks @soapy1. I've expanded it to catch the instance of stopped containers referencing the image. Now LGTM.

PR #11356, opened 2020-01-31
Docker provider: catch container name when using podman.

jackorp:
When user is using podman's docker CLI emulation the containers would
fail to enter running state because the docker driver could not catch
the container name. This commit fixes that by adding a check if podman
docker emulation is used and pick the container hash correctly from the
output.

### DIFF
--- a/plugins/providers/docker/driver.rb
+++ b/plugins/providers/docker/driver.rb
@@ -29,7 +29,16 @@ module VagrantPlugins
         # from standard docker
         matches = result.scan(/writing image .+:([0-9a-z]+) done/i).last
         if !matches
-          matches = result.scan(/Successfully built (.+)$/i).last
+          if podman?
+            # Check for podman format when it is emulating docker CLI.
+            # Podman outputs the full hash of the container on
+            # the last line after a successful build.
+            match = result.split.select { |str| str.match?(/[0-9a-z]{64}/) }.last
+            return match[0..7] unless match.nil?
+          else
+            matches = result.scan(/Successfully built (.+)$/i).last
+          end
+
           if !matches
             # This will cause a stack trace in Vagrant, but it is a bug
             # if this happens anyways.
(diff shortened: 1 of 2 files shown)


pvalena:
LGTM.

pvalena:
_Note: this is using `docker` executable(wrapper) for podman from package `podman-docker` on Fedora / CentOS._

plugins/providers/docker/driver.rb
```diff
@@ -42,6 +52,16 @@ def build(dir, **opts, &block)
         matches[:id]
       end
 
+      # Check if podman emulating docker CLI is enabled.
+      #
+      # @return [Bool]
+      def podman?
+        execute('docker', '-v')
+          .scan(/podman/)
+          .any?
```
chrisroberts: This can be reduced down to: `execute("docker", "-v").include?("podman")`

jackorp: Ah, thanks :)

jackorp:
Deleted extra newline.

plugins/providers/docker/driver.rb
```diff
@@ -42,6 +52,13 @@ def build(dir, **opts, &block)
         matches[:id]
       end
 
+      # Check if podman emulating docker CLI is enabled.
+      #
+      # @return [Bool]
+      def podman?
+        execute('docker', '-v').include?("podman")
```
soapy1: Older versions of podman don't support  the `-v` flag. Eg. if you just run a `sudo yum -y install podman` from centos without adding any more current repositories will leave you with `podman` 1.4.4 which does not have the `-v` option. However they all seem to have the `--version` option. 
```suggestion
        execute('docker', '--version').include?("podman")
```

jackorp: I added this in the newest push, thanks!

PR #11400, opened 2020-02-15
Use systemd-networkd on Archlinux guests

### LINKED ISSUE #11254: [Feature-Request] Change network provider for Arch plugin
Arch Linux builds official Arch Linux Vagrant images under the namespace (Archlinux/Archlinux) on vagrant cloud. We are constantly trying to reduce the image size. Today I've encountered a bug report for our images, where a user tried to use the network segment: https://github.com/archlinux/arch-boxes/issues/70

```ruby
config.vm.box = "archlinux/archlinux"
config.vm.network "private_network", ip: "192.168.0.50"
end
```

This failed, because our image doesn't provide netctl (the next version will possibly do, but I am interested in a change of the Arch plugin).

Arch Linux ships `systemd` in their default installation, with `systemd` `systemd-networkd` is available. I have to admit, that I am not a big fan of netctl, I think most of the features are covered by `systemd-networkd` (Ethernet) and `iwd` (wifi).

Therefore I would like to request a change of the Arch network plugin. I don't know ruby, but the particular lines I would like to change are here:
https://github.com/hashicorp/vagrant/blob/85acf0cac724ef4bfda593a66e0c17f7e4110da0/plugins/guests/arch/cap/configure_networks.rb#L42L49

A quick fix would be to install netctl (either in the image or via the plugin), but on the long term I would like to have `systemd-networkd` been used in this plugin. Is there an easy way to accomplish this? I guess you have already some sort of `systemd-networkd` network provider?

rmunn:
Fixes #11254.

Note that this will only work with boxes that have systemd v244 and above, as I've used `networkctl reload` (https://github.com/systemd/systemd/issues/6654) rather than `systemctl restart systemd-networkd`, since that seems likely to be the least disruptive. If the archlinux box being configured is older than 2019-11-29, though, then it will have systemd v243 (or older) so `networkctl reload` won't exist. So if you want to support archlinux boxes that are more than two months out of date, then edit plugins/guests/arch/cap/configure_networks.rb in this PR and replace the single `networkctl reload` line with `systemctl restart systemd-networkd` instead.


rmunn:
Looks like I need to update the unit tests to expect the systemd unit-file format: I changed the templates, but not the tests that are checking output against what the old templates look like.

shibumi:
@rmunn Just use `networkctl reload`. We don't support older versions of Arch Linux ;-)

rmunn:
@shibumi Then this PR is ready for review as it stands, since that's what I did. :-)

plugins/guests/arch/cap/configure_networks.rb
```diff
@@ -41,10 +41,9 @@ def self.configure_networks(machine, networks)
 
             commands << <<-EOH.gsub(/^ {14}/, '').rstrip
               # Configure #{network[:device]}
-              mv '#{remote_path}' '/etc/netctl/#{network[:device]}' &&
-              ip link set '#{network[:device]}' down &&
-              netctl restart '#{network[:device]}' &&
-              netctl enable '#{network[:device]}'
+              chmod 0644 '#{remote_path}' &&
+              mv '#{remote_path}' '/etc/systemd/network/#{network[:device]}.network' &&
+              networkctl reload
```
soapy1, line 46: I think this might be a pretty destructive change for folks who are using netctl in their boxes. Vagrant could accommodate using either tool by checking what is available on the guest before trying to configure the network. The debian cap does something similar https://github.com/hashicorp/vagrant/blob/master/plugins/guests/debian/cap/configure_networks.rb#L21-L32.

The guest inspection for this is already built out here https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/util/guest_inspection.rb#L22. This could look something like:

```
if systemd_networkd?(comm)
  # do all the bits that use networkctl
else 
  # do all the bits that use netctl
end
```

shibumi: Nice catch. Could be useful for the transition period, too.

rmunn:
I'm not familiar enough with Ruby syntax to fix the two unit test errors. In both, the error is simply that `systemd_networkd?` is not defined/included. It's defined [here](https://github.com/hashicorp/vagrant/blob/4caa1a8ccee8bcc98bee890a020f1a1019eae390/lib/vagrant/util/guest_inspection.rb#L22), and it's used [here](https://github.com/hashicorp/vagrant/blob/a86482b0176fcc395ca40217a16217c8246040ec/plugins/guests/arch/cap/configure_networks.rb#L42). There's an `include Vagrant::Util` line near the top of that file, but that's not enough to pull in the `systemd_networkd?` function. What's the right `include` syntax to fix the two failing tests?

BTW, the failing tests are [here](https://github.com/hashicorp/vagrant/blob/a86482b0176fcc395ca40217a16217c8246040ec/test/unit/plugins/guests/arch/cap/configure_networks_test.rb#L47-L61), in case that matters.

plugins/guests/arch/cap/configure_networks.rb
rmunn: Done in https://github.com/hashicorp/vagrant/commit/a86482b0176fcc395ca40217a16217c8246040ec, though see [my comment](https://github.com/hashicorp/vagrant/pull/11400#issuecomment-635703434) about the two failing unit tests. I'm not familiar enough with Ruby basics to know how to import the `systemd_networkd?` function.

soapy1:
@rmunn I sent you a PR to fix tests https://github.com/rmunn/vagrant/pull/1

rmunn:
@soapy1 - Thanks, looks good. I've pulled in your PR with no changes except to fix one typo I noticed. Looks like tests are passing now, so this is ready to merge unless you want more changes.

ckotte:
Can you please merge this PR?

PR #11405, opened 2020-02-19
Add config files for acceptance tests with Vagrant and GitHub Actions

briancain:
This pull request introduces a couple new workflows for getting started with running the Vagrant blackbox tests in GitHub Actions. Because the repo doesn't exist yet, it has yet to be tested, thus the `[WIP]` in the title. Most of this code is based on other Vagrant GitHub Action workflows.

Relies on https://github.com/hashicorp/vagrant-spec/pull/31


.github/workflows/spectesting.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: .ci/spec/create-packet.sh
+        working-directory: ${{github.workspace}}
+
+  spec-tests:
+    if: github.repository == 'hashicorp/vagrant-blackbox'
+    runs-on: self-hosted
+    strategy:
+      matrix:
+        host_os: ["hashicorp/bionic64"]
+        guest_os: ["hashicorp/bionic64", "hashicorp-vagrant/ubuntu-16.04"]
```
briancain: Not entirely sure if this will work like I hope it might, but :man_shrugging:

.ci/spec/run-test.sh
```diff
@@ -0,0 +1,57 @@
...
+pushd "${root}" > "${output}"
+
+# Assumes packet is already set up
+
+# Define a custom cleanup function to destroy any orphan guests
+# on the packet device
+function cleanup() {
+    (>&2 echo "Cleaning up packet device")
+    unset PACKET_EXEC_PERSIST
+    pkt_wrap_stream "cd vagrant;VAGRANT_CWD=test/vagrant-spec/ VAGRANT_VAGRANTFILE=Vagrantfile.spec vagrant destroy -f" \
```
chrisroberts: The `VAGRANT_CWD` won't be needed here

.github/workflows/spectesting.yml
```diff
@@ -0,0 +1,58 @@
...
+
+  setup-hosts:
+    if: github.repository == 'hashicorp/vagrant-blackbox'
+    runs-on: self-hosted
+    name: Vagrant-Spec Start Hosts
+    needs: setup-packet
+    steps:
+      - name: Code Checkout
+        uses: actions/checkout@v1
+      - name: Run Tests with host ${{ matrix.host_os }} using guest ${{ matrix.guest_os }}
```
chrisroberts: Needs matrix strategy defined, but if testing initially without, will need to remove the matrix usage within the name content.

.ci/spec/create-hosts.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# job_id is provided by common.sh
+export PACKET_EXEC_REMOTE_DIRECTORY="${job_id}"
+
+# spec test configuration, defined by action runners, used by Vagrant on packet
+export PKT_VAGRANT_HOST_BOXES="${VAGRANT_HOST_BOXES}"
+export PKT_VAGRANT_GUEST_BOXES="${VAGRANT_GUEST_BOXES}"
+# other vagrant-spec options
+export PKT_VAGRANT_HOST_MEMORY=10000
+export PKT_VAGRANT_CWD="/test/vagrant-spec/"
```
chrisroberts: Should be a relative path? (test/vagrant-spec)

.github/workflows/spectesting.yml
briancain: Oops, I had it originally, but removed it when I started breaking out the workflow into multiple jobs. I'll put it back. :+1:

.ci/spec/clean-packet.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+csource="${BASH_SOURCE[0]}"
+while [ -h "$csource" ] ; do csource="$(readlink "$csource")"; done
+root="$( cd -P "$( dirname "$csource" )/../" && pwd )"
```
chrisroberts, line 5: Needs to step up one more directory

.ci/spec/clean-packet.sh
```diff
@@ -0,0 +1,23 @@
...
+
+. "${root}/.ci/spec/env.sh"
+. "${root}/.ci/common.sh"
+
+pushd "${root}" > "${output}"
+
+# Ensure we have a packet device to connect
```
chrisroberts, line 12: This comment doesn't look like it applies here.

.ci/spec/create-hosts.sh
```diff
@@ -0,0 +1,38 @@
...
+export PKT_VAGRANT_HOST_BOXES="${VAGRANT_HOST_BOXES}"
+export PKT_VAGRANT_GUEST_BOXES="${VAGRANT_GUEST_BOXES}"
+# other vagrant-spec options
+export PKT_VAGRANT_HOST_MEMORY=10000
+export PKT_VAGRANT_CWD="test/vagrant-spec/"
+export PKT_VAGRANT_VAGRANTFILE=Vagrantfile.spec
+###
+
+# Grab vagrant-spec gem and place inside root dir of Vagrant repo
+download_assets "${ASSETS_PRIVATE_BUCKET}/vagrant-spec/vagrant-spec.gem" "."
```
chrisroberts, line 27: asset will have to be grabbed directly using the `aws` command

.ci/spec/create-hosts.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+csource="${BASH_SOURCE[0]}"
+while [ -h "$csource" ] ; do csource="$(readlink "$csource")"; done
+root="$( cd -P "$( dirname "$csource" )/../" && pwd )"
```
chrisroberts, line 5: Needs to step up one more directory

.ci/spec/create-hosts.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Grab vagrant-spec gem and place inside root dir of Vagrant repo
+download_assets "${ASSETS_PRIVATE_BUCKET}/vagrant-spec/vagrant-spec.gem" "."
+###
+
+# Run the job
+
+echo "Running vagrant spec tests..."
+# Need to make memory customizable for windows hosts
+wrap_stream packet-exec run -upload --  "cd vagrant;vagrant up --no-provision --provider vmware_desktop" \
```
chrisroberts, line 34: The `cd vagrant` should not be required

.ci/spec/create-hosts.sh
```diff
@@ -0,0 +1,38 @@
...
+                                        "Vagrant Blackbox host creation command failed"
+
+
+echo "Finished vagrant spec tests"
```
chrisroberts, line 38: Should be that host guests were started, not that tests are finished.

.ci/spec/create-packet.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+csource="${BASH_SOURCE[0]}"
+while [ -h "$csource" ] ; do csource="$(readlink "$csource")"; done
+root="$( cd -P "$( dirname "$csource" )/../" && pwd )"
```
chrisroberts, line 5: Needs to step up one more directory

.ci/spec/run-test.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+csource="${BASH_SOURCE[0]}"
+while [ -h "$csource" ] ; do csource="$(readlink "$csource")"; done
+root="$( cd -P "$( dirname "$csource" )/../" && pwd )"
```
chrisroberts, line 5: Needs to step up one more directory

.ci/spec/run-test.sh
```diff
@@ -0,0 +1,43 @@
...
+# Assumes packet is already set up
+
+# Define a custom cleanup function to destroy any orphan guests
+# on the packet device
+function cleanup() {
+    (>&2 echo "Cleaning up packet device")
+    unset PACKET_EXEC_PERSIST
+    pkt_wrap_stream "cd vagrant;vagrant destroy -f" \
+                "Vagrant command failed"
+}
```
chrisroberts, line 21: We don't need this cleanup function in the tests script. We can just clean everything at once when we are done.

.ci/spec/run-test.sh
```diff
@@ -0,0 +1,43 @@
...
+export PKT_VAGRANT_HOST_MEMORY=10000
+export PKT_VAGRANT_CWD="test/vagrant-spec/"
+export PKT_VAGRANT_VAGRANTFILE=Vagrantfile.spec
+###
+
+# Run the job
+
+echo "Running vagrant spec tests..."
+# Need to make memory customizable for windows hosts
+pkt_wrap_stream "cd vagrant;vagrant provision --provider vmware_desktop" \
```
chrisroberts, line 39: The `cd vagrant;` should not be required.

.github/workflows/code.yml
```diff
@@ -0,0 +1,48 @@
+on:
+  push:
+    branches:
+      - 'master'
+      - 'spec-test-*'
+
+jobs:
+  sync-vagrant:
+    if: github.repository == 'hashicorp/vagrant-blackbox'
```
chrisroberts, line 9: This section can be removed. We should only be syncing from vagrant and never to vagrant.

.github/workflows/spectesting.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Create packet instance
+        run: .ci/spec/create-packet.sh
+        working-directory: ${{github.workspace}}
+
+  setup-hosts:
+    if: github.repository == 'hashicorp/vagrant-blackbox'
+    runs-on: self-hosted
+    name: Vagrant-Spec Start Hosts
+    needs: setup-packet
+    strategy:
```
chrisroberts, line 24: No strategy should be used in this job

.ci/spec/run-test.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+# job_id is provided by common.sh
+export PACKET_EXEC_REMOTE_DIRECTORY="${job_id}"
+
+# spec test configuration, defined by action runners, used by Vagrant on packet
+export PKT_VAGRANT_HOST_BOXES="${VAGRANT_HOST_BOXES}"
+export PKT_VAGRANT_GUEST_BOXES="${VAGRANT_GUEST_BOXES}"
+# other vagrant-spec options
+export PKT_VAGRANT_HOST_MEMORY=10000
```
chrisroberts, line 30: We shouldn't need the host memory value here.

.ci/spec/clean-packet.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "Cleaning up packet device..."
+
+export PKT_VAGRANT_HOST_BOXES="${VAGRANT_HOST_BOXES}"
+export PKT_VAGRANT_GUEST_BOXES="${VAGRANT_GUEST_BOXES}"
+export PKT_VAGRANT_CWD="test/vagrant-spec/"
+export PKT_VAGRANT_VAGRANTFILE=Vagrantfile.spec
+pkt_wrap_stream "cd vagrant;vagrant destroy -f" \
```
chrisroberts, line 19: The `cd vagrant;` should not be required

.ci/spec/clean-packet.sh
```diff
@@ -0,0 +1,23 @@
...
+                "Vagrant command failed"
```
chrisroberts, line 20: Would be good to make this more descriptive, like: "Failed to destroy guests during cleanup"

.ci/spec/create-hosts.sh
```diff
@@ -0,0 +1,38 @@
...
+# Assumes packet is already set up
+
+# job_id is provided by common.sh
+export PACKET_EXEC_REMOTE_DIRECTORY="${job_id}"
+
+# spec test configuration, defined by action runners, used by Vagrant on packet
+export PKT_VAGRANT_HOST_BOXES="${VAGRANT_HOST_BOXES}"
+export PKT_VAGRANT_GUEST_BOXES="${VAGRANT_GUEST_BOXES}"
+# other vagrant-spec options
+export PKT_VAGRANT_HOST_MEMORY=10000
```
chrisroberts, line 21: We should allow these to be set within the environment, and default to a reasonable value. This way we can easily adjust them using the env vars set in the workflow file.

.ci/spec/create-hosts.sh
```diff
@@ -0,0 +1,38 @@
...
+###
+
+# Grab vagrant-spec gem and place inside root dir of Vagrant repo
+download_assets "${ASSETS_PRIVATE_BUCKET}/vagrant-spec/vagrant-spec.gem" "."
+###
+
+# Run the job
+
+echo "Running vagrant spec tests..."
+# Need to make memory customizable for windows hosts
```
chrisroberts, line 33: This comment doesn't look like it belongs here

.ci/spec/run-test.sh
briancain: We have to here for Windows tests, otherwise there won't be enough memory to run them.

.github/workflows/spectesting.yml
briancain: This is what is used to set up the Vagrant host and guests on packet. The strategy here is passed along into environment variables so vagrant-spec can bring up the right host and guest for the downstream job to provision and run test with.

PR #11411, opened 2020-02-24
Alpine capability for Ansbile support

thunder-spb:
This PR adds capability for Alpine guest for Ansible provisioner installation.

Some notes:
* Ansible version installed via Package manager depends on Alpine guest version
* Use Python3 for any installation mode

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
+++ b/plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
@@ -24,13 +24,12 @@ module VagrantPlugins
             private
 
             def self.ansible_apk_install(machine)
-              machine.communicate.sudo "apk add --update --no-cache python3 ansible"
+              machine.communicate.sudo "apk add --update --no-cache python ansible"
             end
 
             def self.pip_setup(machine, pip_install_cmd = "")
-              machine.communicate.sudo "apk add --update --no-cache python3"
-              machine.communicate.sudo "if [ ! -e /usr/bin/python ]; then ln -sf python3 /usr/bin/python ; fi"
-              machine.communicate.sudo "apk add --update --no-cache --virtual .build-deps python3-dev libffi-dev openssl-dev build-base"
+              machine.communicate.sudo "apk add --update --no-cache python"
+              machine.communicate.sudo "apk add --update --no-cache --virtual .build-deps python-dev libffi-dev openssl-dev build-base"
               Pip::get_pip machine, pip_install_cmd
             end
 
--- a/plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
+++ b/plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
@@ -24,12 +24,15 @@ module VagrantPlugins
             private
 
             def self.ansible_apk_install(machine)
-              machine.communicate.sudo "apk add --update --no-cache python ansible"
+              machine.communicate.sudo "apk add --update --no-cache python3 ansible"
+              machine.communicate.sudo "if [ ! -e /usr/bin/python ]; then ln -sf python3 /usr/bin/python ; fi"
+              machine.communicate.sudo "if [ ! -e /usr/bin/pip ]; then ln -sf pip3 /usr/bin/pip ; fi"
             end
 
             def self.pip_setup(machine, pip_install_cmd = "")
-              machine.communicate.sudo "apk add --update --no-cache python"
-              machine.communicate.sudo "apk add --update --no-cache --virtual .build-deps python-dev libffi-dev openssl-dev build-base"
+              machine.communicate.sudo "apk add --update --no-cache python3"
+              machine.communicate.sudo "if [ ! -e /usr/bin/python ]; then ln -sf python3 /usr/bin/python ; fi"
+              machine.communicate.sudo "apk add --update --no-cache --virtual .build-deps python3-dev libffi-dev openssl-dev build-base"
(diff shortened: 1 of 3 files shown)


thunder-spb:
@briancain hey, any examples to test this because I didn't find any existing tests for capabilities...

briancain:
Hey @thunder-spb ! It should look pretty similar to the existing ansible_install rspec tests...here is debian or arch linux for example: https://github.com/hashicorp/vagrant/blob/master/test/unit/plugins/provisioners/ansible/cap/guest/debian/ansible_install_test.rb
https://github.com/hashicorp/vagrant/blob/master/test/unit/plugins/provisioners/ansible/cap/guest/arch/ansible_install_test.rb

thunder-spb:
Thanks @briancain !

thunder-spb:
@briancain here you go.

plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
```diff
@@ -0,0 +1,45 @@
...
+              machine.communicate.sudo "apk add --update --no-cache python3"
+              machine.communicate.sudo "if [ ! -e /usr/bin/python ]; then ln -sf python3 /usr/bin/python ; fi"
+            end
+
+            def self.ansible_apk_install(machine)
+              machine.communicate.sudo "apk add --update --no-cache ansible"
+            end
+
+            def self.pip_setup(machine)
+              machine.communicate.sudo "pip3 install --upgrade pip"
```
briancain: It looks like we do this for debian (and I also had to do this manually while testing the PR with the guest I was using), we'll want to install some required dev libraries for building ansible with pip:

```ruby
machine.communicate.sudo "apk add build-base libffi-dev python3-dev openssl-dev"
```

plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
```diff
@@ -0,0 +1,45 @@
...
+            def self.python_setup(machine)
+              machine.communicate.sudo "apk add --update --no-cache python3"
+              machine.communicate.sudo "if [ ! -e /usr/bin/python ]; then ln -sf python3 /usr/bin/python ; fi"
+            end
+
+            def self.ansible_apk_install(machine)
+              machine.communicate.sudo "apk add --update --no-cache ansible"
+            end
+
+            def self.pip_setup(machine)
```
briancain: Since we're given a `pip_install_cmd` for `ansible_install`, we should also pass it along and use it if it's set. You can find an example of that in the other AnsibleInstall modules.

plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
thunder-spb: hm, I successfully tested ansible installation via pip without need to install those dev libraries... Interesting...

thunder-spb: so we install those dev libs to install pip via get_pip.py script. In python3 we get pip3 with python3 package. In most cases, this command will run once, on first provision. 
so, yeah, I'll add those back since we have an option to run provision against already provisioned VM and it could fail...

plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
thunder-spb: makes sense... thanks.

plugins/provisioners/ansible/cap/guest/alpine/ansible_install.rb
thunder-spb: sorry, had a tough week :)
So, question still the same. Do we need pip_install_cmd? and do we really need those dev packages? Both required **ONLY** when we use `get_pip.py` script to install `pip`. For Alpine we don't use it and those parameters and packages not required since pip installed via package manager so it's already "compiled" and will install those deps only if required.

If you mean those libraries for later use with ansible then it must be handled by ansible role, not inside this provisioner.

briancain: Hey there, no worries at all. Tough week for many of us, no need to apologize :heart: 

> So, question still the same. Do we need pip_install_cmd? and do we really need those dev packages?

Yep, please add this to your PR. If you want to see what other guest capabilities are doing, they are also adding this command. It makes the pip install more flexible in case users need to modify what pip command is there. We can do what debian and the other guest capabilities to and default it to empty string.

https://github.com/hashicorp/vagrant/blob/8c11489ea73ce2672d9c81a87dd2887eac99f5cd/plugins/provisioners/ansible/cap/guest/debian/ansible_install.rb#L38-L42

For the dev packages, I recommend installing them too. Especially since the most popular alpine boxes on vagrant cloud are missing them and end up failing when trying the pip install method with this PR. So I imagine the first thing users will run into are these missing packages. We also install them with Debian, so we should be good installing them here too.

Hope that makes some more sense.

thunder-spb: Again, your example uses `Pip::get_pip`, because debian's default version is 2.7 and it's required to install pip separately. And only this command consumes `pip_install_cmd`. 
My version does not need `Pip::get_pip` because `pip` installed as a part of python 3. And I don't need `pip_install_cmd` at all. That's why I removed that argument from pip_setup but
 left that in ansible_install to make it backward compatible.

thunder-spb:
@briancain Hey. Added pip installation via `get_pip` script as in Debian. Please review.
Thanks,

PR #11426, opened 2020-03-03
Missed comma in configuration file

bat9r:
Vagrant throws this error when I tried to run this block of code.
```
syntax error, unexpected tIDENTIFIER, expecting keyword_endate_network, type: "dhcp" name: "my-custom-network"
```

After I added a comma, all starts to work correctly.


briancain:
Thanks, nice catch! :pray: :+1:

PR #11430, opened 2020-03-05
Fix issues with Windows SSH provisioner (continued...)

chrisroberts:
This a rebased version of #11310 with some added updates to force
PowerShell environment when running remote ssh commands.


jbonhag:
This fixes #11232.

PR #11436, opened 2020-03-08
Allow to specify python_version

hvnsweeting:
As the feature is supported by bootstrap-salt.sh

https://github.com/saltstack/salt-bootstrap#python-3-support


chrisroberts:
Thanks for this PR. Added some tests to cover the expected behaviors for when the value is and is not set (and rebased it as well). Thanks again!

PR #11453, opened 2020-03-15
Add RHEL/CentOS 8 to list of RHEL flavors

dustymabe:
This will make it so that `:rhel_8` will be returned on el8
guests.

### DIFF
--- a/plugins/guests/redhat/cap/flavor.rb
+++ b/plugins/guests/redhat/cap/flavor.rb
@@ -10,7 +10,9 @@ module VagrantPlugins
           end
 
           # Detect various flavors we care about
-          if output =~ /(CentOS|Red Hat Enterprise|Scientific|Cloud|Virtuozzo)\s*Linux( .+)? release 7/i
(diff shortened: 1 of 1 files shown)


jbonhag:
Hi @dustymabe, thanks for opening the PR! This looks fine, but I have two requests/suggestions:

1. Could you please add some examples of the `:rhel_8` flavors to `test/unit/plugins/guests/redhat/cap/flavor_test.rb`?
2. We have a `:rhel_7` override for installing Docker in `plugins/provisioners/docker/cap/redhat/docker_install.rb` -- I assuming that `:rhel_8` should behave like `:rhel_7`, so would you be willing to add that code?

Thanks again for the PR! 🎉

soapy1:
@dustymabe thank you for the PR! This has been superseded as part of the work in https://github.com/hashicorp/vagrant/pull/11462

dustymabe:
Thanks sorry I didn't have time to get back to this PR

PR #11454, opened 2020-03-16
Fix #11373: Check all interfaces for port collisions

### LINKED ISSUE #11373: Port detection bug with Windows 10
### Vagrant version

Vagrant 2.2.7

### Host operating system

Windows 10

### Guest operating system

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.6 LTS"
```

### Vagrantfile

Full Vagrant file: https://gist.github.com/apyrgio/0369cbb2fa29483baaecb78c0e9c2751#file-vagrantfile

Relevant section from the Vagrantfile:

```ruby
config.vm.network "forwarded_port", guest: 32123, host: 6174, guest_ip: "10.10.10.10", host_ip: "0.0.0.0", auto_correct: false, id: "sync"
```

### Debug output

Gist: https://gist.github.com/apyrgio/0369cbb2fa29483baaecb78c0e9c2751#file-vagrant-log

### Expected behavior

The VM should have started successfully, and the guest port ("10.10.10.10", 32123) should be forwarded to all interfaces in the host ("0.0.0.0", 6174). This works with Vagrant 2.2.4.

### Actual behavior

As can be seen from the logs, the VM failed to start because Vagrant attempted to connect to ("0.0.0.0", 6174) for a port detection check. There are two issues with that:

1. Vagrant should handle `host_ip: "0.0.0.0"` as it would handle an empty `host_ip` option, since they are equivalent.
2. Whatever happens during port detection, it should show up as a warning at most. If the port detection check is advisory, then it shouldn't affect `vagrant up`.

As for what changed from Vagrant 2.2.4: it seems that commit ade076cabbbf072382c9cbff9a9898e5a3ccf037 did not add a case for `host_ip: "0.0.0.0"`, and has caused this regression.

### Steps to reproduce

Install Vagrant 2.2.7 and put in an existing Vagrantfile the following section:

```ruby
config.vm.network "forwarded_port", guest: 32123, host: 6174, host_ip: "0.0.0.0"
```

jbonhag:
This commit changes the behavior of the port check to check all possible
IPv4 network interfaces on a Windows host when the host IP is unset or
`0.0.0.0`.  This means that if the desired port is available on any
network interfaces, a forward from 0.0.0.0 will use that interface.

If the port is open (in use) on all interfaces, then it's treated as a
collision and will either throw an error or auto-correct the port, based
on the Vagrantfile configuration.

### DIFF
--- a/lib/vagrant/action/builtin/handle_forwarded_port_collisions.rb
+++ b/lib/vagrant/action/builtin/handle_forwarded_port_collisions.rb
@@ -243,35 +243,33 @@ module Vagrant
           return extra_in_use.fetch(hostport).include?(hostip)
         end
 
-        def ipv4_addresses
-          ip_addresses = []
-          Socket.getifaddrs.each do |ifaddr|
-            if ifaddr.addr && ifaddr.addr.ipv4?
-              ip_addresses << ifaddr.addr.ip_address
-            end
+        def ipv4_interfaces
+          Socket.getifaddrs.select do |ifaddr|
+            ifaddr.addr && ifaddr.addr.ipv4?
+          end.map do |ifaddr|
+            [ifaddr.name, ifaddr.addr.ip_address]
           end
-          ip_addresses
         end
 
         def port_check(host_ip, host_port)
           # If no host_ip is specified, intention taken to be listen on all interfaces.
           test_host_ip = host_ip || "0.0.0.0"
           if @machine.config.vm.guest == :windows && test_host_ip == "0.0.0.0"
-            @logger.debug("Checking port #{host_port} on all IPv4 addresses...")
-            available_ips = ipv4_addresses.select do |test_host_ip|
-              @logger.debug("Host IP: #{test_host_ip}, port: #{host_port}")
-              !is_port_open?(test_host_ip, host_port)
+            @logger.debug("Testing port #{host_port} on all IPv4 interfaces...")
+            available_interfaces = ipv4_interfaces.select do |interface|
+              @logger.debug("Testing #{interface[0]} with IP address #{interface[1]}")
+              !is_port_open?(interface[1], host_port)
             end
-            if available_ips.empty?
+            if available_interfaces.empty?
               @logger.debug("Cannot forward port #{host_port} on any interfaces.")
               true
             else
-              @logger.debug("These IP addresses will forward to the guest: #{available_ips.join(', ')}")
+              @logger.debug("Port #{host_port} will forward to the guest on the following interfaces: #{available_interfaces}")
               false
             end
           else
             # Do a regular check
-            if test_host_ip == "0.0.0.0" || ipv4_addresses.include?(test_host_ip)
+            if test_host_ip == "0.0.0.0" || ipv4_interfaces.detect { |iface| iface[1] == test_host_ip }
               is_port_open?(test_host_ip, host_port)
             else
               raise Errors::ForwardPortHostIPNotFound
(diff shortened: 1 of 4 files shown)


lib/vagrant/action/builtin/handle_forwarded_port_collisions.rb
```diff
@@ -242,19 +243,34 @@ def is_forwarded_already(extra_in_use, hostport, hostip)
...
-          # If platform is windows, default back to localhost only
-          test_host_ip = host_ip || "0.0.0.0"
-          begin
-            is_port_open?(test_host_ip, host_port)
-          rescue Errno::EADDRNOTAVAIL
-            if !host_ip && test_host_ip == "0.0.0.0"
-              test_host_ip = "127.0.0.1"
-              retry
+          # If no host_ip is specified, intention taken to be listen on all interfaces.
+          if host_ip.nil? || host_ip == "0.0.0.0"
```
chrisroberts: Only run updated check on windows platform

jbonhag:
Example output:

```
DEBUG handle_port_collisions: Testing port 6174 on all IPv4 interfaces...
DEBUG handle_port_collisions: Testing ethernet_32771 with IP address 172.16.0.185
DEBUG handle_port_collisions: Testing ethernet_32772 with IP address 192.168.56.1
DEBUG handle_port_collisions: Testing ethernet_32773 with IP address 10.10.10.1
DEBUG handle_port_collisions: Testing loopback_0 with IP address 127.0.0.1
DEBUG handle_port_collisions: Cannot forward port 6174 on any interfaces.
 INFO handle_port_collisions: Attempting to repair FP collision: 6174
DEBUG handle_port_collisions: Testing port 2200 on all IPv4 interfaces...
DEBUG handle_port_collisions: Testing ethernet_32771 with IP address 172.16.0.185
DEBUG handle_port_collisions: Testing ethernet_32772 with IP address 192.168.56.1
DEBUG handle_port_collisions: Testing ethernet_32773 with IP address 10.10.10.1
DEBUG handle_port_collisions: Testing loopback_0 with IP address 127.0.0.1
DEBUG handle_port_collisions: Port 2200 will forward to the guest on the following interfaces: [["ethernet_32771", "172.16.0.185"], ["ethernet_32772", "192.168.56.
1"], ["ethernet_32773", "10.10.10.1"], ["loopback_0", "127.0.0.1"]]
```

templates/locales/en.yml
```diff
@@ -2321,6 +2321,10 @@ en:
           forwarding: Forwarding ports...
           forwarding_entry: |-
             %{guest_port} (guest) => %{host_port} (host) (adapter %{adapter})
+          host_ip_not_found:
+            You are trying to forward a host IP that does not exist. Please set `host_ip`
+            to the address of an existing IPv4 network interface, or remove the option
+            from your port forward configuration.
```
briancain: It could be useful here to say which guest is setting the option. Sometimes these error messages get lost if the user has a large amount of guests in a Vagrantfile.

briancain: Side note to that, I'm not sure if the guest name is available where this error is being raised, but if we can make it available I think this will be easier for folks to debug.

jbonhag: Thanks for the suggestion, I'll check it out!

PR #11437, opened 2020-03-08
Fix warnings and errors for ruby 2.7

Nautigsam:
Ruby 2.7 causes many depreciation warnings. I tried to remove the majority of them.

I still get some tests failing in `test/unit/plugins/providers/docker/driver_compose_test.rb`. One example of a test failing:
```
$ bundle exec rspec test/unit/plugins/providers/docker/driver_compose_test.rb -e "sets container name"
Run options:
  include {:full_description=>/sets\ container\ name/}
  exclude {:windows=>true}

VagrantPlugins::DockerProvider::Driver::Compose
  #create
    sets container name (FAILED - 1)

Failures:

  1) VagrantPlugins::DockerProvider::Driver::Compose#create sets container name
     Failure/Error: before { expect(subject).to receive(:execute).with(*compose_execute_up) }

       #<VagrantPlugins::DockerProvider::Driver::Compose:0x000056416b3e4ba0 @logger=#<Log4r::Logger:0x000056416b2dd748 @fullname="vagrant::docker::driver::compose", @outputters=[], @additive=true, @name="compose", @path="vagrant::docker::driver", @parent=#<Log4r::Logger:0x000056416b3d2658 @fullname="vagrant::docker::driver", @outputters=[], @additive=true, @name="driver", @path="vagrant::docker", @parent=#<Log4r::RootLogger:0x000056416a3297c0 @level=0, @outputters=[]>, @level=0, @trace=false>, @level=0, @trace=false>, @executor=#<VagrantPlugins::DockerProvider::Executor::Local:0x000056416b325598>, @machine=#<Double "machine">, @data_directory=#<Double "data-directory">, @compose_lock=#<Thread::Mutex:0x000056416b2396c0>> received :execute with unexpected arguments
         expected: ("docker-compose", "-f", "docker-compose.yml", "-p", "cwd", "up", "--remove-orphans", "-d", {})
              got: ("docker-compose", "-f", "docker-compose.yml", "-p", "cwd", "up", "--remove-orphans", "-d") (1 time)
                   ("docker-compose", "-f", "docker-compose.yml", "-p", "cwd", "ps", "-q", "docker_1") (1 time)
     # ./test/unit/plugins/providers/docker/driver_compose_test.rb:88:in `block (3 levels) in <top (required)>'

Finished in 2.02 seconds (files took 0.47252 seconds to load)
1 example, 1 failure
```
I do not know how to fix this. Does someone who is more fluent in Ruby could help?

PR #11459, opened 2020-03-19
Updates for ruby 2.7

terceiro:
In this PR I took 1 commit from #11437 by @Nautigsam, and did another of my
own, plus I'm adding ruby 2.7 to the CI.

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -12,7 +12,7 @@ Gem::Specification.new do |s|
   s.summary       = "Build and distribute virtualized development environments."
   s.description   = "Vagrant is a tool for building and distributing virtualized development environments."
 
-  s.required_ruby_version     = "~> 2.4", "< 2.7"
+  s.required_ruby_version     = "~> 2.4"
   s.required_rubygems_version = ">= 1.3.6"
 
   s.add_dependency "bcrypt_pbkdf", "~> 1.0.0"
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        ruby: [ '2.4.x', '2.5.x', '2.6.x' ]
+        ruby: [ '2.4.x', '2.5.x', '2.6.x', '2.7.x' ]
     name: Vagrant unit tests on Ruby ${{ matrix.ruby }}
     steps:
       - name: Code Checkout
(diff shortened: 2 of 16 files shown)


Nautigsam:
I don't understand why there is no maintainer to communicate on this subject on the numerous issues and PR opened. What's the point in staying quiet?

terceiro:
just be patient and the maintainers will get to it. this is certainly only one of several other things they work on.

vagrant.gemspec
```diff
@@ -12,7 +12,7 @@ Gem::Specification.new do |s|
   s.summary       = "Build and distribute virtualized development environments."
   s.description   = "Vagrant is a tool for building and distributing virtualized development environments."
 
-  s.required_ruby_version     = "~> 2.4", "< 2.7"
+  s.required_ruby_version     = "~> 2.4"
```
dcermak, line 15: I would suggest to keep the restriction on the ruby version, to prevent any breakage that could come with Ruby 2.8.
```suggestion
  s.required_ruby_version     = "~> 2.4", "< 2.8"
```

terceiro: Speaking as a downstream maintainer, I have to say that this adds unnecessary friction to the work or testing vagrant against new ruby versions, and will force downstream to diverge. it's a little thing, but little things here and there accumulate.

besides, vagrant upstream embeds ruby in the omnibus packages, so they have full control over which ruby to include. that upper bound helps nobody.

but if the maintainers insist, by all means I can make this change.

terceiro:
just pushed a rebased branch against master

micheelengronne:
Thanks. As of today, vagrant is impossible to install on Ubuntu Focal without the weight of ruby environments. Ubuntu Focal supports ruby2.7 natively.

chrisroberts:
Hi @terceiro!

Thank you for this PR. I appreciate the time you took in it and my apologies it has taken some time to get back to it. The deprecation and obsolete warnings being generated by 2.7 did end up being a bit involved and took some time to not only chase things down, but also rework some implementation details to properly handle argument and keyword passing on the current non-EOL versions of Ruby.

All of the required updates are included in #11814 which cover the updates you have applied here. As noted in https://github.com/hashicorp/vagrant/pull/11437#issuecomment-672948905 the `URI` -> `CGI` doesn't provide the same results, and #11814 retains the same functionality by using a non-obsolete method.

Thank you again!

terceiro:
thanks @chrisroberts

Issue #11393, opened 2020-02-13
2.2.7 with the docker provider does not obtain the image id if installing awscli


robjay:
### Vagrant version
Vagrant 2.2.7

### Host operating system
Windows 10 1903(OS Build 18362.592)

### Guest operating system
DockerDesktopVM

When using the docker provider to build and image, and the build process outputs a line that starts with Successfully built, Vagrant fails to execute commands on the created image since it picks up the wrong image name from the output.

### Error Output:

        default:  ---> 97044317c1aa
        default: Successfully built 97044317c1aa
        default: SECURITY WARNING: You are building a Docker image from Windows against a non-Windows Docker host. All files and directories added to build context will have '-rwxr-xr-x' permissions. It is recommended to double check and reset permissions for sensitive files and directories.
        default:
        default: Image: PyYAML
    ==> default: Creating the container...
        default:   Name: t_default_1581598305
        default:  Image: PyYAML
    A Docker command executed by Vagrant didn't complete successfully!
    The command run along with the output from the command is shown
    below.

    Command: ["docker", "run", "--name", "t_default_1581598305", "-d", "PyYAML", {:notify=>[:stdout, :stderr]}]

    Stderr: C:\Program Files\Docker\Docker\resources\bin/docker.EXE: invalid reference format: repository name must be lowercase.
    See 'C:\Program Files\Docker\Docker\resources\bin/docker.EXE run --help'.

### Relevant line from the output that is causing the issue:

    default: Building wheels for collected packages: PyYAML
    default:   Running setup.py bdist_wheel for PyYAML: started
    default:   Running setup.py bdist_wheel for PyYAML: finished with status 'done'
    default:   Stored in directory: /root/.cache/pip/wheels/54/b7/c7/2ada654ee54483c9329871665aaf4a6056c3ce36f29cf66e67
    default: Successfully built PyYAML

### Analysis
Looking at the code changes between 2.2.6 and 2.2.7, I've suspect it's to do with the inclusion of stderr when performing a docker build (plugins/providers/docker/driver.rb)

    25 +        opts = {with_stderr: true}
    26 +        result = execute('docker', 'build', *args, opts, &block) 
    27 +        matches = result.match(/Successfully built (?<id>.+)$/i)
    28 +        if !matches
    29 +          # Check for the new output format 'writing image sha256...'
    30 +          # In this case, docker builtkit is enabled. Its format is different
    31 +          # from standard docker
[...]

PR #11461, opened 2020-03-23
Get correct docker image from build output

soapy1:
address https://github.com/hashicorp/vagrant/issues/11393

### DIFF
--- a/plugins/providers/docker/driver.rb
+++ b/plugins/providers/docker/driver.rb
@@ -24,13 +24,12 @@ module VagrantPlugins
         args << dir
         opts = {with_stderr: true}
         result = execute('docker', 'build', *args, opts, &block)
-        matches = result.scan(/Successfully built (?<id>.+)$/i).last
+        # Check for the new output format 'writing image sha256...'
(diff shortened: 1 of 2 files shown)


plugins/providers/docker/driver.rb
```diff
@@ -24,13 +24,13 @@ def build(dir, **opts, &block)
...
         result = execute('docker', 'build', *args, opts, &block)
-        matches = result.match(/Successfully built (?<id>.+)$/i)
+        matches = result.scan(/Successfully built (?<id>.+)$/i).last
         if !matches
           # Check for the new output format 'writing image sha256...'
           # In this case, docker builtkit is enabled. Its format is different
           # from standard docker
           @logger.warn("Could not determine docker container ID. Scanning for buildkit output instead")
-          matches = result.match(/writing image .+:(?<id>[0-9a-z]+) done/i)
+          matches = result.scan(/writing image .+:(?<id>[0-9a-z]+) done/i).last
```
chrisroberts: Since the conversion to `#scan` to replace `#match` doesn't use the match identifier (`?<id>`) it can be removed from the regex here.

Issue #11330, opened 2020-01-21
vm_config.vm.provision :docker fails on CentOS 8


punkrokk:
### Vagrant version
Vagrant 2.2.6

### Host operating system
OSX 10.15.1
### Guest operating system
`centos/8`

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
VAGRANTFILE_API_VERSION = "2"

VIRTUAL_MACHINES = {
  :trusty => {
    :hostname => 'st2-packages-trusty',
    :box => 'ubuntu/trusty64',
    :ip => '192.168.16.20',
  },
  :xenial => {
    :hostname => 'st2-packages-xenial',
    :box => 'ubuntu/xenial64',
    :ip => '192.168.16.21',
  },
  :bionic=> {
    :hostname => 'st2-packages-bionic',
    :box => 'ubuntu/bionic64',
    :ip => '192.168.16.23',
  },
  :el7 => {
    :hostname => 'st2-packages-el7',
    :box => 'centos/7',
    :ip => '192.168.16.22',
  },
  :el8 => {
    :hostname => 'st2-packages-el8',
    :box => 'centos/8',
    :ip => '192.168.16.24',
  },
}

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  VIRTUAL_MACHINES.each do |name, cfg|
    config.vm.define name do |vm_config|
      vm_config.vm.hostname = cfg[:hostname]
      vm_config.vm.box = cfg[:box]

      # Give VM access to all CPU cores on the host
      # docker-compose & rake build can benefit from more CPUs
      host_os = RbConfig::CONFIG['host_os']
      if host_os =~ /darwin/
        cpus = `sysctl -n hw.ncpu`.to_i
      elsif host_os =~ /linux/
        cpus = `nproc`.to_i
      else # sorry Windows folks, I can't help you
        cpus = 2
      end

      # Box Specifications
      vm_config.vm.provider :virtualbox do |vb|
        vb.name = "#{cfg[:hostname]}"
        # NOTE: With 2048MB, system slows due to kswapd. Recommend at least 4096MB.
        vb.customize ['modifyvm', :id, '--memory', '4096']
        vb.customize ["modifyvm", :id, "--cpus", cpus]

        # Docker connection fix
        vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
[...]

PR #11462, opened 2020-03-25
Docker provision centos8

soapy1:
addresses: https://github.com/hashicorp/vagrant/issues/11330


plugins/provisioners/docker/cap/redhat/docker_install.rb
```diff
@@ -7,7 +7,12 @@ def self.docker_install(machine)
             machine.communicate.tap do |comm|
               comm.sudo("yum -q -y update")
               comm.sudo("yum -q -y remove docker-io* || true")
-              comm.sudo("curl -sSL https://get.docker.com/ | sh")
+              if machine.guest.capability("flavor") == :rhel_8
+                # containerd.io is not available on official yum repos
+                # install it directly from docker
+                comm.sudo("yum -y install https://download.docker.com/linux/centos/7/x86_64/stable/Packages/containerd.io-1.2.6-3.3.el7.x86_64.rpm")
+              end
+              comm.sudo("curl -fsSL https://get.docker.com/ | sh")
```
chrisroberts: I'm curious if it would work to just add the docker repository to the system and install the `docker-ce` package instead of just installing the single package and curl/bashing the rest of the installation. The main advantage would be that we wouldn't be referencing a specific package version which will likely change in the future.

plugins/guests/redhat/cap/flavor.rb
```diff
@@ -12,6 +12,8 @@ def self.flavor(machine)
           # Detect various flavors we care about
           if output =~ /(CentOS|Red Hat Enterprise|Scientific|Cloud|Virtuozzo)\s*Linux( .+)? release 7/i
             return :rhel_7
+          elsif output =~ /(CentOS|Red Hat Enterprise|Scientific|Cloud|Virtuozzo)\s*Linux( .+)? release 8/i
+            return :rhel_8
```
jbonhag: This will close #11453 👍

plugins/provisioners/docker/cap/redhat/docker_install.rb
```diff
@@ -4,10 +4,15 @@ module Cap
       module Redhat
         module DockerInstall
           def self.docker_install(machine)
+            if machine.guest.capability("flavor") == :rhel_8
+              machine.ui.warn("Docker is not supported on RHEL 8 machines.")
```
soapy1: Could probably also provide a link to an article that instructs a user how to get around this. Or could suggest they do something like use the podman provisioner https://github.com/hashicorp/vagrant/issues/11464

jbonhag: Borrowing from the get.docker.com script slightly, I think we could say something like:

> RHEL is now only supported by Docker EE. Please install Docker EE or switch to the Podman provisioner to run containers on RHEL.

Does this apply to RHEL < 8 as well?

soapy1: Nice, ya that sounds better!

As far as I can tell, docker isn't supported on RHEL at all. Though the docker docs still have a guide for installing docker ce that works up until (and including) RHEL 7 https://docs.docker.com/install/linux/docker-ce/centos/#install-docker-engine---community.

Maybe the warning should be emitted for all RHEL systems and only hard fail on RHEL 8?

jbonhag: I feel like Vagrant should not attempt to install Docker on RHEL:
1. The docker-install script says it's not supported: https://github.com/docker/docker-install/blob/b50fac9805c3d95fb5e8b16d57b19d333e972e20/install.sh#L318
2. [Docker Engine - Community is not supported on Red Hat Enterprise Linux.](https://docs.docker.com/ee/docker-ee/rhel/)

Folks who want to use Docker on RHEL can still install it themselves using a shell provisioner, since this seems to be pretty well documented.

plugins/guests/centos/plugin.rb
```diff
@@ -0,0 +1,45 @@
...
+      name "CentOS guest"
+      description "CentOS guest support."
+
+      guest(:centos, :linux) do
+        require_relative "guest"
+        Guest
+      end
+
+      guest_capability(:centos, :change_host_name) do
+        require_relative "../redhat/cap/change_host_name"
```
soapy1, line 15: centos should have pretty much all the same capabilities as redhat. I'm not sure if this is the ideal way to de-duplicate code in this case?

chrisroberts: Guests can inherit from a parent when the guest OS is a derivative of an existing distribution. A very simple example of this can be seen in the `ubuntu` guest which really just uses all the `debian` capabilities: https://github.com/hashicorp/vagrant/blob/master/plugins/guests/ubuntu/plugin.rb

Likewise, the `debian` guest inherits from the `linux` guest. The `linux` guest provides mostly generic capabilities, and then the `debian` guest will override specific capabilities which it provides customized implementations: https://github.com/hashicorp/vagrant/blob/master/plugins/guests/debian/plugin.rb

Issue #11414, opened 2020-02-25
vagrant up has not option to silence progress


ssbarnea:
Most CLI tools know to avoid printing progress information when running on non-interactive consoles (no-tty), or at least have an option to disable this.

Apparently there is not cli option for making `vagrant up` less verbose.

### Vagrant version
```
2.2.7
```

### Expected behavior
Be able to control verbosity, especially around progressive commands (those that rewrite current line).

### Actual behavior
```
[K==> default: Successfully added box 'generic/alpine310' (v2.0.6) for 'libvirt'!
2020-02-24 20:28:59.624663 | fedora-30 | ==> default: Uploading base box image as volume into libvirt storage...
2020-02-24 20:29:00.479456 | fedora-30 | 
[KProgress: 0%
[KProgress: 0%
[KProgress: 0%
[KProgress: 0%
[KProgress: 1%
[KProgress: 1%
[KProgress: 1%
[KProgress: 1%
[KProgress: 2%
[KProgress: 2%
[KProgress: 2%
[KProgress: 2%
[KProgress: 3%
[KProgress: 3%
[KProgress: 3%
```
### Steps to reproduce
```
vagran up
```

PR #11465, opened 2020-03-25
Quiet cli option

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11414

Putting this up as an idea. Not sure if I really like the implementation. A couple of questions:
- should vagrant just throw out the whole output when the `--quiet` flag is used?
- are there any instances where the output should be "quiet" but there still needs to be information returned to the user?
- should the `--quiet` flag exist globally, or does it make sense for only some commands to have a quiet option

### DIFF
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -78,12 +78,6 @@ if argv.include?("--debug-timestamp")
   ENV["VAGRANT_LOG_TIMESTAMP"] = "1"
 end
 
-# Setting to enable/disable showing progress bars
-if argv.include?("--quiet-progress")
-  argv.delete("--quiet-progress")
-  ENV["VAGRANT_QUIET_PROGRESS"] = "1"
-end
-
 # Stdout/stderr should not buffer output
 $stdout.sync = true
 $stderr.sync = true
@@ -114,6 +108,7 @@ begin
     o.on("--debug", "Enable debug output")
     o.on("--timestamp", "Enable timestamps on log output")
     o.on("--debug-timestamp", "Enable debug output with timestamps")
+    o.on("--no-tty", "Enable non-interactive output")
   })
 
   # Create a logger right away
@@ -150,6 +145,12 @@ begin
   if argv.include?("--machine-readable")
     argv.delete("--machine-readable")
     opts[:ui_class] = Vagrant::UI::MachineReadable
+  end
+
+  # Setting to enable/disable showing progress bars
+  if argv.include?("--no-tty")
+    argv.delete("--no-tty")
+    opts[:ui_class] = Vagrant::UI::NonInteractive
   end
 
   # Default to colored output
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -78,6 +78,12 @@ if argv.include?("--debug-timestamp")
   ENV["VAGRANT_LOG_TIMESTAMP"] = "1"
 end
 
+# Setting to enable/disable showing progress bars
+if argv.include?("--quiet-progress")
+  argv.delete("--quiet-progress")
+  ENV["VAGRANT_QUIET_PROGRESS"] = "1"
+end
+
 # Stdout/stderr should not buffer output
 $stdout.sync = true
 $stderr.sync = true
--- a/lib/vagrant/ui.rb
+++ b/lib/vagrant/ui.rb
@@ -31,9 +31,14 @@ module Vagrant
       attr_accessor :stderr
 
       def initialize
-        @logger   = Log4r::Logger.new("vagrant::ui::interface")
-        @opts     = { :show_progress => true }
+        if ENV["VAGRANT_QUIET_PROGRESS"]
+          show_progress = false
+        else
+          show_progress = true
+        end
 
+        @logger   = Log4r::Logger.new("vagrant::ui::interface")
+        @opts     = { :show_progress => show_progress }
         @stdin  = $stdin
         @stdout = $stdout
         @stderr = $stderr
--- a/lib/vagrant/ui.rb
+++ b/lib/vagrant/ui.rb
@@ -31,14 +31,8 @@ module Vagrant
       attr_accessor :stderr
 
       def initialize
-        if ENV["VAGRANT_QUIET_PROGRESS"]
-          show_progress = false
-        else
-          show_progress = true
-        end
-
         @logger   = Log4r::Logger.new("vagrant::ui::interface")
-        @opts     = { :show_progress => show_progress }
+        @opts     = {}
         @stdin  = $stdin
         @stdout = $stdout
         @stderr = $stderr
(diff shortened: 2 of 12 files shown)


bin/vagrant
```diff
@@ -78,6 +78,12 @@ if argv.include?("--debug-timestamp")
   ENV["VAGRANT_LOG_TIMESTAMP"] = "1"
 end
 
+# Setting to enable/disable showing progress bars
+if argv.include?("--quiet-progress")
```
soapy1: I don't know about this
- Seems like this is the pattern set for `--debug` and `--timestamp`. I think `--quiet-progress` is in the same class of options as those
- Any other suggestions about naming this option? I think this makes sense since it is not silencing the whole output. But also, I think this name is pretty clunky. I thought against `--quiet` since this options doesn't truly quiet all the output but I'm happy to rethink that

briancain: To your second point, I agree that just `--quiet` will be a bit misleading if we're using this flag to disable progress output. I like `--quiet-progress`, I think it's short and descriptive enough that hopefully folks know what it's for. Unfortunately if we define this flag globally like this, we won't be able to show it easily in any CLI help output, so having it be named something short and descriptive is ideal! And because of that, we'll want to make sure to update the documentation with this flag and the corresponding environment variable.

soapy1: this bit should probably change to follow new global cli options after this PR https://github.com/hashicorp/vagrant/pull/11471 gets merged. Also, rename this to `non-interactive` or something

chrisroberts: After discussion internally I think we landed on `--no-tty`

bin/vagrant
```diff
@@ -78,6 +78,12 @@ if argv.include?("--debug-timestamp")
...
+  argv.delete("--quiet-progress")
+  ENV["VAGRANT_QUIET_PROGRESS"] = "1"
```
chrisroberts: With having a dedicated UI subclass, we won't need to worry about tracking in an environment variable. Instead, just use the option to instantiate the new subclass here: https://github.com/hashicorp/vagrant/blob/e5d4243aecde7526176d16441f2252f744614885/bin/vagrant#L118

lib/vagrant/ui.rb
```diff
@@ -31,9 +31,14 @@ class Interface
...
-        @logger   = Log4r::Logger.new("vagrant::ui::interface")
-        @opts     = {}
+        if ENV["VAGRANT_QUIET_PROGRESS"]
+          show_progress = false
+        else
+          show_progress = true
+        end
 
+        @logger   = Log4r::Logger.new("vagrant::ui::interface")
+        @opts     = { :show_progress => show_progress }
```
chrisroberts: `show_progress` options can be ignored since the subclass usage will implicitly provide the logic

lib/vagrant/ui.rb
```diff
@@ -75,6 +80,12 @@ def color?
       def machine(type, *data)
         @logger.info("Machine: #{type} #{data.inspect}")
       end
+
+      def output_if_showing_progress
```
chrisroberts: Maybe we can rename this to `rewriting`? Then, when used within the code we'd end up with something like:

```ruby
loop do
  @ui.rewriting do |ui|
    ui.detail "Status ..."
    ui.clear_line
  end
  sleep 5
end
```

The `rewriting` method would then simply `yield` self, except in the new non interactive subclass, which would provide a no-op.

lib/vagrant/ui.rb
```diff
@@ -252,6 +265,19 @@ def format_message(type, message, **opts)
...
+    class NonInteractive < Basic
+      def initialize
+        super
+        @opts = { :show_progress => false }
+      end
+
+      def ask(*args)
+        # Non interactive can't ask for input
+        raise Errors::UIExpectsTTY
+      end
```
chrisroberts: Should include a no-op `rewriting` method. 

The `clear_line` method should probably also be overridden. I think the correct behavior would be throwing an error when used, but that would likely introduce unexpected breakages simply for correctness. Might be better to just emit a warning for now, and output a newline instead?

lib/vagrant/ui.rb
```diff
@@ -290,7 +316,11 @@ def #{method}(message, *args, **opts)
 
       [:clear_line, :report_progress].each do |method|
         # By default do nothing, these aren't formatted
-        define_method(method) { |*args| @ui.send(method, *args) }
+        define_method(method) do |*args|
```
chrisroberts: Won't need this with dedicated subclass.

bin/vagrant
chrisroberts: Global options PR got moved to #11473 (just FYI)

lib/vagrant/ui.rb
soapy1: Sounds good. 
When outputting a new line the result of `vagrant up box` is 
```
...
==> vbox: Loading metadata for box 'bento/ubuntu-18.04'
    vbox: URL: https://vagrantcloud.com/bento/ubuntu-18.04
==> vbox: Adding box 'bento/ubuntu-18.04' (v202002.14.0) for provider: virtualbox
    vbox: Downloading: https://vagrantcloud.com/bento/boxes/ubuntu-18.04/versions/202002.14.0/providers/virtualbox.box

    vbox: Download redirected to host: vagrantcloud-files-production.s3.amazonaws.com
...
```


When not outputting a new line the result of `vagrant up box` is 
```
...
==> vbox: Loading metadata for box 'bento/ubuntu-18.04'
    vbox: URL: https://vagrantcloud.com/bento/ubuntu-18.04
==> vbox: Adding box 'bento/ubuntu-18.04' (v202002.14.0) for provider: virtualbox
    vbox: Downloading: https://vagrantcloud.com/bento/boxes/ubuntu-18.04/versions/202002.14.0/providers/virtualbox.box
==> vbox: Box download is resuming from prior download progress
    vbox: Download redirected to host: vagrantcloud-files-production.s3.amazonaws.com
...
```

I think outputting a new line looks a little more messy. But I'm happy to keep it in if you think it's useful for users to see that

lib/vagrant/ui.rb
```diff
@@ -75,6 +75,10 @@ def color?
       def machine(type, *data)
         @logger.info("Machine: #{type} #{data.inspect}")
       end
+
```
chrisroberts, line 78: Would you please add a documentation block for this method including a brief explanation of its purpose.

lib/vagrant/ui.rb
chrisroberts: I agree, it does look messy, but I think that might be the result of `clear_line` not being used correctly in that location. Generally, when it is used, a line is written without a new line attached so the cursor stays on the line. The clear line will remove the content and return the cursor to the start of the line and then write again. If progress output, or some other type of dynamically modified output, is being displayed and is not wrapped with the `rewriting` method, the result will be one extremely long line. So I think the behavior of writing a new line is okay here, but we'll need to cleanup the clear_line usage internally.

PR #11471, opened 2020-03-27
Include default options in option parser

chrisroberts:
Adds method to shared helpers for adding procs to be evaluated
which can add default modifications to the option parser used
by commands. Customized option parser class within Vagrant
handles processing defined procs.

### DIFF
--- a/bin/vagrant
+++ b/bin/vagrant
@@ -99,6 +99,17 @@ begin
     $VERBOSE = nil
   end
 
+  # Add any option flags defined within this file here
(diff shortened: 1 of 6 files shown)


bin/vagrant
```diff
@@ -99,6 +99,17 @@ begin
     $VERBOSE = nil
   end
 
+  # Add any option flags defined within this file here
+  # so they are automatically propagated to all commands
+  Vagrant.add_default_cli_options(proc { |o|
```
soapy1: 😍

Issue #11464, opened 2020-03-25
Add podman provisioner


chrisroberts:
Add podman support via explicit `:podman` provisioner built on top of existing docker provisioner plugin.

PR #11472, opened 2020-03-27
Add podman provisioner

soapy1:
fixes: https://github.com/hashicorp/vagrant/issues/11464


plugins/provisioners/podman/cap/linux/podman_installed.rb
```diff
@@ -0,0 +1,26 @@
...
+              "/usr/bin/podman",
+              "/usr/local/bin/podman",
+              "/usr/sbin/podman",
+            ]
+
+            paths.each do |p|
+              if machine.communicate.test("test -f #{p}", sudo: true)
+                return true
+              end
+            end
```
chrisroberts: Instead of testing a list of explicit paths, a single test with `command -v podman` will validate if podman is installed and available via the configured `PATH`

Issue #11407, opened 2020-02-21
Multiple capabilities do not work with Systemd


AnthonyMastrean:
Some distros are completely switched to Systemd and no longer ship a `reboot` or `shutdown` command... they use the new [systemctl halt service](http://man7.org/linux/man-pages/man8/systemd-halt.service.8.html). I think that several capabilities need to be enhanced.

 * [halt](https://github.com/hashicorp/vagrant/blob/74e54483c6f0ffe2ce9027643517544bbd58acd2/plugins/guests/linux/cap/halt.rb#L7) (and one [halt test](https://github.com/hashicorp/vagrant/blob/11b0d58fa081cd9a27c272244a0d119acc81f32e/test/unit/plugins/guests/linux/cap/halt_test.rb))
 * [reboot](https://github.com/hashicorp/vagrant/blob/74e54483c6f0ffe2ce9027643517544bbd58acd2/plugins/guests/linux/cap/reboot.rb#L11) (there are no Linux reboot tests, only [Windows](https://github.com/hashicorp/vagrant/blob/11b0d58fa081cd9a27c272244a0d119acc81f32e/test/unit/plugins/guests/windows/cap/reboot_test.rb))

It looks like Vagrant already supports [Systemd inspection](https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/util/guest_inspection.rb). And [other caps](https://github.com/hashicorp/vagrant/blob/11b0d58fa081cd9a27c272244a0d119acc81f32e/plugins/guests/redhat/cap/change_host_name.rb#L34-L40) already test for Systemd and choose different commands. So the pattern seems established? The fix _may_ be along these lines...

```ruby
reboot_script = "reboot"
if systemd?(comm)
    reboot_script = "systemctl reboot"
end
comm.sudo(reboot_script)
```

### Vagrant version
Vagrant 2.2.6

### Host operating system
Windows 10

### Guest operating system
Debian 10 (Buster)

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
  config.vm.box = "generic/debian10"
  config.vm.provision "shell", reboot: true
end
```

### Expected behavior
Systemd guests should halt/reboot correctly.

### Actual behavior
Vagrant says

```
Vagrant attempted to execute the capability 'reboot'
on the detect guest OS 'debian', but the guest doesn't
support that capability. This capability is required for your     
[...]

PR #11482, opened 2020-04-01
Use systemd to halt system if available

soapy1:
Addresses the `halt` part of this issue https://github.com/hashicorp/vagrant/issues/11407

### DIFF
--- a/plugins/guests/linux/cap/halt.rb
+++ b/plugins/guests/linux/cap/halt.rb
@@ -1,10 +1,18 @@
+require 'vagrant/util/guest_inspection'
+
 module VagrantPlugins
   module GuestLinux
     module Cap
       class Halt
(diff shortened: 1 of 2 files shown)


plugins/guests/linux/cap/halt.rb
```diff
@@ -1,10 +1,18 @@
...
   module GuestLinux
     module Cap
       class Halt
+        extend Vagrant::Util::GuestInspection::Linux
+
         def self.halt(machine)
           begin
-            machine.communicate.sudo("shutdown -h now")
+            if systemd?(machine.communicate)
+              machine.communicate.sudo("systemctl halt")
```
jbonhag, line 12: I think this should use `systemctl poweroff` because `halt` leaves the machine running.

jbonhag:
🙌

PR #11485, opened 2020-04-01
Add static route support for debian guests

thomasDOTwtf:
It should be possible to confgure static routes for debian guests.
There is an existing implementation to perform static route configuration inside funtoo linux guest systems. This implementation assumes a route property to be existent inside the network options.

The route property is interpreted like the following:
> route
> Specify a semi-colon delimited list of IPv4 routes to apply when this interface is brought up. Will be appended to ip -4 route add.

According to funtoo documentation: [funtoo linux - Networking Documentation](https://www.funtoo.org/Networking#Server_Network_Configuration
)

Currently route configurations aren't passed along from the libvirt provider.
There is a filed pull request: https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1086

### DIFF
--- a/templates/guests/debian/network_static.erb
+++ b/templates/guests/debian/network_static.erb
@@ -7,4 +7,9 @@ iface <%= options[:device] %> inet static
 <% if options[:gateway] %>
       gateway <%= options[:gateway] %>
 <% end %>
+<% if options[:route] %>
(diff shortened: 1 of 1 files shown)


thomasDOTwtf:
Hey, do you see any timeframe for the review needed? @briancain

PR #11487, opened 2020-04-02
Update childprocess gem

soapy1:
childprocess v 1.x includes a fix for encoding on Windows
https://github.com/enkessler/childprocess/pull/134
This was previously causing errors on Windows hosts


soapy1:
I think this also requires an update to `vagrant-spec`
https://github.com/hashicorp/vagrant-spec/pull/33

PR #11494, opened 2020-04-05
add static route configuration for redhat guests

thomasDOTwtf:
It should be possible to confgure static routes for Redhat and CentOS guests.
There is an existing implementation to perform static route configuration inside funtoo linux guest systems. This implementation assumes a route property to be existent inside the network options.

The route property is interpreted like the following:

route
Specify a semi-colon delimited list of IPv4 routes to apply when this interface is brought up. Will be appended to ip -4 route add.

According to funtoo documentation: funtoo linux - Networking Documentation

There is a second pull request for debian OS.

Currently route configurations aren't passed along from the libvirt provider.
There is a filed pull request: vagrant-libvirt/vagrant-libvirt#1086


plugins/guests/redhat/cap/configure_networks.rb
```diff
@@ -48,6 +48,23 @@ def self.configure_networks(machine, networks)
               raise Vagrant::Errors::NetworkManagerNotInstalled, device: network[:device]
             end
 
+            # Render static route file if present
+            if network.key?(:route) && network[:route] != nil
```
briancain, line 52: No need to explicitly check nil, just checking `network[:route]` will be good enough

```suggestion
            if network.key?(:route) && network[:route]
```

plugins/guests/redhat/cap/configure_networks.rb
```diff
@@ -48,6 +48,23 @@ def self.configure_networks(machine, networks)
...
+              Tempfile.open("vagrant-redhat-configure-networks-routes") do |f|
+                f.binmode
+                f.write(routeentry)
+                f.fsync
+                f.close
+                machine.communicate.upload(f.path, route_remote_path)
+              end
+              route_final_path = "#{network_scripts_dir}/route-#{network[:device]}"
+              commands[:middle] << "mv -f '#{route_remote_path}' '#{route_final_path}'"
+            end
```
briancain, line 66: This whole change needs some unit tests inside the redhat configure networks class.

templates/guests/redhat/network_route.erb
```diff
@@ -0,0 +1,7 @@
+#VAGRANT-BEGIN
+# The contents below are automatically generated by Vagrant. Do not modify.
+# <%= routes.inspect %>
+<% routes.split(';').each do |route| %>
```
briancain, line 4: Is this `routes` option coming from a Vagrantfile? What if someone uses something other than `;` to split their IPs? If I remember right this `network` option from `configure_networks` comes from `config.vm.network :network_type, routes: "routes;go;here"` which means someone could split with commas or spaces and this won't work :thinking:

Issue #11378, opened 2020-02-06
"netplan apply" hanged "vagrant up" when starting Ubuntu Desktop with NetworkManager


ruzickap:
### Vagrant version

```
$ vagrant -v
Vagrant 2.2.7
```

### Host operating system

```
$ grep PRETTY_NAME /etc/os-release
PRETTY_NAME="Ubuntu 18.04.4 LTS"
```

### Guest operating system

Ubutnu 19.10 Desktop

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "peru/ubuntu-19.10-desktop-amd64-libvirt-test"
  config.vm.box_version = "0.0.1"
end
```

### Debug output

```
 INFO ssh: SSH is ready!
 INFO ssh: Execute:  (sudo=false)
 INFO guest: Execute capability: insert_public_key [#<Vagrant::Machine: default (VagrantPlugins::ProviderLibvirt::Provider)>, "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC28hHtLOqk+dXxqOQ9Rua550vsxXID8fMtPe6Lw+N/U/8u6rUTamWdykTjBiSZ4ZxGdiZRbJB9gbORo+9RdOEh9Fkq0huLRj2fu25/HNJaBryyqyUPHtc9MomsPQnqOBBUJ14xLcVLNevOZ1oTbP1znE56ijbgZ75792DParPJK/xv3gVNB8FPX5q5OI7n+U8fYBRrcCF1R6z7iUAV6CTWPvmPBft9/Vkxc+B7Fd+syDS0+5n19AwlARUUx/o50RgKSyePWXEj/FwX/HqCJWcr2nFUjF2JcobBl8yaoF45SH/Xpp8MxmuNrQDLYO9JTtxY6Di/d98zi9OZZSHL3u2j vagrant"] (ubuntu)
 INFO ssh: Execute: mkdir -p "/tmp" (sudo=false)
 INFO ssh: Execute: mkdir -p ~/.ssh
chmod 0700 ~/.ssh
cat '/tmp/vagrant-insert-pubkey-1581002099' >> ~/.ssh/authorized_keys && chmod 0600 ~/.ssh/authorized_keys
result=$?
rm -f '/tmp/vagrant-insert-pubkey-1581002099'
exit $result
 (sudo=false)
[...]

PR #11498, opened 2020-04-06
Choose netplan renderer based on network config and available tools

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11378

### DIFF
--- a/plugins/guests/debian/cap/configure_networks.rb
+++ b/plugins/guests/debian/cap/configure_networks.rb
@@ -10,7 +10,6 @@ module VagrantPlugins
         extend Vagrant::Util::GuestInspection::Linux
 
         NETPLAN_DEFAULT_VERSION = 2
-        NETPLAN_DEFAULT_RENDERER = "networkd".freeze
         NETPLAN_DIRECTORY = "/etc/netplan".freeze
         NETWORKD_DIRECTORY = "/etc/systemd/network".freeze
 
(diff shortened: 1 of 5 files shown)


plugins/guests/debian/cap/configure_networks.rb
```diff
@@ -61,12 +61,21 @@ def self.configure_netplan(machine, interfaces, comm, networks)
           # By default, netplan expects the renderer to be systemd-networkd,
           # but if any device is managed by NetworkManager, then we use that renderer
           # ref: https://netplan.io/reference
-          renderer = NETPLAN_DEFAULT_RENDERER
-          ethernets.keys.each do |k|
-            if nm_controlled?(comm, k)
-              renderer = "NetworkManager"
-              break
+          if networkd?(comm)
```
soapy1, line 64: netplan supports `networkd` and `NetworkManager` as renderers (https://netplan.io/), by default it uses `networkd`. If `networkd` is not installed, Vagrant should probably fall back to `NetworkManager` if it is available. If neither of these are available, then netplan will not work (Vagrant should throw an error). 
Additionally, a device can be set to be managed by `NetworkManager`, in which case netplan should use the `NetworkManager` renderer.

templates/locales/en.yml
```diff
@@ -1000,6 +1000,9 @@ en:
         the proper location.
       multi_vm_target_required: |-
         This command requires a specific VM name to target in a multi-VM environment.
+      netplan_no_available_renderers: |-
+        No renderers compatible with netplan are avaialble on guest. Please install
```
chrisroberts, line 1004: Typo on `available`

plugins/guests/debian/cap/configure_networks.rb
```diff
@@ -61,12 +61,21 @@ def self.configure_netplan(machine, interfaces, comm, networks)
           # By default, netplan expects the renderer to be systemd-networkd,
           # but if any device is managed by NetworkManager, then we use that renderer
           # ref: https://netplan.io/reference
-          renderer = NETPLAN_DEFAULT_RENDERER
```
chrisroberts, line 64: The `NETPLAN_DEFAULT_RENDERER` constant can be removed as well since this was the only place it was being used.

Issue #11039, opened 2019-08-25
Hyper-V integration services does not work with non ASCII names


111andre111:
### Vagrant version
`> vagrant -v
Vagrant 2.2.5`

### Host operating system
Windows 10 v1903 build 18362.239 German language version
for sure simiar issues with other languages

### Guest operating system
rhel7

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "generic/rhel7"
  config.vm.network "public_network", bridge: "Default Switch"
config.vm.synced_folder ".", "/vagrant", disabled: true
  config.vm.provider "hyperv" do |h|
    h.enable_virtualization_extensions = true 
    h.linked_clone = true 
    h.vm_integration_services = {
        guest_service_interface: true,
        key_value_pair_exchange: true,
        time_synchronization: true,
        heartbeat: true,
    }
  end
end
```

### Debug output
Provide a link to a GitHub Gist containing the complete debug output:
https://gist.github.com/111andre111/e65f66a90ba129677ad60e4335da0183

### Expected behavior
What should have happened?
Machine should start through, but it stuck with error.

The original reason why I added these fields, especially
`key_value_pair_exchange`, was this error:
`Hyper-V failed to determine your machine's IP address within the configured timeout. Please verify the machine properly booted and the network works. To do this, open the Hyper-V manager, find your virtual machine, and connect to it.
The most common cause for this error is that the running virtual machine doesn't have the latest Hyper-V integration drivers. Please research for your operating system how to install these in order for the VM to properly communicate its IP address to Hyper-V.`
Increase timeout didn't help, so I hoped it is the reason adding field `key_value_pair_exchange`.

### Actual behavior
Stuck with error:
`Failed to enable VM integration service Key-Value Pair Exchange: Es wurde keine Integrationskomponente mit dem angegebenen Namen gefunden.`

### Steps to reproduce
1. added the values from integration services
2. execute `vagrant up --provider=hyperv`
3. found error

### References
Are there any other GitHub issues (open or closed) that should be linked here?
similar issue:
- GH-9679

### fast Solution
I digged a bit into this topic.
At first searching what I get back for integration services with this command:
```
> Get-VMIntegrationService -VMName generic-rhel7-hyperv
             VMName               Name                               Enabled PrimaryStatusDescription SecondaryStatusDescription
             ------               ----                               ------- ------------------------ --------------------------
             generic-rhel7-hyperv Gastdienstschnittstelle            False
[...]

PR #11499, opened 2020-04-07
Use service id for enabling/disabling vm integration services

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11039

ref:
https://github.com/PlagueHO/LabBuilder/issues/183
ref service name to id:
https://social.technet.microsoft.com/Forums/de-DE/154917de-f3ca-4b1e-b3f8-23dd4b4f0f06/getvmintegrationservice-sprachabhngig?forum=powershell_de

### DIFF
--- a/plugins/providers/hyperv/driver.rb
+++ b/plugins/providers/hyperv/driver.rb
@@ -10,15 +10,16 @@ module VagrantPlugins
       ERROR_REGEXP  = /===Begin-Error===(.+?)===End-Error===/m
       OUTPUT_REGEXP = /===Begin-Output===(.+?)===End-Output===/m
 
-      # Name mapping for integration services for nicer keys
+      # Name mapping for integration services for id
+      # https://social.technet.microsoft.com/Forums/de-DE/154917de-f3ca-4b1e-b3f8-23dd4b4f0f06/getvmintegrationservice-sprachabhngig?forum=powershell_de
       INTEGRATION_SERVICES_MAP = {
-        guest_service_interface: "Guest Service Interface",
-        heartbeat: "Heartbeat",
-        key_value_pair_exchange: "Key-Value Pair Exchange",
-        shutdown: "Shutdown",
-        time_synchronization: "Time Synchronization",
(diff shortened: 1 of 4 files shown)


plugins/providers/hyperv/driver.rb
```diff
@@ -10,14 +10,15 @@ class Driver
...
-        key_value_pair_exchange: "Key-Value Pair Exchange",
-        shutdown: "Shutdown",
-        time_synchronization: "Time Synchronization",
-        vss: "VSS",
+        guest_service_interface: "6C09BB55-D683-4DA0-8931-C9BF705F6480",
+        heartbeat: "84EAAE65-2F2E-45F5-9BB5-0E857DC8EB47",
+        key_value_pair_exchange: "2A34B1C2-FD73-4043-8A5B-DD2159BC743F",
+        shutdown: "9F8233AC-BE49-4C79-8EE3-E7E1985B2077",
+        time_synchronization: "2497F4DE-E9FA-4204-80E4-4B75C46419C0",
+        vss: "5CED1297-4598-4915-A5FC-AD21BB4D02A4",
```
chrisroberts, line 21: Please add a `.freeze` on the end of each string value and a `.freeze` the entire constant Hash as well.

test/unit/plugins/providers/hyperv/driver_test.rb
```diff
@@ -91,7 +91,7 @@ def generate_error(msg)
     describe "#set_vm_integration_services" do
       it "should map known integration services names automatically" do
         expect(subject).to receive(:execute) do |name, args|
-          expect(args[:Name]).to eq("Shutdown")
+          expect(args[:Id]).to eq("9F8233AC-BE49-4C79-8EE3-E7E1985B2077")
```
chrisroberts, line 94: Please use the constant hash to match within the `eq` instead of the raw value. This just makes it easier to understand when reading it (instead of just seeing a magic value) and will be automatically updated if the value ever has to be modified for some reason.

PR #11500, opened 2020-04-07
Fix #8704: Raise an exception if the guest IP ends with .1

### LINKED ISSUE #8704: mount.nfs: access denied by server while mounting
https://github.com/mitchellh/vagrant/issues/5424 is not fixed. The cause is the installation of docker on the guest. Please see my comments there.

jbonhag:
If the VirtualBox guest property `/VirtualBox/GuestInfo/Net/1/V4/IP`
returns an IP address ending in .1, raise an exception.

This addresses an issue that was revealed as an NFS error, where Vagrant
was creating an exports file with the wrong IP address. This was thought
to be caused by the presence of a docker0 interface, but it manifested
itself even without Docker installed.

This issue is difficult to reproduce, but hopefully this PR will get us
closer to the root cause.

### DIFF
--- a/plugins/providers/virtualbox/driver/version_5_0.rb
+++ b/plugins/providers/virtualbox/driver/version_5_0.rb
@@ -587,6 +587,10 @@ module VagrantPlugins
 
         def read_guest_ip(adapter_number)
           ip = read_guest_property("/VirtualBox/GuestInfo/Net/#{adapter_number}/V4/IP")
+          if ip.end_with?(".1")
+            @logger.warn("VBoxManage guest property returned: #{ip}. Result resembles IP of DHCP server. Retrying IP lookup.")
+            ip = read_guest_property("/VirtualBox/GuestInfo/Net/#{adapter_number}/V4/IP")
+          end
           if !valid_ip_address?(ip)
             raise Vagrant::Errors::VirtualBoxGuestPropertyNotFound,
               guest_property: "/VirtualBox/GuestInfo/Net/#{adapter_number}/V4/IP"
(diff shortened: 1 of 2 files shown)


plugins/providers/virtualbox/driver/version_5_0.rb
```diff
@@ -587,6 +587,10 @@ def read_guest_additions_version
 
         def read_guest_ip(adapter_number)
           ip = read_guest_property("/VirtualBox/GuestInfo/Net/#{adapter_number}/V4/IP")
+          if ip.end_with?(".1")
+            @logger.warn("VBoxManage guest property returned: #{ip}. Result resembles IP of DHCP server. Retrying IP lookup.")
+            ip = read_guest_property("/VirtualBox/GuestInfo/Net/#{adapter_number}/V4/IP")
```
briancain, line 592: Is this expected to be different than the call on line 589? :thinking:

chrisroberts, line 592: Instead of reading the value again, the log warning should note the value received and that it is being ignored. Then the `ip` value should be set to `nil`(which will result in an exception being raised when the value is validated).

jbonhag: That's sort of my hope... that the `.1` thing is a fluke and the next time you check you get the real value. I haven't been able to reproduce the issue consistently enough to validate that theory. But I think it's a better idea to just bail on `.1` per @chrisroberts's comment.

gurjeet:
Just in case a reproducible Vagrantfile is ever needed, I had developed the following for the PR #11530.

```
  config.vm.synced_folder ".", "/vagrant_data", type: "nfs"

  config.vm.provision "shell", privileged: false, inline: <<-SHELL
sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
curl -sSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
sudo add-apt-repository \
"deb [arch=amd64] https://download.docker.com/linux/ubuntu \
$(lsb_release -cs) \
stable"
sudo apt-get update
sudo apt-get install -y docker-ce
sudo docker run hello-world

# Uncomment the following line to fix the problem
#sudo systemctl disable docker
SHELL
end
```

Issue #11317, opened 2020-01-13
wrong entry in /etc/hosts file


dthpulse:
since Vagrant version 2.2.6 vagrant creates wrong entry on very first line in /etc/hosts:

> 127.0.0.1  myserver1.domain myserver1

briancain:
Hi there,

Thank you for opening an issue. However, there is not enough information here to reproduce or provide a recommendation. When you first open an issue, a template is pre-filled in the box. That template provides instructions for getting your host and guest information, the version of Vagrant, and the debug output. Without that information, it is very challenging for us to assist you.

Please supply that information either by updating this ticket or by closing this ticket and opening a new one. Thank you!

dthpulse:
Please note that the Vagrant issue tracker is in priority reserved for bug reports and enhancements. For general usage questions, please use the Vagrant mailing list:
https://groups.google.com/forum/#!forum/vagrant-up. Thank you!

**Tip:** Before submitting your issue, don't hesitate to remove the above introductory text, possible empty sections (e.g. References), and this tip.

### Vagrant version

Vagrant 2.2.6

### Host operating system
NAME="openSUSE Tumbleweed"
VERSION="20200107"

### Guest operating system
NAME="SLES"
VERSION="15-SP2"

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
    config.vm.provider :libvirt do |libvirt|
        libvirt.driver = 'kvm'
        libvirt.host = 'localhost'
        libvirt.uri = 'qemu:///system'
        libvirt.memory = '2048'
        libvirt.cpus = '1'
        libvirt.storage_pool_name = 'default'
        libvirt.memory = '4096'
        libvirt.cpus = '2'
    end
    config.vm.hostname = 'tester'
    config.vm.box = 'sle15sp1'
    config.vm.network :private_network, :ip => "192.168.122.110"
end
```

### Debug output
https://gist.github.com/dthpulse/71790e4d51c82596738197fcf73b062c

### Expected behavior
VM is created. 
There shouldn't be entry 
> 127.0.0.1       tester  tester

in /etc/hosts file


### Actual behavior
Command `head -1 /etc/hosts` will return

> 127.0.0.1       tester  tester


### Steps to reproduce
1. use Vagrantfile I provided
2. deploy VM by command `vagrant up`

briancain:
Hey there @dthpulse - Thanks for the information.

[...]

Issue #7834, opened 2016-09-27
Vagrant (re-)uses 127.0.0.1 (localhost) on Debian/Ubuntu when config.vm.hostname ist set 


dhs-rec:
When managing the hostname of a Debian or Ubuntu VM via config.vm.hostname, vagrant writes an additional line with address 127.0.0.1 into /etc/hosts, so that it looks like this afterwards:

```
127.0.0.1       myhost.virtual.box   myhost
127.0.0.1       localhost
127.0.1.1       packer-Ubuntu1604-1471511555
::1     localhost ip6-localhost ip6-loopback
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

Debian and Ubuntu use an address of 127.0.1.1 for this, so that the 127.0.0.1 entry can be left untouched. So instead of adding another line with 127.0.0.1, Vagrant should instead replace the line that starts with 127.0.1.1 on Debian or Ubuntu VMs.
### Vagrant version

% vagrant -v
Vagrant 1.8.5
### Host operating system

Windows Server 2012 R2
### Guest operating system

Ubuntu 16.04
### Expected behavior

The line starting with 127.0.1.1 should have been replaced in /etc/hosts (or added if it didn't exist).
### Actual behavior

An additional line starting with 127.0.0.1 has been added to /etc/hosts.
### Steps to reproduce
1. vagrant init <a Debian or Ubuntu box>
2. Edit Vagrantfile and add a config.vm.hostname entry
3. vagrant up
4. Check /etc/hosts on the VM

meshy:
Related to https://github.com/hashicorp/vagrant/issues/7263

chrisroberts:
Closing this as the updates which will be provided to resolve #11508 will fix this issue. Cheers!

PR #11502, opened 2020-04-08
Set hostname for suse

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11317

### DIFF
--- a/plugins/guests/suse/cap/change_host_name.rb
+++ b/plugins/guests/suse/cap/change_host_name.rb
@@ -6,16 +6,16 @@ module VagrantPlugins
           comm = machine.communicate
 
           basename = name.split(".", 2)[0]
-          if !comm.test('test "$(hostnamectl --static status)" = "#{basename}"', sudo: false)
+          if comm.test("hostnamectl --static status", sudo: true)
+            comm.sudo("hostnamectl set-hostname '#{basename}'")
+          else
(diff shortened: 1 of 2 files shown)


soapy1:
This PR changes the way that the hostname is set on SUSE hosts. Though I am not entirely sure this is the right approach to responding to https://github.com/hashicorp/vagrant/issues/11317.

The issue, as I understand it is that the user does not want their hostname on the loopback address. This seems to be an issue for some applications that one might run in a vm. eg https://support.datastax.com/hc/en-us/articles/204675669-Spark-hostname-resolving-to-loopback-address-warning-in-spark-worker-logs
There are also old vagrant issues around manipulating `/etc/hosts`
* https://github.com/hashicorp/vagrant/issues/7263
* https://github.com/hashicorp/vagrant/issues/7834

The rationale behind this PR is, vagrant should be using `hostnamectl` if possible. If that's not possible, then change the hostname more manually by adding it to `/etc/HOSTNAME`. In both cases, append the hostname to the `127.0.0.1` entry instead of prepending a new entry. 

After reading more related issues, I think this solution is also not "right" since the issue seems to arise around the hostname being associated with the loopback address.

soapy1:
superseded by issue https://github.com/hashicorp/vagrant/issues/11508

PR #11517, opened 2020-04-14
Ensure windows files get an extension

soapy1:
When using the winrm provisioner with the inline option, the default script created by Vagrant doesn't always get an extension. This will break execution on windows. For example, with the Vagrantfile:
```
  config.vm.box = 'StefanScherer/windows_10'
  config.vm.provision :shell, :inline => 'New-Item C:\testfile3.txt -ItemType file', :powershell_elevated_interactive => true
end
```

The result of running `vagrant up` is
```
    default: Running: inline PowerShell script

cmd /q /c "C:\tmp\vagrant-shell"

Stdout from the command:



Stderr from the command:

'C:\tmp\vagrant-shell' is not recognized as an internal or external command,
operable program or batch file.
```

This PR resolves this issue by setting a default extension on the uploaded file, if none is set already. The default for winrm is assumed to be `.ps1`

### DIFF
--- a/plugins/provisioners/shell/provisioner.rb
+++ b/plugins/provisioners/shell/provisioner.rb
@@ -4,7 +4,6 @@ require "tempfile"
 require "vagrant/util/downloader"
 require "vagrant/util/retryable"
 
-require "pry-byebug"
 module VagrantPlugins
   module Shell
     class Provisioner < Vagrant.plugin("2", :provisioner)
@@ -143,7 +142,7 @@ module VagrantPlugins
               remote_ext = @machine.config.winssh.shell == "cmd" ? ".bat" : ".ps1"
             end
 
-            remote_path = add_extension(upload_path(), remote_ext)
+            remote_path = add_extension(upload_path, remote_ext)
             if remote_ext == "bat"
               command = "#{env}\n cmd.exe /c \"#{remote_path}\" #{args}"
             else
--- a/plugins/provisioners/shell/provisioner.rb
+++ b/plugins/provisioners/shell/provisioner.rb
@@ -4,10 +4,13 @@ require "tempfile"
 require "vagrant/util/downloader"
 require "vagrant/util/retryable"
 
+require "pry-byebug"
 module VagrantPlugins
   module Shell
     class Provisioner < Vagrant.plugin("2", :provisioner)
       include Vagrant::Util::Retryable
+
+      DEFAULT_WINDOWS_SHELL_EXT = ".ps1".freeze
 
       def provision
         args = ""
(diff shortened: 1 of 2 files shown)


plugins/provisioners/shell/provisioner.rb
```diff
@@ -135,24 +137,22 @@ def provision_winssh(args)
...
             env = config.env.map{|k,v| comm.generate_environment_export(k, v)}.join(';')
-            shell = comm.machine_config_ssh.shell
-            if File.extname(upload_path).empty?
-              remote_ext = File.extname(path)[1..-1]
-              if !remote_ext
-                remote_ext = @machine.config.winssh.shell == "cmd" ? "bat" : "ps1"
-              end
-              upload_path << ".#{remote_ext}"
+            remote_ext = File.extname(path)
+            if remote_ext == ""
```
chrisroberts, line 141:
```ruby
if remote_ext.empty?
```

plugins/provisioners/shell/provisioner.rb
```diff
@@ -135,24 +137,22 @@ def provision_winssh(args)
...
+              remote_ext = @machine.config.winssh.shell == "cmd" ? ".bat" : ".ps1"
             end
+            remote_path = add_extension(upload_path, remote_ext)
```
chrisroberts, line 144: just overwrite `upload_path` here and then keep the name consistent below.

soapy1: The `upload_path` here refers to this function https://github.com/hashicorp/vagrant/blob/master/plugins/provisioners/shell/provisioner.rb#L41. So, setting `upload_path = add_extension(upload_path, remote_ext)` causes some strange behaviors. I think something like `remote_path = add_extension(upload_path(), remote_ext)` is maybe less ambiguous. I'm not sure what the ruby way is here?

chrisroberts: Completely valid assertion and my mistake :slightly_smiling_face:

plugins/provisioners/shell/provisioner.rb
```diff
@@ -135,24 +137,22 @@ def provision_winssh(args)
...
-              if !remote_ext
-                remote_ext = @machine.config.winssh.shell == "cmd" ? "bat" : "ps1"
-              end
-              upload_path << ".#{remote_ext}"
+            remote_ext = File.extname(path)
+            if remote_ext.empty?
+              remote_ext = @machine.config.winssh.shell == "cmd" ? ".bat" : ".ps1"
             end
+
+            remote_path = add_extension(upload_path(), remote_ext)
```
chrisroberts, line 145: The parens on `upload_path()` can be removed. Not included on method calls when no parameters are provided (it was my bad on the earlier comment).

Issue #8987, opened 2017-09-21
Automatically install the Bash completion script


alb-i986:
Would it be possible for the Debian package (and possibly the others too) to have the [Bash completion script](https://github.com/hashicorp/vagrant/blob/master/contrib/bash/completion.sh) automatically installed in the system, i.e. copied into `/etc/bash_completion.d/`?

AFAIU one needs to manually copy it him/herself, isn't it?

Thanks

References:
- #400

rfults:
Terraform includes ```-install-autocomplete``` option. Maybe that can happen with all the Hashicorp tools?

patr00n:
macos user has homebrew package `$ brew install vagrant-completion`

Jeansen:
I'd also be happy to have this one installed automatically or at least have it mentioned in the official documentation. Some automation or option would be nice, especially with respect to the official installation guide that recommends NOT to use any package provided by different Linux distributions.

PR #11520, opened 2020-04-15
Add zsh completion

soapy1:
looks like:
```
% vagrant     
box            -- manages everything related to Vagrant Cloud
destroy        -- stops and deletes all traces of the vagrant machine
global-status  -- outputs status Vagrant environments for this user
halt           -- stops the vagrant machine
help           -- shows the help for a subcommand
init           -- initializes a new Vagrant environment by creating a Vagrantfile
package        -- packages a running vagrant environment into a box
plugin         -- manages plugins: install, uninstall, update, etc.
port           -- displays information about guest port mappings
powershell     -- connects to machine via powershell remoting
provision      -- provisions the vagrant machine
push           -- deploys code in this environment to a configured destination
rdp            -- connects to machine via RDP
reload         -- restarts vagrant machine, loads new Vagrantfile configuration
resume         -- resume a suspended vagrant machine
snapshot       -- manages snapshots; saving, restoring, etc.
ssh            -- connects to machine via SSH
ssh-config     -- outputs OpenSSH valid configuration to connect to the machine
status         -- outputs status of the vagrant machine
suspend        -- suspends the machine
up             -- starts and provisions the vagrant environment
upload         -- upload to machine via communicator
validate       -- validates the Vagrantfile
version        -- prints current and latest Vagrant version
winrm          -- executes commands on a machine via WinRM
winrm-config   -- outputs WinRM configuration to connect to the machine
```

PR #11523, opened 2020-04-16
Install zsh autocomplete

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/8987
relies on https://github.com/hashicorp/vagrant/pull/11520 (or some file in `contrib/zsh`)

### DIFF
--- a/lib/vagrant/cli.rb
+++ b/lib/vagrant/cli.rb
@@ -2,6 +2,7 @@ require 'log4r'
 require 'optparse'
 
 require 'vagrant/util/experimental'
+require 'vagrant/util/install_cli_autocomplete'
 
 module Vagrant
   # Manages the command line interface to Vagrant.
@@ -27,6 +28,12 @@ module Vagrant
         # Help is next in short-circuiting everything. Print
         # the help and exit.
         help
+        return 0
+      end
+
+      if @main_args.include?("-install_autocomplete")
+        # Install zsh autocomplete script
+        Vagrant::Util::InstallCLIAutocomplete.install
         return 0
       end
 
--- a/lib/vagrant/cli.rb
+++ b/lib/vagrant/cli.rb
@@ -2,7 +2,6 @@ require 'log4r'
 require 'optparse'
 
 require 'vagrant/util/experimental'
-require 'vagrant/util/install_cli_autocomplete'
 
 module Vagrant
   # Manages the command line interface to Vagrant.
@@ -28,17 +27,6 @@ module Vagrant
         # Help is next in short-circuiting everything. Print
         # the help and exit.
         help
-        return 0
-      end
-
-      if @main_args.include?("-install-autocomplete")
-        # Install autocomplete scripts
-        written_paths = Vagrant::Util::InstallCLIAutocomplete.install
-        if written_paths && written_paths.length > 0
-          @env.ui.info(I18n.t("vagrant.autocomplete.installed", paths: written_paths.join("\n- ")))
-        else
-          @env.ui.info(I18n.t("vagrant.autocomplete.not_installed"))
-        end
         return 0
       end
 
--- a/lib/vagrant/cli.rb
+++ b/lib/vagrant/cli.rb
@@ -34,13 +34,10 @@ module Vagrant
       if @main_args.include?("-install-autocomplete")
         # Install autocomplete scripts
         written_paths = Vagrant::Util::InstallCLIAutocomplete.install
-        if written_paths.length > 0
-          @env.ui.info("Autocomplete installed at paths:")
-          written_paths.each do |p|
-            @env.ui.info("- #{p}")
-          end
+        if written_paths && written_paths.length > 0
+          @env.ui.info(I18n.t("vagrant.autocomplete.installed", paths: written_paths.join("\n- ")))
         else
-          @env.ui.info("Autocomplete not installed")
+          @env.ui.info(I18n.t("vagrant.autocomplete.not_installed"))
         end
         return 0
       end
--- a/lib/vagrant/cli.rb
+++ b/lib/vagrant/cli.rb
@@ -31,9 +31,17 @@ module Vagrant
         return 0
       end
 
-      if @main_args.include?("-install_autocomplete")
-        # Install zsh autocomplete script
-        Vagrant::Util::InstallCLIAutocomplete.install
+      if @main_args.include?("-install-autocomplete")
+        # Install autocomplete scripts
+        written_paths = Vagrant::Util::InstallCLIAutocomplete.install
+        if written_paths.length > 0
+          @env.ui.info("Autocomplete installed at paths:")
+          written_paths.each do |p|
+            @env.ui.info("- #{p}")
+          end
+        else
+          @env.ui.info("Autocomplete not installed")
+        end
         return 0
       end
 
--- a/lib/vagrant/util/install_cli_autocomplete.rb
+++ b/lib/vagrant/util/install_cli_autocomplete.rb
@@ -100,13 +100,16 @@ module Vagrant
         "bash" => Vagrant::Util::InstallBashShellConfig.new()
       }
 
-      def self.install
+      def self.install(shells=[])
+        shells = shells.length == 0 ? SUPPORTED_SHELLS.keys() : shells
         home = Dir.home
         written_paths = []
         SUPPORTED_SHELLS.each do |k, shell|
-          p = shell.install(home)
-          if p
-            written_paths.push(p)
+          if shells.include?(k)
+            p = shell.install(home)
+            if p
+              written_paths.push(p)
+            end
           end
         end
         return written_paths
--- a/lib/vagrant/util/install_cli_autocomplete.rb
+++ b/lib/vagrant/util/install_cli_autocomplete.rb
@@ -1,14 +1,13 @@
 module Vagrant
   module Util
-    class ZSHShell
-      PREPEND = "# >>>> Vagrant zsh completion (start)".freeze
-      STRING_INSERT = """fpath=(#{File.join(Vagrant.source_root, "contrib", "zsh")} $fpath)\ncompinit""".freeze
-      APPEND = "# <<<<  Vagrant zsh completion (end)".freeze
-
-      CONFIG_PATHS = [".zshrc"].freeze
+    class Shell
+      PREPEND = "".freeze
+      STRING_INSERT = "".freeze
+      APPEND = "".freeze
+      CONFIG_PATHS = [""].freeze
 
       def self.shell_installed(home)
-        CONFIG_PATHS.each do |path|
+        self::CONFIG_PATHS.each do |path|
           config_file = File.join(home, path)
           if File.exists?(config_file)
             return config_file
@@ -19,7 +18,7 @@ module Vagrant
 
       def self.is_installed(path)
         File.foreach(path) do |line|
-          if line.include?(PREPEND)
+          if line.include?(self::PREPEND)
             return true
           end
         end
@@ -31,15 +30,22 @@ module Vagrant
         if path && !is_installed(path)
           File.open(path, "a") do |f|
             f.write("\n")
-            f.write(PREPEND)
+            f.write(self::PREPEND)
             f.write("\n")
-            f.write(STRING_INSERT)
+            f.write(self::STRING_INSERT)
             f.write("\n")
-            f.write(APPEND)
+            f.write(self::APPEND)
             f.write("\n")
           end
         end
       end
+    end
+
+    class ZSHShell < Shell
+      PREPEND = "# >>>> Vagrant zsh completion (start)".freeze
+      STRING_INSERT = """fpath=(#{File.join(Vagrant.source_root, "contrib", "zsh")} $fpath)\ncompinit""".freeze
+      APPEND = "# <<<<  Vagrant zsh completion (end)".freeze
+      CONFIG_PATHS = [".zshrc"].freeze
     end
 
     # Install autocomplete script for supported shells
--- a/lib/vagrant/util/install_cli_autocomplete.rb
+++ b/lib/vagrant/util/install_cli_autocomplete.rb
@@ -2,40 +2,52 @@ module Vagrant
   module Util
     # Generic installation of content to shell config file
     class InstallShellConfig
-      PREPEND = "".freeze
-      STRING_INSERT = "".freeze
-      APPEND = "".freeze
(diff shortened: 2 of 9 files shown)


lib/vagrant/util/install_cli_autocomplete.rb
```diff
@@ -0,0 +1,65 @@
+module Vagrant
+  module Util
+    class Shell
+      PREPEND = "".freeze
+      STRING_INSERT = "".freeze
+      APPEND = "".freeze
+      CONFIG_PATHS = [""].freeze
+
+      def self.shell_installed(home)
+        self::CONFIG_PATHS.each do |path|
```
soapy1, line 10: I assume there is a better way to this `self::<variable>` bit. This is a gap in my ruby-foo, would appreciate a nudge in the right direction

lib/vagrant/cli.rb
```diff
@@ -30,6 +31,20 @@ def execute
...
+        written_paths = Vagrant::Util::InstallCLIAutocomplete.install
+        if written_paths.length > 0
+          @env.ui.info("Autocomplete installed at paths:")
+          written_paths.each do |p|
+            @env.ui.info("- #{p}")
+          end
+        else
+          @env.ui.info("Autocomplete not installed")
+        end
+        return 0
```
soapy1, line 45: output looks like:
```
$ dvag -install_autocomplete               

Autocomplete installed at paths:
- /redacted/.zshrc
- /redacted/.bash_profile
```

lib/vagrant/cli.rb
```diff
@@ -30,6 +31,20 @@ def execute
...
+      if @main_args.include?("-install-autocomplete")
+        # Install autocomplete scripts
+        written_paths = Vagrant::Util::InstallCLIAutocomplete.install
+        if written_paths.length > 0
+          @env.ui.info("Autocomplete installed at paths:")
+          written_paths.each do |p|
+            @env.ui.info("- #{p}")
+          end
+        else
+          @env.ui.info("Autocomplete not installed")
```
briancain, line 43: We should use the `I18n` library here to abstract out the strings into the translations file.

lib/vagrant/cli.rb
```diff
@@ -30,6 +31,20 @@ def execute
         return 0
       end
 
+      if @main_args.include?("-install-autocomplete")
+        # Install autocomplete scripts
+        written_paths = Vagrant::Util::InstallCLIAutocomplete.install
+        if written_paths.length > 0
+          @env.ui.info("Autocomplete installed at paths:")
+          written_paths.each do |p|
+            @env.ui.info("- #{p}")
```
briancain, line 40: Instead of printing each line by itself, we should build up the string here and print it all at once afterwords. Otherwise the debug output or machine readable output will end up with a bunch of extra stuff in between each line from invoking `@env.ui.info` for each path.

lib/vagrant/util/install_cli_autocomplete.rb
briancain: Maybe instead of these global constants you could have an `initialize` class and set them as class variables instead? Then you could simply access them with `@config_paths`, etc.

lib/vagrant/util/install_cli_autocomplete.rb
```diff
@@ -0,0 +1,82 @@
+module Vagrant
+  module Util
+    # Generic installation of content to shell config file
+    class InstallShellConfig
+      PREPEND = "".freeze
+      STRING_INSERT = "".freeze
+      APPEND = "".freeze
+      CONFIG_PATHS = [""].freeze
+
+      def self.shell_installed(home)
```
briancain, line 10: These should all have method docs for the parameters expected and the return type expected

lib/vagrant/cli.rb
```diff
@@ -30,6 +31,20 @@ def execute
         return 0
       end
 
+      if @main_args.include?("-install-autocomplete")
+        # Install autocomplete scripts
+        written_paths = Vagrant::Util::InstallCLIAutocomplete.install
+        if written_paths.length > 0
```
briancain, line 37: Hmm, it seems like it's possible written_paths could return nil? I think this line would fail if so with no method `length` for nil class.

lib/vagrant/util/install_cli_autocomplete.rb
briancain: And if we can get an initialize method, it would also be good to have a `logger` with `log4r` with this class. Then we can add some more extra context in the log around setting these up which will help with bug reports.

lib/vagrant/cli.rb
```diff
@@ -30,6 +31,20 @@ def execute
...
+          @env.ui.info("Autocomplete installed at paths:")
```
briancain, line 38: Same note here about using the `I18n` library to abstract out these strings

lib/vagrant/cli.rb
```diff
@@ -30,6 +31,20 @@ def execute
         return 0
       end
 
+      if @main_args.include?("-install-autocomplete")
```
briancain, line 34: Oops...just thought of this. We should guard against this flag being valid for anything other than something like `vagrant --install-autocomplete` Otherwise this will work if you do stuff like `vagrant up --install-autocomplete`?

Also this should start with at `--` instead of a single `-`.

test/unit/vagrant/util/install_cli_autocomplete_test.rb
```diff
@@ -0,0 +1,59 @@
...
+  let(:home) { "#{Dir.tmpdir}/not-home" }
+  let(:target_file) { "#{home}/.zshrc" }
+
+  subject { described_class }
+
+  before do
+    Dir.mkdir(home)
+    File.open(target_file, "w") do |f|
+      f.write("some content")
+    end
```
briancain, line 17: One thing that I think we can do here is mock up the File and Dir classes to "pretend" to create these files....just so we aren't actually creating files on the CI system or users computers while running the unit tests.

lib/vagrant/cli.rb
briancain: Also, based in what was mentioned in slack, it seems like this would be better implemented as its own Command class, rather than here.

plugins/commands/autocomplete/command/root.rb
```diff
@@ -0,0 +1,64 @@
...
+          return help if !command_class || !@sub_command
+          @logger.debug("Invoking command class: #{command_class} #{@sub_args.inspect}")
+
+          # Initialize and execute the command class
+          command_class.new(@sub_args, @env).execute
+        end
+
+        def help
+          opts = OptionParser.new do |opts|
+            opts.banner = "Usage: vagrant autoinstall <subcommand>"
```
briancain, line 42: Oops, I think this should read `Usage: autocomplete <subcommand>`

```suggestion
            opts.banner = "Usage: vagrant autocomplete <subcommand>"
```

plugins/commands/autocomplete/command/root.rb
```diff
@@ -0,0 +1,64 @@
...
+            # out as well.
+            keys = []
+            @subcommands.each { |key, value| keys << key.to_s }
+
+            keys.sort.each do |key|
+              opts.separator "     #{key}"
+            end
+
+            opts.separator ""
+            opts.separator "For help on any individual subcommand run `vagrant autoinstall <subcommand> -h`"
```
briancain, line 56:
```suggestion
            opts.separator "For help on any individual subcommand run `vagrant autocomplete <subcommand> -h`"
```

plugins/commands/autocomplete/command/root.rb
```diff
@@ -0,0 +1,64 @@
+require "optparse"
+require 'vagrant/util/install_cli_autocomplete'
+
+module VagrantPlugins
+  module CommandAutocomplete
+    module Command
+      class Root < Vagrant.plugin("2", :command)
+        def self.synopsis
+          "manages autocomplete: installation, removal, etc."
```
briancain, line 9: Maybe we should only say "installation on host" or something until we get a remove sub command :thinking: Thoughts?

plugins/commands/autocomplete/command/install.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module VagrantPlugins
+  module CommandAutocomplete
+    module Command
+      class Install < Vagrant.plugin("2", :command)
+        def execute
+          options = {}
+
+          opts = OptionParser.new do |o|
+            o.banner = "Usage: vagrant autocomplete install [-h]"
```
briancain, line 13: Is it possible for us to have flags based on which shell to install? Not sure how hard that would be to add.

soapy1: I think it's possible. What do you envision being different between shells?
I was thinking options would be like: `--append` where it appends the blrub, `--prepend` where it prepends the blurb, `--shell` that allows you to choose which shell you want to install.

briancain: I'm mostly thinking people might want an option to only install one of the available shells, instead of every possible one. So maybe it'd be like `autocomplete install --bash` for example.

soapy1: Ya, I like that. Should users be allowed to specify multiple shells like: `autocomplete install --bash --zsh`
Then the output of `autocomplete install [-h]` looks something like 
```
Usage: vagrant autocomplete install [-h]

Options:

    -b, --bash                       Install bash autocomplete
    -z, --zsh                        Install zsh autocomplete
        --[no-]color                 Enable or disable color output
        --machine-readable           Enable machine readable output
    -v, --version                    Display Vagrant version
        --debug                      Enable debug output
        --timestamp                  Enable timestamps on log output
        --debug-timestamp            Enable debug output with timestamps
        --no-tty                     Enable non-interactive output
```

Another option in the same line of thinking is something like `autocomplete install <shell>`
Then the output of `autocomplete install [-h]` looks something like :
```
Usage: vagrant autocomplete install [-h] [shell name]

Available shells: zsh bash

Options:

        --[no-]color                 Enable or disable color output
        --machine-readable           Enable machine readable output
    -v, --version                    Display Vagrant version
        --debug                      Enable debug output
        --timestamp                  Enable timestamps on log output
        --debug-timestamp            Enable debug output with timestamps
        --no-tty                     Enable non-interactive output
```
I like (2) since it doesn't require developers to change the cli when adding more shells, and it more closely resembles other commands in vagrant like `box add <name>` Though `vagrant autocomplete install zsh` does seem pretty clunky. Any thoughts on this?

briancain: Yeah! I think it makes sense to keep it a flag, unless each shell could have additional configuration options? :thinking: Otherwise I think what you have on that first output with the shells being flags looks good to me! :+1:

lib/vagrant/util/install_cli_autocomplete.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    # Install autocomplete script for supported shells
+    class InstallCLIAutocomplete
+      SUPPORTED_SHELLS = {
+        "zsh" => Vagrant::Util::InstallZSHShellConfig.new(),
+        "bash" => Vagrant::Util::InstallBashShellConfig.new()
+      }
+
+      def self.install(shells=[])
+        shells = shells.length == 0 ? SUPPORTED_SHELLS.keys() : shells
```
chrisroberts, line 104:
```ruby
shells = SUPPORTED_SHELLS.keys() if shells.empty?
```

lib/vagrant/util/install_cli_autocomplete.rb
```diff
@@ -0,0 +1,119 @@
...
+        written_paths = []
+        SUPPORTED_SHELLS.each do |k, shell|
+          if shells.include?(k)
+            p = shell.install(home)
+            if p
+              written_paths.push(p)
+            end
+          end
+        end
+        return written_paths
```
chrisroberts, line 115: This can be compacted down to:

```ruby
shells.map do |shell|
  next if !SUPPORTED_SHELLS[shell]
  SUPPORTED_SHELLS[shell].install(home)
end.compact
```

I'm also wondering if instead of silently ignoring requested shells that are not supported if we should be raising an `ArgumentError`.

plugins/commands/autocomplete/command/root.rb
```diff
@@ -0,0 +1,64 @@
...
+        def help
+          opts = OptionParser.new do |opts|
+            opts.banner = "Usage: vagrant autocomplete <subcommand>"
+            opts.separator ""
+            opts.separator "Available subcommands:"
+
+            # Add the available subcommands as separators in order to print them
+            # out as well.
+            keys = []
+            @subcommands.each { |key, value| keys << key.to_s }
```
chrisroberts, line 49:
```ruby
keys = @subcommands.keys.map(&:to_s)
```

lib/vagrant/util/install_cli_autocomplete.rb
```diff
@@ -0,0 +1,119 @@
...
+            f.write("\n")
+          end
+        end
+        return path
+      end
+    end
+
+    # Install autocomplete script to zsh config located as .zshrc
+    class InstallZSHShellConfig < InstallShellConfig
+      def initialize()
```
chrisroberts, line 76: For methods with no arguments, the parenthesis can be left off

lib/vagrant/util/install_cli_autocomplete.rb
```diff
@@ -0,0 +1,119 @@
...
+        prepend_string = "# >>>> Vagrant zsh completion (start)".freeze
```
chrisroberts, line 77: These `prepend_string` and `append_string` would be nice to have defined as constants at the top, without the specific shell included. That way our file tagging will just be a generic value.

plugins/commands/autocomplete/command/install.rb
```diff
@@ -0,0 +1,49 @@
...
+
+require 'vagrant/util/install_cli_autocomplete'
+
+module VagrantPlugins
+  module CommandAutocomplete
+    module Command
+      class Install < Vagrant.plugin("2", :command)
+        def execute
+          options = {
+            :shells => []
```
chrisroberts, line 11:
```ruby
shells: []
```

Just for consistency as that is style that has been most widely adopted throughout the codebase.

lib/vagrant/util/install_cli_autocomplete.rb
soapy1: I think that sounds good!

soapy1:
re-based to fix docs merge conflicts. Read to merge once https://github.com/hashicorp/vagrant/pull/11520 is merged

PR #11530, opened 2020-04-19
Use VBox VM's MAC address to find the IP address for NFS mounts

gurjeet:
The network adapter numbers used by Vagrant is not a reliable method of
finding guest-side IP addresses. This is because any virtual network
interface created inside the VM by the guest will affect the order of
in which the network adapters are listed.

We now use a network adapter's MAC address to uniquely identify it,
since it's guaranteed to not change after VM starts.

Following is the error that prompted troubleshooting and this fix:

    NFS requires a host-only network to be created.
    Please add a host-only network to the machine (with either DHCP or a
    static IP) for NFS to work.

This prevented the shared-folders feature to stop mounting my
host-directories.

I frequently noticed this error in my development VBox VMs, but not very
reliably. Restarting the VM about 5 times would make the problem go
away.

After significant troubleshooting I noticed this was caused by Docker
daemon installed in the VM. Disabling Docker's auto-start-on-boot would
resolve this problem for good.

The following is the minimal Vagrantfile that reproduces the problem.
When trying to reproduce this problem, remember that you _may_ not see
the error message once in a while since this error is sensitive to when
Docker creates its virtual NIC.

      config.vm.synced_folder ".", "/vagrant_data", type: "nfs"

      config.vm.provision "shell", privileged: false, inline: <<-SHELL
    sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
    curl -sSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
    sudo add-apt-repository \
    "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
    $(lsb_release -cs) \
    stable"
    sudo apt-get update
    sudo apt-get install -y docker-ce
    sudo docker run hello-world

    # Uncomment the following line to fix the problem
    #sudo systemctl disable docker
    SHELL
    end

I discovered this bug early last year on VBox version 5.0, and developed
a fix and have been using the fix for a while. I have now tested this
fix on versions 5.2 and 6.0 as well.

### DIFF
--- a/plugins/providers/virtualbox/action/prepare_nfs_settings.rb
+++ b/plugins/providers/virtualbox/action/prepare_nfs_settings.rb
@@ -37,18 +37,20 @@ module VagrantPlugins
         #
         # The ! indicates that this method modifies its argument.
         def add_ips_to_env!(env)
-          adapter, host_ip = find_host_only_adapter
-          machine_ip       = read_static_machine_ips
+          # Find the MAC address of guest-side NIC associated with the host-only
+          # adapter of the guest
+          mac_address, host_ip = find_host_only_mac_address
+          machine_ip           = read_static_machine_ips
 
           if !machine_ip
             # No static IP, attempt to use the dynamic IP.
-            machine_ip = read_dynamic_machine_ip(adapter)
+            machine_ip = read_dynamic_machine_ip(mac_address)
           else
             # We have static IPs, also attempt to read any dynamic IPs.
             # If there is no dynamic IP on the adapter, it doesn't matter. We
             # already have a static IP.
             begin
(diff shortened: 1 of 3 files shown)


chrisroberts:
Hi @gurjeet. Thanks so much for creating this PR! We appreciate you taking the time to do this. The errors encountered when detecting the IP address for NFS mounts has been fixed via #11500 in a slightly different way.

This PR itself looked pretty good, but was missing test coverage for the new behaviors introduced as well as breaking existing tests. Including those in the future will make it much easier for us to verify and ultimately accept PRs.

Again, thanks so much for creating and submitting this PR.

Cheers!

gurjeet:
I can confirm that #11500 does fix the issue when using the Vagrantfile above.

PR #11538, opened 2020-04-22
Shell provisioner: Change double-quoted values of environment variables to single-quotes

simplytech:
Please refer to: https://discuss.hashicorp.com/t/shell-provisioner-how-to-pass-env-var-containing-dollar-sign/7924/3


soapy1:
Thanks again of the PR! After exploring this some more it seems this change will be pretty destructive for some users who may be relying on the expansion of environment variables. You could get around your variable expansion issue by escaping `$`. For example `DOCKERHUB_PASSWORD='Pas\$word'`.

Issue #11539, opened 2020-04-22
WinRM communicator: defaults values are wrong


nqb:
### Vagrant version
2.2.7

### Host operating system
Debian 10 (Buster)

### Guest operating system
Windows 2016 (jborean93/WindowsServer2016 Vagrant box)

### Expected behavior
According to [Vagrant's documentation (config.winrm)](https://github.com/hashicorp/vagrant/blob/master/website/source/docs/vagrantfile/winrm_settings.html.md), default values for following parameters should be:

- `config.winrm.ssl_peer_verification`: not mentioned
- `config.winrm.timeout`: 60 seconds
- `config.winrm.retry_limit`: 3
- `config.winrm.retry_delay`: 2
- `config.winrm.codepage`: 65001

### Actual behavior
Using box mentioned above, I took a look at https://github.com/hashicorp/vagrant/blob/21822e6c46ac3220f99d2180223019d991d22287/plugins/communicators/winrm/config.rb#L34-L48

and noticed that default values are completely different (specially `config.winrm.timeout` with a default value of 1800 seconds in place of 60 seconds).

There is also a missing parameter: `config.winrm.execution_time_limit` (see https://github.com/hashicorp/vagrant/pull/6213)

How do you want to fix that bug ? Should we need to update documentation or defaults values of WinRM communicator ?

PR #11540, opened 2020-04-22
Update winrm config docs

soapy1:
Fixes https://github.com/hashicorp/vagrant/issues/11539

### DIFF
--- a/website/source/docs/vagrantfile/winrm_settings.html.md
+++ b/website/source/docs/vagrantfile/winrm_settings.html.md
@@ -33,7 +33,8 @@ what most public boxes are made as.
 By default this is empty, because the provider usually figures this out for
 you.
 
-* `config.winrm.port` (integer) - The WinRM port to connect to, by default 5985.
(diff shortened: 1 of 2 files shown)


website/source/docs/vagrantfile/winrm_settings.html.md
```diff
@@ -51,11 +51,13 @@ the Windows machine must be configured appropriately.
     **Note:** It is strongly recommended that you only use basic authentication for
     debugging purposes. Credentials will be transferred in plain text.
 
+* `config.winrm.execution_time_limit` (string) - The amount of time that is allowed to complete task. This defaults to "PT2H", that is 2 hours.
```
jbonhag, line 54: Would you mind wrapping this line? It doesn't look like we are super strict about this, but most lines are wrapped to between 80-90 characters.

nqb:
Hello @soapy1,

Could you also add:
- when SSL is enabled, port will be 5986 in place of 5985.
- default value of `config.winrm.ssl_peer_verification`

and fix default value of `config.winrm.codepage` ?

Thanks.

PR #11541, opened 2020-04-22
[Feature] Hyper-V Virtual Hard Disk Management

### LINKED ISSUE #8648: Feature request: attach secondary disk on Hyper-V

briancain:
This pull request builds off of the VirtualBox provider and introduces a new enhancement to the Hyper-V provider. It leverages the generic disk config implementation and adds new driver methods to create, resize, attach, and remove disks from Hyper-V guests.

Fixes #8648


plugins/providers/hyperv/cap/configure_disks.rb
```diff
@@ -0,0 +1,235 @@
...
+        end
+
+        # Creates and attaches a disk to a machine
+        #
+        # @param [Vagrant::Machine] machine
+        # @param [Kernel_V2::VagrantConfigDisk] disk_config
+        def self.create_disk(machine, disk_config)
+          machine.ui.detail(I18n.t("vagrant.cap.configure_disks.create_disk", name: disk_config.name))
+          disk_provider_config = {}
+          disk_provider_config = disk_config.provider_config[:hyperv] if disk_config.provider_config
```
soapy1, line 147: I think these two lines can be condensed to 
```suggestion
          disk_provider_config = disk_config.provider_config ? disk_config.provider_config[:hyperv] : {}

```

plugins/providers/hyperv/cap/configure_disks.rb
```diff
@@ -0,0 +1,235 @@
...
+              if k.is_a? Integer
+                # Assume it is bytes
+                bytes = disk_provider_config[k]
+              elsif k == :BlockSizeBytes
+                bytes = Vagrant::Util::Numeric.string_to_bytes(disk_provider_config[k])
+              elsif k == :LogicalSectorSizeBytes || k == :PhysicalSectorSizeBytes
+                # Logical and Physical can only be these exact values, so converting it
+                # won't work
+                case disk_provider_config[k]
+                  when "4096MB"
```
soapy1, line 197: I'm not super familiar with hyper-v, are these the only valid options? Will more ever be added?

briancain: It is totally possible that more could be added....but yeah these parameters are literally "only" accepting these two values for `PhysicalSectorSizeBytes` and `LogicalSectorSizeBytes` right now, and any other value ends up being an exception: 

> Accepted values: | 512, 4096

https://docs.microsoft.com/en-us/powershell/module/hyper-v/new-vhd?view=win10-ps

Which is why I didn't use the Numeric class here to convert them...they won't end up "exactly" those values so it ends up with a powershell exception.

This entire method is unfortunately a hack to get Vagrant to properly pass along these parameters into the New-VHD powershell command :cry: They're technically expecting a `UInt32`, but they expect units to go along with it, but it also can't be a string like `"129MB"`..it has to be a literal like `-PhysicalSectorSizeBytes 512mb`.

briancain: Now that I re-read the docs, it looks like maybe we could just delete these switch statements, and expect people to just set it to 512 or 4096. If they don't set it properly, they will see the error from Hyper-V :shrug: Then the only thing we should be converting is BlockSizeBytes.

plugins/providers/hyperv/cap/configure_disks.rb
```diff
@@ -0,0 +1,235 @@
...
+
+            current_disk = all_disks.select { |d| d["ControllerLocation"] == 0 && d["ControllerNumber"] == 0 }.first
+
+            # Need to get actual disk info to obtain UUID instead of what's returned
+            #
+            # This is not required for newly created disks, as its metadata is
+            # set when creating and attaching the disk. This is only for the primary
+            # disk, since it already exists.
+            real_disk_info = machine.provider.driver.get_disk(current_disk["Path"])
+            current_disk = real_disk_info
```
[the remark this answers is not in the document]
briancain: Yep :+1: I had some extra code in between these two lines which didn't make the final cut! So that's what happened here

plugins/kernel_v2/config/disk.rb
```diff
@@ -107,21 +113,20 @@ def finalize!
...
-        @disk_ext = "vdi" if @disk_ext == UNSET_VALUE
-
         if @primary == UNSET_VALUE
           @primary = false
         end
 
-        if @name == UNSET_VALUE
+        if @name.is_a?(String) && @name.match(FILE_CHAR_REGEX)
+            @logger.warn("Vagrant will remove detected invalid characters in '#{@name}' and convert the disk name into something usable for a file")
+            @name.gsub!(FILE_CHAR_REGEX, "")
```
chrisroberts, line 122: When invalid characters are encountered, I think that we would be better to replace them rather than remove them (like with `"_"`). This will prevent things like empty names being generated.

plugins/kernel_v2/config/disk.rb
```diff
@@ -135,7 +140,18 @@ def validate(machine)
                            types: DEFAULT_DISK_TYPES.join(', '))
         end
 
-        if @disk_ext
+        if @disk_ext == UNSET_VALUE
+          # Work around to finalize disk_ext with a valid default per-provider
```
chrisroberts, line 144: Instead of putting provider logic here, it would be better to just have a default defined (in the case here `"vdi"`) and have a provider capability for `default_disk_ext` be checked to determine an acceptable default.

plugins/providers/hyperv/cap/configure_disks.rb
```diff
@@ -0,0 +1,210 @@
...
+        # Creates and attaches a disk to a machine
+        #
+        # @param [Vagrant::Machine] machine
+        # @param [Kernel_V2::VagrantConfigDisk] disk_config
+        def self.create_disk(machine, disk_config)
+          machine.ui.detail(I18n.t("vagrant.cap.configure_disks.create_disk", name: disk_config.name))
+          disk_provider_config = {}
+          disk_provider_config = disk_config.provider_config[:hyperv] if disk_config.provider_config
+
+          if disk_provider_config
```
chrisroberts, line 148: This conditional isn't needed as is since `disk_provider_config` will always be a truthy value

plugins/providers/hyperv/cap/configure_disks.rb
```diff
@@ -0,0 +1,210 @@
...
+        end
+
+        # Converts any "shortcut" options such as "123MB" into its byte form. This
+        # is due to what parameter type is expected when calling the `New-VHD`
+        # powershell command
+        #
+        # @param [Hash] disk_provider_config
+        # @return [Hash] disk_provider_config
+        def self.convert_size_vars!(disk_provider_config)
+          if disk_provider_config.keys.include?(:BlockSizeBytes)
```
chrisroberts, line 183: `if disk_provider_config.key?(:BlockSizeBytes)`

plugins/providers/hyperv/driver.rb
```diff
@@ -218,6 +218,68 @@ def set_name(vmname)
...
+      def attach_disk(disk_file_path,  **opts)
+        execute(:attach_disk_drive, VmId: @vm_id, Path: disk_file_path, ControllerType: opts[:ControllerType],
+                ControllerNumber: opts[:ControllerNumber], ControllerLocation: opts[:ControllerLocation])
+      end
+
+      # @param [String] path
+      # @param [Int] size_bytes
+      # @param [Hash] opts
+      def create_disk(path, size_bytes, **opts)
+        # ensure size_bytes is a uint64
```
chrisroberts, line 238: Is a type check required here? Or if this is a note on why it's a separate parameter, maybe relocate it to the doc block above with a little context.

plugins/providers/hyperv/driver.rb
```diff
@@ -218,6 +218,68 @@ def set_name(vmname)
...
+      # @param [String] controller_location
+      # @param [Hash] opts
+      def attach_disk(disk_file_path,  **opts)
+        execute(:attach_disk_drive, VmId: @vm_id, Path: disk_file_path, ControllerType: opts[:ControllerType],
+                ControllerNumber: opts[:ControllerNumber], ControllerLocation: opts[:ControllerLocation])
+      end
+
+      # @param [String] path
+      # @param [Int] size_bytes
+      # @param [Hash] opts
```
chrisroberts, line 236: Add `@option` entries for the `opts` hash on what options are supported.

plugins/providers/hyperv/driver.rb
```diff
@@ -218,6 +218,68 @@ def set_name(vmname)
...
+      def resize_disk(disk_file_path, size_bytes, **opts)
+        execute(:resize_disk_drive, VmId: @vm_id, DiskFilePath: disk_file_path,
+                DiskSize: size_bytes)
+      end
+
+      ########
+      ########
+      ########
```
chrisroberts, line 280: Remove extra comments

plugins/providers/hyperv/driver.rb
```diff
@@ -227,6 +289,7 @@ def execute_powershell(path, options, &block)
         options = options || {}
         ps_options = []
         options.each do |key, value|
+          next if value == '' || value.nil?
```
chrisroberts, line 292: `next if !value || value.to_s.empty?`

plugins/providers/hyperv/driver.rb
briancain: This was mostly a note to myself that I missed when cleaning up the PR.

plugins/providers/hyperv/cap/configure_disks.rb
```diff
@@ -0,0 +1,200 @@
...
+        # @param [Array] all_disks - A list of all currently defined disks in VirtualBox
+        # @return [Hash] current_disk - Returns the current disk. Returns nil if it doesn't exist
+        def self.get_current_disk(machine, disk, all_disks)
+          current_disk = nil
+          if disk.primary
+            # Ensure we grab the proper primary disk
+            # We can't rely on the order of `all_disks`, as they will not
+            # always come in port order, but primary should always be Location 0 Number 0.
+
+            current_disk = all_disks.select { |d| d["ControllerLocation"] == 0 && d["ControllerNumber"] == 0 }.first
```
jbonhag, line 55: You can replace the `select...first` expression with a `detect`

plugins/providers/hyperv/cap/configure_disks.rb
```diff
@@ -0,0 +1,200 @@
...
+          if !disk_provider_config.empty?
+            disk_provider_config = convert_size_vars!(disk_provider_config)
+          end
+
+          # Get the machines data dir, that will now be the path for the new disk
+          guest_disk_folder = machine.data_dir.join("Virtual Hard Disks")
+
+          if disk_config.file
+            disk_file = disk_config.file
+            LOGGER.info("Disk already defiend by user at '#{disk_file}'. Using this disk instead of creating a new one...")
```
jbonhag, line 147: Tiny typo: "defined"

plugins/providers/hyperv/driver.rb
```diff
@@ -218,6 +218,76 @@ def set_name(vmname)
...
+      # @param [String] disk_file_path
+      def dismount_disk(disk_file_path)
+        execute(:dismount_vhd, DiskFilePath: disk_file_path)
+      end
+
+      # @param [String] disk_file_path
+      def get_disk(disk_file_path)
+        execute(:get_vhd, DiskFilePath: disk_file_path)
+      end
+
```
jbonhag, line 264: Could use an RDoc comment here

plugins/providers/hyperv/cap/configure_disks.rb
briancain: I think that might make things more complicated, as `all_disks` is an array of hashes. I'd have to map a `detect` on each hash in the array, right? `select` will apply the enumerator block to each hash and return the result. The `.first` is kinda gross, but there should only be 1 disk that matches (and select always returns the results in an array).

briancain: Hmm, it's not working like I expected it would in pry, so I'll bring up a vm and try it.

briancain: Alright, switched it over to detect. Dunno why it was working differently in a pry session but whatever! haha

Issue #11546, opened 2020-04-23
box_url not using system ssh config for settings


peledies:
Im trying to distribute a packaged base box and I want to use one of my existing servers that global developers have access to with a pem file.

I am aware of vagrant cloud but that is not an option in this case.

It looks like the packaged curl command does indeed support `scp` but for some reason it is not respecting the values I have in my `~/.ssh/config` and it's not providing the identity file and username when vagrant up is ran.

I can confirm that the following command does work when specifying the identity file with the `--key` option while using the version of `curl` that is packaged with vagrant.

```bash
/opt/vagrant/embedded/bin/curl --verbose \
  "scp://redacteddomain.com:/path/to/custom_box.box" --insecure \
  --key ~/.ssh/my-key.pem --user admin: --output test.box
```

How can i pass the required options to the `base_url` or make the packaged version of curl respect my ssh config file?

Here is the config block im using in my `vagrantfile`
```ruby
  config.vm.define :devbox, primary: true do |devbox|
    devbox.vm.box_download_insecure = true
    devbox.vm.box = "base-box"
    devbox.vm.box_url = "scp://redacteddomain.com:/path/to/custom_box.box"
  end
```

Any guidance would be greatly appreciated, I have wasted about 10 hours on this so far.

### Vagrant version

Vagrant 2.2.7

### Host operating system
macOS Catalina 10.15.4

### Vagrantfile

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
  config.vm.define :example, primary: true do |example|
    example.vm.box_download_insecure = true
    example.vm.box = "custom_box"
    example.vm.box_url = ["scp://myurl.com:/opt/vagrant/custom_box.box"]

    example.vm.network "private_network", ip: "192.168.33.10"
    example.vm.network :forwarded_port, guest: 22, host: 2222, id: 'ssh'
    example.vm.synced_folder "./", "/example", :mount_options => ["dmode=777","fmode=777"]
    example.vm.provider "virtualbox" do |v|
      v.memory = 4096
      v.customize ["modifyvm", :id, "--cpus", "2"]
    end
  end
end
```

### Debug output

https://gist.github.com/peledies/2baab7f568a7377e6de748049e92841f

### Expected behavior

Vagrant should use the `~/.ssh/config` file to make its curl request when using the `scp` option.
OR
[...]

PR #11560, opened 2020-04-28
Add option `box_download_options`

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11546
Allow users to specify a map of extra options to pass to the downloader

### DIFF
--- a/lib/vagrant/util/downloader.rb
+++ b/lib/vagrant/util/downloader.rb
@@ -70,7 +70,7 @@ module Vagrant
           :sha384 => options[:sha384],
           :sha512 => options[:sha512]
         }.compact
-        @extra_download_options = options[:box_download_options]
+        @extra_download_options = options[:box_extra_download_options] || []
       end
 
       # This executes the actual download, downloading the source file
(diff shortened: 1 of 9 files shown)


plugins/kernel_v2/config/vm.rb
```diff
@@ -467,6 +469,7 @@ def finalize!
         @box_download_location_trusted = false if @box_download_location_trusted == UNSET_VALUE
         @box_url = nil if @box_url == UNSET_VALUE
         @box_version = nil if @box_version == UNSET_VALUE
+        @box_download_options = {} if @box_download_options == UNSET_VALUE
```
briancain, line 472: It would be good to have some validation of this option in the `validate` step! Then we can give some good error messages if this option is incorrectly set to the wrong type, or whatever users might end up doing.

lib/vagrant/util/downloader.rb
```diff
@@ -245,6 +247,9 @@ def options
         options << "-u" << @auth if @auth
         options << "--location-trusted" if @location_trusted
 
+        options.concat(
+          Vagrant::Util::MapCommandOptions.map_to_command_options(@extra_download_options))
```
briancain, line 251: Another thing we should do with this config is call this conversion map in the `finalize!` method of `#Config::Vm` instead of here. Then we could give people some warning in the config validate step in case their command options are wrong before they go and use them, and just pass along the `@extra_download_options` here since it'll already be converted. This will also make it easier to use else where if we have to since it'll already be the right format.

Issue #11508, opened 2020-04-10
Modification of change_host_name capability


chrisroberts:
## Changes to Vagrantfile configuration

* ~`config.vm.disable_hosts_modification`~ `config.vm.allow_hosts_modification`

This is a new configuration option. It is a boolean value that when ~enabled~ disabled will prevent Vagrant from making modifications to the guest's `/etc/hosts` file.

* `config.vm.network` - `:hostname`

Add a new `:hostname` option to the `config.vm.network` method. This will be a boolean value that, when set to `true`, will signal that a specific address should be used for configuring a hostname within the `/etc/hosts` file. There are a couple of restrictions that come with using this option:

1. The network configuration _must_ provide a static IP address
2. Only one network configuration for the guest can set this value to `true`

## Changes to `/etc/hosts` modifications

The `change_host_name` capabilities need to be updated to follow:

* If `config.vm.disable_hosts_modifications` is `true`, no modifications are performed
* If `config.vm.disable_hosts_modifications` is `false`
  * If a network configuration has been flagged for hostname
    * Search guest /etc/hosts file for existing entry.
    * Remove entry if found
    * Add new entry using address from network configuration
  * If no network configuration has been flagged for hostname
    * Search guest /etc/hosts file for existing entry
    * If entry is found, make no modifications
    * If no entry is found, add entry for hostname on loopback interface with address not in use

## Updates required

Vagrant configuration:

- [x] Add `config.vm.disable_hosts_modification` configuration
- [x] Add :hostname option to `config.vm.network`

Guest capabilities:

- [x] alpine
- [x] alt
- [x] arch
- [x] atomic
- [ ] coreos
- [x] darwin
- [x] debian
- [x] freebsd
- [x] gentoo
- [x] omnios
- [x] openbsd
- [x] photon
- [x] pld
- [x] redhat
- [x] slackware
- [x] suse

soapy1:
Seems like coreOS doesn't really use `/etc/hosts`, so I think this issue is resolved.

PR #11565, opened 2020-04-29
Add option to disable modification of /etc/hosts on guest

soapy1:
Add `config.vm.allow_hosts_modification` configuration as per https://github.com/hashicorp/vagrant/issues/11508

### DIFF
--- a/lib/vagrant/action/builtin/set_hostname.rb
+++ b/lib/vagrant/action/builtin/set_hostname.rb
@@ -17,12 +17,12 @@ module Vagrant
           @app.call(env)
 
           hostname = env[:machine].config.vm.hostname
-          disable_hosts_modification = env[:machine].config.vm.disable_hosts_modification
-          if !hostname.nil? && !disable_hosts_modification
+          allow_hosts_modification = env[:machine].config.vm.allow_hosts_modification
+          if !hostname.nil? && allow_hosts_modification
             env[:ui].info I18n.t("vagrant.actions.vm.hostname.setting")
             env[:machine].guest.capability(:change_host_name, hostname)
           else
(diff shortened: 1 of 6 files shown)


lib/vagrant/action/builtin/set_hostname.rb
```diff
@@ -16,7 +16,8 @@ def call(env)
           @app.call(env)
 
           hostname = env[:machine].config.vm.hostname
-          if !hostname.nil?
+          disable_hosts_modification = env[:machine].config.vm.disable_hosts_modification
+          if !hostname.nil? && !disable_hosts_modification
```
briancain, line 20: Looks good! Only suggestion I have is add a `@logger` in the initialize method, and then log a simple `info` level message saying that hosts modification has been disabled and will be skipped since `disable_hosts_modification` is true.

plugins/kernel_v2/config/vm.rb
```diff
@@ -473,6 +475,7 @@ def finalize!
         @box_version = nil if @box_version == UNSET_VALUE
         @box_download_options = {} if @box_download_options == UNSET_VALUE
         @box_extra_download_options = Vagrant::Util::MapCommandOptions.map_to_command_options(@box_download_options)
+        @disable_hosts_modification = false if @disable_hosts_modification == UNSET_VALUE
```
briancain, line 478: Ahhh, sorry I didn't catch this sooner. But we should have a validation step that ensures this value is true/false and not something like a string or integer. (Like if someone set `config.vm.disable_hosts_modification = "true"`). Just so we can give good error messages if someones config is wrong. I notice it's being auto-set to false if it's anything other than true, but it might be nice to have a warning if they've set this value incorrectly.

soapy1:
Going to merge this since the last pushes were to fix a merge conflict, and add some forgotten punctuation

PR #11566, opened 2020-04-30
Add :hostname option to network config

soapy1:
addresses adding hostname option to network config as per https://github.com/hashicorp/vagrant/issues/11508 

* hostname is a boolean
* a network that sets hostname should have a static ip address
* only one network may set hostname

can be set at `config.vm.network :public_network, hostname: true, ip: "192.168.0.1"`


soapy1:
Going to merge this since the last pushes were to fix a merge conflict, and add some forgotten punctuation

PR #11567, opened 2020-04-30
Set hostname according to hostname flag and ip for suse

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11317
depends on https://github.com/hashicorp/vagrant/pull/11566
addresses the suse host cap as per https://github.com/hashicorp/vagrant/issues/11508

### DIFF
--- a/plugins/guests/suse/cap/change_host_name.rb
+++ b/plugins/guests/suse/cap/change_host_name.rb
@@ -4,18 +4,37 @@ module VagrantPlugins
       class ChangeHostName
         def self.change_host_name(machine, name)
           comm = machine.communicate
-
           basename = name.split(".", 2)[0]
           if !comm.test('test "$(hostnamectl --static status)" = "#{basename}"', sudo: false)
             comm.sudo <<-EOH.gsub(/^ {14}/, '')
-              hostnamectl set-hostname '#{basename}'
-
-              # Prepend ourselves to /etc/hosts
-              grep -w '#{name}' /etc/hosts || {
-                sed -i'' '1i 127.0.0.1\\t#{name}\\t#{basename}' /etc/hosts
-              }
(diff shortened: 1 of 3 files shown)


plugins/guests/suse/cap/change_host_name.rb
```diff
@@ -4,19 +4,38 @@ module Cap
...
+              add_hostname_to_loopback(comm, name, basename)
+            end
           end
         end
+
+        def self.add_hostname_to_loopback(comm, name, basename)
+          # Add hostname to /etc/hosts if not already there
+          comm.sudo <<-EOH.gsub(/^ {14}/, '')
+          grep -w '#{name}' /etc/hosts || {
+            sed -i'' '1i 127.0.0.1\\t#{name}\\t#{basename}' /etc/hosts
```
chrisroberts, line 26: When doing the default add to loopback, we want to use an unused loopback address. Debian does this using `127.0.1.1` which is reasonable but we'll want to check the `/etc/hosts` file and confirm it's not already in use, and if it is, increment until an unused one is found.

plugins/guests/suse/cap/change_host_name.rb
```diff
@@ -4,19 +4,46 @@ module Cap
...
+          # the loop_bound
+          loop_bound = 5
+          comm.sudo <<-EOH.gsub(/^ {14}/, '')
+          grep -w '#{name}' /etc/hosts || {
+            for i in {1..#{loop_bound}}; do
+              grep -w "127.0.${i}.1" /etc/hosts || {
+                echo "127.0.${i}.1 #{name} #{basename}" >> /etc/hosts
+                break
+              }
+            done
```
soapy1, line 34: I don't know how I feel about this.... It seems to work consistently, but also its a little janky?

Issue #11368, opened 2020-02-03
After reboot, Vagrant will not mount shared folders.  Only 'vagrant reload' works.


queglay:
### Vagrant version
Vagrant 2.2.6

### Host operating system
CENTOS 7

### Guest operating system
ubuntu 16

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
  config.vm.define "firehawkgateway" do |firehawkgateway|
    firehawkgateway.vm.box = "bento/ubuntu-16.04"
  end
end
```

### Debug output
https://gist.github.com/queglay/3b5c6f923ad5c9d3f55ee896cf5f0ecc

Also a log of the shell operations to show the missing folder-
```
vagrant@vagrant:~$ ls /vagrant/
Vagrantfile
vagrant@vagrant:~$ touch /vagrant/test
vagrant@vagrant:~$ ls /vagrant/
test  Vagrantfile
vagrant@vagrant:~$ sudo reboot
Connection to 127.0.0.1 closed by remote host.
Connection to 127.0.0.1 closed.
[user@workstation test-vagrant-mount]$ vagrant ssh
Welcome to Ubuntu 16.04.6 LTS (GNU/Linux 4.4.0-170-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

 * Overheard at KubeCon: "microk8s.status just blew my mind".

     https://microk8s.io/docs/commands#microk8s.status

0 packages can be updated.
0 updates are security updates.



This system is built by the Bento project by Chef Software
More information can be found at https://github.com/chef/bento
Last login: Mon Feb  3 11:08:42 2020 from 10.0.2.2
vagrant@vagrant:~$ ls /vagrant/
vagrant@vagrant:~$ ls /vagrant/
vagrant@vagrant:~$ ls /vagrant/
vagrant@vagrant:~$ exit
logout
Connection to 127.0.0.1 closed.
[user@workstation test-vagrant-mount]$ vagrant reload
==> firehawkgateway: Attempting graceful shutdown of VM...
==> firehawkgateway: Checking if box 'bento/ubuntu-16.04' version '201912.15.0' is up to date...
==> firehawkgateway: Clearing any previously set forwarded ports...
==> firehawkgateway: Fixed port collision for 22 => 2222. Now on port 2204.
==> firehawkgateway: Clearing any previously set network interfaces...
[...]

PR #11570, opened 2020-05-01
Automatically mount virtual box shared folder when machine reboots

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11368
maybe supersedes https://github.com/hashicorp/vagrant/pull/11417

This is done by adding the mount to fstab
ref: https://askubuntu.com/questions/252853/how-to-mount-a-virtualbox-shared-folder-at-startup


Jasper-Ben:
Hi @soapy1, thanks for the work! :+1: 
I cannot help but wonder however, if this is the best strategy for archiving persistent mounting, as it only works for Linux. My proposed solution in https://github.com/hashicorp/vagrant/pull/11417 works guest OS independent, as it uses VirtualBoxes build-in functionality for persistent volume mounting.

soapy1:
@Jasper-Ben good point! I spent some time trying to evaluate both methods of auto mounting after a reboot initiated from the host. While the upside of using the automounting feature on virtualbox is being OS independent, the downside is loosing control over the mount options of the folder. Using the virtual box flag does not preserve the owner/group of the folder between reboots. Users are currently able to specify these parameters and these setting should be respected between reboots that are initiated from inside the guest. These setting are able to be preserved using the fstab approach, so I think it should be used going forward. Note, that for windows auto mounting after reboot currently works, and for Darwin, only nfs and rsync synced folders are supported.

Jasper-Ben:
Hi @soapy1, I was not aware that auto-mounting does already work within Windows. Although I personally never really saw the need for assigning UID/GID for virtualboxes shared folders (I always used them the "virtualbox way" by adding the guests user(s) to the vboxsf group), I definitely can see that this feature might come in handy in advanced use cases. So I guess you are right, fstab is the way to go :smile:

plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
```diff
@@ -61,6 +58,32 @@ def self.mount_virtualbox_shared_folder(machine, name, guestpath, options)
...
+              :mount_options => mount_options,
+            })
+          end
+
+          fstab_entry = Vagrant::Util::TemplateRenderer.render('guests/linux/etc_fstab', folders: export_folders)
+          # Replace existing vagrant managed fstab entry
+          machine.communicate.sudo("sed -i '/\#VAGRANT-BEGIN/,/\#VAGRANT-END/d' /etc/fstab")
+          machine.communicate.sudo("echo '#{fstab_entry}' >> /etc/fstab")
+
+          fstab_valid = machine.communicate.sudo("mount -a", error_check: false)
```
chrisroberts, line 80:
```ruby
machine.communicate.test("mount -a", sudo: true)
```

will return a `true`/`false` that can be tested.

plugins/guests/linux/cap/mount_virtualbox_shared_folder.rb
```diff
@@ -61,6 +58,32 @@ def self.mount_virtualbox_shared_folder(machine, name, guestpath, options)
...
+            guest_path = Shellwords.escape(data[:guestpath])
+            mount_options, mount_uid, mount_gid  =  self.mount_options(machine, name, guest_path, data)
+            mount_options = "#{mount_options},nofail"
+            export_folders.push({
+              :name => name,
+              :mount_point => guest_path,
+              :mount_type => VB_MOUNT_TYPE,
+              :mount_options => mount_options,
+            })
+          end
```
chrisroberts, line 73: You can use a `#map` here to compact this a little:

```ruby
export_folders = fstab_folders.map do |name, data|
  ...
  {name: name, mount_point: guest_path, mount_type: VB_MOUNT_TYPE, mount_options: mount_options}
end
```

plugins/guests/linux/cap/persist_mount_shared_folder.rb
```diff
@@ -0,0 +1,39 @@
...
+              mount_options: mount_options,
+            }
+          end
+
+          fstab_entry = Vagrant::Util::TemplateRenderer.render('guests/linux/etc_fstab', folders: export_folders)
+          # Replace existing vagrant managed fstab entry
+          machine.communicate.sudo("sed -i '/\#VAGRANT-BEGIN/,/\#VAGRANT-END/d' /etc/fstab")
+          machine.communicate.sudo("echo '#{fstab_entry}' >> /etc/fstab")
+
+          fstab_valid = machine.communicate.test("mount -a", sudo: true)
```
jbonhag, line 29: I think we should mount each share individually if possible rather than doing a `mount -a`.

plugins/providers/virtualbox/synced_folder.rb
```diff
@@ -52,12 +53,16 @@ def enable(machine, folders, _opts)
             machine.guest.capability(
               :mount_virtualbox_shared_folder,
               os_friendly_id(id), data[:guestpath], data)
+            fstab_folders.push([os_friendly_id(id), data])
```
jbonhag, line 56: Using the "shovel" operator `<<` is usually more common in Ruby, e.g.:

```suggestion
            fstab_folders << [os_friendly_id(id), data]
```

PR #11584, opened 2020-05-05
Seperate podman centos and rhel install

soapy1:


jbonhag:
This is great! 🎉 One suggestion, what do you think about `install_mode` (like [ansible_local](https://www.vagrantup.com/docs/provisioning/ansible_local.html#install_mode)) for activating the kubic install?

```ruby
config.vm.provision "podman" do |p| 
  p.install_mode = :kubik
end
```

Issue #11577, opened 2020-05-03
[Docker provider] Enhancement Request: Private networking using --inhibitipv4


terryburton:
Future releases of Docker will allow a Docker network to be created using the bridge driver without assigning an IP address, using the `--inhibitipv4` option:

https://github.com/moby/moby/issues/37430

The resulting situation will be much more similar in terms of network isolation to the private-type networks created by other providers.

The Docker provider should either use the `--inhibitipv4` option by default (where this exists) for private network, or express this functionality as an option.

#### Context

At the moment a "private" network created by the existing Docker provider (using the existing bridge driver machinery) is always assigned a router IP address with an entry planted in the main routing table. Therefore the Docker containers are directly accessible from the host through this private network, although not from other Docker networks due to the iptables entries that Docker's previous concept of a private, bridged networking introduces.

The new `--inhipbtipv4` option creates a construction that is effectively a disconnected Layer 2 segment to which the containers are joined.

Aligning the behaviour of the Docker provider with that of other providers would be especially useful when you are using Docker as a form of lightweight virtualisation for mocking production environments, including production IP addresses. Isolation is better preserved since the normal routes to the production environment's IP addresses are not replaced when starting the mock environment.

soapy1:
I believe this option can be set with the current docker provider. The Vagrantfile should look something like:

```
Vagrant.configure("2") do |config|
  config.vm.provider "docker" do |d|
    d.image = "ubuntu"
    d.cmd = ["tail", "-f", "/dev/null"]
  end

  config.vm.network :private_network, ip: "172.20.128.2", docker_network__opt: "com.docker.network.bridge.inhibit_ipv4=true"
end
```

The command that gets generated looks something like: `docker network create vagrant_network_172.20.128.0/24 --internal --opt com.docker.network.bridge.inhibit_ipv4=true --subnet 172.20.128.0/24`

When I inspect the network created using docker I get:
```
...
"Options": {
            "com.docker.network.bridge.inhibit_ipv4": "true"
        },
...
```

terryburton:
@soapy1 I've just tested this and you're absolutely right. Good catch!

PR #11585, opened 2020-05-06
Docker provider: bridge_inhibitipv4 option to prevent IP assignment

terryburton:
Make the scoped option `docker_network__bridge_inhibitipv4` pass an argument to
`docker network create` that inhibits IP address assignemnt to the bridge:

    -o com.docker.network.bridge.inhibit_ipv4=true

This patch allows the upcoming Docker functionality to be access using a
network option. In future there might be a case for switching this to true by
default for private: networks for parity with other providers.

Rationale and context is provided in issue #11577.


soapy1:
Thanks for the contribution 😸 ! Since the related issue has been resolved I'm going to close this PR.

PR #11587, opened 2020-05-06
esxi guest: only create the ssh directory when needed

### LINKED ISSUE #11574: Cannot insert ssh public key in ESXi 7.0
### Vagrant version

2.2.7

### Guest operating system

ESXi 7.0

### Expected behavior

SSH keys should be inserted without errors.

### Actual behavior

The code at https://github.com/hashicorp/vagrant/blob/v2.2.7/plugins/guests/esxi/cap/public_key.rb#L23-L31

Fails with:

```
==> esxi: Waiting for SSH to become available...
    esxi: 
    esxi: Inserting generated public key within guest...

set -e
SSH_DIR="$(grep -q '^AuthorizedKeysFile */etc/ssh/keys-%u/authorized_keys$' /etc/ssh/sshd_config && echo -n /etc/ssh/keys-${USER} || echo -n ~/.ssh)"
mkdir -p "${SSH_DIR}"
chmod 0700 "${SSH_DIR}"
cat '/tmp/vagrant-insert-pubkey-1588517835' >> "${SSH_DIR}/authorized_keys"
chmod 0600 "${SSH_DIR}/authorized_keys"
rm -f '/tmp/vagrant-insert-pubkey-1588517835'


Stdout from the command:



Stderr from the command:

chmod: /etc/ssh/keys-root: Operation not permitted

```

Here are the ESXi 7.0 filesystem permissions:

```
[root@localhost:~] ll /etc/ssh/
total 596
drwxr-xr-x    1 root     root           512 May  3 14:56 .
-r-------T    1 root     root             0 Mar 16 10:34 .#ssh_host_rsa_key
-r--r--r-T    1 root     root             0 Mar 16 10:34 .#ssh_host_rsa_key.pub
-r-------T    1 root     root          1396 Mar 16 10:34 .#sshd_config
drwxr-xr-x    1 root     root           512 May  3 14:57 ..
drwxr-xr-x    1 root     root           512 May  3 14:56 keys-root
-r--r--r--    1 root     root        577388 Mar 16 10:34 moduli
-rw------T    1 root     root           551 Mar 16 10:34 ssh_config
-rw------T    1 root     root             0 Mar 16 10:34 ssh_host_dsa_key
-rw-r--r-T    1 root     root             0 Mar 16 10:34 ssh_host_dsa_key.pub
-r--------    1 root     root          1704 May  3 14:56 ssh_host_rsa_key
-rw-------    1 root     root           381 May  3 14:56 ssh_host_rsa_key.pub
-rw-------    1 root     root          1397 May  3 14:56 sshd_config
```

And:

```
[root@localhost:~] ll /etc/ssh/keys-root/
total 8
drwxr-xr-x    1 root     root           512 May  3 14:56 .
drwxr-xr-x    1 root     root           512 May  3 14:56 ..
-rw------T    1 root     root             0 Mar 16 10:34 authorized_keys
```

In this new ESXi 7 version something is now preventing the `chmod` to change the directory permissions.

To fix this one just needs to remove the first `chmod` line.

**NB** the second `chmod` works correctly. thou... it will remove the `T` permission bit from the `authorized_keys` file.

Can I submit the PR with that or do you have any other suggestion?

rgl:
Close #11574

### DIFF
--- a/plugins/guests/esxi/cap/public_key.rb
+++ b/plugins/guests/esxi/cap/public_key.rb
@@ -24,7 +24,9 @@ module VagrantPlugins
             set -e
             SSH_DIR="$(grep -q '^AuthorizedKeysFile\s*\/etc\/ssh\/keys-%u\/authorized_keys$' /etc/ssh/sshd_config && echo -n /etc/ssh/keys-${USER} || echo -n ~/.ssh)"
             mkdir -p "${SSH_DIR}"
-            chmod 0700 "${SSH_DIR}"
+            # NB in ESXi 7.0 we cannot change the /etc/ssh/keys-root directory
+            #    permissions, so ignore any errors.
+            chmod 0700 "${SSH_DIR}" || true
             cat '#{remote_path}' >> "${SSH_DIR}/authorized_keys"
             chmod 0600 "${SSH_DIR}/authorized_keys"
             rm -f '#{remote_path}'


plugins/guests/esxi/cap/public_key.rb
```diff
@@ -23,8 +23,10 @@ def self.insert_public_key(machine, contents)
           comm.execute <<-EOH.gsub(/^ {12}/, "")
             set -e
             SSH_DIR="$(grep -q '^AuthorizedKeysFile\s*\/etc\/ssh\/keys-%u\/authorized_keys$' /etc/ssh/sshd_config && echo -n /etc/ssh/keys-${USER} || echo -n ~/.ssh)"
-            mkdir -p "${SSH_DIR}"
-            chmod 0700 "${SSH_DIR}"
+            if [ ! -d "${SSH_DIR}" ]; then
```
soapy1, line 26: Since this is only checking for the existence of `$SSH_DIR`, if the dir exists but doesn't have the correct permissions the `chmod 0700 "${SSH_DIR}"` will not get run possibly resulting in unexpected behavior. So, I think this should be broken out into two checks like:
```
if [ ! -d "${SSH_DIR}" ]; then mkdir -p "${SSH_DIR}"; fi                  # Ensure dir exists
if [ $(stat -c '%a' $SSH_DIR) -lt 700 ]; chmod 0700 "${SSH_DIR}" ; fi     # Ensure dir has appropriate permissions
```

rgl: I don't think it needs to do that, because whatever created the directory before, must do it correctly, or else that part is broken and should be fixed upstream.

soapy1: >  whatever created the directory before, must do it correctly, or else that part is broken and should be fixed upstream.

yup, I agree with you here. Though, Vagrant should probably be a bit defensive and give changing the permissions on the directory a try. Maybe something like`set +e; chmod 0700 "${SSH_DIR}"; set -e;`. In the case of an upstream issue or a user error, this will maybe remedy the situation so that the Vagrant environment is still usable. If the `chmod` command is unsuccessful like in the case of ESXI7 it will at least not error out here. However, if the `$SSH_DIR` does have bad permissions after trying to fix them (due to an upstream issue or user error) ssh-ing into the machine will fail. This failure should be dealt with upstream, or by the user.

rgl: Will do this:

```diff
--- a/plugins/guests/esxi/cap/public_key.rb
+++ b/plugins/guests/esxi/cap/public_key.rb
@@ -23,10 +23,10 @@ module VagrantPlugins
           comm.execute <<-EOH.gsub(/^ {12}/, "")
             set -e
             SSH_DIR="$(grep -q '^AuthorizedKeysFile\s*\/etc\/ssh\/keys-%u\/authorized_keys$' /etc/ssh/sshd_config && echo -n /etc/ssh/keys-${USER} || echo -n ~/.ssh)"
-            if [ ! -d "${SSH_DIR}" ]; then
-              mkdir -p "${SSH_DIR}"
-              chmod 0700 "${SSH_DIR}"
-            fi
+            mkdir -p "${SSH_DIR}"
+            # NB in ESXi 7.0 we cannot change the /etc/ssh/keys-root directory
+            #    permissions, so ignore any errors.
+            chmod 0700 "${SSH_DIR}" || true
             cat '#{remote_path}' >> "${SSH_DIR}/authorized_keys"
             chmod 0600 "${SSH_DIR}/authorized_keys"
             rm -f '#{remote_path}'
```

PR #11602, opened 2020-05-08
Fixes #9067: Ensure new containers don't grab existing bound ports

### LINKED ISSUE #9067: Port collision not fixed when using Docker provider
### Vagrant version
```
$ vagrant -v
Vagrant 2.0.0
```

### Host operating system
Mac OSX 10.11.6

### Guest operating system
Ubuntu

### Vagrantfile
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

  config.ssh.username = "root"
  config.vm.provider "docker" do |docker|
    docker.image = "nineseconds/docker-vagrant"
    docker.has_ssh = true
    docker.remains_running = true
  end
end
```

### Debug output
https://gist.github.com/tjoelsson/fca760f3d0257bb31d76dc8a682db0c8

### Expected behavior
Should be able to run two containers simultaneously.

### Actual behavior
Port collision stops container from starting.

### Steps to reproduce
1. Prepare two folders (a and b) with identical Vagrantfile
2. Run `vagrant up` in folder a
3. Run `vagrant halt` in folder a
4. Run `vagrant up` in folder b
5. Run `vagrant up` in folder a

### References

briancain:
Prior to this commit, if a created but exited container bound a port,
and a new container grabed that same port (say for an ssh port forward),
when the initial container came back up it would fail because the port
also got bound to the second container. This commit fixes that behavior
by first looking at what containers are already bound prior to creating
a container.

### DIFF
--- a/lib/vagrant/util/ipv4_interfaces.rb
+++ b/lib/vagrant/util/ipv4_interfaces.rb
@@ -1,0 +1,13 @@
+module Vagrant
+  module Util
+    module IPv4Interfaces
+      def self.ipv4_interfaces
+        Socket.getifaddrs.select do |ifaddr|
+          ifaddr.addr && ifaddr.addr.ipv4?
+        end.map do |ifaddr|
+          [ifaddr.name, ifaddr.addr.ip_address]
+        end
+      end
+    end
+  end
+end
--- a/lib/vagrant/util/ipv4_interfaces.rb
+++ b/lib/vagrant/util/ipv4_interfaces.rb
@@ -1,13 +1,15 @@
 module Vagrant
   module Util
     module IPv4Interfaces
-      def self.ipv4_interfaces
+      def ipv4_interfaces
         Socket.getifaddrs.select do |ifaddr|
           ifaddr.addr && ifaddr.addr.ipv4?
         end.map do |ifaddr|
           [ifaddr.name, ifaddr.addr.ip_address]
         end
       end
+
+      extend IPv4Interfaces
     end
   end
 end
--- a/lib/vagrant/util/is_port_open.rb
+++ b/lib/vagrant/util/is_port_open.rb
@@ -13,7 +13,7 @@ module Vagrant
       # @param [Integer] port Port to check.
       # @return [Boolean] `true` if the port is open (listening), `false`
       #   otherwise.
-      def self.is_port_open?(host, port)
+      def is_port_open?(host, port)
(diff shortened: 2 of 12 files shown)


plugins/providers/docker/driver.rb
```diff
@@ -126,6 +126,46 @@ def image?(id)
...
+
+        all_containers.each do |c|
+          container_info = inspect_container(c)
+
+          if container_info["HostConfig"]["PortBindings"]
+            # We remove the first character because docker inspect adds a '/' to
+            # the beginning every container name to match the "internal"
+            # implementation of docker:
+            #   https://github.com/moby/moby/issues/6705
+            container_name = container_info["Name"][1..-1]
```
briancain, line 148: Oops, thought I was going to use this and I guess I didn't need it.

lib/vagrant/util/is_port_open.rb
```diff
@@ -13,7 +13,7 @@ module IsPortOpen
       # @param [Integer] port Port to check.
       # @return [Boolean] `true` if the port is open (listening), `false`
       #   otherwise.
-      def is_port_open?(host, port)
+      def self.is_port_open?(host, port)
```
chrisroberts, line 16: With this being an already existing module, this would be a breaking change. Instead, just add an `extend` to the end:

```ruby
module Vagrant
  module Util
    module IsPortOpen
      def is_port_open?
      ...
      end
      extend IsPortOpen
    end
  end
end
```

lib/vagrant/util/ipv4_interfaces.rb
```diff
@@ -0,0 +1,13 @@
+module Vagrant
+  module Util
+    module IPv4Interfaces
+      def self.ipv4_interfaces
```
chrisroberts, line 4: Lets remove the `self.` so we can include this module if we want

lib/vagrant/util/ipv4_interfaces.rb
```diff
@@ -0,0 +1,13 @@
...
+        Socket.getifaddrs.select do |ifaddr|
+          ifaddr.addr && ifaddr.addr.ipv4?
+        end.map do |ifaddr|
+          [ifaddr.name, ifaddr.addr.ip_address]
+        end
+      end
```
chrisroberts, line 10: And then add an `extend IPv4Interfaces` so you can still call it like you are currently.

PR #11613, opened 2020-05-12
Feature: ISO attachment for VirtualBox

jbonhag:
This builds on the existing disk functionality, and adds some special
IDE controller-related flavor.

Considerations for IDE controllers:
- Primary/secondary attachments, so that each port can have two devices
  attached
- Adding the ability to address a specific controller name for disk
  attachment

This also prevents a user from attaching multiple instances of the same
ISO file, because VirtualBox will assign each of these the same UUID
which makes disconnection difficult. However, if multiple copies of the
ISO are attached to different devices, removing the DVD config will
cause the duplicate devices to be removed.

We may want to consider additional work to make the storage controllers
truly generic.


plugins/kernel_v2/config/disk.rb
```diff
@@ -164,6 +165,10 @@ def validate(machine)
           end
         end
 
+        if @type == :dvd && !@file
+          errors << I18n.t("vagrant.config.disk.dvd_type_file_required", name: @name, machine: machine.name)
+        end
```
briancain, line 170: I think we also need a validation step to check if type is DVD but has been set to `primary`. Since `primary` is intended for hard disks rather than DVDs.

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -108,6 +108,45 @@ def self.handle_configure_disk(machine, disk, all_disks)
           disk_metadata
         end
 
+        # Helper method to get the UUID of a specific controller attachment
+        #
+        # @param [Vagrant::Machine] machine - the current machine
+        # @param [String] controller_name - the name of the controller to examine
+        # @param [String] port - port to look up
+        # @param [String] device - device to look up
+        def self.attachment(machine, controller_name, port, device)
```
briancain, line 117: This could probably be a driver method rather than a helper method here. And it could be called something like `get_disk_uuid` so it's a bit more descriptive to what it's doing. The method docs should also list the `# @return` value too!

briancain:
Oh! I forgot to mention this yesterday, but we should also update the docs for this feature :smile:

plugins/providers/virtualbox/cap/cleanup_disks.rb
```diff
@@ -28,23 +29,52 @@ def self.handle_cleanup_disk(machine, defined_disks, disk_meta)
...
           primary_disk = vm_info["SATA Controller-ImageUUID-0-0"]
 
-          disk_meta.each do |d|
-            dsk = defined_disks.select { |dk| dk.name == d["name"] }
-            if !dsk.empty? || d["uuid"] == primary_disk
-              next
-            else
-              LOGGER.warn("Found disk not in Vagrantfile config: '#{d["name"]}'. Removing disk from guest #{machine.name}")
-              disk_info = machine.provider.driver.get_port_and_device(d["uuid"])
+          unless disk_meta.nil?
```
chrisroberts, line 32:
```ruby
if disk_meta
```

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -167,19 +206,38 @@ def self.create_disk(machine, disk_config)
...
+            # with an empty device
+            (0..1).each do |port|
+              break if dsk_info[:port] && dsk_info[:device]
+
+              (0..1).each do |device|
+                if vm_info["#{controller}-ImageUUID-#{port}-#{device}"].to_s.empty?
+                  dsk_info[:port] = port.to_s
+                  dsk_info[:device] = device.to_s
+                  break
+                end
```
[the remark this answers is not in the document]
jbonhag: 👍 I abstracted that a bit using the StorageController object, for example:
- https://github.com/hashicorp/vagrant/pull/11613/files#diff-432b13780e5f879a3792dd0fd9cec2e1R23

plugins/providers/virtualbox/cap/configure_disks.rb
jbonhag: This was moved into the driver as part of the storage controller refactor.

jbonhag:
Update: pretty serious refactoring, I basically removed all hard-coded values and added some logic for dealing with 1..n storage controllers.

- With a single controller, everything gets attached to that controller
- With multiple controllers, we'll attach disks to the SATA and DVDs to the IDE

plugins/providers/virtualbox/cap/cleanup_disks.rb
```diff
@@ -25,26 +26,77 @@ def self.cleanup_disks(machine, defined_disks, disk_meta_file)
...
+              raise Vagrant::Errors::VirtualBoxDisksControllerNotFound, storage_bus: "SATA"
+            end
+          end
 
-              machine.ui.warn("Disk '#{d["name"]}' no longer exists in Vagrant config. Removing and closing medium from guest...", prefix: true)
+          primary = primary_controller.attachments.detect { |a| a[:port] == "0" && a[:device] == "0" }
+          if primary.nil?
+            raise Vagrant::Errors::VirtualBoxDisksPrimaryNotFound
+          end
+          primary_uuid = primary[:uuid]
```
briancain, line 43: I think all the code above here detecting the proper controller would make sense to be its own method.

plugins/providers/virtualbox/cap/cleanup_disks.rb
```diff
@@ -25,26 +26,77 @@ def self.cleanup_disks(machine, defined_disks, disk_meta_file)
...
 
-              if disk_info.empty?
-                LOGGER.warn("Disk '#{d["name"]}' not attached to guest, but still exists.")
+          if disk_meta
```
briancain, line 45: Hm, I don't think this if is required. It should be an empty hash if there's no meta data, so the each will exit without enumerating over anything already.

plugins/providers/virtualbox/cap/cleanup_disks.rb
```diff
@@ -25,26 +26,77 @@ def self.cleanup_disks(machine, defined_disks, disk_meta_file)
...
-                machine.provider.driver.remove_disk(disk_info[:port], disk_info[:device])
+                LOGGER.warn("Found disk not in Vagrantfile config: '#{d["name"]}'. Removing disk from guest #{machine.name}")
+                machine.ui.warn("Disk '#{d["name"]}' no longer exists in Vagrant config. Removing and closing medium from guest...", prefix: true)
+
+                controller = storage_controllers.detect { |c| c.name == d["controller"] }
+                attachment = controller.attachments.detect { |a| a[:port] == d["port"] &&
+                                                                 a[:device] == d["device"] }
+
+                # Retry lookup by UUID
+                if attachment.nil?
```
briancain, line 59: I think this can be simplified:

```suggestion
                if !attachment
```

plugins/providers/virtualbox/cap/cleanup_disks.rb
```diff
@@ -25,26 +26,77 @@ def self.cleanup_disks(machine, defined_disks, disk_meta_file)
...
+                  attachment = controller.attachments.detect { |a| a[:uuid] == d["uuid"] }
+                end
+
+                if attachment.nil?
```
briancain, line 63:
```suggestion
                if !attachment
```

plugins/providers/virtualbox/cap/cleanup_disks.rb
```diff
@@ -25,26 +26,77 @@ def self.cleanup_disks(machine, defined_disks, disk_meta_file)
...
               end
+            end
+          end
+        end
+
+        # @param [Vagrant::Machine] machine
+        # @param [VagrantPlugins::Kernel_V2::VagrantConfigDisk] defined_dvds
+        # @param [Hash] dvd_meta - A hash of all the previously defined dvds from the last configure_disk action
+        def self.handle_cleanup_dvd(machine, defined_dvds, dvd_meta)
+          if dvd_meta
```
briancain, line 79: I think this should be true here too. dvd_meta should default to an empty hash, so the `.each` enumerator will cover this case without the extra if-statement.

plugins/providers/virtualbox/cap/cleanup_disks.rb
```diff
@@ -25,26 +26,77 @@ def self.cleanup_disks(machine, defined_disks, disk_meta_file)
...
+                machine.ui.warn("DVD '#{d["name"]}' no longer exists in Vagrant config. Removing medium from guest...", prefix: true)
+
+                storage_controllers = machine.provider.driver.read_storage_controllers
+                controller = storage_controllers.detect { |c| c.name == d["controller"] }
+                attachment = controller.attachments.detect { |a| a[:port] == d["port"] &&
+                                                                 a[:device] == d["device"] }
 
-              machine.provider.driver.close_medium(d["uuid"])
+                # Retry lookup by UUID
+                if attachment.nil?
```
briancain, line 94:
```suggestion
                if !attachment
```

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -20,27 +17,74 @@ def self.configure_disks(machine, defined_disks)
...
+                name: controller.name
+            else
+              disk_controller = controller
+              dvd_controller = controller
+            end
+          else
+            disks_defined = defined_disks.select { |d| d.type == :disk }
+            if disks_defined.any?
+              disk_controller = storage_controllers.detect { |c| c.storage_bus == "SATA" }
+              if disk_controller.nil?
```
briancain, line 47:
```suggestion
              if !disk_controller
```

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -52,19 +96,24 @@ def self.configure_disks(machine, defined_disks)
...
+        def self.get_current_disk(machine, disk, all_disks, controller)
           current_disk = nil
           if disk.primary
             # Ensure we grab the proper primary disk
             # We can't rely on the order of `all_disks`, as they will not
             # always come in port order, but primary is always Port 0 Device 0.
-            vm_info = machine.provider.driver.show_vm_info
-            primary_uuid = vm_info["SATA Controller-ImageUUID-0-0"]
+            primary = controller.attachments.detect { |a| a[:port] == "0" && a[:device] == "0" }
+            if primary.nil?
```
briancain, line 109:
```suggestion
            if !primary
```

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -52,19 +96,24 @@ def self.configure_disks(machine, defined_disks)
...
+              raise Vagrant::Errors::VirtualBoxDisksPrimaryNotFound
+            end
+            primary_uuid = primary[:uuid]
 
             current_disk = all_disks.select { |d| d["UUID"] == primary_uuid }.first
           else
-            current_disk = all_disks.select { |d| d["Disk Name"] == disk.name}.first
+            current_disk = all_disks.select { |d| d["Disk Name"] == disk.name }.first
```
briancain, line 116: Hm, maybe this could turn into a `.detect`?
```suggestion
            current_disk = all_disks.detect { |d| d["Disk Name"] == disk.name }
```

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -75,39 +124,75 @@ def self.get_current_disk(machine, disk, all_disks)
...
+          # Refresh the controller information
+          storage_controllers = machine.provider.driver.read_storage_controllers
+          controller = storage_controllers.detect { |c| c.name == controller.name }
+          attachment = controller.attachments.detect { |a| a[:port] == port && a[:device] == device }
+
+          if attachment
+            { uuid: attachment[:uuid], name: dvd.name, controller: controller.name, port: port, device: device }
+          else
+            {}
+          end
```
briancain, line 193: Hm, maybe we could be a bit more explicit here and define a `dvd_meta = {}` at the top. Then change this `if` to set `dvd_meta` to the hash above, then remove the else, and always return `dvd_meta` at the bottom.

plugins/providers/virtualbox/driver/version_5_0.rb
```diff
@@ -22,20 +22,22 @@ def initialize(uuid)
...
-          execute('storageattach', @uuid, '--storagectl', controller, '--port',
-                  port.to_s, '--device', device.to_s, '--type', type, '--medium',
-                  file, '--comment', comment)
+          execute('storageattach', @uuid,
+                  '--storagectl', controller_name,
+                  '--port', port.to_s,
+                  '--device', device.to_s,
+                  '--type', type,
+                  '--medium', file,
+                  '--comment', comment)
```
briancain, line 40: Nice cleanup! :tada:

plugins/providers/virtualbox/driver/version_5_0.rb
```diff
@@ -923,6 +929,32 @@ def vdi_to_vmdk(defined_disk_path)
...
 
+        # Helper method to get a list of storage controllers added to the
+        # current VM
+        #
+        # @return [Array<VagrantPlugins::ProviderVirtualBox::Model::StorageController>]
+        def read_storage_controllers
+          vm_info = show_vm_info
+          count = vm_info.count { |key, value| key.match(/^storagecontrollername\d+$/) }
+
+          (0..count - 1).map do |n|
```
briancain, line 940: Hmm, at first glance it seems like this should be a `.each` instead of a `.map`. We're just iterating over a sequence `n` here I think?

templates/locales/en.yml
```diff
@@ -1663,11 +1663,22 @@ en:
         4.2.14 contains a critical bug which prevents it from working with
         Vagrant. VirtualBox 4.2.16+ fixes this problem. Please upgrade
         VirtualBox.
+      virtualbox_disks_controller_not_found: |-
+        The current disk configuration requires a controller with storage bus
+        '%{storage_bus}', but no such controller was found. Please add the
+        proper controller to your guest using provider-specific
```
briancain, line 1669: We could probably be specific here about the provider since the error is `virtualbox_disks`. Is there any info we could give to make it easier to solve this issue?

templates/locales/en.yml
```diff
@@ -1954,6 +1970,12 @@ en:
           Ignoring provider config for validation...
         multiple_primary_disks_error: |-
           There are more than one primary disks defined for guest '%{name}'. Please ensure that only one disk has been defined as a primary disk.
+        multiple_disk_files_error: |-
+          The following disk file is used multiple times in the disk configuration:
+
+          %{file}
+
+          Each disk file may only be attached to a VM once.
```
briancain, line 1978: Add the guest name here if possible!

website/pages/docs/disks/configuration.mdx
```diff
@@ -14,13 +14,21 @@ Vagrant Disks has several options that allow users to define and attach disks to
 - `disk_ext` (string) - Optional argument that defines what kind of file
   extension a disk should have. Defaults to `"vdi"` if unspecified. For a list of
   supported disk extensions, please check the specific provider being used.
-- `file` (string) - Optional argument that defines a path on disk pointing to
-  the location of a disk file that already exists.
+  Not used for type `:dvd.`
+
+- `file` (string) - Ffor type `:dvd`, this is a required argument that should
```
briancain, line 19:
```suggestion
- `file` (string) - For type `:dvd`, this is a required argument that should
```

website/pages/docs/disks/configuration.mdx
```diff
@@ -14,13 +14,21 @@ Vagrant Disks has several options that allow users to define and attach disks to
...
+  point to an .iso file on the host machine. For type `:disk`, this is an
```
briancain, line 20:
```suggestion
  point to an `.iso` file on the host machine. For type `:disk`, this is an
```

website/pages/docs/disks/usage.mdx
```diff
@@ -89,12 +89,32 @@ Vagrant.configure("2") do |config|
...
 ```
 
-Note: Virtualbox only allows for up to 30 disks to be attached to a given SATA Controller,
+Note: VirtualBox only allows for up to 30 disks to be attached to a given SATA Controller,
 and this number includes the primary disk! Attempting to configure more than 30 will
 result in a Vagrant error.
 
+### Attaching optical drives
+
+Vagrant can attach .iso files as optical drives using the VirtualBox provider.
```
briancain, line 98:
```suggestion
Vagrant can attach `.iso` files as optical drives using the VirtualBox provider.
```

website/pages/docs/disks/configuration.mdx
soapy1, line 19:
```suggestion
- `file` (string) - For type `:dvd`, this is a required argument that should
```

website/pages/docs/disks/configuration.mdx
```diff
@@ -14,13 +14,21 @@ Vagrant Disks has several options that allow users to define and attach disks to
...
+  optional argument that can point to the location of a disk file that already
+  exists.
+
 - `name` (string) - Required option to give the disk a name. This name will be
```
soapy1, line 24:
```suggestion
- `name` (string) - Required option to give the disk a name. This name will
```

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -167,20 +280,47 @@ def self.create_disk(machine, disk_config)
...
+
+            dsk_info[:port] = next_available_port.to_s
+            dsk_info[:device] = "0"
+          elsif controller.storage_bus == "IDE"
+            # IDE Controllers have primary/secondary devices, so find the first port
+            # with an empty device
+            (0..(controller.maxportcount - 1)).each do |port|
+              # Skip this port if it's full
+              port_attachments = controller.attachments.select { |a| a[:port] == port.to_s }
+              next if port_attachments.count == 2
```
chrisroberts, line 301: The `2` should be assigned to a descriptive constant, and the constant used here instead.

plugins/providers/virtualbox/cap/configure_disks.rb
```diff
@@ -167,20 +280,47 @@ def self.create_disk(machine, disk_config)
...
+
+              dsk_info[:port] = port.to_s
+
+              # Check for a free device
+              if port_attachments.detect { |a| a[:device] == "0" }
```
chrisroberts, line 306: `#any?` instead of `#detect`

plugins/providers/virtualbox/driver/version_5_0.rb
chrisroberts: Since the result of the map is what is being returned from the method, it should be `#map` as that's building the `Array` result :smiley:

plugins/providers/virtualbox/cap/cleanup_disks.rb
```diff
@@ -28,23 +29,52 @@ def self.handle_cleanup_disk(machine, defined_disks, disk_meta)
...
+          unless dvd_meta.nil?
+            dvd_meta.each do |d|
+              dsk = defined_dvds.select { |dk| dk.name == d["name"] }
+              if !dsk.empty?
+                next
+              else
+                LOGGER.warn("Found dvd not in Vagrantfile config: '#{d["name"]}'. Removing dvd from guest #{machine.name}")
+                (0..1).each do |port|
+                  (0..1).each do |device|
+                    if d["uuid"] == vm_info["#{controller}-ImageUUID-#{port}-#{device}"]
```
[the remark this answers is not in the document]
jbonhag: This is now possible (and implemented!) because I gave up on trying to support duplicate ISO attachments on a single guest. https://github.com/hashicorp/vagrant/pull/11613/files#diff-3494e15e76d59a15896961f458368a54R76

plugins/providers/virtualbox/cap/cleanup_disks.rb
jbonhag: Sounds good! I added these checks originally because I didn't want to have to mock out an empty dvd_meta hash for the disk tests, and vice versa. But adding the type check seems to have revealed some errors in the tests anyways, so I'm happy to continue down that road.

plugins/kernel_v2/config/vm.rb
```diff
@@ -849,6 +849,13 @@ def validate(machine, ignore_provider=nil)
                            name: duplicate_names)
         end
 
+        disk_files = @disks.map { |d| d.file }
+        duplicate_files = disk_files.detect { |d| disk_files.count(d) > 1 }
```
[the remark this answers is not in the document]
jbonhag: Makes sense, thanks.

plugins/providers/virtualbox/cap/cleanup_disks.rb
jbonhag: Thanks for the suggestion! Added `StorageControllerArray#get_primary_attachment`.

plugins/providers/virtualbox/model/storage_controller_array.rb
```diff
@@ -4,9 +4,17 @@ module Model
       # A collection of storage controllers. Includes finder methods to look
       # up a storage controller by given attributes.
       class StorageControllerArray < Array
+        SATA_TYPE = "SATA".freeze
+        IDE_TYPE = "IDE".freeze
+        SUPPORTED_TYPES = [SATA_TYPE, IDE_TYPE].freeze
+
+        # TODO: hook into ValidateDiskExt capability
+        DEFAULT_DISK_EXT = [".vdi", ".vmdk", ".vhd"].map(&:freeze).freeze
```
chrisroberts, line 12: You can just grab the values from the constant directly:

```ruby
DEFAULT_DISK_EXT = Cap::ValidateDiskExt::DEFAULT_DISK_EXT
```

briancain: It would also be good to merge this soon: https://github.com/hashicorp/vagrant/pull/11541
I've renamed that to more accurately reflect its value. There's `DEFAULT_DISK_EXT_LIST` and `DEFAULT_DISK_EXT`.

plugins/providers/virtualbox/cap/configure_disks.rb
jbonhag: I moved this into an reader in the StorageController class, which gets set when the storage controller is initialized.

plugins/providers/virtualbox/model/storage_controller_array.rb
jbonhag: Cool! I just hooked into `VagrantPlugins::ProviderVirtualBox::Cap::ValidateDiskExt.validate_disk_ext` so it should not be affected by the rename.

templates/locales/en.yml
```diff
@@ -1663,11 +1663,30 @@ en:
...
+      virtualbox_disks_controller_not_found: |-
+        Vagrant expected to find a storage controller called '%{name}',
+        but there is no controller with this name attached to the current VM.
+        If you have changed or removed any storage controllers, please restore
+        them to their previous configuration.
+      virtualbox_disks_no_supported_controllers: |-
+        Vagrant was unable to detect a disk controller with any of the
+        following supported types:
+
+        {supported_types}
```
soapy1, line 1675:
```suggestion
        %{supported_types}
```

jbonhag: Ah, thanks!

plugins/providers/virtualbox/model/storage_controller.rb
```diff
@@ -0,0 +1,135 @@
+module VagrantPlugins
+  module ProviderVirtualBox
+    module Model
+      # Represents a storage controller for VirtualBox. Storage controllers
+      # have a type, a name, and can have hard disks or optical drives attached.
+      class StorageController
+        IDE_CONTROLLER_TYPES = ["PIIX4", "PIIX3", "ICH6"].map(&:freeze).freeze
+        SATA_CONTROLLER_TYPES = ["IntelAhci"].map(&:freeze).freeze
+        SCSI_CONTROLLER_TYPES = [ "LsiLogic", "BusLogic"].map(&:freeze).freeze
```
soapy1, line 9: 😻

PR #11614, opened 2020-05-12
Fix #9552: Install rsync on Haiku guests

### LINKED ISSUE #9552: Haiku capability: install rsync
The Haiku plugin should be capable of installing rsync: The command is `installoptionalpackage -a rsync`. This works for Haiku OS version alpha4.

Another package manager is being developed for Haiku OS beta1, though it is not yet available.

https://www.haiku-os.org/guides/daily-tasks/installoptionalpackage/

https://www.haiku-os.org/guides/daily-tasks/install-applications

jbonhag:
This PR creates guest capabilities for HaikuOS to install rsync and perform basic synchronization. I created a Vagrant box for testing here: https://app.vagrantup.com/jbonhag/boxes/haiku-x86_64

Caveat:
Running rsync against a destination under the root directory (`/vagrant`)will create directories but not files, so you will see errors like this:

```
rsync: open "/vagrant/Vagrantfile" failed: Invalid Argument (-2147483643)
```

As a workaround, we can create a symlink to e.g. `/boot/home/vagrant` from `/vagrant` using a similar design to the way we create synthetic firmlinks for macOS 10.15 guests. This can be done under a different PR.

Update: Testing box has been updated with the following default synced folder:

```ruby
config.vm.synced_folder ".", "/boot/home/vagrant", type: "rsync"
```

### DIFF
--- a/plugins/guests/haiku/cap/rsync.rb
+++ b/plugins/guests/haiku/cap/rsync.rb
@@ -1,0 +1,15 @@
+module VagrantPlugins
+  module GuestHaiku
+    module Cap
+      class RSync
+        def self.rsync_installed(machine)
+          machine.communicate.test("test -f /bin/rsync")
+        end
+
+        def self.rsync_install(machine)
+          machine.communicate.execute("pkgman install -y rsync")
+        end
+      end
+    end
+  end
+end
--- a/plugins/guests/haiku/cap/rsync.rb
+++ b/plugins/guests/haiku/cap/rsync.rb
@@ -9,6 +9,10 @@ module VagrantPlugins
         def self.rsync_install(machine)
           machine.communicate.execute("pkgman install -y rsync")
         end
+
+        def self.rsync_command(machine)
+          "rsync -zz"
+        end
       end
     end
   end
--- a/plugins/guests/haiku/plugin.rb
+++ b/plugins/guests/haiku/plugin.rb
@@ -40,6 +40,11 @@ module VagrantPlugins
         require_relative "cap/rsync"
         Cap::RSync
       end
+
+      guest_capability(:haiku, :rsync_command) do
+        require_relative "cap/rsync"
+        Cap::RSync
+      end
     end
   end
 end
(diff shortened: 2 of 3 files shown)


chrisroberts:
Should the root directory support files being written to it in this way and it's a bug, or does the system actually implement something like firmlinks that's causing the problem?

jbonhag:
I don't think you're really supposed to touch the root directory, reading between the lines on https://www.haiku-os.org/docs/userguide/en/filesystem-layout.html:

> Files that you'd like to share with other users in a future multi-user environment have do be put outside /boot/home/. For example, you could create a folder /boot/all-users/ and put the stuff there.

Running an rsync command right on the guest, like `rsync -azv /boot/home /home-backup`, produces the same "Invalid Argument (-2147483643)" error, but `rsync -azv /boot/home /boot/home-backup` works fine.

We could also avoid a special case for Haiku, and leave it up to the Vagrantfile to use a "safe" synced folder destination:

```ruby
config.vm.synced_folder ".", "/boot/home/vagrant", type: "rsync"
```

kallisti5:
Hello! Haiku developer here.   / is read-only by design.  Haiku works by the following model:

1. Everything is read-only
2. Disks are mounted under /Name
3. /boot is always a symlink to the boot disk    /boot -> /Name
4. Some pre-configured symlinks exist at / for compatibility   (/bin -> /boot/system/bin)
5. Your home directory is writeable, as well as some special locations (placed for software packaged (eg /boot/system/package), places you configure the os, any directories marked as "non-packaged" (eg /boot/system/non-packaged/lib, /boot/system/non-packaged/bin, etc)

Software packages are "mounted over" the read-only system as read-only.

As a side note, i'm preparing official updated boxes for Haiku R1/Beta2 as I type this.  haiku-os/r1beta2-x86_64 , haiku-os/r1beta2-x86_gcc2

kallisti5:
Also, side note, as the person who first PR'ed Haiku support into vagrant.. these changes all look great to me :-) :+1:

jbonhag:
Hi @kallisti5,

Thank you so much for the explanation, and for taking the time to check out the PR. As a Haiku user from back in the day, I'm  excited to see the Vagrant guest plugin getting some attention. 😄 

Cheers!

Issue #10663, opened 2019-02-11
Malformed version number string trusty64


philippgille:
### Vagrant version

Vagrant 2.2.3

### Host operating system

Ubuntu 14.04

### Expected behavior

A list of the boxes on my machine should be shown.

### Actual behavior

The following error occurs:

```
/opt/vagrant/embedded/lib/ruby/2.4.0/rubygems/version.rb:208:in `initialize': Malformed version number string trusty64 (ArgumentError)
	from /opt/vagrant/embedded/lib/ruby/2.4.0/rubygems/version.rb:199:in `new'
	from /opt/vagrant/embedded/lib/ruby/2.4.0/rubygems/version.rb:199:in `new'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/lib/vagrant/box_collection.rb:251:in `block in all'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/lib/vagrant/box_collection.rb:250:in `each'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/lib/vagrant/box_collection.rb:250:in `sort_by'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/lib/vagrant/box_collection.rb:250:in `sort_by!'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/lib/vagrant/box_collection.rb:250:in `all'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/plugins/commands/box/command/list.rb:25:in `execute'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/plugins/commands/box/command/root.rb:66:in `execute'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/lib/vagrant/cli.rb:58:in `execute'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/lib/vagrant/environment.rb:291:in `cli'
	from /opt/vagrant/embedded/gems/2.2.3/gems/vagrant-2.2.3/bin/vagrant:182:in `<main>'
```

### Steps to reproduce

1. `vagrant box list`

### References

There are 14 related issues, you can find them with this search:

- https://github.com/hashicorp/vagrant/search?o=desc&q="Malformed+version+number+string"&s=created&type=Issues

[...]

PR #11618, opened 2020-05-13
Treat an empty box value as invalid

jbonhag:
This fixes an issue where having a box name set to an empty string will
cause all Vagrant commands to fail with an error like:

> ArgumentError: Malformed version number string (random box name)

This may be related to #10663.

### DIFF
--- a/plugins/kernel_v2/config/vm.rb
+++ b/plugins/kernel_v2/config/vm.rb
@@ -511,7 +511,7 @@ module VagrantPlugins
         @base_mac = nil if @base_mac == UNSET_VALUE
         @base_address = nil if @base_address == UNSET_VALUE
         @boot_timeout = 300 if @boot_timeout == UNSET_VALUE
-        @box = nil if @box == UNSET_VALUE
+        @box = nil if @box == UNSET_VALUE || @box.to_s.empty?
         @ignore_box_vagrantfile = false if @ignore_box_vagrantfile == UNSET_VALUE
 
         if @box_check_update == UNSET_VALUE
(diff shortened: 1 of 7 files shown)


plugins/kernel_v2/config/vm.rb
```diff
@@ -455,7 +455,7 @@ def finalize!
         @base_mac = nil if @base_mac == UNSET_VALUE
         @base_address = nil if @base_address == UNSET_VALUE
         @boot_timeout = 300 if @boot_timeout == UNSET_VALUE
-        @box = nil if @box == UNSET_VALUE
+        @box = nil if @box == UNSET_VALUE || @box.to_s.empty?
```
briancain, line 458: Instead, we should check the value of `@box` inside the `#validate` method, and then we can give the user a proper error if they've set the value to an empty string.

plugins/kernel_v2/config/vm.rb
```diff
@@ -732,6 +732,10 @@ def validate(machine, ignore_provider=nil)
           errors << I18n.t("vagrant.config.vm.clone_and_box")
         end
 
+        if box && box.empty?
+          errors << I18n.t("vagrant.config.vm.box_empty")
```
briancain, line 736:
```suggestion
          errors << I18n.t("vagrant.config.vm.box_empty", machine_name: machine.name)
```

templates/locales/en.yml
```diff
@@ -1980,6 +1980,7 @@ en:
           Checksum type specified but "box_download_checksum" is blank
         box_download_checksum_notblank: |-
           Checksum specified but must also specify "box_download_checksum_type"
+        box_empty: "Box value is an empty string."
```
briancain, line 1983:
```suggestion
        box_empty: "Box value for guest '%{machine_name}' is an empty string."
```

jbonhag:
@briancain Thanks for doing the work for me! 🎉 📈 I added one more "nil or empty" check but this should be good to go now.

Issue #9495, opened 2018-02-21
NFS synced_folder with mount_options doesn't work on Windows machine


DzeryCZ:
I have an issue with `nfs` synced_folder on Windows 10 machine. 

If I have configuration like this:
```config.vm.synced_folder "../myproject", "/myproject", type: "nfs"```

Everything works fine. But when I want to try play around performance and do some mount_options changes. E.g. 
```config.vm.synced_folder "../myproject", "/myproject", type: "nfs", mount_options: ['rw', 'vers=3', 'tcp']```

It always returns:

```
default: Mounting shared folders...
default: /myproject => C:/Users/j.zivny/projects/myproject
Vagrant was unable to mount VirtualBox shared folders. This is usually
because the filesystem "vboxsf" is not available. This filesystem is
made available via the VirtualBox Guest Additions and kernel module.
Please verify that these guest additions are properly installed in the
guest. This is not a bug in Vagrant and is usually caused by a faulty
Vagrant box. For context, the command attempted was:

mount -t vboxsf -o rw,vers=3,tcp,dmode=777,fmode=777,uid=1000,gid=1000 myproject  /myproject 

The error output from the command was:

  fmask     =<arg> umask of regular files
```

When I delete `mount_options` parameter, everything works.

My software: 
- Vagrant 2.0.1
- VirtualBox 5.2.4 r119785
- Host: Windows 10 1709
- Guest: Ubuntu 14.04

Any idea?

adamreid:
The documentation for the [NFS](https://www.vagrantup.com/docs/synced-folders/nfs.html) synchronized folders states that 

> Windows users: NFS folders do not work on Windows hosts. Vagrant will ignore your request for NFS synced folders on Windows.

So what you were observing was that the sync folder was working with `vboxsf` without options. When you added options intended for `nfs` those got passed to `vboxsf` instead which did not understand them.

chrisroberts:
When synced folder type is not valid given the host/guest platforms, and error should be raised instead of silently ignoring.

PR #11631, opened 2020-05-21
Error nfs on win

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/9495

### DIFF
--- a/plugins/kernel_v2/config/vm.rb
+++ b/plugins/kernel_v2/config/vm.rb
@@ -601,11 +601,6 @@ module VagrantPlugins
             options[:type] = :nfs
           end
 
-          # Ignore NFS on Windows
-          if options[:type] == :nfs && Vagrant::Util::Platform.windows?
-            options.delete(:type)
-          end
-
           if options[:hostpath]  == '.'
             current_dir_shared = true
           end


plugins/kernel_v2/config/vm.rb
```diff
@@ -601,11 +601,6 @@ def finalize!
             options[:type] = :nfs
           end
 
-          # Ignore NFS on Windows
-          if options[:type] == :nfs && Vagrant::Util::Platform.windows?
-            options.delete(:type)
-          end
```
soapy1, line 607: bringing a machine up on windows with nfs will have have an output something like

```
C:\Users\vagrant\test>vagrant up ubuntu
Bringing machine 'ubuntu' up with 'hyperv' provider...
==> ubuntu: Verifying Hyper-V is enabled...
==> ubuntu: Verifying Hyper-V is accessible...
    ubuntu: Configuring the VM...
==> ubuntu: Starting the machine...
==> ubuntu: Waiting for the machine to report its IP address...
    ubuntu: Timeout: 120 seconds
    ubuntu: IP: 192.168.64.237
==> ubuntu: Waiting for machine to boot. This may take a few minutes...
    ubuntu: SSH address: 192.168.64.237:22
    ubuntu: SSH username: vagrant
    ubuntu: SSH auth method: private key
==> ubuntu: Machine booted and ready!
It appears your machine doesn't support NFS, or there is not an
adapter to enable NFS on this machine for Vagrant. Please verify
that `nfsd` is installed on your machine, and try again. If you're
on Windows, NFS isn't supported. If the problem persists, please
contact Vagrant support.
```

should Vagrant be able to identify that a synced folder is not supported before the machine boots as part of validating the config?

chrisroberts: Yes, I do think that it should be running a precheck prior to this point, but for now this will keep things consistent and working as expected.

PR #11645, opened 2020-05-26
[docker provider] When using podman docker emulation, correctly parse ssh info

phil-lopreiato:
I believe I've run into another slight incompatibility when using podman's docker emulation on fedora. 

The output of `docker inspect` (which the ssh command uses to get the port over which to connect) returns a slightly different format. 

This diff will teach `vagrant` to understand this new format and correctly parse out the port.

I've tested this on my environment locally, as well as written a unit test.

Please forgive any awful code, this is my first time writing ruby :)

### DIFF
--- a/plugins/providers/docker/provider.rb
+++ b/plugins/providers/docker/provider.rb
@@ -146,18 +146,26 @@ module VagrantPlugins
         # If the container isn't running, we can't SSH into it
         return nil if state.id != :running
 
-        port_name = "#{@machine.config.ssh.guest_port}/tcp"
+        guest_port = @machine.config.ssh.guest_port
+        port_name = "#{guest_port}/tcp"
         network   = driver.inspect_container(@machine.id)['NetworkSettings']
 
-        if network["Ports"][port_name].respond_to?(:first)
+        if driver.podman?
+          port = network["Ports"].detect {|p| p["protocol"] == "tcp" and p["containerPort"] == guest_port }
+          if port
+            port_info = {
+              "HostIp" => port["hostIP"],
+              "HostPort" => port["hostPort"]
+            }
+          end
+        elsif network["Ports"][port_name].respond_to?(:first)
           port_info = network["Ports"][port_name].first
         else
           ip = network["IPAddress"]
-          port = @machine.config.ssh.guest_port
           if !ip.to_s.empty?
             port_info = {
               "HostIp" => ip,
-              "HostPort" => port
+              "HostPort" => guest_port
             }
           end
         end
--- a/test/unit/plugins/providers/docker/provider_test.rb
+++ b/test/unit/plugins/providers/docker/provider_test.rb
@@ -18,6 +18,7 @@ describe VagrantPlugins::DockerProvider::Provider do
   before do
     stub_const("Vagrant::Util::Platform", platform)
     allow(machine).to receive(:id).and_return("foo")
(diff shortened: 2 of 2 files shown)


plugins/providers/docker/provider.rb
```diff
@@ -146,18 +146,26 @@ def ssh_info
...
         return nil if state.id != :running
 
-        port_name = "#{@machine.config.ssh.guest_port}/tcp"
+        guest_port = @machine.config.ssh.guest_port
+        port_name = "#{guest_port}/tcp"
         network   = driver.inspect_container(@machine.id)['NetworkSettings']
 
-        if network["Ports"][port_name].respond_to?(:first)
+        if driver.podman?
+          port = network["Ports"].select {|p| p["protocol"] == "tcp" and p["containerPort"] == guest_port } .first
```
chrisroberts, line 154: You can drop the `.first` by using a `.detect` instead of `.select` here:

```ruby
port = network["Ports"].detect {|p| p["protocol"] == "tcp" and p["containerPort"] == guest_port }
```

phil-lopreiato: amazing!

phil-lopreiato:
forced pushed with a rebase + @chrisroberts's suggestion

test/unit/plugins/providers/docker/provider_test.rb
```diff
@@ -127,5 +130,17 @@
...
+      allow(provider_config).to receive(:compose).and_return(false)
+      allow(platform).to receive(:windows?).and_return(false)
+      allow(platform).to receive(:darwin?).and_return(false)
+      allow(driver_obj).to receive(:podman?).and_return(true)
+      allow(driver_obj).to receive(:created?).and_return(true)
+      allow(driver_obj).to receive(:state).and_return(:running)
+      allow(driver_obj).to receive(:inspect_container).and_return(podman_network_settings)
+
+      expect(subject.ssh_info).to eq({:host=> "127.0.0.1", :port => 2222})
+    end
```
soapy1, line 144: looks great! could you add a test to ensure that with podman emulation `nil` is returned if the port info is empty. It should be similar to https://github.com/hashicorp/vagrant/blob/7c1997e366317a79da0a90f6e641f04297497aa9/test/unit/plugins/providers/docker/provider_test.rb#L110-L120

one way to group the podman related tests is to use `context`, something like:
```
  context "podman emulation" do
    before do
      allow(driver_obj).to receive(:podman?).and_return(true)
    end
 
    it "should receive a valid address" do ..... end

    it "returns nil if a port info is nil from the drive" do ..... end
  end 
```

phil-lopreiato:
rebased + forced pushed with @soapy1's comments

test/unit/plugins/providers/docker/provider_test.rb
phil-lopreiato: you got it!

phil-lopreiato:
I think we may also need @chrisroberts to take another look here, since I can't dissmiss his prior review even after updating the PR?

phil-lopreiato:
rebase + ping @chrisroberts

phil-lopreiato:
So, turns out this was fixed upstream in https://github.com/containers/libpod/commit/f586c006f8428a04fce5a5c7ae6b921e6337ebe6 which is a strictly better solution than what I've got here :)

PR #11650, opened 2020-05-27
Fixes #11594: Clarify provisioning guide

### LINKED ISSUE #11594: DOCS Getting started - Provisioning, minor change.
Not sure where to file this. But the docs are wrong.

https://www.vagrantup.com/intro/getting-started/provisioning.html

bootstrap.sh
apache default folder is /var/www/html not /var/www.

briancain:
This commit adds some clarification to the provisioning guide by
explaining ahead of time that an html folder will be created that will
be used by the provisioning script.


jbonhag:
What do you think about creating the `html` content before doing anything else? This might mirror a practitioner experience of having written some web content and then wanting to test it out in a Vagrant box.

briancain:
@jbonhag - I thought about that too but initially didn't want to change too much, but I think it makes sense. I updated the PR now.

Issue #11214, opened 2019-11-21
Vagrant should recover from missing global data in machine index file


briancain:
In some cases, it's possible to have Vagrant fail and not be able to add a local guests metadata to the global state file that keeps track of all running guests on a host machine (at `~/.vagrant.d/data/machine-index/index`). Although the guest still exists and is being tracked locally inside the project dirs `.vagrant` folder, it's missing from `global-status`. This issue should tackle trying to recover from this bad state so that the guest is added back into the global state file.

### Vagrant version

Any version

### Host operating system

n/a

### Guest operating system

n/a

### Vagrantfile

n/a

### Debug output

n/a

### Expected behavior

Vagrant should attempt to re-add the locally defined guest into the global state file if it is still valid locally.

### Actual behavior

The guest remains missing from `global-status`.

### Steps to reproduce

1. Bring up a guest
2. Delete its entry or the entire global index file at `~/.vagrant.d/data/machine-index/index`
3. Run `vagrant global-status` and see that it's missing

### References
Are there any other GitHub issues (open or closed) that should be linked here?
For example:
- GH-11187

mprey:
I can try working on this issue. Do you have any pointers for getting started?

PR #11656, opened 2020-05-29
Add local machine to global index of machines if it does not exist

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11214

### DIFF
--- a/lib/vagrant/plugin/v2/command.rb
+++ b/lib/vagrant/plugin/v2/command.rb
@@ -230,6 +230,10 @@ module Vagrant
           color_index = 0
 
           machines.each do |machine|
+            if machine.state && machine.state.id != :not_created && !@env.machine_index.include?(machine.index_uuid)
+              machine.recover_machine(machine.state.id)
+            end
+
             # Set the machine color
             machine.ui.opts[:color] = color_order[color_index % color_order.length]
             color_index += 1
--- a/lib/vagrant/machine.rb
+++ b/lib/vagrant/machine.rb
@@ -556,6 +556,41 @@ module Vagrant
       result
     end
 
+    # Returns the state of this machine. The state is queried from the
+    # backing provider, so it can be any arbitrary symbol.
+    #
+    # @param [Symbol] state of machine
+    # @return [Entry] entry of recovered machine
+    def recover_machine(state)
+      entry = @env.machine_index.get(index_uuid)
+      if entry
+        @env.machine_index.release(entry)
+        return entry
+      end
+
+      entry = MachineIndex::Entry.new(id=index_uuid, {})
+      entry.local_data_path = @env.local_data_path
+      entry.name = @name.to_s
+      entry.provider = @provider_name.to_s
+      entry.state = state
+      entry.vagrantfile_path = @env.root_path
+      entry.vagrantfile_name = @env.vagrantfile_name
+
+      if @box
+        entry.extra_data["box"] = {
+          "name"     => @box.name,
+          "provider" => @box.provider.to_s,
+          "version"  => @box.version.to_s,
+        }
+      end
+
+      @state_mutex.synchronize do
+        entry = @env.machine_index.recover(entry)
+        @env.machine_index.release(entry)
+      end
+      return entry
+    end
+
     # Returns the user ID that created this machine. This is specific to
     # the host machine that this was created on.
     #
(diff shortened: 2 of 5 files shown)


lib/vagrant/plugin/v2/command.rb
```diff
@@ -230,6 +230,10 @@ def with_target_vms(names=nil, options=nil)
           color_index = 0
 
           machines.each do |machine|
+            if machine.state && machine.state.id != :not_created
```
soapy1, line 233: I don't know about this part. This will update the global index for pretty much any Vagrant command that is run. For large environments this might be non-negligibly large?

briancain: Hmm, it seems like it would be better to look at what state is currently in the global index and compare it to what machines we know about, then adjust with the delta there. Then we wouldn't be doing this operation for every machine, only ones missing from the index file.

briancain: And also for 99% of the time when there isn't anything missing, this will just become a noop which should save us any compute time

lib/vagrant/plugin/v2/command.rb
```diff
@@ -230,6 +230,10 @@ def with_target_vms(names=nil, options=nil)
           color_index = 0
 
           machines.each do |machine|
+            if machine.state && machine.state.id != :not_created && !@env.machine_index.include?(machine.id)
```
briancain, line 233: The last part of this if-statement needs to be updated. I had a running machine that existed in the global machine index, but it kept being called to recover despite being in the json file already. The actual id `machine.id` doesn't match with how we store it in the global index file, so that check will always be true and we'll always recover:

```ruby
[2] pry(#<VagrantPlugins::CommandStatus::Command>)> @env.machine_index
=> #<Vagrant::MachineIndex:0x0000561c5563fb38
 @data_dir=#<Pathname:/home/brian/.vagrant.d/data/machine-index>,
 @index_file=#<Pathname:/home/brian/.vagrant.d/data/machine-index/index>,
 @lock=
  #<Monitor:0x0000561c5563ed50
   @mon_count=0,
   @mon_mutex=#<Thread::Mutex:0x0000561c5563eaf8>,
   @mon_mutex_owner_object_id=47339845842600,
   @mon_owner=nil>,
 @machine_locks={},
 @machines=
  {"a8756e374a7745c08aac72797a60522e"=>
    {"local_data_path"=>"/home/brian/code/vagrant-sandbox/vagrant-installers/.vagrant",
     "name"=>"ubuntu-14.04",
     "provider"=>"vmware_desktop",
     "state"=>"running",
     "vagrantfile_name"=>nil,
     "vagrantfile_path"=>"/home/brian/code/vagrant-sandbox/vagrant-installers",
     "updated_at"=>nil,
     "extra_data"=>
      {"box"=>
        {"name"=>"hashicorp-vagrant/ubuntu-14.04", "provider"=>"vmware_desktop", "version"=>"1.0.20180822143433"}}},
   "ae12c6cd2b7a4d0f8c3c758e2db80570"=>
    {"local_data_path"=>"/home/brian/code/vagrant-sandbox/.vagrant",
     "name"=>"bork",
     "provider"=>"virtualbox",
     "state"=>"running",
     "vagrantfile_name"=>nil,
     "vagrantfile_path"=>"/home/brian/code/vagrant-sandbox",
     "updated_at"=>nil,
     "extra_data"=>
      {"box"=>{"name"=>"elastic/ubuntu-16.04-x86_64", "provider"=>"virtualbox", "version"=>"20200607.0.0"}}}}>
[9] pry(#<VagrantPlugins::CommandStatus::Command>)> !@env.machine_index.include?(machine.id)
=> true
[10] pry(#<VagrantPlugins::CommandStatus::Command>)> machine.id
=> "f6ded812-5402-44ec-956e-8493109ed32e"
```

I'm not sure why there's a difference here :thinking:

briancain: I think it's because `machine.id` != the machines uuid?  I think maybe we'd want `machine.index_uuid` here

briancain: This looks right:

```ruby
[10] pry(#<VagrantPlugins::CommandStatus::Command>)> machine.id
=> "f6ded812-5402-44ec-956e-8493109ed32e"
[11] pry(#<VagrantPlugins::CommandStatus::Command>)> machine.index_uuid
=> "ae12c6cd2b7a4d0f8c3c758e2db80570"
```

soapy1: good catch, thanks!

PR #11676, opened 2020-06-05
Introduce cloud_init config for Vagrant

briancain:
This commit adds the basic config class for defining cloud_init for a
given guest.


website/pages/docs/Cloud-Init/usage.mdx
```diff
@@ -0,0 +1,72 @@
...
+supported providier for this feature.
+
+Below are some very simple examples of how to use Vagrant Disks with the VirtualBox provider.
+
+## Basic Examples
+
+Here is an example containing two parts of user data:
+
+```ruby
+config.vm.cloud_init do |cloud_init|
```
briancain, line 34: So, although the RFC mentions having a method that is just called `add_user_config` that simply sets values for the cloud_init config class. I think this is an anti-pattern here. Current `config.vm` options such as `provision`, `synched_folder`, and `disk` configs are contained to a single ruby block, rather than multiple configs inside one ruby block. Not using a `add_user_config` here falls in line with the existing pattern Vagrant has defined for storing configuration. And having a config method/option called `add_user_config` seems redundant here, and makes the implementation for storing these config values more complicated than it needs to be if it's going to stay a `config.vm` option. Hopefully that makes sense for why I did that!

jbonhag: Thanks for the explanation! The method should actually have been called `add_user_data`, which makes it a little less redundant. I saw this as somewhat akin to the Docker provisioner, which has a single configuration block that performs the same operation multiple times with different arguments:

```ruby
  config.vm.provision "docker" do |d|
    d.pull_images "ubuntu"
    d.pull_images "vagrant"
  end
end
```

The single block form also makes it possible to add other types of things to the cloud-init configuration in the future, like explicitly defining metadata:

```ruby
config.vm.cloud_init do |cloud_init|
  cloud_init.add_metadata instance_id: "i-95576fda42cf750a28"
end
```

In general, though, since we are producing a single cloud-init configuration, I think we should try to define that within the scope of a single config block.

I'm also sort of assuming that people are much less likely to use `inline` than `path`. What if we simplified the configuration to something like this?

```ruby
config.vm.cloud_init do |cloud_init|
  cloud_init.add_user_data "provision.sh", content_type: "text/x-shellscript"
  cloud_init.add_user_data "cloud.cfg", content_type: "text/cloud-config"
end
```

briancain: > Thanks for the explanation! The method should actually have been called `add_user_data`, which makes it a little less redundant. I saw this as somewhat akin to the Docker provisioner, which has a single configuration block that performs the same operation multiple times with different arguments:
> 
> ```ruby
> Vagrant.configure("2") do |config|
>   config.vm.provision "docker" do |d|
>     d.pull_images "ubuntu"
>     d.pull_images "vagrant"
>   end
> end
> ```

Ah, sorry! I mistyped the name from the RFC. But I think my point still stands, we're already adding the data by defining it and storing it into the config, so I guess I just don't see it as a useful thing here and feels kind of redundant to me.

> 
> The single block form also makes it possible to add other types of things to the cloud-init configuration in the future, like explicitly defining metadata:
> 
> ```ruby
> config.vm.cloud_init do |cloud_init|
>   cloud_init.add_metadata instance_id: "i-95576fda42cf750a28"
> end
> ```

This could be done with a metadata hash that takes key=>values such as instance_id:

```ruby
config.vm.cloud_init do |cloud_init|
  cloud_init.metadata = {instance_id: "i-95576fda42cf750a28"}
end
```

> 
> In general, though, since we are producing a single cloud-init configuration, I think we should try to define that within the scope of a single config block.
> 
> I'm also sort of assuming that people are much less likely to use `inline` than `path`. What if we simplified the configuration to something like this?

I think it's better to be explicit when setting the path key rather than leaving it off.

> 
> ```ruby
> config.vm.cloud_init do |cloud_init|
>   cloud_init.add_user_data "provision.sh", content_type: "text/x-shellscript"
>   cloud_init.add_user_data "cloud.cfg", content_type: "text/cloud-config"
> end
> ```

plugins/kernel_v2/config/disk.rb
```diff
@@ -10,7 +10,7 @@ class VagrantConfigDisk < Vagrant.plugin("2", :config)
       # Config class for a given Disk
       #-------------------------------------------------------------------
 
-      DEFAULT_DISK_TYPES = [:disk, :dvd, :floppy].freeze
+      DEFAULT_DISK_TYPES = [:disk, :dvd, :floppy].map(&:freeze).freeze
```
chrisroberts, line 13: The symbols themselves don't need to be frozen as they already are.

plugins/kernel_v2/config/vm_trigger.rb
```diff
@@ -9,7 +9,7 @@ class VagrantConfigTrigger < Vagrant.plugin("2", :config)
       # Defaults
       DEFAULT_ON_ERROR = :halt
       DEFAULT_EXIT_CODE = 0
-      VALID_TRIGGER_TYPES = [:command, :action, :hook].freeze
+      VALID_TRIGGER_TYPES = [:command, :action, :hook].map(&:freeze).freeze
```
chrisroberts, line 12: Symbols don't need to be frozen

plugins/kernel_v2/config/disk.rb
briancain: Oooh alright, makes sense! I thought maybe we had just missed these.

website/pages/docs/cloud-init/usage.mdx
```diff
@@ -0,0 +1,72 @@
...
+supported or functional.
+
+This feature currently reqiures the experimental flag to be used. To explicitly enable this feature, you can set the experimental flag to:
+
+```
+VAGRANT_EXPERIMENTAL="cloud_init"
+```
+
+Please note that `VAGRANT_EXPERIMENTAL` is an environment variable. For more
+information about this flag visit the [Experimental docs page](/docs/experimental/)
```
soapy1, line 21: the experimental link goes to 404, I think the last `/` needs to be dropped
```suggestion
information about this flag visit the [Experimental docs page](/docs/experimental)
```

plugins/kernel_v2/config/vm.rb
```diff
@@ -153,6 +156,28 @@ def merge(other)
           end
           result.instance_variable_set(:@disks, new_disks)
 
+          # Merge defined cloud_init_configs
+          other_cloud_init_configs = other.instance_variable_get(:@cloud_init_configs)
```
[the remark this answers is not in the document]
briancain: It should, as it's also identical to the trigger and provisioner merge. But I think there's probably a larger issue here that needs to be addressed for how Vagrant parses config options that I think should probably happen in its own PR :thinking:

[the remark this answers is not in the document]
briancain: I made an internal issue for tracking this work.

website/pages/docs/cloud-init/configuration.mdx
```diff
@@ -0,0 +1,34 @@
+---
+layout: docs
+page_title: Vagrant Cloud-Init Configuration
+sidebar_title: Configuration
+description: Documentation of various configuration options for Vagrant cloud-init
+---
+
+# Configuration
+
+Vagrant cloud-init has several options that allow users to define and attach disks to guests.
```
chrisroberts: Just the last bit of this line.

website/pages/docs/cloud-init/index.mdx
```diff
@@ -0,0 +1,19 @@
...
+page_title: Cloud-Init
+sidebar_title: Cloud-Init
+description: Introduction to using cloud-init with Vagrant
+---
+
+# Vagrant cloud-init
+
+~> **Warning!** This feature is experimental and may break or
+change in between releases. Use at your own risk. It currently is not officially
+supported or functional. Please refer to the providier specific disk documentation
```
chrisroberts, line 12: `providier` -> `provider`

website/pages/docs/cloud-init/index.mdx
```diff
@@ -0,0 +1,19 @@
...
+for more information on how to use and enable this feature.
+
+For examples on how to achieve this, among other use cases, please refer to the [usage](/docs/cloud-init/usage)
+guide for more information!
+
+For more information about what options are available for configuring disks, see the
```
chrisroberts, line 18: Disks listed here.

website/pages/docs/cloud-init/usage.mdx
```diff
@@ -0,0 +1,77 @@
...
+```
+VAGRANT_EXPERIMENTAL="cloud_init"
+```
+
+Please note that `VAGRANT_EXPERIMENTAL` is an environment variable. For more
+information about this flag visit the [Experimental docs page](/docs/experimental)
+for more info. Without this flag enabled, any disks defined will not be configured.
+
+Also note that the examples below use the VirtualBox provider, which is the current
+supported providier for this feature.
```
chrisroberts, line 25: `providier` -> `provider`

website/pages/docs/cloud-init/usage.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+This feature currently reqiures the experimental flag to be used. To explicitly enable this feature, you can set the experimental flag to:
+
+```
+VAGRANT_EXPERIMENTAL="cloud_init"
+```
+
+Please note that `VAGRANT_EXPERIMENTAL` is an environment variable. For more
+information about this flag visit the [Experimental docs page](/docs/experimental)
+for more info. Without this flag enabled, any disks defined will not be configured.
```
chrisroberts, line 22: > any disks...

website/pages/docs/cloud-init/usage.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+Also note that the examples below use the VirtualBox provider, which is the current
+supported providier for this feature.
+
+Below are some very simple examples of how to use Vagrant Disks with the VirtualBox provider.
```
chrisroberts, line 27: > Vagrant Disks...

plugins/kernel_v2/config/cloud_init.rb
```diff
@@ -49,6 +65,13 @@ def finalize!
       def validate(machine)
         errors = _detected_errors
 
+        if @type && @type != DEFAULT_CONFIG_TYPE
```
soapy1, line 68: Will other types be supported at some point as well?

briancain: It sounds like it's a maybe, but for the most part this type is to make it so that a new type _could_ exist in the future without breaking older Vagrantfiles if we do end up introducing a new kind of config. So for now there's really only one type, thus the check here! :smile:

Issue #11636, opened 2020-05-24
Vagrant OSX 2.2.9 issue: invalid mask 0.0.0.24 (IPAddr::InvalidPrefixError)


flavio-fernandes:
### Vagrant version

$ vagrant version
Installed Version: 2.2.9
Latest Version: 2.2.9

Also broken on 2.2.8

### Host operating system

OS-X
macOS Catalina Version 10.15.4

### Guest operating system

It does not matter

### Vagrantfile

https://github.com/flavio-fernandes/mqtt2cmd/blob/96db9fc96d52f2949a23eb21b3cb44c0c313d2ff/Vagrantfile

```ruby
# coding: utf-8
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Vagrantfile API/syntax version.
VAGRANTFILE_API_VERSION = "2"
Vagrant.require_version ">=1.7.0"

Vagrant.configure(2) do |config|
    config.vm.hostname = "foo"
    config.vm.box = "centos/7"
    config.vm.network 'private_network', ip: "192.168.123.123"
end
```

### Debug output

https://gist.github.com/flavio-fernandes/54efab0832c011621e3e1cc1fe1d4740

https://gist.github.com/flavio-fernandes/ca8872fecdc72f0aa03e6ca5df507306

### Expected behavior

```shell
from /opt/vagrant/embedded/lib/ruby/2.6.0/ipaddr.rb:161:in `mask'
/opt/vagrant/embedded/lib/ruby/2.6.0/ipaddr.rb:509:in `mask!': invalid mask 0.0.0.24 (IPAddr::InvalidPrefixError)
```

Should be able to handle default mask value.

What should have happened?

It should have worked, as it does on version 2.2.7

### Actual behavior

What actually happened?

### Steps to reproduce

on osx system

git clone https://github.com/flavio-fernandes/mqtt2cmd.git && \
cd mqtt2cmd && vagrant up

### References

broken in osx only?
works fine on libvirt provider

chrisroberts:
Thanks for reporting this issue. There appears to be an issue with the netmask value that's being read from VirtualBox. The version of Ruby used prior to the 2.2.8 release did not have any issue accepting the value. On the newer versions of Ruby, it is now throwing this error. We'll need to add a bit of value sanitizing on the netmask value in the VirtualBox driver to resolve this.

flavio-fernandes:
[...]

PR #11679, opened 2020-06-05
Validate netmask value from VirtualBox

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11636


lib/vagrant/util/network_ip.rb
```diff
@@ -3,11 +3,23 @@
 module Vagrant
   module Util
     module NetworkIP
+
+      DEFAULT_MASK = "255.255.255.0".freeze
+
+      def initialize
+        @logger = Log4r::Logger.new("vagrant::util::NetworkIP")
```
briancain, line 10: Quick question about this....it looks like the way this is used is by calling `include` on this class and invoking the class method `network_address` directly rather than on a new `NetworkIP` object. I'm not 100% sure about this, but would that bypass the initialize here and make `@logger` nil below?

soapy1: Manually testing it out seems to work. I also updated the tests so the logger bit gets run and it seems to work. That being said, when I included this module using `extend` instead of `include` then `@logger` will be `nil`. I think it has to do with how Ruby inserts classes into the object hierarchy when using `include` vs `extend`. 
I think if `network_address` is meant to be a static method then this will break? Maybe I'll chuck this in `network_address`?

briancain: Ok! That's what I wasn't sure about, the extend versus include. I always confuse the two! I'd say it doesn't hurt to move it into the `network_address` method to be safe, in case a caller ends up `extend`ing the library instead of `include`.

briancain: Also, in the past for classes like this we've ended up defining a "global" logger here. Not ideal, but then we could define the logger once at the top to be used for any method in the class. Something like this at the top:
```ruby
LOGGER = Log4r::Logger.new(...)
```

Then we can use `LOGGER` in the class without defining it multiple times :thinking:

lib/vagrant/util/network_ip.rb
```diff
@@ -3,11 +3,21 @@
...
+
       # Returns the network address of the given IP and subnet.
       #
       # @return [String]
       def network_address(ip, subnet)
-        IPAddr.new(ip).mask(subnet).to_s
+        begin
+          IPAddr.new(ip).mask(subnet).to_s
+        rescue IPAddr::InvalidPrefixError
+          LOGGER.info("Provided mask '#{subnet}' is invalid. Falling back to using mask '#{DEFAULT_MASK}'")
```
chrisroberts, line 18: Can you update this from `info` to `warn`

Issue #9163, opened 2017-11-09
vagrant rdp does not terminate until rdp session ends


DDSGooch:
When running _vagrant rdp_, the command hangs until the rdp session is completed or the command is terminated with _CTRL-C_. This means that you cannot do anything else in this command console, which is very annoying and I'm sure it isn't designed to work this way, nor how it used to work.

In the example below (debug log attached), the rdp session started, then was terminated with _CTRL-C_.
```
>vagrant rdp
==> default: Detecting RDP info...
    default: Address: 127.0.0.1:33389
    default: Username: vagrant
==> default: Vagrant will now launch your RDP client with the connection parameters
==> default: above. If the connection fails, verify that the information above is
==> default: correct. Additionally, make sure the RDP server is configured and
==> default: running in the guest machine (it is disabled by default on Windows).
==> default: Also, verify that the firewall is open to allow RDP connections.

>Exiting due to interrupt.
```

More detailed debug log also attached.

### Vagrant version
Vagrant 2.0.1

### Host operating system
Windows 7 x64

### Guest operating system
Windows Server 2012

### Vagrantfile
```ruby
Vagrant.require_version ">= 1.8.7"

ip = '192.168.56.3'

Vagrant.configure("2") do |config|
  config.vm.box = "test_box_prereqs"
  config.vm.synced_folder "share", "/vagrant"

  config.vm.network "forwarded_port", guest: 80, host: 8010,
    auto_correct: true

  config.vm.network "private_network", ip: ip, auto_config: false

  config.vm.network "public_network", auto_config: true, bridge: "Intel(R) Dual Band Wireless-AC 8260"

  config.vm.provider "virtualbox" do |vb, override|
    override.vm.provision "shell", inline: "New-NetIPAddress #{ip} -PrefixLength 24 -InterfaceAlias 'Ethernet 2'"
  end

  config.vm.provider "vmware_desktop" do |vmware, override|
    override.vm.provision "shell", inline: "New-NetIPAddress #{ip} -PrefixLength 24 -InterfaceAlias 'Ethernet1'"
  end

  config.vm.provider "virtualbox" do |v|
    v.memory = 4096
    v.cpus = 2
  end
  config.vm.provider "vmware_workstation" do |v|
    v.vmx["memsize"] = "4096"
    v.vmx["numvcpus"] = "2"
  end

end
```
### Debug output
[...]

PR #11691, opened 2020-06-09
Allow Vagrant to exit independently from RDP

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/9163

### DIFF
--- a/plugins/hosts/windows/cap/rdp.rb
+++ b/plugins/hosts/windows/cap/rdp.rb
@@ -21,8 +21,10 @@ module VagrantPlugins
           if rdp_info[:extra_args]
             args = rdp_info[:extra_args] + args
           end
-          # Launch it
-          Vagrant::Util::Subprocess.execute("mstsc", *args)
+          # Spawn it and detach from process. This will allow
+          # Vagrant to exit independent from the rdp session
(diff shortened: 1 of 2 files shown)


plugins/hosts/windows/cap/rdp.rb
```diff
@@ -21,8 +21,10 @@ def self.rdp_client(env, rdp_info)
           if rdp_info[:extra_args]
             args = rdp_info[:extra_args] + args
           end
-          # Launch it
-          Vagrant::Util::Subprocess.execute("mstsc", *args)
+          # Spawn it and detach from process. This will allow
+          # Vagrant to exit independent from the rdp session
+          pid = spawn("mstsc", *args)
```
soapy1, line 26: [childprocess 4.0.0](https://github.com/enkessler/childprocess/releases/tag/v4.0.0) supports detaching from a process in Windows. Might be nice to update that gem and make Vagrant detach from the process that way?

chrisroberts: I think that sounds like the best path, and once the dependency constraint is updated, adding in functionality to the subprocess util to handle detaching from the created process would be much nicer.

PR #11694, opened 2020-06-10
Build iso for Darwin host

soapy1:


plugins/hosts/darwin/cap/fs_iso.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+
+        # Check if source directory has any new updates
+        #
+        # @param [Pathname] iso_path Path to ISO file
+        # @param [Pathname] dir_path Path to source directory
+        # @return [Boolean]
+        def self.iso_update_required?(iso_path, dir_path)
+          Dir.glob(dir_path.join("**/**/*")).each do |path|
+            if File.mtime > iso_path.mtime
```
jbonhag: I think we need to add another boolean clause here. If iso_path is empty (i.e. newly created), it will have a later mtime than the source directory, so the initial ISO won't get built.

plugins/hosts/darwin/cap/fs_iso.rb
```diff
@@ -0,0 +1,63 @@
...
+            tmpfile = Tempfile.new("vagrant-iso")
+            file_destination = Pathname.new(tmpfile.path)
+            tmpfile.delete
+          else
+            file_destination = Pathname.new(file_destination.to_s)
+          end
+          source_directory = Pathname.new(source_directory)
+          if iso_update_required?(file_destination, source_directory)
+            # Ensure destination directory is available
+            FileUtils.mkdir_p(file_destination.to_s)
```
jbonhag, line 40:
```suggestion
            FileUtils.mkdir_p(File.dirname(file_destination.to_s))
```

plugins/hosts/darwin/cap/fs_iso.rb
soapy1: oo right. I think I'm going to extract this into some utility too.

test/unit/vagrant/util/directory_changed_test.rb
```diff
@@ -0,0 +1,40 @@
...
+  let(:subject){ Vagrant::Util::DirectoryChanged }
+  
+  describe ".directory_changed?" do
+
+    let(:dir_paths) { ["/one/two.ext", "/one"] }
+    
+    before do
+      allow(Dir).to receive(:glob).and_return(dir_paths)
+      allow(Pathname).to receive(:new) do |arg|
+        FakePathname.new(Time.now)
```
soapy1, line 26: I suspect there is a better way to mock `Pathname` in order to set the mtime to something predictable to test against... 🤔

plugins/hosts/darwin/cap/fs_iso.rb
```diff
@@ -0,0 +1,62 @@
...
+        #
+        # @param [Vagrant::Environment] env
+        # @param [String] source_directory Contents of ISO
+        # @param [String, nil] file_destination Location to store ISO
+        # @param [Map] extra arguments to pass to the iso building command
+        # @return [Pathname] ISO location
+        # @note If file_destination exists, source_directory will be checked
+        #       for recent modifications and a new ISO will be generated if requried.
+        def self.create_iso(env, source_directory, file_destination=nil, extra_opts={})
+          file_destination = ensure_output_file(ext=".iso", file_destination=file_destination)
```
jbonhag, line 35: I think there's an advantage to keeping this code in the cap, at least until we need a similar check some place else. If there's a concern that the `file_destination` doesn't end in `.iso`, we could also check `File.extname(file_destination)` here.

```suggestion
          if file_destination.nil?
            tmpfile = Tempfile.new(["vagrant", ".iso"])
            file_destination = Pathname.new(tmpfile.to_path)
            tmpfile.delete
          else
            file_destination = Pathname.new(file_destination.to_s)
            FileUtils.mkdir_p(File.dirname(file_destination))
          end
```

plugins/hosts/darwin/cap/fs_iso.rb
```diff
@@ -0,0 +1,62 @@
...
+          source_directory = Pathname.new(source_directory)
+
+          # If the destrination does not exist or there have been changes in the source directory since the last build, then build
+          if !file_destination.exist? || Vagrant::Util::DirectoryChanged.directory_changed?(source_directory, file_destination.mtime)
```
jbonhag, line 39: Thinking about this a little more, there are two distinct conditions when we need to build the ISO:

1. If the destination file doesn't exist
2. If the destination file exists, and source files have changed -- in this case we need to also delete the destination file so that `hdiutil` doesn't throw an error.

```suggestion
          if !file_destination.exist?
            build_required = true
          elsif file_destination.exist? && Vagrant::Util::DirectoryChanged.directory_changed?(source_directory, file_destination.mtime)
            File.delete(file_destination)
            build_required = true
          end

          if build_required
            @@logger.info("Building ISO from source #{source_directory}")
```

plugins/hosts/darwin/cap/fs_iso.rb
```diff
@@ -0,0 +1,62 @@
...
+        @@logger = Log4r::Logger.new("vagrant::host::darwin::fs_iso")
+
+        BUILD_ISO_CMD = "hdiutil".freeze
+
+        # Check that the host has the ability to generate ISOs
+        #
+        # @param [Vagrant::Environment] env
+        # @return [Boolean]
+        def self.isofs_available(env)
+          !!Vagrant::Util::Which.which("hdiutil")
```
jbonhag, line 22: This can use `BUILD_ISO_CMD` as well.

plugins/hosts/darwin/cap/fs_iso.rb
soapy1: My reasoning for pulling this out into a util was that a similar check will be needed for the linux and windows iso caps. What do you think about extracting this bit of code into this util (to be used by the darwin, linux and windows caps) and dropping the bit about ensuring the file ends in some extension?

plugins/hosts/darwin/cap/fs_iso.rb
soapy1: oh nice catch! I think this can be expressed more succinctly using the `hdiutil` option `-ov` which will overwrite the destination path if it already exists. I think that will cover both conditions.

plugins/hosts/darwin/cap/fs_iso.rb
jbonhag: The way I read it, the first check `if file_destination.nil?` is providing a default value for the `file_destination` argument to the method. The second clause is the one that could be better candidate to be extracted out, conceptually ("make sure my target destination exists"), but since that's not very much code I don't see a lot of value in making that a helper method.

lib/vagrant/util/directory_changed.rb
```diff
@@ -0,0 +1,23 @@
...
+        #               changed since this time
+        # @return [Boolean]
+        def self.directory_changed?(dir_path, threshold_time)
+          dir_path = Pathname.new(dir_path)
+          Dir.glob(dir_path.join("**")).each do |path|
+            if Pathname.new(path).mtime > threshold_time
+              return true
+            end
+          end
+          false
```
chrisroberts, line 19: This can be reduced down to:

```ruby
Dir.glob(Pathname.new(dir_path).join("**", "*")).any? do |path|
  Pathname.new(path).mtime > threshold_time
end
```

lib/vagrant/util/directory_changed.rb
```diff
@@ -0,0 +1,23 @@
+require 'pathname'
+
+module Vagrant
+  module Util
+    class DirectoryChanged
```
chrisroberts, line 5: It would be nice to have this something a little more generic, like just `Directory`, so that other directory related utility methods could be included here as well.

lib/vagrant/util/directory_changed.rb
```diff
@@ -0,0 +1,23 @@
...
+        # @return [Boolean]
+        def self.directory_changed?(dir_path, threshold_time)
+          dir_path = Pathname.new(dir_path)
+          Dir.glob(dir_path.join("**")).each do |path|
+            if Pathname.new(path).mtime > threshold_time
+              return true
+            end
+          end
+          false
+        end
```
chrisroberts, line 20: Looks like the indenting on this file is a little off for the class definition

lib/vagrant/util/directory.rb
```diff
@@ -0,0 +1,20 @@
...
+  module Util
+    class Directory
+      # Check if directory has any new updates
+      #
+      # @param [Pathname, String] Path to directory
+      # @param [Time] time to compare to eg. has any file in dir_path
+      #               changed since this time
+      # @return [Boolean]
+      def self.directory_changed?(dir_path, threshold_time)
+        dir_path = Pathname.new(dir_path)
```
chrisroberts, line 13: This `dir_path =` line can get removed

plugins/hosts/darwin/cap/fs_iso.rb
```diff
@@ -0,0 +1,76 @@
...
+
+          # If the destrination does not exist or there have been changes in the source directory since the last build, then build
+          if !file_destination.exist? || Vagrant::Util::Directory.directory_changed?(source_directory, file_destination.mtime)
+            @@logger.info("Building ISO from source #{source_directory}")
+            iso_command = [BUILD_ISO_CMD, "makehybrid"]
+            iso_command << "-hfs"
+            iso_command << "-iso"
+            iso_command << "-joliet"
+            iso_command << "-ov"
+            iso_command.concat(Vagrant::Util::MapCommandOptions.map_to_command_options(extra_opts, cmd_flag="-"))
```
jbonhag, line 59: One thing that just came to mind as I was reviewing https://github.com/hashicorp/vagrant/pull/11706 -- we can't pass extra_opts directly to the ISO command because the volume ID argument will be different depending on which command is being invoked. So maybe extra_opts should have a standard key like `extra_opts[:volume_id]` for setting the volume ID.

plugins/hosts/darwin/cap/fs_iso.rb
```diff
@@ -0,0 +1,78 @@
...
+          else
+            file_destination = Pathname.new(file_destination.to_s)
+            # If the file destination path is a folder, target the output to a randomly named
+            # file in that dir
+            if file_destination.extname != ".iso"
+              file_destination = file_destination.join("#{rand(36**6).to_s(36)}_vagrant.iso")
+            end
+          end
+          # Ensure destination directory is available
+          FileUtils.mkdir_p(File.dirname(file_destination.to_s))
```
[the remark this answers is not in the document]
soapy1: How do you feel about changing this to `FileUtils.mkdir_p(file_destination.dirname)`. Using the `file_destination.dirname.mkpath` requires building out far more extensive mocks in the tests. I think it makes the tests way harder to read and debug. For example, in order to test this function with `file_destination.dirname.mkpath` I tried to do something like:

```
let(:file_destination) { "/woo/out.iso" }
let(:file_destination_path) {Pathname.new( "/woo/out.iso" )}

before do
  allow(Pathname).to receive(:new).with(file_destination).and_return(file_destination_path) 
  # make sure no files are actually created for test
  allow(file_destination_path).to receive_message_chain(:dirname, :mkpath)
end
```

maybe my approach to testing this is wrong though?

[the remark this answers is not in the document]
chrisroberts: I think you could do something like this:

```ruby
require "tmpdir"

let(:file_destination) { @fd ||= Pathname.new(Dir.mktmpdir).join("out.iso").to_s }
let(:file_destination_path) { Pathname.new(file_destination) }

before { allow(Pathname).to receive(:new).with(file_destination).and_return(file_destination_path) }
after { file_destination_path.dirname.delete }
```

which would allow you to have a proper pathname instance that wouldn't need anything mocked, but you could include checks that expected behaviors were occurring, like:

```
expect(file_destination_path).to receive_message_chain(:dirname, :mkpath)
```

However, keeping it with the `#mkdir_p` is fine as they are functionally equivalent. Within the tests, though, we should be checking the the parameter provided to the mock (`expect(FileUtils).to receive(:mkdir_p).with("expected/prefix")`) so it's only matching what we expect to be matched.

PR #11704, opened 2020-06-15
Make alt + arch + atomic-host +debian + gentoo + slackware + pld + photon + rhel +suse guests respect hostname network config option

soapy1:
ref: https://github.com/hashicorp/vagrant/issues/11508

### DIFF
--- a/plugins/guests/linux/cap/change_host_name.rb
+++ b/plugins/guests/linux/cap/change_host_name.rb
@@ -4,34 +4,42 @@ module VagrantPlugins
   module GuestLinux
     module Cap
       module ChangeHostName
-        include Vagrant::Util::GuestHosts::Linux
-
-        def change_name_command(name)
-          return <<-EOH.gsub(/^ {14}/, '')
-              # Set the hostname
-              echo '#{name}' > /etc/hostname
-              hostname '#{name}'
-            EOH
-        end
-
-        def change_host_name?(comm, name)
-          !comm.test("hostname -f | grep '^#{name}$'", sudo: false)
-        end
-
-        def change_host_name(machine, name)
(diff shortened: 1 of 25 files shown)


plugins/guests/linux/cap/change_host_name.rb
```diff
@@ -0,0 +1,38 @@
...
+          if network_with_hostname
+            replace_host(comm, name, network_with_hostname[:ip])
+          else
+            add_hostname_to_loopback_interface(comm, name)
+          end
+
+          if change_host_name?(comm, name)
+            comm.sudo(change_name_command(name))
+          end
+        end
```
chrisroberts, line 34: Wrap these methods in a module, add an `.extended` method, and then extend in the `ChangeHostName` module:

```ruby
module ChangeHostName
  module Methods
    def change_name_command(name)
      ...
    end
    ...
    def self.extended(klass)
      klass.extend(Vagrant::Util::GuestHosts::Linux)
    end
  end

  extend Methods
end
```

test/unit/plugins/guests/linux/cap/change_host_name_test.rb
```diff
@@ -0,0 +1,77 @@
+require_relative "../../../../base"
+
+describe "VagrantPlugins::GuestLinux::Cap::ChangeHostName" do
+  let(:described_class) do
```
chrisroberts, line 4: This should be grabbing the capability directly from the plugin, not creating a new class to extend the module into:

```ruby
let(:described_class) do
  VagrantPlugins::GuestLinux::Plugin.
    components.guest_capabilities[:linux].get(:change_host_name)
end
```

PR #11706, opened 2020-06-17
Initial commit of cloud_init setup action

briancain:
This commit introduces a new builtin action for preparing and
configuring guests cloud_init configs as an iso to be attached later by
the disk actions.

It also sorts the gemspec dependencies alphabetically and adds a new rubygem `mime` for handling MIME text.

Note: Currently this PR only adds the setup to the VirtualBox provider, since there isn't a PR in the works for supporting DVD disks yet for Hyper-V.


lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,124 @@
...
+        #
+        # @param [Vagrant::Machine] machine
+        # @param [MIME::Multipart::Mixed] msg
+        # @return [String] iso_path
+        def write_cfg_iso(machine, env, msg)
+          iso_path = nil
+
+          if env[:env].host.capability?(:create_iso)
+            # TODO: make temp_root configurable?
+            source_dir = Pathname.new(Dir.mktmpdir(TEMP_PREFIX, TEMP_ROOT))
```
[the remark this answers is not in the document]
briancain: I think we still have to use mktmpdir instead, because the iso host cap is expecting a directory. I can just remove the TEMP_PREFIX though and it should do the right thing on which ever platform it runs on.

lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,130 @@
...
+        #
+        # @param [Vagrant::Machine] machine
+        # @param [MIME::Multipart::Mixed] user_data
+        # @param [Hash] meta_data
+        def write_cfg_iso(machine, env, user_data, meta_data)
+          iso_path = nil
+
+          if env[:env].host.capability?(:create_iso)
+            begin
+              source_dir = Pathname.new(Dir.mktmpdir(TEMP_PREFIX))
```
soapy1, line 99: `Dir.mktmpdir` doesn't exist in [ruby 2.6](https://ruby-doc.org/core-2.6/Dir.html) 😭

lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,130 @@
...
+              File.open("#{source_dir}/user-data", 'w') { |file| file.write(user_data.to_s) }
+
+              File.open("#{source_dir}/meta-data", 'w') { |file| file.write(meta_data.to_s) }
+
+              iso_path = env[:env].host.capability(:create_iso, env[:env],
+                                                   source_dir, volume_id: "cidata")
```
soapy1, line 105: the current pr for this host cap accepts an opts hash like `extra_opts: {:volume_id => "cidata"}`. I'm happy to change that pr if it makes more sense that way.

lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,130 @@
...
+
+        # Adds a new :dvd disk config with the given iso_path to be attached
+        # to the guest later
+        #
+        # @param [Vagrant::Machine] machine
+        # @param [Vagrant::Environment] env
+        # @param [String] iso_path
+        def attach_disk_config(machine, env, iso_path)
+          @logger.info("Adding cloud_init iso '#{iso_path}' to disk config")
+          machine.config.vm.disk :dvd, file: iso_path, name: "vagrant-cloud_init-disk"
```
soapy1, line 124: can there be multiple disk attachments created by cloud-init? If so, I think the names might collide. Could append/prepend a uid type thing to the name?

briancain: I'm assuming 1 cloud_init disk per guest. All cloud_init configs for a single guest get combined into a single user-data file, and thats the iso that gets attached. I don't think you'd end up attaching multiple discs to a single guest but I could be wrong.

lib/vagrant/action/builtin/cloud_init_setup.rb
briancain: @soapy1 I see! Hmm, yeah, I think it would be cleaner if the `create_iso` cap defined the extra options with something like:

```ruby
def create_iso(env, dir, **extra_opts)
 # stuff
end
```

Then it can work like how it's defined above where the final param is a hash of options rather than needing to pass a key that has its value as a hash.

lib/vagrant/action/builtin/cloud_init_setup.rb
briancain: It will exist when you require it from stdlib :smile: https://ruby-doc.org/stdlib-2.6.6/libdoc/tmpdir/rdoc/Dir.html

lib/vagrant/action/builtin/cloud_init_setup.rb
briancain: Also I think code wise nothing should change, you should be able to access the keys inside `extra_opts` the same way that you are now (like `extra_opts[:key]`), so the only thing that changes is the parameter!

lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,130 @@
...
+              File.open("#{source_dir}/user-data", 'w') { |file| file.write(user_data.to_s) }
+
+              File.open("#{source_dir}/meta-data", 'w') { |file| file.write(meta_data.to_s) }
+
+              iso_path = env[:env].host.capability(:create_iso, env[:env],
+                                                   source_dir, volume_id: "cidata")
+
+              attach_disk_config(machine, env, iso_path)
+            ensure
+              FileUtils.remove_entry source_dir
```
chrisroberts, line 109: Styling nit: `remove_entry(source_dir)`

lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,130 @@
...
+        # @param [Vagrant::Environment] env
+        # @param [Array<#VagrantPlugins::Kernel_V2::VagrantConfigCloudInit>] user_data_cfgs
+        # @return [MIME::Text] user_data
+        def setup_user_data(machine, env, user_data_cfgs)
+          machine.ui.info(I18n.t("vagrant.actions.vm.cloud_init_user_data_setup"))
+
+          text_cfgs = []
+          user_data_cfgs.each do |cfg|
+            text_cfgs << read_text_cfg(machine, cfg)
+          end
```
chrisroberts, line 42: Can just set this with a `#map`:

```ruby
text_cfgs = user_data_cfgs.map { |cfg| read_text_cfg(machine, cfg) }
```

lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,130 @@
...
+            text = File.read(Pathname.new(cfg.path).expand_path(machine.env.root_path))
+          else
+            text = cfg.inline
+          end
+
+          # Note: content_type must remove the leading `text/` because
+          # the MIME::Text initializer hardcodes `text/` already to the type.
+          # We assume content_type is correct due to the validation step
+          # in VagrantConfigCloudInit.
+          content_type = cfg.content_type.split('/')
```
chrisroberts, line 65: `content_type = cfg.content_type.split("/", 2).last`

Will ensure it only splits once if there are multiple slashes (there shouldn't be) and still returns the value if the `text/` is not provided (which it shouldn't be, but provides safety for if it is not).

lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,130 @@
...
+        # Adds a new :dvd disk config with the given iso_path to be attached
+        # to the guest later
+        #
+        # @param [Vagrant::Machine] machine
+        # @param [Vagrant::Environment] env
+        # @param [String] iso_path
+        def attach_disk_config(machine, env, iso_path)
+          @logger.info("Adding cloud_init iso '#{iso_path}' to disk config")
+          machine.config.vm.disk :dvd, file: iso_path, name: "vagrant-cloud_init-disk"
+          machine.config.vm.disks.map { |d| d.finalize! if d.type == :dvd && d.file == iso_path }
```
chrisroberts, line 125: Can use an `#each` here in place of the `#map` since the result is not needed.

lib/vagrant/action/builtin/cloud_init_setup.rb
```diff
@@ -0,0 +1,130 @@
...
+
+        def call(env)
+          machine = env[:machine]
+
+          user_data_configs = machine.config.vm.cloud_init_configs
+                                                  .select { |c| c.type == :user_data }
+
+          if !user_data_configs.empty?
+            user_data = setup_user_data(machine, env, user_data_configs)
+            meta_data = { "instance-id": "i-#{machine.id.split('-').join}" }
```
jbonhag, line 23: Nice! Good thinking reusing the machine id here.

Issue #11589, opened 2020-05-06
Rebooting Ubuntu guest after provisioning kills /vagrant mount point


kthy:
If I reboot my Ubuntu guest after provisioning, the `/vagrant` mount point is empty. If I comment out the line rebooting the guest, the folder contains the contents of the current folder on the host as expected.

### Vagrant version

```
C:\> vagrant -v
Vagrant 2.2.8
```

### Host operating system

* Windows 10 Pro Version 1903 (Build 18362.778)
* Hyper-V provider

### Guest operating system

Ubuntu 18.04 LTS (`hashicorp/bionic64`)

### Vagrantfile

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "hashicorp/bionic64"
  config.vm.provision :shell do |s|
    s.privileged = false
    s.reboot = true
    s.inline = <<-SHELL
      echo "Moo"
    SHELL
  end
end
```

### Debug output

* [vagrant-reboot.log](https://gist.github.com/kthy/22c64b58cc6a3b3e4620036c6e585491)
* [vagrant-noreboot.log](https://gist.github.com/kthy/4b14cbab247860f801fc886d49c23e27) (`s.reboot` line in `Vagrantfile` commented out)

### Expected behavior

I should see the contents of the host directory in `/vagrant`:

```bash
vagrant@ubuntu-18:~$ ls /vagrant/
Vagrantfile
vagrant@ubuntu-18:~$
```

### Actual behavior

```bash
vagrant@ubuntu-18:~$ ls /vagrant/
vagrant@ubuntu-18:~$
```

### Steps to reproduce

1. `vagrant up`
1. `vagrant ssh`
1. `ls /vagrant/`
1. `exit`
1. `vagrant destroy`
1. Comment out `s.reboot = true` and repeat

PR #11708, opened 2020-06-18
Add smb synced folders to fstab

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11589


plugins/guests/linux/cap/mount_smb_shared_folder.rb
```diff
@@ -82,27 +63,14 @@ def self.mount_smb_shared_folder(machine, name, guestpath, options)
           ensure
             # Always remove credentials file after mounting attempts
             # have been completed
-            machine.communicate.sudo("rm /etc/smb_creds_#{name}")
+            if !machine.config.vm.allow_fstab_modification
+              machine.communicate.sudo("rm /etc/smb_creds_#{name}")
```
soapy1, line 67: In order for the folder to be mounted the creds file need to exist on the guest. I think it makes sense to only remove the creds if fstab modification is turned off in this case. In order to make this safer, could make this file only readable by root?

soapy1: When a synced folder is removed these creds should probably be cleaned up as well

plugins/synced_folders/smb/synced_folder.rb
```diff
@@ -168,6 +174,10 @@ def cleanup(machine, opts)
         if machine.env.host.capability?(:smb_cleanup)
           machine.env.host.capability(:smb_cleanup, machine, opts)
         end
+        # Try to unmount all smb mounts, ignore if unmount fails for any reason
+        machine.communicate.sudo("mount -v | grep 'type cifs' | awk '{print $3}' | xargs umount || true")
+        # Clean up any credentials files that might exist
+        machine.communicate.sudo("rm /etc/smb_creds_vgt-*")
```
soapy1, line 180: I'm not entirely sure if this is an appropriate place to do this bit of clean up since it cleans up config on the guest

plugins/guests/linux/cap/mount_smb_shared_folder.rb
chrisroberts: Yes, setting the file mode to `0600` would be good, and removal when the fstab modification is not allowed is :+1:

soapy1: it gets set here https://github.com/hashicorp/vagrant/blob/1d602283a716e5514d20786bb4def9c60eb6a904/plugins/guests/linux/cap/mount_smb_shared_folder.rb#L42 🎉

soapy1:
Need to revisit this. Found a bug around fstab entries being overwritten when using multiple types of synced folders. For example, for the Vagrantfile
```
  config.vm.box = "hashicorp/bionic64"
  config.vm.synced_folder ".", "/vagrant", type: "smb"
  config.vm.synced_folder "../other", "/other"
end
```

the expected fstab entry by Vagrant is something like
```
#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
other /other vboxsf uid=1000,gid=1000,nofail 0 0
<smb mount location> /vagrant cifs <smb mount options> 0 0
#VAGRANT-END
```

the actual fstab entry is
```
#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
other /other vboxsf uid=1000,gid=1000,nofail 0 0
#VAGRANT-END
```

lib/vagrant/machine.rb
```diff
@@ -616,6 +617,25 @@ def with_ui(ui)
       end
     end
 
+    # Returns the SyncedFolder object associated with this machine.
+    #
+    # @return [Class]
+    def synced_folders
+      return @synced_folders if defined?(@synced_folders)
+      @synced_folders = @config.vm.synced_folders.map { |id, opts|
+        if opts[:type]
```
soapy1, line 626: I don't really like this bit. But, if the default mount is not in the Vagrantfile it will get added to the synced folders without a type. When the syncing action happens there is a bit that [sets the synced folder type](https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/action/builtin/mixin_synced_folders.rb#L14) if not already set.

lib/vagrant/synced_folder.rb
```diff
@@ -0,0 +1,16 @@
+require "vagrant/capability_host"
+
+module Vagrant
+  # This class handles host-OS specific interactions. It is responsible for
+  # detecting the proper host OS implementation and delegating capabilities
+  # to plugins.
+  #
+  # See {Guest} for more information on capabilities.
+  class SyncedFolder
```
soapy1, line 9: might be able to pop this into the [synced folder plugin bit](https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/plugin/v2/synced_folder.rb)

lib/vagrant/action/builtin/mixin_synced_folders.rb
```diff
@@ -190,6 +188,12 @@ def synced_folders(machine, **opts)
             folders[impl_name] = new_fs
           end
 
+          folders.each do |impl_name, fs|
+            fs.each do |id, data|
+              data[:impl] = plugins[impl_name][0].new()
```
soapy1, line 193: hmmmmm, it would be neat to have the synced folder object attached to the synced folder instead of just a hash of information. Would also cut down of the number of times the synced folders classes get instantiated. It seems like a new object is instantiated for about every action with the current implementation. 
I'm not completely convinced by this approach though. I don't like that each folder get's its own `:impl`. It would be better if each folder type got it's own `:impl` that is shared across folders of the same type.

lib/vagrant/machine.rb
```diff
@@ -10,6 +10,8 @@ module Vagrant
   # API for querying the state and making state changes to the machine, which
   # is backed by any sort of provider (VirtualBox, VMware, etc.).
   class Machine
+    extend Vagrant::Action::Builtin::MixinSyncedFolders
```
soapy1, line 13: I think I'm not a huge fan of extending from a module in `Vagrant::Action` here.
I think what I really want here is to have the bit of `Vagrant::Action::Builtin::MixinSyncedFolders.synced_folders` be defined here so that you can do a `machine.synced_folders`.

lib/vagrant/plugin/v2/synced_folder.rb
```diff
@@ -54,6 +56,12 @@ def disable(machine, folders, opts)
         # @param [Hash] opts
         def cleanup(machine, opts)
         end
+
+        def _initialize(machine, synced_folder_type)
+          plugins = Vagrant.plugin("2").manager.synced_folders
+          capabilities = Vagrant.plugin("2").manager.synced_folder_capabilities
+          initialize_capabilities!(synced_folder_type, plugins, capabilities, machine)
```
soapy1, line 63: would be neat if this was defined by the plugin. Something like a constant `SYNCED_FOLDER_TYPE` and then this could be a `initialize_capabilities!(self::SYNCED_FOLDER_TYPE, plugins, capabilities, machine)`

plugins/synced_folders/unix_mount_helpers.rb
```diff
@@ -111,6 +111,46 @@ def mount_options(machine, name, guest_path, options)
         mount_options = mount_options.join(',')
         return mount_options, mount_uid, mount_gid
       end
+
+      def smb_mount_options(machine, name, guest_path, options)
```
soapy1, line 115: maybe these mount options can be defined as synced folder capabilities?

plugins/guests/linux/cap/persist_mount_shared_folder.rb
```diff
@@ -14,23 +14,43 @@ class PersistMountSharedFolder
...
+          export_folders = folders.map { |type, folder|
+            folder.map { |name, data|
+              data[:owner] ||= ssh_info[:username]
+              data[:group] ||= ssh_info[:username]
+              guest_path = Shellwords.escape(data[:guestpath])
+              case type
+              when :virtualbox
+                mount_options, mount_uid, mount_gid = machine.synced_folders[:virtualbox].capability(
+                  :mount_options, name, guest_path, data)
+                mount_type = "vboxsf"
```
soapy1, line 33: maybe `mount_type` can be a synced_folder cap?

PR #11712, opened 2020-06-22
Update bootstrap-salt.ps1 to archive saltstack repo

prfj:
All the new versions that are patched are on the previous repo.

https://github.com/saltstack/salt-enhancement-proposals/blob/master/accepted/0022-old-releases.md


UtahDave:
I prefer the fix in https://github.com/hashicorp/vagrant/pull/12127 over this fix. By using the upstream windows bootstrap script, the latest versions will be used by default and will be updated when Salt has releases.

The "archive" section of the Salt repo was created because some of those older versions have CVEs that are fixed in later versions. I would recommend not using the archived versions unless there's a compelling reason.

soapy1:
Hi there,

Cheers!

PR #11719, opened 2020-06-25
Make darwin + openbsd +freebsd guest respect 'hostname' network config option

soapy1:
ref: https://github.com/hashicorp/vagrant/issues/11508

### DIFF
--- a/lib/vagrant/util/guest_hosts.rb
+++ b/lib/vagrant/util/guest_hosts.rb
@@ -2,11 +2,10 @@ module Vagrant
   module Util
     # Helper methods for modfiying guests /etc/hosts file
     module GuestHosts
-      # Linux specific inspection helpers
-      module Linux
 
+      module Unix
         DEAFAULT_LOOPBACK_CHECK_LIMIT = 5.freeze
-       
+
         # Add hostname to a loopback address on /etc/hosts if not already there
         # Will insert name at the first free address of the form 127.0.X.1, up to
         # the loop_bound
@@ -27,7 +26,11 @@ module Vagrant
           }
           EOH
         end
+      end
 
+      # Linux specific inspection helpers
+      module Linux
+        include Unix
         # Remove any line in /etc/hosts that contains hostname,
         # then add hostname with associated ip 
         #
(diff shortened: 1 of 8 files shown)


lib/vagrant/util/guest_hosts.rb
```diff
@@ -42,6 +45,24 @@ def replace_host(comm, name, ip)
           EOH
         end
       end
+
+      # BSD specific inspection helpers
+      module BSD
```
soapy1, line 50: BSD has a different `sed` 😭

lib/vagrant/util/guest_hosts.rb
```diff
@@ -2,11 +2,10 @@ module Vagrant
   module Util
     # Helper methods for modfiying guests /etc/hosts file
     module GuestHosts
-      # Linux specific inspection helpers
-      module Linux
 
+      module Unix
```
soapy1, line 6: maybe this naming can be better?

plugins/guests/darwin/cap/change_host_name.rb
```diff
@@ -21,14 +25,14 @@ def self.change_host_name(machine, name)
...
               hostname '#{name}'
-
-              # Prepend ourselves to /etc/hosts - sed on bsd is sad
-              grep -w '#{name}' /etc/hosts || {
-                echo -e '127.0.0.1\\t#{name}\\t#{basename}' | cat - /etc/hosts > /tmp/tmp-hosts &&
-                  mv /tmp/tmp-hosts /etc/hosts
-              }
             EOH
           end
+          network_with_hostname = machine.config.vm.networks.map {|t, c| c if c[:hostname] }.compact[0]
```
chrisroberts, line 30: Since the `t` variable isn't being used, you can "throw it away" for the loop:

```ruby
.map { |_, c| ... }
```

lib/vagrant/util/guest_hosts.rb
chrisroberts: We can definitely see if any other opinions show up. I'm leaning towards just having the modules named after the host (Linux, Darwin, Solaris, etc.) so we can provide overrides that are host specific where required and make usage of the modules a little more natural.

PR #11727, opened 2020-06-29
Add Algolia index script to CI

pruett:
Adds job to CI that indexes and pushes content to Algolia via Node.js script

- Added a `ALGOLIA_API_KEY` env var to CCI (All other env vars are non-sensitive and provided within `.env` file)


website/.env
```diff
@@ -0,0 +1,3 @@
+NEXT_PUBLIC_ALGOLIA_APP_ID=YY0FFNI7MF
+NEXT_PUBLIC_ALGOLIA_INDEX=product_VAGRANT
+NEXT_PUBLIC_ALGOLIA_SEARCH_ONLY_API_KEY=5037da4824714676226913c65e961ca0
```
pruett, line 3: Safe to add to VC, according to https://www.algolia.com/doc/guides/security/api-keys/#search-only-api-key

briancain:
@alvin-huang Yeah...I think the website refresh was setup to use CircleCI but eventually we will move it to GitHub actions along with the rest of our CI.

PR #11739, opened 2020-07-03
haiku: Add network_interfaces and configure_networks caps

kallisti5:


plugins/guests/haiku/cap/configure_networks.rb
```diff
@@ -0,0 +1,29 @@
...
+            entry = TemplateRenderer.render("guests/haiku/network_#{network[:type]}",
+                                            options: network)
+            Tempfile.open("vagrant-haiku-configure-networks") do |f|
+              f.binmode
+              f.write(entry)
+              f.fsync
+              f.close
+              machine.communicate.upload(f.path, "/tmp/vagrant-network-entry")
+            end
+            machine.communicate("mv /tmp/vagrant-network-entry /boot/system/settings/network/interfaces")
```
jbonhag, line 22:
```suggestion
            machine.communicate.execute("mv /tmp/vagrant-network-entry /boot/system/settings/network/interfaces")
```

Should there be multiple interfaces in `/boot/system/settings/network/interfaces`? With multiple networks defined, I ended up with only a single interface:

Vagrantfile:

```ruby
  config.vm.network "public_network", ip: "192.168.127.2"
```

/boot/system/settings/network/interfaces:

```
interface  {
  address inet {
    address 192.168.127.2
    mask 255.255.255.0
  }
}
```

I think you'll want to move the `Tempfile` creation outside of the loop so we keep writing to the same interfaces file.

plugins/guests/haiku/cap/configure_networks.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require_relative "../../../../lib/vagrant/util/template_renderer"
+
+module VagrantPlugins
+  module GuestHaiku
+    module Cap
+      class ConfigureNetworks
+        include Vagrant::Util
+
+        def self.configure_networks(machine, networks)
```
jbonhag, line 11: You'll want to get the list of interfaces here using the :network_interfaces capability, so you can set the device names, something like:

```ruby
interfaces = machine.guest.capability(:network_interfaces)
networks.each do |network|
  network[:device] = interfaces[network[:interface]]
  ...
```
```

plugins/guests/haiku/cap/configure_networks.rb
kallisti5: Ah. good catch. yeah, that'll only ever template out the first interface.

PR #11752, opened 2020-07-08
🌷 july docs website platform updates

jescalan:
Just a standard maintenance round here! In this round of updates, we have greatly improved formatting and linting consistency - all of our standard linting and formatting scripts now run across a wider range of files, and automatically run as a precommit hook to ensure that they are being evenly applied no matter who is contributing. This PR also hard pins all dependencies to prevent any accidental bugs due to background patch updates.

This should be a great benefit for markdown formatting specifically, as our formatter now covers all markdown files 😀


briancain:
If this is ready to go @jescalan I can hit the merge button :smile:

jescalan:
Ready indeed - thank you @briancain !

PR #11759, opened 2020-07-10
Closes #11758

### LINKED ISSUE #11758: Add guest capability for solaris11 guests: shell_expand_guest_path
When using the file provisioner on solaris 11 guests, I can't use the ~ in the destination path because it doesn't get expanded to the correct home dir. The solaris11 guest doesn't have the capability for this as it doesn't inherit from the Linux guest plugin (which is sensible).

My first solution would be to just reuse the capability from some other guest (generic Linux, for example), as it seems to do the trick. 

(A better solution seems to be to offer a helper class that encapsulates that generic method because many of the guests just seem to reuse the code for the expansion...)

**Additional context**
Nothing really. I'll open a PR for this soon...

danowar2k:
- added shell_expand_guest_path capability for solaris11
- added tests for solaris11 shell_expand_guest_path capability


danowar2k:
By the way, I'm not understanding the way for example solaris11 guests are unit tested now. There is so much ci configuration. Is there some loop that loops through all supported guests and has boxes for all of the guest OS?

briancain:
Hey there @danowar2k - This looks identical to the linux capability. I think we should just have the solaris plugin use that instead.

danowar2k:
Nothing against that. I was just copying other people who just copied the capability code, too (or so it seems).
I'll try to just reuse the linux function...

danowar2k:
Let's see, I don't know much about Ruby best practices, but this is my intuitive try to reuse the linux capability class...

danowar2k:
I could also do the inheritance route if that is preferred

```
class ShellExpandGuestPath < VagrantPlugins::GuestLinux::Cap::ShellExpandGuestPath
```

danowar2k:
Okay, I'll remove the spec test. How about I move the ~ test to the linux tests, there seems to be none there and real path tests don't seem to be complete without one ;-)

briancain:
@danowar2k Ahhh, yes! Yeah that would belong in that linux spec test. That would be a good case to cover in there, thanks! :+1:

briancain:
> EDIT: Although isn't the spec test done on a Solaris 11 VM? Wouldn't it be better to at least test the stuff on that OS?

So these spec tests are just unit tests that Ruby runs, they aren't something higher level like acceptance tests or blackbox tests which would run real code on a system and test the result.

danowar2k:
> > EDIT: Although isn't the spec test done on a Solaris 11 VM? Wouldn't it be better to at least test the stuff on that OS?
> 
> So these spec tests are just unit tests that Ruby runs, they aren't something higher level like acceptance tests or blackbox tests which would run real code on a system and test the result.

Ah that explains why I didn't find any code that spins up Vagrant guests.

briancain:
We have a different repo that has those kinds of tests: https://github.com/hashicorp/vagrant-spec

danowar2k:
I don't know if my action was necessary or not, but I still had the requested changes notice and I finally found a button to address that (I think)

test/unit/plugins/guests/linux/cap/shell_expand_guest_path_test.rb
```diff
@@ -25,6 +25,14 @@
       cap.shell_expand_guest_path(machine, path)
     end
 
+    it "expands a path with tilde" do
+      path = "~/folder"
+      allow(machine.communicate).to receive(:execute).
```
soapy1, line 30:
```suggestion
      expect(machine.communicate).to receive(:execute).
```

danowar2k: If this is RSpec and this https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/basics/expecting-messages explains it right, a failed expect will fail the test...which should be the preferred action, shouldn't it?

danowar2k:
Sorry, I was sick for a few days and couldn't answer...

briancain:
Hey there @danowar2k - no worries about being sick, hope you feel better now! I ended up merging it because it's similar to the other surrounding tests. Having another `expect` there would make the tests better. I'm fine with opening a PR today unless you are itching to write some more tests! :)

PR #11773, opened 2020-07-16
Add cloud init wait action

soapy1:


plugins/providers/virtualbox/action.rb
```diff
@@ -58,7 +58,7 @@ module Action
 
       # This action boots the VM, assuming the VM is in a state that requires
       # a bootup (i.e. not saved).
-      def self.action_boot
+      def self.action_boot(cloud_init=false)
```
briancain, line 61: Small nitpick on these, but I think it'll be more maintainable if this is something like `**opts` instead, just in case more options like this are ever introduced.

lib/vagrant/action/builtin/cloud_init_wait.rb
```diff
@@ -0,0 +1,22 @@
...
+      class CloudInitWait
+
+        def initialize(app, env)
+          @app    = app
+          @logger = Log4r::Logger.new("vagrant::action::builtin::cloudinitwait")
+        end
+
+        def call(env)
+          machine = env[:machine]
+          @logger.info("Waiting for cloud init")
```
briancain, line 13: Maybe this should be moved into the if-statement below just so that it only logs that if we run the wait command?

lib/vagrant/action/builtin/cloud_init_wait.rb
```diff
@@ -0,0 +1,22 @@
...
+          if !machine.config.vm.cloud_init_configs.empty?
+            machine.communicate.sudo("cloud-init status --wait", {:error_check => false})
```
briancain, line 15: Hmm, I wonder if it's worth it to `Which::which` on the `cloud-init` executable just so we can be sure we're getting the right path on the guest.

lib/vagrant/action/builtin/cloud_init_wait.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        def initialize(app, env)
+          @app    = app
+          @logger = Log4r::Logger.new("vagrant::action::builtin::cloudinitwait")
+        end
+
+        def call(env)
+          machine = env[:machine]
+          if !machine.config.vm.cloud_init_configs.empty?
+            if machine.communicate.test("cloud-init -v", error_check: false)
```
chrisroberts, line 14: can use this instead to check for the command and not rely on exit code to be zero:

```ruby
if machine.communicate.test("command -v cloud-init")
```

plugins/providers/virtualbox/action.rb
```diff
@@ -58,7 +58,7 @@ module Action
 
       # This action boots the VM, assuming the VM is in a state that requires
       # a bootup (i.e. not saved).
-      def self.action_boot
+      def self.action_boot(**opts)
```
chrisroberts, line 61: Options should not be added to the method here.

plugins/providers/virtualbox/action.rb
```diff
@@ -317,7 +318,7 @@ def self.action_ssh_run
 
       # This action starts a VM, assuming it is already imported and exists.
       # A precondition of this action is that the VM exists.
-      def self.action_start
+      def self.action_start(**opts)
```
chrisroberts, line 321: Options should not be added here

plugins/providers/virtualbox/action.rb
```diff
@@ -344,7 +345,8 @@ def self.action_start
 
                 # The VM is not saved, so we must have to boot it up
                 # like normal. Boot!
-                b4.use action_boot
+                run_cloud_init = opts[:cloud_init] || false
+                b4.use action_boot(cloud_init: run_cloud_init)
```
chrisroberts, line 349: the `b4.use action_boot` doesn't require any modifications

plugins/providers/virtualbox/action.rb
```diff
@@ -412,7 +414,7 @@ def self.action_up
             end
           end
 
-          b.use action_start
+          b.use action_start(cloud_init: true)
```
chrisroberts, line 417: The `cloud_init: true` can be removed

plugins/providers/virtualbox/action.rb
```diff
@@ -412,7 +414,7 @@ def self.action_up
             end
           end
```
chrisroberts, line 416: Here we can add:

```ruby
b.use EnvSet, cloud_init: true
```

plugins/providers/virtualbox/action.rb
```diff
@@ -79,13 +79,14 @@ def self.action_boot
           b.use ForwardPorts
           b.use SetHostname
           b.use SaneDefaults
-          b.use CloudInitSetup
```
chrisroberts, line 82: The need for cloud init can be set into the env in the `.action_up`. Here in the `.action_boot` the `IsEnvSet` action can be used to check for it, and then add the cloud init related actions if needed.

```ruby
b.use Call, IsEnvSet, :cloud_init do |env, b2|
  if env[:result]
    b2.use CloudInitSetup
  end
  b2.use CleanupDisks
  b2.use Disk
  ...
  b2.use WaitForCommunicator, [:starting, :running]
  if env[:result]
    b2.use CloudInitWait
  end
  b2.use Customize, "post-comm"
  b2.use CheckGuestAdditions
end
```

lib/vagrant/action/builtin/cloud_init_wait.rb
```diff
@@ -0,0 +1,26 @@
...
+        def initialize(app, env)
+          @app    = app
+          @logger = Log4r::Logger.new("vagrant::action::builtin::cloudinitwait")
+        end
+
+        def call(env)
+          machine = env[:machine]
+          if !machine.config.vm.cloud_init_configs.empty?
+            if machine.communicate.test("command -v cloud-init")
+              @logger.info("Waiting for cloud init")
```
jbonhag, line 15: I think we might want to use the regular UI output for this so users know what the machine is doing.

```suggestion
              env[:ui].output(I18n.t("vagrant.cloud_init_waiting"))
```

lib/vagrant/action/builtin/cloud_init_wait.rb
```diff
@@ -0,0 +1,26 @@
...
+              machine.communicate.sudo("cloud-init status --wait")
```
jbonhag, line 16: It might be nice to pop up a message that says that cloud-init has finished. We should also check the result of this command and raise a cloud-init specific error, rather than the generic SSH error:


```suggestion
              result = machine.communicate.sudo("cloud-init status --wait", error_check: false)
```
(etc.)

PR #11774, opened 2020-07-16
Redirect getting started pages to learn. 

judithpatudith:
This PR redirects all the getting started pages (formerly at `/docs` and `/intro` to the newly revamped learn guides. 

It removes the files for the guides so that there is a single source of truth for the content.

It removes the menu items for each guide and leaves the get started link, which points directly to the first tutorial in the track.

It also adds `301!` to all the redirects to comply with the instructions in the README.


website/data/intro-navigation.js
```diff
@@ -8,21 +8,14 @@
...
-      'boxes',
-      'up',
-      'synced_folders',
-      'provisioning',
-      'networking',
-      'share',
-      'teardown',
-      'rebuild',
-      'providers',
+      { title: 'Getting Started', href: 'https://learn.hashicorp.com/vagrant' },
```
briancain, line 14: Hmm, I'm not sure if this is the navigation file that's doing this...or if it's something that hasn't been changed yet, but when I'm on the `intro` page I see a `Getting Started` link in the nav bar. However it takes me to `http://localhost:3000/intro/getting-started` which ends up being a 404, rather than directing to the learning site :thinking:

judithpatudith: Oh weird. Have you tried it on the [Netlify build](https://deploy-preview-11774--vagrant-docs.netlify.app/intro)? The website's README file says that redirects don't work correctly on the local build, but I think it's ok.

briancain: Yeah! Ok the Netlify build works :) I didn't realize the redirects won't work locally.

judithpatudith:
I don't have permission to merge this but @topfunky do you, and are you ready for the vagrant redirects to kick in?

briancain:
Let me know when you're ready, and I'd be happy to merge it!

judithpatudith:
@briancain Just checked with Geoffrey yesterday and we're ready 🎉 Thank you!

PR #11787, opened 2020-07-28
Use the correct powershell executable for privileged commands

fliam:
For the other occurrences of the powershell executable in `lib/vagrant/util/powershell.rb` this change has already been implemeted in back in 2018: https://github.com/hashicorp/vagrant/pull/9300/commits/ed686ce724f2b6117b03a3267f26d9f66286b387.


fliam:
Sure!
I think I did it correctly, but these are my first few lines of Ruby (outside of a Vagrantfile) 😉.

PR #11794, opened 2020-07-29
Gem::Specification.default_specifications_dir is deprecated.

pvalena:
Use Gem.default_specifications_dir instead.

_ _ _ _
When run with Ruby 2.7 (Rubygems 3.1).

Can be seen in vagrant-libvirt test suite f.e.: https://gist.github.com/pvalena/ddeafe6a0e44f5443ef96aed6678fc8b

### DIFF
--- a/lib/vagrant/bundler.rb
+++ b/lib/vagrant/bundler.rb
@@ -326,7 +326,7 @@ module Vagrant
     # @return [Array<[Gem::Specification, String]>] spec and directory pairs
     def vagrant_internal_specs
       list = {}
-      directories = [Gem::Specification.default_specifications_dir]
+      directories = [Gem.default_specifications_dir]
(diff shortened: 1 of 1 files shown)


briancain:
Hey there @pvalena - Thanks for the PR: However currently we have about 3 PRs that all are trying to fix the same thing: https://github.com/hashicorp/vagrant/pull/11459, https://github.com/hashicorp/vagrant/pull/11437, https://github.com/hashicorp/vagrant/pull/11307

This change won't be made until we can upgrade the packaging Vagrant using to support Ruby 2.7 which is depending on the Ruby encoder supporting Ruby version 2.7. It looks like it's recently been added in the ruby encoder, so fixing this should be coming soon. I'm going to go ahead and close this due to the other PRs we already have, and the missing CLA signature. Thanks for taking the time to open a PR though!

Issue #11588, opened 2020-05-06
When mounting SMB, Vagrant does not ask for re-entry of Windows credentials


kthy:
Looking at [`synced_folder.rb`](https://github.com/hashicorp/vagrant/blob/c1b695c47119f4294d92152f0c8f8da27f657660/plugins/synced_folders/smb/synced_folder.rb#L51) lines 50-80, I'd expect to be asked to re-enter my Windows password up to 5 times when I mistype it while mounting the standard SMB share. However, Vagrant errors out instead.

I have checked that the `check_credentials.ps1` script referenced in [`smb.rb`](https://github.com/hashicorp/vagrant/blob/eeecd258f8dc4376ae5e0e08cf10685753aaa184/plugins/hosts/windows/cap/smb.rb#L26) is available at `C:\HashiCorp\Vagrant\embedded\gems\2.2.8\gems\vagrant-2.2.8\plugins\hosts\windows\scripts` on my machine.

### Vagrant version

```
C:\> vagrant -v
Vagrant 2.2.8
```

### Host operating system

* Windows 10 Pro Version 1903 (Build 18362.778)
* Hyper-V provider

### Guest operating system

Ubuntu 18.04 LTS (`hashicorp/bionic64`)

### Vagrantfile

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "hashicorp/bionic64"
end
```

### Debug output

[vagrant.log](https://gist.github.com/kthy/a27e59754bde944bf3d0668d3c51eebe)

### Expected behavior

I should be asked to re-enter my Windows credentials.

### Actual behavior

An error is thrown:

```bash
Failed to mount folders in Linux guest. This is usually because
the "vboxsf" file system is not available. Please verify that
the guest additions are properly installed in the guest and
can work properly. The command attempted was:

mount -t cifs -o vers=2.0,credentials=/etc/smb_creds_vgt-db75744b775209a18a00f908efff2de3-6ad5fdbcbf2eaa93bd62f92333a2e6e5,uid=1000,gid=1000,mfsymlinks //172.17.205.97/vgt-db75744b775209a18a00f908efff2de3-6ad5fdbcbf2eaa93bd62f92333a2e6e5 /vagrant

The error output from the last command was:

mount error(13): Permission denied
Refer to the mount.cifs(8) manual page (e.g. man mount.cifs)
```

### Steps to reproduce

1. `vagrant up`
2. Enter wrong password when asked for SMB credentials.

PR #11795, opened 2020-07-30
Fix SMB credential checks

LachlanArthur:
The credential checking capability was not added to the Windows host plugin. This prevented the capability from running, and also caused the tests to be skipped.

The function signature for the `smb_validate_password` method was also different to where it was being used in both the synced_folder plugin and the tests.

The unit tests succeed, but as I'm very unfamiliar with Ruby, I don't really understand them (for example, it looks to me like the `should retry when validation fails` test is sending the same username and password each time, but expecting different results)


LachlanArthur:
This PR should fix issue #11588

PR #11797, opened 2020-07-30
Synced folder caps for mount options and default fstab modification behaviour

soapy1:


lib/vagrant/machine.rb
```diff
@@ -616,6 +619,29 @@ def with_ui(ui)
...
+    #
+    # @return [Hash<Symbol, Hash<String, Hash>>]
+    def synced_folders
+      self.class.synced_folders(self)
+    end
+
+    # Returns the SyncedFolder object associated with this machine.
+    #
+    # @return [List<Class>]
+    def synced_folder_types
```
soapy1, line 634: I'm not super sure about this. In order to use synced folder capabilities the synced folder class needs to be  initialized. However, the pattern for using synced folders currently is based one passing around a hash map of folder paths with their implementation.
Having this list of synced_folders allows Vagrant to keep one set of initialized synced folders that can be used to access capabilities for the configured folder types. Could make this part of the synced_folders hash? Maybe something else?

Issue #11705, opened 2020-06-16
Support host shared folders when Vagrant is executed inside WSL2


jstangroome:
### Vagrant version

```
$ vagrant -v
Vagrant 2.2.9
```

### Host operating system

Windows 10 Version 2004 (OS Build 19041.329)

### Expected behavior

From within WSL2, with the working directory set to a path on the host filesystem.

```
$ export VAGRANT_WSL_ENABLE_WINDOWS_ACCESS=1
$ cd /mnt/c/example
$ vagrant up
...
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The host path of the shared folder is not supported from WSL. Host
path of the shared folder must be located on a file system with
DrvFs type. Host path: .
```

In WSL1 the `mount` command returns:
```
C:\ on /mnt/c type drvfs (rw,noatime,uid=1000,gid=1000,umask=22,metadata,case=off)
```

But in WSL2 the `mount` command returns:
```
C:\ on /mnt/c type 9p (rw,noatime,dirsync,aname=drvfs;path=C:\;uid=1000;gid=1000;metadata;umask=22;symlinkroot=/mnt/,mmap,access=client,msize=65536,trans=fd,rfd=8,wfd=8)
```

Modifying [platform.rb:564](https://github.com/hashicorp/vagrant/blob/6efdcd2068a97411240878318a360e65b0237067/lib/vagrant/util/platform.rb#L564) from:
```
if info && info[:type] == "drvfs"
```
to
```
if info && (info[:type] == "drvfs" || info[:type] == "9p")
```
solves the issue and specifies the correct host-relative Shared Folder paths in the Virtualbox VM configuration.

### Actual behavior

```
* The host path of the shared folder is not supported from WSL. Host
path of the shared folder must be located on a file system with
DrvFs type. Host path: .
```

### References

Not to be confused with:
- GH-11623
- GH-11311

asciifaceman:
I'm going to wager there is no workaround for this issue without building from source?

strarsis:
@jstangroome: Thanks for this fix! After each Vagrant update I have to reapply it. It would be nice to have this fixed in a new Vagrant release, so synced folders can indeed be used in WSL 2.

cristian-hostpapa:
Any ETA on when this will be available ?  right now, vagrant is useless on WSL2.

chrisroberts:
We are tossing this patch into the next release (#12056) so you won't need to patch locally.

[...]

Issue #10576, opened 2019-01-09
Vagrant up (2.2.2) not working in WSL


ivan-radic:
Hi there,

On fresh install of Windows 10, fresh install of latest Vagrant, trying to create Ubuntu 18.04 VM, I get:

```
vagrant up
/opt/vagrant/embedded/gems/2.2.2/gems/vagrant-2.2.2/lib/vagrant/util/which.rb:37: warning: Insecure world writable dir /mnt/c in PATH, mode 040777
Bringing machine 'default' up with 'virtualbox' provider...
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The host path of the shared folder is not supported from WSL. Host
path of the shared folder must be located on a file system with
DrvFs type. Host path: .

```
### Vagrant version
vagrant -v
Vagrant 2.2.2

### Host operating system
Windows 10 Enterprise 1709

### Guest operating system
Ubuntu 18.04
https://app.vagrantup.com/ubuntu/boxes/bionic64

### Vagrantfile
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "ubuntu/bionic64"
  config.vm.provider "virtualbox" do |vb|
    vb.memory = "1024"
  # tried below line, as suggested in another bug report, but no luck
  # vb.customize [ "modifyvm", :id, "--uartmode1", "disconnected" ]
  end
end
```

### Debug output
https://gist.github.com/ivan-radic/9545bf66b793dff0c6794619cc7c6cd9

### Expected behavior
I am looking for a workaround, if there is one.
Note that Vagrant for Windows in powershell console works fine, and successfully creates Ubuntu 18.04 machine.

If you need more info, just ask.

Thanks,
Ivan

chrisroberts:
Hi there,

This is because you are attempting to use vagrant project directory located within the WSL and synced folders with paths inside the WSL are not supported. Moving the project directory to a DrvFs mounted file system will fix the issue (this is the mounted local `C:\` drive which can be found either at `/c` or `/mnt/c`). If you don't need the synced folder you can disable it by adding this to your Vagrantfile:

```ruby
config.vm.synced_folder '.', '/vagrant', disabled: true
```

Cheers!

ivan-radic:
Hi Chris,

thanks a lot, I can confirm that both methods work.
I guess this thread can serve as documentation of sorts, but may I suggest a small improvement.

Rephrase the wording a bit:

```
vm:
[...]

PR #11803, opened 2020-08-04
Fix WSL 2 Docker Sync Issue

thed24:
By adjusting the conditional check for the file system, we can also check for 9P systems, which is what WSLs server is classed as. After changing this, we no longer get the error regarding the project being stored on the incorrect filesystem, however there still is an issue with pathing, as it formats the source location according to Windows pathing standards, although we're using WSL. If WSL is enabled, we want to use standard unix pathing as Docker recognizes that when mapping it's volumes. These changes allow for dockers folder sync to work, if the source if an absolute path.

### DIFF
--- a/lib/vagrant/util/platform.rb
+++ b/lib/vagrant/util/platform.rb
@@ -561,7 +561,7 @@ module Vagrant
               result = Util::Subprocess.execute("mount")
               result.stdout.each_line do |line|
                 info = line.match(MOUNT_PATTERN)
-                if info && info[:type] == "drvfs"
+                if info && (info[:type] == "drvfs" || info[:type] == "9p")
                   @_wsl_drvfs_mounts << info[:mount]
                 end
               end
--- a/plugins/providers/docker/driver.rb
+++ b/plugins/providers/docker/driver.rb
@@ -87,6 +87,8 @@ module VagrantPlugins
               # means that there's no long path support). Hopefully this
               # will be fixed someday and the gsub below can be removed.
               host, guest = v.split(":", 2)
+              # If using WSL within a drvfs file system
+              # we want to continue to use linux pathing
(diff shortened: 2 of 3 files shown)


thed24:
Adjusted to use masters Ruby version. This should solve issue #11705 and #10576!

vafschoen:
Thanks for this pr. I was struggling with getting vagrant to work in wsl2 as well.
I've went a step further and added "ext4" to the filesystem list in lib/vagrant/util/platform.rb.
This allows us to use docker and a wsl2 ubuntu-vm according to the docker "best practices" https://docs.docker.com/docker-for-windows/wsl/. this provides a nice performance boost.

As I'm not sure about the implications this might have to others I'm not going to add to this pr.

thed24:
@vafschoen I'm glad this PR was able to help. I might attempt to test adding that file system myself as that definitely does sound like a nice change!

tobias-af (approved):

DominicCodespoti:
Has this functionality or similar changes been incorporated into master yet? @chrisroberts

vafhornung:
Bump.
Can someone please merge and release this ?

I'am using this pr as custom based on 2.2.10 version on my windows, no issues so far.

Thanks 😊

phinze:
This change is going to be reviewed as part of a batch of proposed WSL changes in an upcoming release. It's good to hear it has worked for at least one user, but we want to make sure it doesn't break anything else. Labeling this so it gets picked up when we make our WSL sweep.

thed24:
> This change is going to be reviewed as part of a batch of proposed WSL changes in an upcoming release. It's good to hear it has worked for at least one user, but we want to make sure it doesn't break anything else. Labeling this so it gets picked up when we make our WSL sweep.

Any progress on this? :)

PR #11829, opened 2020-08-17
Replace unmaintained Erubis by Erubi.

voxik:
Erubis is not maintained since 2011 and it has some unfixed security issues. Erubi is used by Ruby on Rails, so it should be much better choice.

### DIFF
--- a/lib/vagrant/util/template_renderer.rb
+++ b/lib/vagrant/util/template_renderer.rb
@@ -1,7 +1,7 @@
 require 'ostruct'
 require "pathname"
 
-require 'erubis'
+require 'erubi'
 
 module Vagrant
   module Util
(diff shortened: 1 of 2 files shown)


voxik:
BTW I'm going to use this patch for Fedora, because it does not make sense to maintain Erubis on Fedora, if upstream is dead.

And as usual, I'm not going to sign CLA, so feel free whatever you want to do with this PR.

voxik:
I should note that this break vagrant-libvirt, because vagrant-libvirt uses [Erubis](https://github.com/vagrant-libvirt/vagrant-libvirt/blob/d7debb68c634c80fbdae6e8bd4e3187581e80f8f/lib/vagrant-libvirt/util/erb_template.rb#L17), but it doesn't specify the Erubis dependency, because Vagrant always loaded Erubis previously.

chrisroberts:
Thanks for the notification :slightly_smiling_face:

Issue #11828, opened 2020-08-17
Vagrantfile intermittently fails with winrm_port undefined method


shoddyguard:
Occasionally when trying to `vagrant up` from the below vagrantfile we're getting the error: 
`C:/HashiCorp/Vagrant/embedded/gems/2.2.7/gems/vagrant-2.2.7/plugins/communicators/winrm/helper.rb:61:in winrm_port': undefined method each for nil:NilClass (NoMethodError)`

This happens randomly and re-running `vagrant up` will result in the machines getting built correctly. This is happening across multiple hosts.

I've tried to remove a lot of the fluff from the included vagrantfile while still including the parts relevant to our build process, hopefully it contains all the relevant info.

### Vagrant version

2.2.7

### Host operating system

Windows Server 2019

### Guest operating system

Various Windows - though in the specific example provided Windows Server 2019

### Vagrantfile

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :
require 'getoptlong'

opts = GetoptLong.new(
  [ '--noDevTools', GetoptLong::OPTIONAL_ARGUMENT ],
  [ '--nodevtools', GetoptLong::OPTIONAL_ARGUMENT ]
)

installDevToolsLocally = true

opts.ordering=(GetoptLong::REQUIRE_ORDER)

opts.each do |opt, arg|
  case opt
    when '--noDevTools'
        installDevToolsLocally=false
    when '--nodevtools'
        installDevToolsLocally=false
  end
end

machines_for_application_tests=[
   {
    :name => "sql2019", # tests a clean installation
    :box_name => "red-gate/windows-2019-sql2019",
    :box_version => '>= 2020.03.24',
    :sql_server_instance_name => "sql2019",
    :ip => "192.168.254.119"
   },
]

Vagrant.configure(2) do |config|
  machines_for_application_tests.each do |machine|

    config.vm.define machine[:name] do |node|

      node.vm.box = machine[:box_name]
      node.vm.box_version = machine[:box_version]
      config.vm.network "private_network", ip: machine[:ip]
      node.vm.boot_timeout = 600

      node.vm.provider "virtualbox" do |vb|
        vb.customize 'pre-boot', ["setextradata", :id, "GUI/LastGuestSizeHint", "1152,864"]
        vb.gui = ENV['TEAMCITY_VERSION'].nil?
        vb.linked_clone = true
        vb.memory = "2048"
        vb.cpus = 2
      end
  end
end
```

### Expected behavior

[...]

PR #11831, opened 2020-08-17
Check for nil return from forwarded ports cap

soapy1:
Nil may be returned if the forwarded ports cap is called while the machine is not in a running state.

Fixes https://github.com/hashicorp/vagrant/issues/11828


chrisroberts (dismissed):
:+1:

PR #11835, opened 2020-08-18
Download a box by setting auth headers

soapy1:


lib/vagrant/box.rb
```diff
@@ -58,12 +58,13 @@ class Box
     # @param [Pathname] directory The directory where this box exists on
     #   disk.
     # @param [String] metadata_url Metadata URL for box
```
chrisroberts, line 60: Can you add a doc line for the new hook parameter?

lib/vagrant/util/downloader.rb
```diff
@@ -29,8 +29,9 @@ class Downloader
         "vagrantup.com".freeze
       ].freeze
 
-      attr_reader :source
+      attr_accessor :source
       attr_reader :destination
+      attr_accessor :headers
```
chrisroberts, line 34: In the initializer we should verify that the headers are an array (`@headers = Array(@options[:headers])`)

plugins/commands/cloud/auth/middleware/add_downloader_authentication.rb
```diff
@@ -0,0 +1,54 @@
...
+          if target_url.host == server_uri.host
+            if server_uri.host != TARGET_HOST && !self.class.custom_host_notified?
+              env[:ui].warn(I18n.t("cloud_command.middleware.authentication.different_target",
+                custom_host: server_uri.host, known_host: TARGET_HOST) + "\n")
+              sleep CUSTOM_HOST_NOTIFY_WAIT
+              self.class.custom_host_notified!
+            end
+
+            if !Array(env[:downloader].headers).any? { |h| h.include?("Authorization") }
+              env[:downloader].headers ||= []
```
chrisroberts, line 42: With the downloader initializer setting up the `headers` variable as an array type, setting it shouldn't be necessary here.

plugins/commands/cloud/auth/middleware/add_downloader_authentication.rb
```diff
@@ -0,0 +1,54 @@
...
+        if token && !server_uri.host.to_s.empty?
+          if target_url.host == server_uri.host
+            if server_uri.host != TARGET_HOST && !self.class.custom_host_notified?
+              env[:ui].warn(I18n.t("cloud_command.middleware.authentication.different_target",
+                custom_host: server_uri.host, known_host: TARGET_HOST) + "\n")
+              sleep CUSTOM_HOST_NOTIFY_WAIT
+              self.class.custom_host_notified!
+            end
+
+            if !Array(env[:downloader].headers).any? { |h| h.include?("Authorization") }
```
chrisroberts, line 41: Would likely be useful to have an `else` on this that logs authentication header isn't being added due to one already being set.

chrisroberts (changes requested):
Looks great, just a few minor things

chrisroberts (approved):
:+1:

Issue #11833, opened 2020-08-18
Credential scrubber can corrupt the synced_folders JSON cache


LachlanArthur:
### Vagrant version

2.2.9

### Host operating system

Windows 10

### Guest operating system

Ubuntu

### Vagrantfile

```ruby
Vagrant.configure('2') do |config|
	config.vm.box = 'bento/ubuntu-16.04'
	config.vm.synced_folder __dir__, '/home/vagrant/example', type: 'smb'
end
```

### Steps to reproduce

**First stage:** ([debug.log](https://gist.github.com/LachlanArthur/7994102e21fed4680f2ff717a997e483))

1. Run `vagrant up` with an SMB share.
2. Provide an incorrect password `a`.
3. An error is displayed and you are prompted to retry.
4. Provide the correct password.

**Second stage:** ([debug.log](https://gist.github.com/LachlanArthur/f2d8b5792dd5d818a75e5b6976333427))

1. Run `vagrant reload`

### Expected behaviour

No error messages are displayed, and the cached JSON config file `.vagrant/machines/default/<provider>/synced_folders` is valid.

### Actual behaviour

After providing the password "a" in the first set of steps, the cached JSON file is now invalid as all instances of the letter `a` have been changed to `*****`. This breaks the unquoted boolean value `false`, as it becomes `f*****lse`.

Running `vagrant reload` now breaks when the invalid JSON file is parsed. The error message is opaque and does not mention the path to the broken file.

### Solution

The credential scrubber should traverse the config and only run on string values before the structure is serialised to JSON.

Also, this error should be caught and the user told where the invalid JSON file is so they can fix/delete it.

PR #11837, opened 2020-08-18
Scrub credentials as whole words, don't capture matching substrings

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11833

### DIFF
--- a/lib/vagrant/util/credential_scrubber.rb
+++ b/lib/vagrant/util/credential_scrubber.rb
@@ -32,7 +32,7 @@ module Vagrant
       def self.desensitize(string)
         string = string.to_s.dup
         sensitive_strings.each do |remove|
-          string.gsub!(remove, REPLACEMENT_TEXT)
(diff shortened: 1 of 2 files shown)


lib/vagrant/action/builtin/mixin_synced_folders.rb
```diff
@@ -98,11 +98,6 @@ def save_synced_folders(machine, folders, opts={})
           end
 
           folder_data = JSON.dump(folders)
-
-          # Scrub any register credentials from the synced folders
-          # configuration data to prevent accidental leakage
-          folder_data = Util::CredentialScrubber.desensitize(folder_data)
```
soapy1, line 104: I think this PR should also maybe remove this line that desensitizes the synced_folder export. Passwords are (and should) never stored be as part of folder data. I think there is also a case to be made about desensitizes information here can lead to leaking of credentials. For example if an exported folder is named "vagrant" and the users password is "vagrant", the synced_folder cache will show "****" in place of the folder name, indicating that it is also password.

But, maybe I'm missing some other consideration here?

PR #11848, opened 2020-08-25
Allow dependabot to update GitHub action versions

jlosito:
https://docs.github.com/en/github/administering-a-repository/keeping-your-actions-up-to-date-with-github-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


chrisroberts:
Hi there,

Thanks for creating this PR, however this isn't something that we are going to be enabling at this point. We use an extremely minimal number of internally accepted actions within our workflows. Because of this, we aim to keep these actions consistent across all our repositories and will update them together once a new version has been accepted and verified.

Thanks again for taking the time to create this PR!

Cheers!

Issue #11852, opened 2020-08-26
SSH connection failed after provisioning step with reboot


skudra:
### Vagrant version 

2.2.10
Also happens on 2.2.8 and 2.2.9
Works fine on 2.2.7

### Host operating system 

Windows 10, 2004

### Guest operating system

Ubuntu 18.04.3 LTS

### Vagrantfile

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
 
  config.vm.box = "hashicorp/bionic64"
  
  config.vm.provision "Before", type: "shell",  run: "always" do |s|
    s.inline="echo 'Before reboot' | tee before.txt"
    s.reboot = true
  end
  
  config.vm.provision "After", type: "shell",  run: "always" do |s|
    s.inline="echo 'After reboot' | tee after.txt"
  end
  
end
```

### Debug output

https://gist.github.com/skudra/9b9e5777bbffbd1c804cd2171c0700c6

### Expected behavior

Vagrant should wait some time before checking if guest is available. This would give time for guest to properly reboot.

### Actual behavior

Looks like this code: https://github.com/hashicorp/vagrant/blob/ce1fb54a04f6c85e0a3b8c0427a3a5530ebced48/plugins/guests/linux/cap/reboot.rb#L41
detected that ssh is available *before* guest stopped sshd and proceeded with provisioning. Moments later guest actually rebooted and ssh connection was lost.
After looking at the code I noticed that this issue started to happen when systemctl support was added. If I modify reboot.rb to always use "reboot" as reboot_script the issue does not happen. Also, if I modify  reboot.rb for Vagrant 2.2.7 - the issue appears.

### Steps to reproduce

1. Launch provided Vagrantfile
2. Observe error:

> SSH connection was reset! This usually happens when the machine is
> taking too long to reboot. First, try reloading your machine with
> `vagrant reload`, since a simple restart sometimes fixes things.
> If that doesn't work, destroy your machine and recreate it with
> a `vagrant destroy` followed by a `vagrant up`. If that doesn't work,
> contact support.

This issue seems to related to time it takes for sshd to stop, before Vagrant checks for host. Reproduces in about 4 runs out of 5.

### References

Merge request that added systemd reboot support:
- https://github.com/hashicorp/vagrant/pull/11483

irfanjunaid:
Hi,
I am facing the same exact issue on the latest version of vagrant. The command prompt immediately exits with connection abort error.

[...]

PR #11860, opened 2020-08-27
Reboot now

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11852

Here is some useful docs: 
* https://wiki.debian.org/systemd
*https://www.commandlinux.com/man-page/man1/systemctl.1.html

### DIFF
--- a/test/unit/plugins/guests/linux/cap/reboot_test.rb
+++ b/test/unit/plugins/guests/linux/cap/reboot_test.rb
@@ -68,7 +68,7 @@ describe "VagrantPlugins::GuestLinux::Cap::Reboot" do
         it "reboots the vm" do
           allow(communicator).to receive(:execute)
 
-          expect(communicator).to receive(:execute).with(/systemctl reboot/, nil).and_return(0)
+          expect(communicator).to receive(:execute).with(/systemctl.*?reboot/, nil).and_return(0)
           expect(described_class).to receive(:wait_for_reboot)
 
           described_class.reboot(machine)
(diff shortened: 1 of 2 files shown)


test/unit/plugins/guests/linux/cap/reboot_test.rb
```diff
@@ -68,7 +68,7 @@
         it "reboots the vm" do
           allow(communicator).to receive(:execute)
 
-          expect(communicator).to receive(:execute).with(/systemctl reboot/, nil).and_return(0)
+          expect(communicator).to receive(:execute).with(/systemctl/, nil).and_return(0)
```
chrisroberts, line 71: I think it would be useful to retain the `reboot` bit just to ensure that's what the command is doing. You can update the pattern to allow other stuff and still match: `/systemctl.*?reboot/`

chrisroberts (changes requested):
Just a small change on the test update

chrisroberts (approved):

soapy1:
Testing this out a bit more, this change mostly fixes the issue. However, this issue still persists for older versions of fedora and centos. eg. `bento/centos-6.7`, `bento/fedora-32`

i-ky:
@soapy1 @chrisroberts What is the status of this PR? Any updates?

soapy1:
Hmm, it looks like this PR can actually be closed. https://github.com/hashicorp/vagrant/pull/12302 resolves #11852 🙌

Issue #11856, opened 2020-08-27
Strange error after 2.2.9 -> 2.2.10


dburba:
After updating to v2.2.10 I get error on vagrant up:

```
Traceback (most recent call last):
	17: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/bin/vagrant:205:in `<main>'
	16: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/environment.rb:290:in `cli'
	15: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/cli.rb:67:in `execute'
	14: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/plugins/commands/up/command.rb:87:in `execute'
	13: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/plugin/v2/command.rb:232:in `with_target_vms'
	12: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/plugin/v2/command.rb:232:in `each'
	11: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/plugin/v2/command.rb:233:in `block in with_target_vms'
	10: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/machine_index.rb:150:in `include?'
	 9: from /opt/vagrant/embedded/lib/ruby/2.6.0/monitor.rb:235:in `mon_synchronize'
	 8: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/machine_index.rb:151:in `block in include?'
	 7: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/machine_index.rb:347:in `with_index_lock'
	 6: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/machine_index.rb:347:in `open'
	 5: from /opt/vagrant/embedded/gems/2.2.10/gems/vagrant-2.2.10/lib/vagrant/machine_index.rb:349:in `block in with_index_lock'
[...]

PR #11863, opened 2020-08-28
Include index

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11856

### DIFF
--- a/lib/vagrant/machine_index.rb
+++ b/lib/vagrant/machine_index.rb
@@ -263,7 +263,7 @@ module Vagrant
     #
     # @return [Hash]
     def find_by_prefix(prefix)
-      return nil if prefix.nil?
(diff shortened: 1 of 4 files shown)


lib/vagrant/machine_index.rb
```diff
@@ -263,6 +263,7 @@ def recover(entry)
     #
     # @return [Hash]
     def find_by_prefix(prefix)
+      return nil if prefix.nil?
```
chrisroberts, line 266: This can be reduced to `return if !prefix`

chrisroberts (dismissed):
Just a styling nit, but this looks great. Feel free to merge.

soapy1:
Merging, since last update was just to address styling 🚀

PR #11867, opened 2020-09-01
Avoid sorting of controllers with nil boot_priority

marxarelli:
Some unsupported storage controllers (e.g. floppy) report a nil
boot_priority which results in a failed sort. Select only supported
storage controllers to avoid that case.


chrisroberts (approved):
:+1: thanks!

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11883

PR #11872, opened 2020-09-02
Run vagrant-spec tests

soapy1:


.ci/spec/clean-packet.sh
```diff
@@ -4,18 +4,28 @@ csource="${BASH_SOURCE[0]}"
...
+. "${root}/.ci/spec/env.sh"
 
 pushd "${root}" > "${output}"
 
 echo "Cleaning up packet device..."
 
+unset PACKET_EXEC_PERSIST
+unset PACKET_EXEC_PRE_BUILTINS
+
+export PACKET_EXEC_REMOTE_DIRECTORY="run-${GITHUB_RUN_ID}"
```
soapy1, line 17: For some reason, by the time that the ci gets to this part the `job_id` changes from `run-${GITHUB_RUN_ID}` to `run1-${GITHUB_RUN_ID}`. This forces it back.

briancain (dismissed):
Approving just cuz I am excited to see this happen! :tada: Thanks for finishing it out. :pray: :chart_with_upwards_trend:

chrisroberts (changes requested):

.ci/spec/run-test.sh
```diff
@@ -4,31 +4,27 @@ csource="${BASH_SOURCE[0]}"
...
 . "${root}/.ci/common.sh"
+. "${root}/.ci/spec/env.sh"
 
 pushd "${root}" > "${output}"
 
 # Assumes packet is already set up
-
-# job_id is provided by common.sh
-export PACKET_EXEC_REMOTE_DIRECTORY="${job_id}"
+unset PACKET_EXEC_PRE_BUILTINS
```
soapy1, line 13: When this is set, it re-installs all the vmware tools and causes `vagrant --provision` to fail

chrisroberts: :+1: though we should have a look at that helper and see why it's not a no-op at this point.

chrisroberts (approved):
:tada:

Issue #11875, opened 2020-09-03
Check if /etc/sysconfig/network exists before trying to edit


konstruktoid:
At https://github.com/hashicorp/vagrant/blob/master/plugins/guests/redhat/cap/change_host_name.rb#L14-L21 `sed` tries to edit `/etc/sysconfig/network` and related files, but if these doesn't exist in the box Vagrant fails.

```
==> fedora: Setting hostname...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

# Update sysconfig
sed -i 's/\(HOSTNAME=\).*/\1fedora/' /etc/sysconfig/network
# Update DNS
sed -i 's/\(DHCP_HOSTNAME=\).*/\1"fedora"/' /etc/sysconfig/network-scripts/ifcfg-*

Stdout from the command:

Stderr from the command:

sed: can't read /etc/sysconfig/network-scripts/ifcfg-*: No such file or directory
```

`vagrant ssh fedora` works after a while:

```
[vagrant@localhost ~]$ sudo grep -Ri hostname /etc/sysconfig/
[vagrant@localhost ~]$ ls /etc/sysconfig/
anaconda  authconfig  chronyd  console  firewalld  kernel  man-db  modules  network  network-scripts  nftables.conf  rpcbind  selinux  sshd
[vagrant@localhost ~]$ ls /etc/sysconfig/network-scripts/
[vagrant@localhost ~]$ cat /etc/redhat-release 
Fedora release 32 (Thirty Two)
[vagrant@localhost ~]$ sudo hostnamectl set-hostname fedora
[vagrant@localhost ~]$ hostname
fedora
[vagrant@localhost ~]$
```

### Vagrant version

Vagrant 2.2.10

### Guest operating system

`bento/fedora-32`

### Vagrantfile

```ruby
  config.vm.define "fedora" do |fedora|
    fedora.vm.box = "bento/fedora-32"
    fedora.ssh.insert_key = true
    fedora.vm.network "private_network", ip: "10.2.3.44"
    fedora.vm.provider "virtualbox" do |f|
      f.default_nic_type = "virtio"
    end
    fedora.vm.hostname = "fedora"
    fedora.vm.provision "shell",
      inline: "dnf install -y ansible"
    fedora.vm.provision "ansible" do |a|
      a.verbose = "v"
      a.limit = "all"
      a.playbook = "tests/test.yml"
      a.extra_vars = {
[...]

Issue #11873, opened 2020-09-03
Not Been able to Bring up the centos VM on Mac OS Catalina 


prem548:
Vagrant Version --> Vagrant 2.2.10
Virtual Box --> 6.1.12r139181
Host Operating System: Mac OS Catalina version 10.15.1
Guest Operating System --> "vitals/centos-6.10"

**Vagrant File:** 
   require 'yaml'

if Vagrant.plugins_enabled?
  if !Vagrant.has_plugin?('vagrant-vbguest')
    raise "Guest Additions not installed. To fix this please run:\n\n\tvagrant plugin install vagrant-vbguest\n"
  end
end

VAGRANTFILE_API_VERSION = "2"
os_name = `sw_vers -productName`
os_version = `sw_vers -productVersion`.to_f
ROOT_DIR = (os_name.include?('Mac OS') && os_version >= 10.15) ? "/System/Volumes/Data#{Dir.pwd}" : '/System/Volumes/Data/Users/UserName/Documents'
BOOTSTRAP_ENVIRONMENT = 'some_environment'
DEBUG = %w(1 true).include?(ENV['DEBUG'])
DEFAULTS = {
  name: "testone.vitals.vm",
  hostname: "testone.vitals.vm",
  box: "vitals/centos-6.6",
  box_version: "1.0.0",
  centos_version: 6,
  organization: ENV.fetch('ORGANIZATION', "engineering"),
  client: ENV.fetch('CLIENT', "engineering"),
  env: ENV.fetch('ENV', "localdev"),
  variant: ENV.fetch('VARIANT', nil),
  role: ENV.fetch('ROLE', 'myrole'),
  sub_role: ENV.fetch('SUB_ROLE', nil),
  ip: "IP Goes here"
  login_as_deploy: false,
  modules_dir: ["#{ROOT_DIR}/modules/"],
  manifests_path: "#{ROOT_DIR}/manifests/",
  manifest_file: 'site.pp',
  memory: 1024,
  cpus: 1,
  forward_agent: false,
  auto_nat_dns_proxy: 'false',
  natdnshostresolver1: 'on',
  natdnsproxy1: 'on'
}

boxes = [{}]

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|

  boxes.each do |box_conf|
    instance = DEFAULTS.merge(box_conf)

    config.vm.define instance[:name] do |config|
      config.vm.box = "vitals/centos-6.10"
[...]

PR #11877, opened 2020-09-03
Check for existence of network files before trying to update them

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11875 and https://github.com/hashicorp/vagrant/issues/11873

### DIFF
--- a/plugins/guests/redhat/cap/change_host_name.rb
+++ b/plugins/guests/redhat/cap/change_host_name.rb
@@ -15,9 +15,13 @@ module VagrantPlugins
             basename = name.split('.', 2)[0]
             comm.sudo <<-EOH.gsub(/^ {14}/, '')
               # Update sysconfig
-              sed -i 's/\\(HOSTNAME=\\).*/\\1#{name}/' /etc/sysconfig/network
+              if [ -f /etc/sysconfig/network ]; then
+                sed -i 's/\\(HOSTNAME=\\).*/\\1#{name}/' /etc/sysconfig/network
+              fi
               # Update DNS
-              sed -i 's/\\(DHCP_HOSTNAME=\\).*/\\1\"#{basename}\"/' /etc/sysconfig/network-scripts/ifcfg-*
+              find /etc/sysconfig/network-scripts -maxdepth 1 -type f -name 'ifcfg-*' | xargs sed -i 's/\\(DHCP_HOSTNAME=\\).*/\\1\"#{basename}\"/'
+              # Set the hostname - use hostnamectl if available
+              echo '#{name}' > /etc/hostname
             EOH
 
             if hostnamectl?(comm)


plugins/guests/redhat/cap/change_host_name.rb
```diff
@@ -15,16 +15,20 @@ def self.change_host_name(machine, name)
             basename = name.split('.', 2)[0]
             comm.sudo <<-EOH.gsub(/^ {14}/, '')
               # Update sysconfig
-              sed -i 's/\\(HOSTNAME=\\).*/\\1#{name}/' /etc/sysconfig/network
+              if ls /etc/sysconfig | grep network; then
```
chrisroberts, line 18: we can do a file check with:

`if [ -f /etc/sysconfig/network ]; then`

ikari7789: Also safer as grepping on the result of `ls` would match any file with "network" in it's name inside of the folder leading to possible false positives.

chrisroberts (approved):
:+1: looks good

brad2014:
@soapy1, @chrisroberts, I suspect this is still not quite right, my notes [here](https://github.com/hashicorp/vagrant/commit/7370f24318a4e3ced61922886f164834c7728f7e#r42356848).

plugins/guests/redhat/cap/change_host_name.rb
```diff
@@ -15,9 +15,13 @@ def self.change_host_name(machine, name)
             basename = name.split('.', 2)[0]
             comm.sudo <<-EOH.gsub(/^ {14}/, '')
               # Update sysconfig
-              sed -i 's/\\(HOSTNAME=\\).*/\\1#{name}/' /etc/sysconfig/network
+              if [ -f /etc/sysconfig/network ]; then
+                sed -i 's/\\(HOSTNAME=\\).*/\\1#{name}/' /etc/sysconfig/network
+              fi
               # Update DNS
-              sed -i 's/\\(DHCP_HOSTNAME=\\).*/\\1\"#{basename}\"/' /etc/sysconfig/network-scripts/ifcfg-*
+              find /etc/sysconfig/network-scripts -maxdepth 1 -type f -name 'ifcfg-*' | xargs sed -i 's/\\(DHCP_HOSTNAME=\\).*/\\1\"#{basename}\"/'
```
brad2014, line 22: This probably needs an xargs -r flag.

soapy1:
@brad2014 thanks for following up on this!
Here is what I found, looking into it a little more. At first, it seems like this PR fixes the issue since the hostname is successfully updated without any errors. However, this is due to the addition of this line https://github.com/hashicorp/vagrant/pull/11877/files#diff-fd473b135e71706459b47b51cb86e249R24 which has a successful output. Since `comm.sudo` is only checking the return code of the last command this appears to works. Removing that line exposes the failure of `find /etc/sysconfig/network-scripts -maxdepth 1 -type f -name 'ifcfg-*' | xargs sed -i 's/\\(DHCP_HOSTNAME=\\).*/\\1\"#{basename}\"/'` which has a return code of `123`. Adding the `-r` argument to `xargs` fixes this. Here is a PR for this https://github.com/hashicorp/vagrant/pull/11910. 
Thanks again for your contribution ❤️

Issue #11802, opened 2020-08-03
Vagrant 2.2.9 : The executable 'docker' Vagrant is trying to run was not found in the %PATH% variable.


lpsguilloul:
### Vagrant version
Vagrant 2.2.9

### Host operating system

Windows 10 x64

### Guest operating system

centos/7

### Vagrantfile
I use two files : Vagrantfile and Host_Vagrantfile

**Vagrantfile**
```ruby
ENV['VAGRANT_DEFAULT_PROVIDER'] = 'docker'
VAGRANTFILE_API_VERSION = "2"
Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  config.vm.define "jenkins" do |v|
    v.vm.provider "docker" do |d|
	  d.force_host_vm = true
      d.image = "jenkins/jenkins:2.190.3-centos"
      d.ports = ["8080:8080"]
      d.vagrant_machine = "dockerhost"
      d.vagrant_vagrantfile = "./Host_Vagrantfile"
	end
  end
end
```
**Host_Vagrantfile**
```ruby
VAGRANTFILE_API_VERSION = "2"
Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  config.vm.define "dockerhost"
  config.vm.box = "centos/7"
  config.vm.box_version = "2004.01"
  config.vm.provision "docker"
  config.vm.network "forwarded_port", guest: 8080, host: 8080,
      auto_correct: true
  config.vm.network "private_network", ip: "172.28.128.2"
  config.vm.provider :virtualbox do |vb|
    vb.name = "dockerhost"
  end
  config.vm.provision "docker"
  config.vm.hostname = "dockerhost"
end
```
### Debug output

https://gist.github.com/lpsguilloul/7baf85c37d184b7930f44ce97862f0e4

### Expected behavior

I'm using docker provider.
When vagrant up is executed, vagrant should spin up the VM host and install Docker before executing any docker commands to set up the containers.

### Actual behavior

Vagrant doesn't spin up the VM host.  It's as if, it's ignoring the vagrant_vagrantfile setting passed to the docker provider.

When I remove "ENV['VAGRANT_DEFAULT_PROVIDER'] = 'docker'" and execute vagrant up, I get another error :

 

> _INFO interface: error: There are errors in the configuration of this machine. Please fix
> the following errors and try again:
> 
> vm:
> * A box must be specified.
> 
> 
> There are errors in the configuration of this machine. Please fix
[...]

PR #11892, opened 2020-09-09
Docker usable check

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11802


soapy1:
Here is one idea about how to solve the issue about checking for docker on the host system. The idea is that users are able to "turn off" checking if their provider is usable on the host system. This would be useful in the case of using a Vagrant host to manipulate docker. Not sure if this is an ideal solution though, very open to suggestions 🌵

PR #11893, opened 2020-09-10
Switch from unmaintained gem erubis to erubi

dcermak:
Rails switched to erubi as well, as erubis has not seen any new releases since 2011.


chrisroberts (approved):
:+1: thanks!

Issue #11894, opened 2020-09-10
sed error when there's no /etc/fstab


marga-kinvolk:
### Issue description

When trying to start a vagrant machine using the latest Flatcar image, it fails with this error:
```
==> default:   - Exclude: [".vagrant/", ".git/"]
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

sed -i '/#VAGRANT-BEGIN/,/#VAGRANT-END/d' /etc/fstab

Stdout from the command:

Stderr from the command:
sed: can't read /etc/fstab: No such file or directory
```

Flatcar indeed doesn't ship an `/etc/fstab` file, because all mounts are managed by systemd. This is becoming common with modern Linux distributions and should be supported by vagrant.

The problematic code is here:
https://github.com/hashicorp/vagrant/blob/master/plugins/guests/linux/cap/persist_mount_shared_folder.rb

And it seems likely that the problem was introduced with the refactor:
https://github.com/hashicorp/vagrant/commit/1dc761a6c423f1288bfe2333738695a22be1f685

### Vagrant version

2.2.10 is affected. 2.2.9 doesn't have the issue.

### Guest operating system

Flatcar Container Linux

### Vagrantfile

```
# -*- mode: ruby -*-
# vi: set ft=ruby :
Vagrant.configure("2") do |config|
  channel = ENV["TF_VAR_flatcar_channel"] || "alpha"

  # Box setup.
  config.vm.box     = "flatcar-#{channel}"
  config.vm.box_url = "https://#{channel}.release.flatcar-linux.net/amd64-usr/current/flatcar_production_vagrant.box"

  # Virtualbox resources.
  config.vm.provider :virtualbox do |v|
    v.check_guest_additions = false
    v.functional_vboxsf     = false
    v.cpus                  = 2
    v.memory                = 2048
    v.customize ['modifyvm', :id, '--paravirtprovider', 'kvm']
  end

  # SSH.
  config.ssh.username = 'core'

  # Sync using rsync, but don't copy locally built binaries and don't remove Terraform files from virtual machine.
  config.vm.synced_folder ".", "/home/core/test", type: "rsync", rsync__exclude: [".git/"]

  # Read content of Vagrant SSH private key.
  ssh_private_key = File.read(ENV['HOME'] + "/.vagrant.d/insecure_private_key")

  # Primary node provisioning.
  config.vm.provision "shell", inline: <<-EOF
    set -e
[...]

PR #11897, opened 2020-09-10
Add check for /etc/fstab

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11894

### DIFF
--- a/plugins/guests/linux/cap/persist_mount_shared_folder.rb
+++ b/plugins/guests/linux/cap/persist_mount_shared_folder.rb
@@ -8,6 +8,8 @@ module VagrantPlugins
       class PersistMountSharedFolder
         extend SyncedFolder::UnixMountHelpers
 
+        @@logger = Log4r::Logger.new("vagrant::guest::linux::persist_mount_shared_folders")
+
         # Inserts fstab entry for a set of synced folders. Will fully replace
         # the currently managed group of Vagrant managed entries. Note, passing
         # empty list of folders will just remove entries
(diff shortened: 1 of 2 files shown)


dcermak (approved):
I have tested this locally with a box that lacked a /etc/fstab and the issue is resolved by that.

plugins/guests/linux/cap/persist_mount_shared_folder.rb
```diff
@@ -53,6 +61,10 @@ def self.persist_mount_shared_folder(machine, folders)
 
         private
 
+        def self.fstab_exists?(machine)
+          machine.communicate.test("stat /etc/fstab")
```
chrisroberts, line 65: Can we simplify this down to just `test -f /etc/fstab` ?

plugins/guests/linux/cap/persist_mount_shared_folder.rb
```diff
@@ -8,14 +8,22 @@ module Cap
...
+        @@logger = Log4r::Logger.new("vagrant::guest::linux::persist_mount_shared_folders")
+
         # Inserts fstab entry for a set of synced folders. Will fully replace
         # the currently managed group of Vagrant managed entries. Note, passing
         # empty list of folders will just remove entries
         #
         # @param [Machine] machine The machine to run the action on
         # @param [Map<String, Map>] A map of folders to add to fstab
         def self.persist_mount_shared_folder(machine, folders)
+          unless fstab_exists?(machine)
```
chrisroberts, line 20: Update the `unless` to an `if !fstab_exists?` simply due to the overall preference for `if` based conditionals

chrisroberts (changes requested):

chrisroberts (approved):

marga-kinvolk:
Thanks for working on this fix!

I do find the logic of this change a bit strange. If `/etc/fstab` doesn't exist, it bails out of doing the additions that it was planning to do. It would be fine to create an `/etc/fstab` that doesn't exist. At least in Flatcar it doesn't exist because all the OS mounts are managed by systemd, not because creating it doesn't work.

In other words, I think the check should be inside `remove_vagrant_managed_fstab` rather than on the whole function.

Thanks again.

soapy1:
@marga-kinvolk thanks for following up about this!
It seems I have misunderstood the interactions between `/etc/fstab` and systemd. It seems desirable to allow Vagrant to modify fstab regardless of if systemd is being used to manage mounts. Going to push up a change in a little bit 🚀 

ref: https://docs.flatcar-linux.org/os/kernel-modules/#create-a-writable-overlay

PR #11899, opened 2020-09-10
Don't try to "fix" a /etc/fstab that does not exist

dcermak:
`remove_vagrant_managed_fstab()` gets called in `self.persist_mount_shared_folder` when `folders` is `nil`. This is an issue for boxes that don't have a `/etc/fstab`, as they can boot but will break once the `sed` call is executed.

Thus we guard it with a check whether `/etc/fstab` exists and only try to remove the modifications then.


dcermak:
This is an alternative fix for #11894

chrisroberts:
@dcermak Thank you for this PR (and for testing out the other one!). I'm going to move forward with the other one only because it isolates things out a bit so we can implement more test coverage on it. But thank you very much for this and the help!

Cheers

PR #11912, opened 2020-09-16
--yes options added to CloudCommand::VersionCommand::Command::Release

kipal:
Hi everyone,

this is my first contribution here (and the very first contribution with ruby).
I ran into a little problem at my workplace during automating a vagrant box build process namely that there's no 'assume yes' option at the ```vagrant cloud version release``` command which requires a 'y' to release the version.
I know that I could use programs like 'expect' and 'yes' to solve this but I've always found it more elegant to use the offered ```--yes``` option.

P.S: I hope I didn't violate any of the rules of this community. If I did, please let me know and I'll try my best.

Thank you in advance for looking into my PR.

### DIFF
--- a/plugins/commands/cloud/version/release.rb
+++ b/plugins/commands/cloud/version/release.rb
@@ -18,6 +18,10 @@ module VagrantPlugins
               o.on("-u", "--username USERNAME_OR_EMAIL", String, "Vagrant Cloud username or email address") do |u|
                 options[:username] = u
               end
+              options[:assume_yes] = false
+              o.on("-y", "--yes", "Automatic yes to prompts") do |y|
+                options[:assume_yes] = y
+              end
             end
 
             # Parse the options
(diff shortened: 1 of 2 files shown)


plugins/commands/cloud/version/release.rb
```diff
@@ -18,6 +18,10 @@ def execute
               o.on("-u", "--username USERNAME_OR_EMAIL", String, "Vagrant Cloud username or email address") do |u|
                 options[:username] = u
               end
+              options[:assume_yes] = false
+              o.on("-y", "--yes", "Automatic yes to prompts") do |y|
```
chrisroberts, line 22: Update to: 

```
o.on("-f", "--force", "Release without confirmation") do |f|
  options[:force] = f
end
```

plugins/commands/cloud/version/release.rb
```diff
@@ -28,9 +32,11 @@ def execute
                 help: opts.help.chomp
             end
 
-            @env.ui.warn(I18n.t("cloud_command.version.release_warn", version: argv[1], box: argv.first))
-            cont = @env.ui.ask(I18n.t("cloud_command.continue"))
-            return 1 if cont.strip.downcase != "y"
+            if not options[:assume_yes]
```
chrisroberts, line 35:
```ruby
if !options[:force]
```

test/unit/plugins/commands/cloud/version/release_test.rb
```diff
@@ -41,9 +39,36 @@
...
+        and_return(version)
+
+      allow(version).to receive(:release).
+        and_raise(VagrantCloud::ClientError.new("Fail Message", "Message", 404))
+      expect(subject.execute).to eq(1)
+    end
+  end
+
+  context "non-interactive mode with arguments" do
+    let (:argv) { ["--yes", "vagrant/box-name", "1.0.0"] }
```
chrisroberts, line 70: `--force`

chrisroberts (changes requested):
@kipal Hi! Thanks for this contribution, it looks great. Just a request for a few modifications, mainly to use the `--force` flag for this behavior. It is what is used elsewhere within Vagrant for bypassing user interaction for confirmation so using it here as well would be ideal for consistency.

Thanks!

chrisroberts (approved):
awesome! thanks for the PR and the quick update! :tada:

PR #11924, opened 2020-09-24
Add guest definitions for Oracle Linux

AmedeeBulle:
As of today, Oracle Linux guests are seen as RedHat guests.

This PR defines Oracle Linux as guest, allowing specific handling where OL differs from RH.

From an implementation perspective it inherits from RedHat, that is: by default it will behave as a RedHat guest.

Using the above definition, we add the following provisioners:
- Ansible: using the Oracle provided EPEL repository instead of the Fedora one
- Docker: which is not available for RedHat
- Podman: the repository layout is different for OL.

--
Signed-off-by: Philippe Vanhaesendonck <philippe.vanhaesendonck@oracle.com>

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/oraclelinux/ansible_install.rb
+++ b/plugins/provisioners/ansible/cap/guest/oraclelinux/ansible_install.rb
@@ -1,0 +1,80 @@
+# Copyright (c) 2020, Oracle and/or its affiliates.
+# Licensed under the MIT License.
+
+require_relative "../facts"
+require_relative "../pip/pip"
+
+module VagrantPlugins
+  module Ansible
+    module Cap
+      module Guest
+        module OracleLinux
+          module AnsibleInstall
+
+            def self.ansible_install(machine, install_mode, ansible_version, pip_args, pip_install_cmd = "")
+              case install_mode
+              when :pip
+                pip_setup machine, pip_install_cmd
+                Pip::pip_install machine, "ansible", ansible_version, pip_args, true
+              when :pip_args_only
+                pip_setup machine, pip_install_cmd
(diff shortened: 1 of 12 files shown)


plugins/provisioners/ansible/cap/guest/oraclelinux/ansible_install.rb
```diff
@@ -0,0 +1,80 @@
...
+
+require_relative "../facts"
+require_relative "../pip/pip"
+
+module VagrantPlugins
+  module Ansible
+    module Cap
+      module Guest
+        module OracleLinux
+          module AnsibleInstall
```
soapy1, line 12: it would be sweet if this module gets a few tests, similar to https://github.com/hashicorp/vagrant/blob/master/test/unit/plugins/provisioners/ansible/cap/guest/alpine/ansible_install_test.rb

soapy1 (commented):
This looks great 🐧 , thanks for the addition! 
Just a couple of small suggestions

soapy1:
Hi there,
thanks for opening up an issue for adding support to a guest plugin. The fastest way to make your plugin available is to build it as an external plugin! We have some template repositories setup so building a new guest plugin should take you no time, and you get to keep control over it's own release cycle. Check out the [Ruby template](https://github.com/soapy1/vagrant-guest-plugin-skeleton) or the [Golang template](https://github.com/hashicorp/vagrant-guest-plugin-skeleton-go).

For more information on writing plugins see the docs on [plugin development](https://www.vagrantup.com/docs/plugins/development-basics). For more information on managing Vagrant plugins see the docs on [plugin usage](https://www.vagrantup.com/docs/plugins/usage).

Issue #11857, opened 2020-08-27
Please consider not using the mime gem


dcermak:
Starting with Vagrant 2.2.10 there is a new dependency: [mime](https://rubygems.org/gems/mime). Please consider rolling back the change that introduced the dependency on this gem, as it is effectively unmaintained (upstream repository is gone, last release in 2017). This makes packaging vagrant more complicated and also means that any issues that you encounter in that gem will probably never be fixed.

PR #11961, opened 2020-10-14
Update mime gem

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11857

### DIFF
--- a/lib/vagrant/util/mime.rb
+++ b/lib/vagrant/util/mime.rb
@@ -1,0 +1,72 @@
+require 'mime/types'
+require 'securerandom'
+
+module Vagrant
+  module Util
+    module Mime
+      class Multipart
+
+        attr_accessor :content
+
+        attr_accessor :content_type
+
+        attr_accessor :headers
+
+        attr_accessor :mime_version
+
+        def initialize(content_type="multipart/mixed", mime_version="1.0")
+          @content_id = "#{SecureRandom.alphanumeric(24)}@#{SecureRandom.alphanumeric(24)}.local"
+          @boundary = "Boundary_#{SecureRandom.alphanumeric(24)}"
+          @content_type = MIME::Types[content_type].first
+          @content = []
+          @mime_version = mime_version 
+          @headers = {
+            "Content-ID"=> "<#{@content_id}>",
+            "Content-Type"=> "#{content_type}; boundary=#{@boundary}",
+          }
+        end
+
+        def add(entry)
+          content << entry
+        end
+
+        # Output MimeEntity as a string
+        def to_s
+          output_string = ""
+          headers.each do |k, v|
+            output_string += "#{k}: #{v}\n"
+          end
+          output_string += "\n--#{@boundary}\n"
+          @content.each do |entry|
+            output_string += entry.to_s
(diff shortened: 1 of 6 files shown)


lib/vagrant/util/mime.rb
```diff
@@ -0,0 +1,69 @@
...
+        attr_accessor :content
+
+        attr_accessor :content_type
+
+        attr_accessor :headers
+
+        attr_accessor :mime_version
+
+        def initialize(content_type="multipart/mixed", mime_version="1.0")
+          @content_id = "#{SecureRandom.alphanumeric(24)}@#{SecureRandom.alphanumeric(24)}.local"
```
soapy1, line 18: not too sure about this `content_id` bit
https://tools.ietf.org/html/rfc2392

lib/vagrant/util/mime.rb
```diff
@@ -0,0 +1,86 @@
+require 'mime/types'
+require 'securerandom'
+
+module Vagrant
+  module Util
+    module Mime
+      class Multipart
+
+        # [List] collection of content part of the multipart mime
```
chrisroberts, line 9: Docs here should list it as a `@return` and the `[List]` should be an `Array`, ideally of the type inside:

```ruby
# @return [Array<String>] collection of ...
```

lib/vagrant/util/mime.rb
chrisroberts: The format is fine since it adheres to the url-addr-spec. To ensure uniqueness it might not be a bad idea to swap out the first section with a `Time.now.to_i` so you end up with something like: 

```ruby
"#{Time.now.to_i}@#{SecureRandom.alphanumeric(24)}.local"
```

lib/vagrant/util/mime.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        # @param [String] entity content
+        # @param [String] type of the entity content
+        def initialize(content, content_type)
+          if !MIME::Types.include?(content_type)
+            MIME::Types.add(MIME::Type.new(content_type))
+          end
+          @content = content
+          @content_type = MIME::Types[content_type].first
+          @content_id = "#{SecureRandom.alphanumeric(24)}@#{SecureRandom.alphanumeric(24)}.local"
```
chrisroberts, line 71: Same note as above with using `Time`

test/unit/vagrant/util/mime_test.rb
```diff
@@ -0,0 +1,73 @@
...
+
+describe Vagrant::Util::Mime::Multipart do
+
+  subject { described_class }
+
+  it "can add headers" do
+    allow(SecureRandom).to receive(:alphanumeric).and_return("123qwe")
+    mime = subject.new()
+    mime.headers["Mime-Version"] = "1.0"
+    expected_string = """Content-ID: <123qwe@123qwe.local>
```
chrisroberts, line 14: These `"""` in the string assignment can just be reduced to a single `"`

chrisroberts (changes requested):
Implementation looks good. Couple small notes.

chrisroberts (approved):

PR #11965, opened 2020-10-14
Add ability to specify target for cap command

soapy1:


plugins/commands/cap/command.rb
```diff
@@ -27,6 +27,10 @@ def execute
           o.on("--check", "Only checks for a capability, does not execute") do |f|
             options[:check] = f
           end
+
+          o.on("-t", "--target=TARGET", "Target guest to run against (if applicable)") do |t|
```
soapy1, line 31: I'm not super thrilled about providing this as a flag. The other ways that this could be done is like

`vagrant cap vmware_try provider delete_all_snapshots`
I didn't like this one because `vmware_try` doesn't really need to be specified

`vagrant cap provider delete_all_snapshots vmware_try`
I didn't like this one because currently the last bits of argv are options that get passed to the cap

Open to suggestions

chrisroberts (approved):
:+1: looks great

Issue #11966, opened 2020-10-15
`which.rb` fails if `PATH` is not set


dariocutillas:
### Vagrant version

I'm running 2.2.9 but I've checked the sources of `which.rb` in 2.2.10 and it is the same.

### Host operating system

Linux (Fedora 32).

### Guest operating system

ubuntu/bionic64.

### Vagrantfile
Not needed to reproduce the issue.

### Description

The implementation of `which.rb` fails if `env['PATH']` is `nil` on line 41:

```
search_path.encode(...)
```

### Expected behaviour

`which` shall not find whatever it is trying to find, that is ok, but we should not see an attempt to access `encode` on a nil reference because it is difficult to understand what is going on.

Assuming `PATH` is set is a wrong assumption and `env['PATH']` will return `nil` if the variable isn't set.

PR #11967, opened 2020-10-15
Check search_path for nil in which.rb

dariocutillas:
Related issue: #11966.

`search_path` can be `nil` because `PATH` might have not been set, which results in an exception when `search_path.encode` is exercised.

If `PATH` is not set, `which.rb` should probably simply consider that the executable isn't found.

Please note I don't develop in Ruby so edit as necessary.

### DIFF
--- a/lib/vagrant/util/which.rb
+++ b/lib/vagrant/util/which.rb
@@ -37,11 +37,13 @@ module Vagrant
           search_path = ENV['PATH']
         end
 
-        SilenceWarnings.silence! do
-          search_path.encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '').split(File::PATH_SEPARATOR).each do |path|
-            exts.each do |ext|
-              exe = "#{path}#{File::SEPARATOR}#{cmd}#{ext}"
-              return exe if File.executable? exe
+        if !search_path.nil?
+          SilenceWarnings.silence! do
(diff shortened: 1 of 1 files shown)


soapy1 (changes requested):
Thanks, this looks great! 🎉 
If you have some time, could you also add a little test for this. Some [tests already exist](https://github.com/hashicorp/vagrant/blob/master/test/unit/vagrant/util/which_test.rb#L42) for this module, they can provide a good example. 🦘

dariocutillas:
Thanks for feedback. I'll have a look during the weekend to see if I can add a relevant test.

lib/vagrant/util/which.rb
```diff
@@ -37,11 +37,13 @@ def self.which(cmd, **opts)
           search_path = ENV['PATH']
         end
 
-        SilenceWarnings.silence! do
-          search_path.encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '').split(File::PATH_SEPARATOR).each do |path|
-            exts.each do |ext|
-              exe = "#{path}#{File::SEPARATOR}#{cmd}#{ext}"
-              return exe if File.executable? exe
+        if !search_path.nil?
```
chrisroberts, line 40: I think that instead of including a conditional here, just forcing the type would be a bit cleaner:

```ruby
search_path.to_s.encode('UTF-8', ...
```

chrisroberts (changes requested):

dariocutillas:
I'm unfortunately not going to be able to sign the CLA given that my employer hasn't given me permission to do it. I guess someone else will have to submit the fix.

PR #11978, opened 2020-10-19
Add TPM functionality for Gen 2 Hyper-V machines

### LINKED ISSUE #11861: Enhancement Request: Add support for vTPM/TPM in Hyper-V Generation 2 VMs

I want to have a vTPM/TPM device enabled in a Hyper-V virtual machine.

Add a provider boolean flag, e.g., `hyperv.tpm = true` (the default should be `false`).

This should internally use the [Enable-VMTPM](https://docs.microsoft.com/en-us/powershell/module/hyper-v/enable-vmtpm?view=win10-ps)/[Disable-VMTPM](https://docs.microsoft.com/en-us/powershell/module/hyper-v/disable-vmtpm?view=win10-ps) cmdlets.

We can do this with a trigger, but having a boolean flag at the provider level is much more straitforward.

**Additional context**

This probably should also be added to packer and its box.xml (similar to what was done in https://github.com/hashicorp/vagrant/pull/11756).

grahhhhhhham:
This allows for the [Trusted Platform Module](https://docs.microsoft.com/en-us/powershell/module/hyper-v/enable-vmtpm?view=win10-ps) to be enabled for Hyper-V machines. I only have Windows 10 Home, so I can't actually test this out beyond unit testing in RSpec, and would appreciate if someone is able to do that.

Also, in general it's my first time contributing here so any pointers / changes just let me know.

Fixes #11861

### DIFF
--- a/plugins/providers/hyperv/scripts/configure_vm.ps1
+++ b/plugins/providers/hyperv/scripts/configure_vm.ps1
@@ -20,7 +20,9 @@ param(
     [parameter (Mandatory=$false)]
     [switch] $EnableCheckpoints,
     [parameter (Mandatory=$false)]
-    [switch] $EnableAutomaticCheckpoints
+    [switch] $EnableAutomaticCheckpoints,
+    [parameter (Mandatory=$false)]
+    [switch] $EnableTrustedPlatformModule
 )
 
 $ErrorActionPreference = "Stop"
@@ -124,4 +126,18 @@ try {
 } catch {
     Write-ErrorMessage "Failed to ${AutoAction} automatic checkpoints on VM: ${PSItem}"
     exit 1
+}
+
+if($VM.Generation -gt 1) {
+    try {
+        if($EnableTrustedPlatformModule) {
+            Hyper-V\Enable-VMTPM -VM $VM
+            $tpmAction = "enable"
+        } else {
+            Hyper-V\Disable-VMTPM -VM $VM
+            $tpmAction = "disable"
+        }
+    } catch {
+        Write-ErrorMessage "Failed to ${tpmAction} Trusted Platform Module on VM: ${PSItem}"
+    }
 }
--- a/Gemfile
+++ b/Gemfile
@@ -5,5 +5,5 @@ gemspec
 if File.exist?(File.expand_path("../../vagrant-spec", __FILE__))
   gem 'vagrant-spec', path: "../vagrant-spec"
 else
-  gem 'vagrant-spec', git: "https://github.com/hashicorp/vagrant-spec.git", branch: :main
+  gem 'vagrant-spec', git: "https://github.com/hashicorp/vagrant-spec.git"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -5,5 +5,5 @@ gemspec
 if File.exist?(File.expand_path("../../vagrant-spec", __FILE__))
   gem 'vagrant-spec', path: "../vagrant-spec"
 else
-  gem 'vagrant-spec', git: "https://github.com/hashicorp/vagrant-spec.git"
+  gem 'vagrant-spec', git: "https://github.com/hashicorp/vagrant-spec.git", branch: :main
 end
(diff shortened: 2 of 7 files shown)


soapy1 (changes requested):
Looking good so far!

Testing this out some I get this error
```
Failed to  Trusted Platform Module on VM: The operation failed.

The selected security settings of a virtual machine cannot be changed without a valid key protector configured.

The operation failed.

The selected security settings of virtual machine 'ubuntu-18.04-amd64' cannot be changed without a valid key protector configured. Configure a valid key protector and try again. (Virtual machine ID C3F39E68-1813-4C4A-9CB8-29BF837F5FCF)
```

It looks like in addition to enabling the TPM module, a [key protector](https://docs.microsoft.com/en-us/powershell/module/hyper-v/set-vmkeyprotector) also needs to be setup for the VM. From the docs, the key protector can be generated for the vm or one can be specified. I think Vagrant should probably support both options. Maybe, like adding another option `key_protector` as a reference to key to use. If no `key_protector` is specified, then Vagrant can go ahead and generate one.

Also, it looks like [Isolated User Mode](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/learn-more/generation-2-virtual-machine-security-settings-for-hyper-v#enable-isolated-user-mode) needs to be enabled for windows 10. It might be nice to output some warning about that so users aren't so surprised about the state of their machine.

grahhhhhhham:
@soapy1 Thanks for the review! I'll take a look at these soon

rgl:
@grahamhub did you had a chance to look into this? :-)

grahamhub:
@rgl between the winter holidays and school I totally forgot about this! thanks for reminding me, I should have some time to dig back into this in the next week or so

plugins/providers/hyperv/scripts/configure_vm.ps1
```diff
@@ -125,3 +127,17 @@ try {
     Write-ErrorMessage "Failed to ${AutoAction} automatic checkpoints on VM: ${PSItem}"
     exit 1
 }
+
+if($VM.Generation -gt 1) {
+    try {
+        if($EnableTrustedPlatformModule) {
+            Hyper-V\Enable-VMTPM -VM $VM
+            $tpmAction = "enable"
```
rgl, line 135: this should be done before calling the `Hyper-V\Enable-VMTPM` cmdlet. Or else the `catch` block will try to use an uninitialized variable.

grahamhub: thanks for catching that!

grahamhub:
@soapy1 it's in a state where I'm almost ready to push my new commit, but I did have one question regarding the Isolated User Mode: the cmdlets that enable this setting require elevation (i.e. need to be run as admin). I checked the codebase and couldn't find any current scripts that handle elevation dynamically - how should I proceed?

I have a branch in `configure_vm.ps1` that checks to see if the cmdlets need to be run, and if they do then it runs the necessary script (`enable_isolated_user_mode.ps1`, which has the `#Requires -RunAsAdministrator` flag so that the primary config script doesn't need it). I'm still a powershell novice so I'm not even sure if it's possible to elevate without the user confirming, or if there's a way to have a "custom" elevation confirmation (like having them confirm in the CLI, then run as admin vs. the usual route of right-clicking on something then running it again). If you have any ideas or if this has already been addressed somewhere else and I just missed it, I'd appreciate the guidance!

soapy1:
Heya @grahamhub, thanks for sorting this out 🎉 
It looks like it'll be quite an ordeal to do this kind of privileged escalation. Maybe correct me if I'm wrong, but the only part that needs admin access is enabling the isolated user mode. So, I think the best course of action here is for Vagrant to not try to enable that. But, rather kick back an error to the user saying that this windows mode needs to be enabled in order to use this feature. The error message can also tell the user what command they need to run to do that.

VishnuJin:
interested to know the state of this PR ? I am also looking for such a solution to deal with `The key protector could not be unwrapped` issue

PR #11979, opened 2020-10-19
Update default branch to main

soapy1:


scripts/website_push_docs.sh
```diff
@@ -34,7 +34,7 @@ git add .
 git commit -q -m "Deploy by $USER"
 
 git remote add heroku git@heroku.com:vagrantup-docs-2.git
-git push -f heroku master
+git push -f heroku main
```
soapy1, line 37: I think this should work (https://devcenter.heroku.com/changelog-items/1829) but should give it a go anyway

chrisroberts: This script is deprecated anyway and we can just remove :smiley:

chrisroberts (approved):
:tada:

scripts/website_push_docs.sh
soapy1: oooo! how about `scripts/website_push_www.sh`?

Issue #11944, opened 2020-10-08
Vagrant does not override synced folder definition as expected


Wenzel:
### Vagrant version

`Vagrant 2.2.10`

### Host operating system

Windows 10

### Guest operating system

`generic/debian10`

### Vagrantfile

```ruby
require 'rubygems'

Vagrant.configure(2) do |config|
    config.vm.box = "generic/debian10"
    config.vm.box_version = "3.0.32"
    config.vm.define "kvmi"

    # configuration
    root_dir = '/vagrant'
    enabled_vms = {
        'winxp': true,
        'win7': false,
        'ubuntu': false,
    }

    # synced_folder_hash is a convenient way to define multiple
    # local directory mappings that should be exposed to the guest
    # [local path] => [remote path]
    # if Nil, remote path is /vagrant/{basename}
    synced_folder_hash = {
        "" => nil,
    }

    # disable exposing current dir
    config.vm.synced_folder ".", "/vagrant", disabled: true

    # expose parent dir
    # default: use NFS
    config.vm.synced_folder "..", root_dir,
        :nfs => true,
        :nfs_version => 4,
        :nfs_udp => false,
        # speedup NFS with custom options
        :linux__nfs_options => ["rw", "no_subtree_check", "all_squash"]

    config.vm.provider "hyperv" do |hyperv, override|
        hyperv.cpus = 4
        hyperv.memory = 2048
        hyperv.enable_virtualization_extensions = true

        # on Windows we will use rsync
        # excluding .git directories is recommended of course, but here we need the kvm/.git directory
        # since we want to do the checkout from inside the VM
        # We cannot do an initial checkout on the Windows host because of filesystem limitations.
        override.vm.synced_folder "..", root_dir,
            :type => "rsync"
    end
end
```

### Debug output

[vagrant.log](https://gist.github.com/Wenzel/d3b4ff0032d94ce7b3621aee28b199db)

### Expected behavior

Vagrant should have overriden the default shared folder configuration for `..` from `nfs` type to `rsync` type.

### Actual behavior

Vagrant complains that NFS is not available on Windows.

### Steps to reproduce

1. vagrant up

![Sans titre](https://user-images.githubusercontent.com/964610/95479585-3c32d300-098b-11eb-8677-059de5f30ef9.png)

Thanks !

soapy1:
Looking into this a bit there are 2 workarounds

1 Use `:type` instead of `:nfs` setting up the synced folder
[...]

PR #11988, opened 2020-10-22
Replace :nfs option with :type for synced folders

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11944


chrisroberts (approved):
:+1:

Issue #11983, opened 2020-10-21
SLES11 change_host_name.rb hostnamectl: command not found


TheKrizz:
### Vagrant version

2.2.10  (since 2.2.7)

### Host operating system

Windows 10

### Guest operating system

SLES11
### Vagrantfile

```ruby
(part of vagrant-file)
    sles11.vm.hostname = "sles11-vm"
    sles11.vm.network 'private_network', ip: '192.168.0.111'
    sles11.vm.guest = 'suse'

```

### Debug output

 INFO interface: info: Setting hostname...
 INFO interface: info: ==> sles11: Setting hostname...
==> sles11: Setting hostname...
DEBUG ssh: Checking whether SSH is ready...
DEBUG ssh: Re-using SSH connection.
 INFO ssh: SSH is ready!
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute:  (sudo=false)
DEBUG ssh: stderr: 41e57d38-b4f7-4e46-9c38-13873d338b86-vagrant-ssh
DEBUG ssh: Exit status: 0
DEBUG guest: Searching for cap: change_host_name
DEBUG guest: Checking in: suse
DEBUG guest: Found cap: change_host_name in suse
 INFO guest: Execute capability: change_host_name [#<Vagrant::Machine: sles11 (VagrantPlugins::ProviderVirtualBox::Provider)>, "sles11-vm"] (suse)
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute:           grep -w 'sles11-vm' /etc/hosts || {
            for i in {1..5}; do
grep -w "127.0.${i}.1" /etc/hosts || {
  echo "127.0.${i}.1 sles11-vm.local.vm sles11-vm" >> /etc/hosts
  break
}
            done
          }
 (sudo=true)
DEBUG ssh: stderr: 41e57d38-b4f7-4e46-9c38-13873d338b86-vagrant-ssh
DEBUG ssh: Exit status: 0
DEBUG ssh: Re-using SSH connection.
 INFO ssh: Execute: test "$(hostnamectl --static status)" = "sles11-vm" (sudo=false)
DEBUG ssh: stderr: 41e57d38-b4f7-4e46-9c38-13873d338b86-vagrant-ssh
DEBUG ssh: stderr: bash: line 4: hostnamectl: command not found

DEBUG ssh: Exit status: 1
DEBUG ssh: Re-using SSH connection.
[...]

PR #11996, opened 2020-10-26
Don't use hostnamectl if not on system

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/11983

### DIFF
--- a/plugins/guests/suse/cap/change_host_name.rb
+++ b/plugins/guests/suse/cap/change_host_name.rb
@@ -1,22 +1,43 @@
-require_relative '../../linux/cap/change_host_name'
+require 'vagrant/util/guest_hosts'
+require 'vagrant/util/guest_inspection'
 
 module VagrantPlugins
   module GuestSUSE
     module Cap
       class ChangeHostName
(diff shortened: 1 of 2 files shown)


plugins/guests/suse/cap/change_host_name.rb
```diff
@@ -1,22 +1,43 @@
...
+          if hostnamectl?(comm)
+            if !comm.test("test \"$(hostnamectl --static status)\" = \"#{basename}\"", sudo: false)
+              cmd = <<-EOH.gsub(/^ {14}/, "")
+              hostnamectl set-hostname '#{basename}'
+              echo #{name} > /etc/HOSTNAME
+              EOH
+              comm.sudo(cmd)
+            end
+          else
+           if !comm.test("hostname -f | grep '^#{name}$'", sudo: false)
```
chrisroberts, line 32: The indentation here seems to be a bit off (only single character instead of two)

chrisroberts (approved):
Just a formatting thing to be addressed. Looks good otherwise so can merge after adjusting indentation.

PR #12003, opened 2020-10-29
Enhance synopsis of up command

jlusiardi:
Just to make clear, this can also handle single machines and not just
complete environments.


soapy1:
Thank you for trying to make the Vagrant docs better! We appreciate the time you took to submit this. At this time we don't think this change improves the clarity of the message so I'm going to go ahead and close this PR.

PR #12011, opened 2020-10-31
Make the max reboot retry duration configurable

### LINKED ISSUE #11695: Enhancement Request: Make windows guest reboot timeout configurable
When provisioning multiple Windows machines in parallel on a VMware provider, many times I've experienced that after host renaming, the [default value of 2 minutes for reboot](https://github.com/hashicorp/vagrant/blob/aef97fa061240fc47304fe35d5283fc01b91b2d5/plugins/guests/windows/cap/reboot.rb#L7 ) is not enough.

Make the variable `MAX_REBOOT_RETRY_DURATION` to be configurable, see:
* https://github.com/hashicorp/vagrant/blob/aef97fa061240fc47304fe35d5283fc01b91b2d5/plugins/guests/windows/cap/reboot.rb#L7 

Not really an alternative, but using the Ruby ability to "Monkey Patch", the `MAX_REBOOT_RETRY_DURATION` value can be changed and workaround the timeout issue.

amaltson:
Previously the maximum amount of time Vagrant would poll for whether a VM has successfully reboot was hard coded to 120 seconds.

This change introduces the `VAGRANT_MAX_REBOOT_RETRY_TIMEOUT` environment variable to allow users to configure the maximum time to support OS configurations that happen to take longer than 120 seconds to reboot.

This PR also adds RSpec tests for the maximum retry logic. I wanted to add some tests around the actual environment variable, but since it's set to a constant on the class it'll require unloading and loading the class back up which was fairly ugly to do in the test. You can see below the environment variable in action (and failing the test because the number of retries don't match).

Screen Shot 2020-10-31 at 6 18 56 PM

Hopefully this is useful, thank you!

Fixes #11695

### DIFF
--- a/plugins/guests/linux/cap/reboot.rb
+++ b/plugins/guests/linux/cap/reboot.rb
@@ -6,7 +6,7 @@ module VagrantPlugins
     module Cap
       class Reboot
         extend Vagrant::Util::GuestInspection::Linux
-        MAX_REBOOT_RETRY_DURATION = 120
+        MAX_REBOOT_RETRY_DURATION = ENV.fetch('VAGRANT_MAX_REBOOT_RETRY_TIMEOUT', 120).to_i
 
         def self.reboot(machine)
           @logger = Log4r::Logger.new("vagrant::linux::reboot")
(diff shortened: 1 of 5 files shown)


chrisroberts (approved):
Thanks for the PR. I made a couple small adjustments but otherwise looks great. Thanks again!

Issue #12012, opened 2020-11-01
Enhancement Request: Support experimental features within packer vagrant builder 


fr123k:
**Is your feature request related to a problem? Please describe.**
I am try to build a vagrant box with enabled expiremental features as part of a packer build.
The vagrant up command is successful but the packer vagrant driver consider
the vagrant build as failed because the enabled feature message.
'You have enabled the experimental flag with all features enabled' was written to stderr
and not stdout. 

**Describe the solution you'd like**
This message shouldn't be send to stderr because it's more like a warning or notice.

**Describe alternatives you've considered**
Make it configurable to receive this message on stderr at all or to just silent this particular message
by setting a environment variable like VAGRANT_EXPERIMENTAL_SILENT.

**Additional context**

Use the following code to reproduce the described problem.
vagrant.pkr.hcl file
```
source "vagrant" "vm" {
  add_force    = true
  communicator = "ssh"
  provider     = "virtualbox"
  source_path  = "ubuntu/bionic64"
}

build {
  sources = ["source.vagrant.vm"]
}
```
The following packer build will be successful.
```
PACKER_LOG=1 packer build vagrant.pkr.hcl
```
The same packer build will fail because of the described problem from above.
```
VAGRANT_EXPERIMENTAL=1 PACKER_LOG=1 packer build vagrant.pkr.hcl
```

chrisroberts:
Hi there,

Would you please provide a gist of the output from packer of the error you are encountering.

Thanks!

fr123k:
Here is the gist with the full verbose output from a packer build.

https://gist.github.com/fr123k/ba6a362251d52843505e2db85c4d1fea

fr123k:
I guess with the fix in the packer repository this issue can be closed.
https://github.com/hashicorp/packer/pull/10213

Thanks a lot for taking care.

PR #12013, opened 2020-11-01
Use stdout for the experimental log message

fr123k:
Use stdout instead of stderr fix the failing packer vagrant builds.

This pull request implements the enhancement to support experimental
features for packer vagrant builder.
https://github.com/hashicorp/vagrant/issues/12012


fr123k:
This PR can be closed. The fix is implemented in packer.

PR #12015, opened 2020-11-03
Provide linux command to display UUID's

khelllendros:
For those not familiar with vbox* commands, adding this small note may be helpful for people working in multi-machine environments.


soapy1:
Thank you for trying to make the Vagrant docs better! We appreciate the time you took to look at our documentation and submit this. Since this is pretty out of date and needs to be rebased on main I've included this change on to https://github.com/hashicorp/vagrant/pull/13085. So, I'm going to go ahead an close this pr and the change will be included when the linked pr is merged. Thanks 🎆

PR #12023, opened 2020-11-06
Check dependency list on initialization

chrisroberts:
When initializing for internal plugin resolution inspect contraints
on all defined dependencies. If a prerelease constraint is detected,
automatically enable prerelease resolution.

Also includes option to enable prerelease resolution via environment
variable when custom resolution behavior is desired.


soapy1 (approved):
☄️

PR #12053, opened 2020-11-17
Mount vmware synced folders for big sur guests

soapy1:


plugins/guests/darwin/cap/flavor.rb
```diff
@@ -0,0 +1,24 @@
+module VagrantPlugins
+  module GuestDarwin
+    module Cap
+      class Flavor
+        def self.flavor(machine)
+          # Read the version file
+          output = ""
+          machine.communicate.sudo("sw_vers -productVersion", error_check: false) do |_, data|
```
soapy1, line 8: I'm pretty sure `sw_vers` is only a macos thing, but I couldn't find a more generic darwin command to get the flavor for the darwin guest. In this case, I figure it can be lax on error checking and an ok default return value could be `:darwin`

plugins/guests/darwin/cap/mount_vmware_shared_folder.rb
```diff
@@ -81,9 +81,15 @@ def self.mount_vmware_shared_folder(machine, name, guestpath, options)
...
+                  if @apply_firmlinks[machine.id][:bootstrap]
+                    case machine.guest.capability("flavor")
+                    when :big_sur
+                      apfs_bootstrap_flag = "-t"
+                    else
+                      apfs_bootstrap_flag = "-B"
+                    end
                     # Re-bootstrap the root container to pick up firmlink updates
-                    comm.sudo("/System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B")
+                    comm.sudo("/System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util #{apfs_bootstrap_flag}", good_exit: [0, 253])
```
soapy1, line 92: if `apfs.util -t` has returns `253` it seems to be successful?

plugins/guests/darwin/cap/mount_vmware_shared_folder.rb
```diff
@@ -81,9 +81,15 @@ def self.mount_vmware_shared_folder(machine, name, guestpath, options)
                   content = @apply_firmlinks[machine.id][:content].join("\n")
                   # Write out the synthetic file
                   comm.sudo("echo -e #{content.inspect} > /etc/synthetic.conf")
-                  if @apply_firmlinks[:bootstrap]
+                  if @apply_firmlinks[machine.id][:bootstrap]
+                    # require "pry-byebug"; binding.pry
```
chrisroberts, line 85: Remove debug code

plugins/guests/darwin/cap/mount_vmware_shared_folder.rb
```diff
@@ -81,9 +81,15 @@ def self.mount_vmware_shared_folder(machine, name, guestpath, options)
...
+                    if machine.guest.capability("darwin_major_version").to_i < 20
```
chrisroberts, line 86: Set the `20` value as a constant (like `MACOS_BIGSUR_DARWIN_VERSION` or w/e) and use it in place of the static value. It's just helpful as it provides better context on what the value means.

plugins/guests/darwin/cap/mount_vmware_shared_folder.rb
chrisroberts: I think it would be good to set the valid exit values in the conditionals as well, since with `-B` we know `0` is expected but with `-t` it could be `253` as well.

chrisroberts (approved):
Overall looks great. Just a couple small adjustment requests.

chrisroberts (approved):
:tada:

PR #12055, opened 2020-11-18
Add nfs exports troubleshoot hint

revoltek-daniel:
Add nfs exports troubleshoot hint to help people to deal with nfs problems on linux.


soapy1 (approved):
Thanks for the addition! 🎆

PR #12099, opened 2020-12-04
scripts/check_credentials.ps1: fix for domain accounts

andriytk:
Currently, the `Machine` context type for the `AccountManagement.PrincipalContext()`
instance fails to check the domain usernames specified in User Principal Name (UPN)
(user@domain) format:

```
    cmu: Username (user[@domain]): user@domain
    cmu: Password (will be hidden):
==> cmu: Credentials incorrect. Please try again.
```

Solution: determine automatically the UPN username format (by checking for '@') and
provide the `Domain` context type into `PrincipalContext()` for them.

Note: NTLM username format (domain\user) still works fine with the `Machine`
context type. Nevertheless, the API documentation [discourages ](https://docs.microsoft.com/en-us/dotnet/api/system.directoryservices.accountmanagement.principalcontext.validatecredentials?view=dotnet-plat-ext-5.0) from specifying the
domain name in any username formats into `ValidateCredentials()`.

Tested on Windows 10 (1909, build 18363.1198).


andriytk:
@soapy1, could you take a look at this patch, please?

soapy1:
So, I think this change is maybe not required anymore. Recently there was a change to check the `Domain` context type if the `Machine` context fails. https://github.com/hashicorp/vagrant/pull/12099/files#diff-bd06477b1f03fce48a72b47419172c8634d05ef148982f9f0b3f4e8efd490382L18

I think this should resolve the issue this PR is trying to solve. If you agree, then I'm ok with closing this PR. Otherwise, I'm open to suggestions.

andriytk:
@soapy1 sorry, where do you mean the change was made? If it really resolves the issue, I'm OK with closing this PR.

soapy1:
Here is the PR where the change I'm referring to were made https://github.com/hashicorp/vagrant/pull/12428. In this change set the check credentials script with try to use the domain context if the machine context fails.

andriytk:
Thanks. Agree, closing this one.

PR #12104, opened 2020-12-10
Add disks flag in cloud-init usage

ryuheechul:
As pointed out here, https://www.grzegorowski.com/how-to-test-cloud-init-locally-with-vagrant, `cloud-init` doesn't work without `disks` flag, so it's being added it.


soapy1:
fixes https://github.com/hashicorp/vagrant/issues/12207

soapy1 (approved):
thanks!

Issue #11394, opened 2020-02-14
New versioning scheme - Salt 3000


max-arnold:
SaltStack has changed its versioning scheme from date-based (e.g., 2019.2.3) to 3000, 3001 , etc for major versions and 3000.1, 3000.2, etc for minior ones (see the [SEP 14](https://github.com/saltstack/salt-enhancement-proposals/blob/master/accepted/0014-dev-overhaul.md) for details).

The powershell bootstrap script in Salt provisioner doesn't support this scheme:

https://github.com/hashicorp/vagrant/blob/master/plugins/provisioners/salt/bootstrap-salt.ps1#L13-L21

This is going to be fixed upstream:

* Issue: https://github.com/saltstack/salt-bootstrap/issues/1394
* Fix: https://github.com/saltstack/salt-bootstrap/pull/1430

Vagrant needs to update the bundled bootstrap script, or maybe just always fetch the upstream one (as it does with `bootstrap-salt.sh`)

P.S. The upstream linux bootstrap script also can't install Salt 3000 and a fix is also in the works: https://github.com/saltstack/salt-bootstrap/issues/1429

jbonhag:
Hi @max-arnold -- thanks for letting us know!

dafyddj:
> Vagrant needs to update the bundled bootstrap script, or maybe just **always fetch the upstream one** (as it does with `bootstrap-salt.sh`)

This!

s0undt3ch:
FYI, the latest stable version for Linux/macOS can be fetched from https://bootstrap.saltstack.com/
The latest stable version for windows can be fetched from https://winbootstrap.saltstack.com/

More info [here](https://github.com/saltstack/salt-bootstrap/issues/1404#issuecomment-586042121)

max-arnold:
Update: the bundled Windows Salt bootstrap script also fails on default version (2018.3.3) because it is 404 now (it was archived because of critical CVEs).

max-arnold:
Could someone review this fixed Windows minion bootstrap script for Salt provisioner in Vagrant (before I submit a PR)? To try it, place it alongside with your Vagrantfile and specify salt.bootstrap_script = "./bootstrap-salt.ps1".

```powershell
# Powershell supports only TLS 1.0 by default. Add support for TLS 1.2
[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]'Tls12'

# Download the upstream bootstrap script
[...]

PR #12127, opened 2021-01-04
Always use upstream Salt bootstrap script on Windows

max-arnold:
Get rid of outdated `bootstrap-salt.ps1` script and instead always use the upstream one (in the same way as for Linux bootstrap script). Fixes https://github.com/hashicorp/vagrant/issues/11394

The upstream script is available at https://github.com/saltstack/salt-bootstrap/blob/stable/bootstrap-salt.ps1,  https://winbootstrap.saltstack.com and https://winbootstrap.saltproject.io


dafyddj:
Can I suggest downloading `upstream_bootstrap.ps1` to a temporary directory?
Otherwise it downloads and runs from `C:\Users\<user>\Documents`.

max-arnold:
My goal was to download it to the same directory where Vagrant uploads its embedded bootstrap script (for easier debugging). Tomorrow I'll double check that.

UtahDave (dismissed):
Looks great to me. This will ensure the latest Salt version is always installed.

max-arnold:
@dafyddj I can confirm that right now the upstream script is downloaded to Documents. I'm going to update the PR so the download folder will be `C:\tmp` (the same where other Vagrant files are uploaded). Thanks for noticing this!

max-arnold:
I updated the built-in bootstrap script so it downloads the upstream one into the same directory (`C:\tmp`) without polluting the `Documents` folder.

A couple of things for reviewers to double-check:

1. Is `Tls12,Tls13` line really necessary and will it work on all PowerShell versions?
2. Is it worth replacing custom `$ScriptRoot` variable with the built-in `$PSScriptRoot` that is only available on PowerShell 3.0+?  In other words, can we assume that any Windows version supported by Salt ships with PowerShell 3.0 or greater?

UtahDave:
> 2. Is it worth replacing custom `$ScriptRoot` variable with the built-in `$PSScriptRoot` that is only available on PowerShell 3.0+?  In other words, can we assume that any Windows version supported by Salt ships with PowerShell 3.0 or greater?

Supported Operating Systems:
https://get.saltstack.com/rs/304-PHQ-615/images/SaltStack-Supported-Operating-Systems.pdf

According to this wikipedia article, Powershell 3 comes with Windows 8 and Windows Server 2012 by default:
https://en.wikipedia.org/wiki/PowerShell#Windows_PowerShell_3.0

So it looks like it might be possible that Windows Server 2008 might not have Powershell 3.0 unless the user installs it first. I haven't tested this theory, though.

plugins/provisioners/salt/bootstrap-salt.ps1
```diff
@@ -1,151 +1,11 @@
...
-$startupType = "Manual"
+# Define script root for PowerShell 2.0
+$ScriptRoot = Split-Path $script:MyInvocation.MyCommand.Path
 
-# Version to install - default to latest if there is an issue
-If ($version -notmatch "2\d{3}\.\d{1,2}\.\d+(\-\d{1})?"){
-  $version = '2018.3.3'
-}
+# Download the upstream bootstrap script
+(New-Object System.Net.WebClient).DownloadFile('https://winbootstrap.saltstack.com', "${ScriptRoot}\bootstrap_salt_upstream.ps1")
```
twangboy, line 8: Please use `winbootstrap.saltproject.io`

twangboy (changes requested):

twangboy (approved):

soapy1:
> Is it worth replacing custom $ScriptRoot variable with the built-in $PSScriptRoot that is only available on PowerShell 3.0+? In other words, can we assume that any Windows version supported by Salt ships with PowerShell 3.0 or greater?

Vagrant's minimum powershell version is 3. So, yep this is a safe assumption

max-arnold:
@soapy1 It's been a while since I submitted this PR. Is there anything I can do to get it merged?

soapy1 (approved):

soapy1:
Looks good, thanks for the gentle nudge!

leon-kleyn:
@max-arnold I am not sure if this is the right place to ask, but how can I go about implementing this? Currently using latest version of Vagrant and Vagrantfile to provision salt on a Windows box. It keeps pulling C:\tmp/bootstrap_salt.ps1 and Salt-Minion-2018.3.3-Py2-AMD64-Setup.exe, which then promptly fails.

max-arnold:
@leon-kleyn You still see the problem because this fix is not yet released. It will be available in Vagrant 2.2.17.

If you can't wait, grab the [fixed script](https://github.com/hashicorp/vagrant/blob/main/plugins/provisioners/salt/bootstrap-salt.ps1), place it alongside your Vagrantfile, and then use like this:

```ruby
  config.vm.define "minion" do |minion|
    minion.vm.provision :salt do |salt|
      salt.minion_config = "salt/etc/minion"
      salt.bootstrap_script = "./bootstrap-salt.ps1"
      salt.verbose = true
      salt.colorize = true
      salt.version = "3003"
      salt.python_version = "3"
    end
  end
```

leon-kleyn:
@max-arnold thank you, will give that a shot! On https://www.vagrantup.com/docs/provisioning/salt, it states bootstrap_script is not supported on Windows, I assume these docs are out of date?

max-arnold:
It is (and was) supported. I fixed the docs as part of this PR as well.

leon-kleyn:
@max-arnold all that worked for me, thanks again. I do get an error in the middle of bootstrapping Salt on Windows box using this script locally:

`Bootstrapping Salt... (this may take a while)

powershell.exe -NonInteractive -NoProfile -executionpolicy bypass -file C:\tmp/bootstrap_salt.ps1  -version 3003 -pythonVersion 3

Stdout from the command:

Downloading Salt minion installer Salt-Minion-3003-Py3-AMD64-Setup.exe
Installing Salt minion
Starting the Salt minion service
Salt minion successfully installed


Stderr from the command:

powershell.exe : Cannot convert value "Tls12,Tls13" to type "System.Net.SecurityProtocolType". Error: "Unable to
    + CategoryInfo          : NotSpecified: (Cannot convert ...or: "Unable to :String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
match the identifier name Tls12,Tls13 to a valid enumerator name. Specify one of the following
enumerator names and try again:
SystemDefault, Ssl3, Tls, Tls11, Tls12"
At C:\tmp\bootstrap_salt.ps1:2 char:1
+ [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.Secu ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [], RuntimeException
    + FullyQualifiedErrorId : SubstringDisambiguationEnumParseThrewAnException`
    
Vagrant and Salt appear to be functioning properly despite this error. The only thing this affects is if I am trying to 'vagrant up' multiple minions, the process stops after this error and does not move on to the next minion. I am using latest Vagrant and latest Virtualbox, using a Windows-2019 box.

max-arnold:
I only tested this line on Windows 10:

```
[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]'Tls12,Tls13'
```

Apparently, `Tls13` enum is not available in your Windows-2019 box. This constant was added in .NET Framework 4.8. If you (or other folks who are reading this) have some time, could you please try running the following command in PowerShell (as administrator) on different Windows and .NET Framework versions:

```
(New-Object System.Net.WebClient).DownloadFile('https://winbootstrap.saltstack.com', "C:\bootstrap_salt_upstream.ps1")
```

It works for me on Windows 10 just fine without the problematic `Tls12,Tls13` line. If you can confirm that `DownloadFile` works on reasonably wide range of supported WIndows versions, we can just drop this line.

https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager.securityprotocol?view=net-5.0
https://docs.microsoft.com/en-us/dotnet/api/system.net.securityprotocoltype?view=net-5.0#System_Net_SecurityProtocolType_SystemDefault
https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed

dafyddj:
@max-arnold the line `(New-Object System.Net.WebClient).DownloadFile('https://winbootstrap.saltstack.com', "C:\bootstrap_salt_upstream.ps1)` fails on Windows 8.1, which is still supported until Jan 2023.

In addition, the enumerations are limited to `Tls12` at a maximum. (With the out of the box .Net 4.5.1)
```
PS C:\Users\vagrant> [System.Net.ServicePointManager]::SecurityProtocol = 15360
Exception setting "SecurityProtocol": "Cannot convert value "15360" to type "System.Net.SecurityProtocolType" due to
enumeration values that are not valid. Specify one of the following enumeration values and try again. The possible
enumeration values are "Ssl3, Tls, Tls11, Tls12"."
```

max-arnold:
@dafyddj Thanks for testing this!

I'm going to submit another PR that will only use Tls12 enum

PR #12145, opened 2021-01-20
Use test doubles instead of `true`.

voxik:
This solves possible issues such as:

~~~
  1) VagrantPlugins::DockerProvider::Provider#host_vm returns the host machine object
     Failure/Error: allow(machine.env).to receive(:root_path).and_return("/.vagrant.d")
     ArgumentError:
       Cannot proxy frozen objects, rspec-mocks relies on proxies for method stubbing and expectations.
     # ./test/unit/plugins/providers/docker/provider_test.rb:82:in `block (3 levels) in <top (required)>'
~~~

This is not good idea, because frozen object should not receive modifications, because these possibly can't be undone. This is prohibited since rspec-mock 3.10.1.

Signed-off-by: Vít Ondruch <vondruch@redhat.com>

As always, I'm not going to sign CLA, so do whatever you have to do.


soapy1:
Thanks for the PR! Incorporating these changes in https://github.com/hashicorp/vagrant/pull/12223

PR #12147, opened 2021-01-22
Remove not required bcrypt_pbkdf gem

dcermak:
This gem is no longer required by vagrant and was forgotten to get removed.

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -15,7 +15,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = "~> 2.5", "< 2.8"
   s.required_rubygems_version = ">= 1.3.6"
 
-  s.add_dependency "bcrypt_pbkdf", "~> 1.0.0"
   s.add_dependency "childprocess", "~> 4.0.0"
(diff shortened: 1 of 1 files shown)


soapy1:
Thanks for the PR 😄  It seems this dependency was added along with `ed25519` in order support ed25519 keys.

> The net-ssh library supports ed25519 keys but only if the required
dependencies are available. This adds the required dependencies.

ref: https://github.com/hashicorp/vagrant/pull/10365

So, I think this dependency is still required.

dcermak:
Sophia Castellarin <notifications@github.com> writes:

> Thanks for the PR 😄  It seems this dependency was added along with `ed25519` in order support ed25519 keys.
>
>> The net-ssh library supports ed25519 keys but only if the required
>> dependencies are available. This adds the required dependencies.
>
> ref: https://github.com/hashicorp/vagrant/pull/10365
>
> So, I think this dependency is still required. 

Interesting, I did not catch that. Thanks for double checking!

I would propose to add a test for this though, because the test suite
happily accepted my PR.

PR #12148, opened 2021-01-22
Bump listen gem version and remove ruby_dep

dcermak:
`listen` no longer depends on `ruby_dep` and thus this pointlessly drags in another dependency.


dcermak:
@soapy1 @chrisroberts  friendly ping: this PR should be really trivial.

soapy1 (approved):

PR #12150, opened 2021-01-23
Update Vagrantfile.min.erb add vi mode for ruby

kikitux:
This PR is to add 

```
# -*- mode: ruby -*-
# vi: set ft=ruby :
``` 

to the Vagrantfile created when using `init -m`

this is very convenient to have in the file to improve the experience of advanced users that would like minimal Vagrantfile


soapy1 (approved):
💎

PR #12157, opened 2021-01-26
website: add new redirects based on recent 404s

BRKalow:
Adding some new redirects based on recent 404s.


jescalan (approved):
Looks great, thank you for knocking this out!

chrisroberts (approved):

Issue #12113, opened 2020-12-16
Vagrant 2.2.14 fails box checksums that match


martin-langhoff:
Vagrant version: 2.2.14
OS: OSX Catalina 10.15.7
VBox 6.1.16r140961

```
vagrant init aaronsachs/fedora-32-workstation
vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Box 'aaronsachs/fedora-32-workstation' could not be found. Attempting to find and install...
    default: Box Provider: virtualbox
    default: Box Version: >= 0
==> default: Loading metadata for box 'aaronsachs/fedora-32-workstation'
    default: URL: https://vagrantcloud.com/aaronsachs/fedora-32-workstation
==> default: Adding box 'aaronsachs/fedora-32-workstation' (v0.0.2) for provider: virtualbox
    default: Downloading: https://vagrantcloud.com/aaronsachs/boxes/fedora-32-workstation/versions/0.0.2/providers/virtualbox.box
Download redirected to host: vagrantcloud-files-production.s3.amazonaws.com
    default: Calculating and comparing box checksum...
The checksum of the downloaded box did not match the expected
value. Please verify that you have the proper URL setup and that
you're downloading the proper file.

Expected: 3fc983e937e7d7a95fae90f20d7ff48a8c1e09dd02f6da7c4aadaded9fea10f1699f1fbe3e32e039f41d113fda34173a3f126e4d39a66a8c5ac77afe0d747e78 
Received: 3fc983e937e7d7a95fae90f20d7ff48a8c1e09dd02f6da7c4aadaded9fea10f1699f1fbe3e32e039f41d113fda34173a3f126e4d39a66a8c5ac77afe0d747e78

```

Observe that in the last 2 lines, the checksums _match_.

aprilFlowers:
How to solve the same problem，please

Rockin2:
There is a space after the expected checksum. Maybe the problem is there?

cjmartian:
@soapy1 I'd like to give this a shot if you don't mind

PR #12168, opened 2021-02-02
remove all space from checksums

cjmartian:
This is related to #12113 
I'm not sure if we would prefer to use `.strip()` to remove only leading and trailing space instead of removing all of it the way I do in this PR.


chrisroberts:
Hi @cjmartian and thanks for this PR! I think that using the `#strip` method would be best here (even though checksums shouldn't include space characters anywhere). We only want to ensure any leading/trailing space characters are removed. Adding a test case for this would be appreciated as well. 

Please note that before we can merge we'll need the CLA signed too. Thanks so much!

chrisroberts (changes requested):
As noted in comment, `#strip` usage and test coverage :slightly_smiling_face:

cjmartian:
Hey @chrisroberts, 

Thank you for the feedback, this is my first time writing ruby so the feedback is more than welcome! I've made changes to use strip() instead. I'm writing a test for this now, I'd like to get feedback on that once I push.

Thank you again for looking at this.

soapy1 (approved):
Thanks!

PR #12169, opened 2021-02-02
update sudo usage example

sassman:
the example was not more working, I suspect the option `rsync__rsync_path` got introduced at a later point and the example was just not updated.


sassman:
@soapy1 is there anything missing in order to land this PR that I can help with?

PR #12173, opened 2021-02-04
website: update next and nextjs-scripts

BRKalow:
Upgrading next to v10 and nextjs-scripts to v15.

### DIFF
--- a/website/next.config.js
+++ b/website/next.config.js
@@ -5,7 +5,7 @@ console.log(`Environment: ${process.env.HASHI_ENV}`)
 
 module.exports = withHashicorp({
   defaultLayout: true,
-  transpileModules: ['is-absolute-url'],
+  transpileModules: ['is-absolute-url', '@hashicorp/react-image'],
 })({
   svgo: { plugins: [{ removeViewBox: false }] },
   redirects: () => redirects,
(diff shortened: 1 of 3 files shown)


website/package.json
```diff
@@ -16,8 +16,8 @@
     "@hashicorp/react-section-header": "3.0.1",
     "@hashicorp/react-subnav": "7.1.0",
     "@hashicorp/react-vertical-text-block-list": "3.0.1",
-    "next": "9.5.5",
-    "next-mdx-remote": "^1.0.1",
+    "next": "10.0.6",
+    "next-mdx-remote": "^2.1.3",
```
BRKalow, line 20:
```suggestion
    "next-mdx-remote": "2.1.3",
```

website/next.config.js
```diff
@@ -5,7 +5,7 @@ console.log(`Environment: ${process.env.HASHI_ENV}`)
 
 module.exports = withHashicorp({
   defaultLayout: true,
-  transpileModules: ['is-absolute-url'],
+  transpileModules: ['is-absolute-url', '@hashicorp/react-image'],
```
jescalan, line 8: Confused about this one, don't we transpile all the global components by default now?

jescalan (approved):

website/next.config.js
BRKalow: We should, but this dependency was nested so it didn't get picked up by my solution.  I'll add a ticket to handle this case within scripts.

BRKalow: Ticket: https://app.asana.com/0/1100423001970639/1199916284212909/f

Issue #12141, opened 2021-01-14
`vagrant destroy -f` does not return exit code 0 even though no errors occured


vilhelmprytz:
### Vagrant version

```bash
$ vagrant -v
Vagrant 2.2.14
```

### Host operating system

Debian 10 and tested on Ubuntu 20.04.

### Guest operating system

Not related to the issue, but in my example, I use Ubuntu 20.04 and Ubuntu 18.04 as example boxes. I've tried with many other operating systems so it's not related.

### Vagrantfile

```ruby
Vagrant.configure("2") do |config|
  config.vm.define "ubuntu_focal" do |ubuntu_focal|
    ubuntu_focal.vm.box = "ubuntu/focal64"
  end

  config.vm.define "ubuntu_bionic" do |ubuntu_bionic|
    ubuntu_bionic.vm.box = "ubuntu/bionic64"
  end
end
```

### Debug output

I doubt that the debug information would help since I don't think this is at all related. If you believe that is not the case, let me know and I'll attach the debug output ASAP.

### Expected behavior

I'd expect the command to return exit code 0 since nothing went wrong.

### Actual behavior

`vagrant destroy -f` returned exit code 2.

### Steps to reproduce

So using the Vagrantfile listed above, I first only boot one of the VMs.

```bash
vagrant up ubuntu_focal
```

When I then want to destroy all VMs, the `vagrant destroy` will return exit code 2.

```bash
$ vagrant destroy -f
==> ubuntu_bionic: VM not created. Moving on...
==> ubuntu_focal: Forcing shutdown of VM...
==> ubuntu_focal: Destroying VM and associated drives...
$ echo $?
2
```

PR #12186, opened 2021-02-11
Don't count not created machines as declined when destroying

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/12141

### DIFF
--- a/plugins/commands/destroy/command.rb
+++ b/plugins/commands/destroy/command.rb
@@ -56,17 +56,13 @@ module VagrantPlugins
         end
 
         machines.each do |m|
-          if m.state.id == init_states[m.name]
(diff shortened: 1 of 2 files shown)


tasks/test.rake
```diff
@@ -3,7 +3,7 @@ require 'rspec/core/rake_task'
 
 namespace :test do
   RSpec::Core::RakeTask.new(:unit) do |t|
-    t.pattern = "test/unit/**/*_test.rb"
+    t.pattern = "test/unit/plugins/commands/destroy/command_test.rb"
```
chrisroberts, line 6: This change should be reverted back to run the entire suite of tests

chrisroberts (changes requested):
Just needs the rake config modification reverted

chrisroberts (approved):

PR #12216, opened 2021-03-01
Bump bcrypt_pbkdf version

dcermak:
There were no functional changes between 1.0.0 and 1.1.0:
https://github.com/net-ssh/bcrypt_pbkdf-ruby/compare/b5d1a921fcec8daa368236f5b4c469e08200fd5c...v1.1.0
so this update should be safe


soapy1 (approved):
🎉

PR #12223, opened 2021-03-08
Use test doubles instead of 'true'

soapy1:


chrisroberts (changes requested):
For the doubles being returned, we should include a name so if it is used in an unexpected way it's easier to determine the source. For example:

```ruby
allow(machine).to receive(:guest).and_return(double("guest"))
```

chrisroberts (approved):
thanks!

PR #12232, opened 2021-03-13
fix the hyperv guest ip detection by limiting the search to the vagra…

### LINKED ISSUE #12231: mismatched guest ip address reported for hyperv vm with multiple network interfaces
There's a mismatch between the guest IP reported by the hyperv provider at:

https://github.com/hashicorp/vagrant/blob/e963774cebedab92c09dfcc687f00f27a67e6a8d/plugins/providers/hyperv/scripts/get_network_config.ps1#L18

and what WinRM address reports (192.168.56.10 vs 10.0.0.126):

```
==> wds: Starting the machine...
==> wds: Waiting for the machine to report its IP address...
    wds: Timeout: 120 seconds
    wds: IP: 192.168.56.10
==> wds: Waiting for machine to boot. This may take a few minutes...
    wds: WinRM address: 10.0.0.126:5985
    wds: WinRM username: vagrant
    wds: WinRM execution_time_limit: PT2H
    wds: WinRM transport: negotiate
Timed out while waiting for the machine to boot. This means that
Vagrant was unable to communicate with the guest machine within
the configured ("config.vm.boot_timeout" value) time period.
...
```

This makes vagrant fail to connect to the proper guest IP address.

This is happening because the referenced hyper-v provider powershell code is not limiting itself to the vagrant management interface (the first network interface). And in my particular case this fails my vagrant up because the secondary network interface is assigned an IP address faster than the first one, and vagrant wrongly chooses it, and hence the problem.

rgl:
…nt management network interface (the first network interface of the vm)

this fixes #12231.

### DIFF
--- a/plugins/providers/hyperv/scripts/get_network_config.ps1
+++ b/plugins/providers/hyperv/scripts/get_network_config.ps1
@@ -15,7 +15,7 @@ try {
 }
 
 try {
-    $networks = Hyper-V\Get-VMNetworkAdapter -VM $vm
+    $networks = Hyper-V\Get-VMNetworkAdapter -VM $vm | Select-Object -First 1
     foreach ($network in $networks) {
         if ($network.IpAddresses.Length -gt 0) {
             foreach ($ip_address in $network.IpAddresses) {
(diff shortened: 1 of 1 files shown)


chrisroberts (approved):
Thanks!

PR #12235, opened 2021-03-15
Prevent notification on default store, fix client end point

chrisroberts:
Prevents redirect notification when getting asset from
the default store. Also fixes some places where the
cloud client was not using the helper method for the
server url. Custom server urls also have api path appended
automatically when not detected.


soapy1 (dismissed):
🎉

Issue #12277, opened 2021-04-05
SMB mount broken since 2.2.15


McBochi:
### Vagrant version

Vagrant 2.2.15

### Host operating system

Windows 10 - Version 20H2 (Build 19042.906)

### Guest operating system

hashicorp/bionic64

### Vagrantfile

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "hashicorp/bionic64"
  config.vm.box_check_update = false

  config.vm.provider "hyperv" do |h|
    h.vmname = "myapplication"
    h.cpus = 1
    h.memory = 1536
    h.enable_checkpoints = false
  end

  config.vm.synced_folder ".", "/vagrant", type: "smb", owner: "www-data", group: "www-data"

  config.vm.provision "shell", inline: <<-SHELL
    # Some installation / setup stuff
  SHELL

end
```

Pretty basic Vagrantfile, which is wasn't changed lately and ran without issues in Vagrant 2.2.14.

### Debug output

https://gist.github.com/McBochi/06cfe6aaf2a7b0d17a4c4486f0a6633c

### Expected behavior

After the prompt for username and password for the SMB share, the folder should be mounted and usable in the guest system.

### Actual behavior

Starting with Vagrant 2.2.15, it fails when trying to mount the share with:
```
C:/HashiCorp/Vagrant/embedded/gems/2.2.15/gems/vagrant-2.2.15/plugins/synced_folders/smb/cap/mount_options.rb:48:in 'mount_name': wrong number of arguments (given 2, expected 3) (ArgumentError)
```

### Steps to reproduce

1. Take the above Vagrantfile.
2. Call `vagrant up`
3. Provide credentials as usual.

### References

It might be related to the latest changes for vbox shared folders:
https://github.com/hashicorp/vagrant/commit/b30a87ce6cb7e5a4398f236baa5574b2d5e6143f#diff-befa3878a0f7ad26bb3669d9239b56d17d90bf3b99a9f3dbd07b57c929c0876f

lexxxel:
I do have the same issue.

McBochi:
Just applied the diff and it works like a charm. Thanks for the quick and easy fix, @soapy1 ! :)

ElijahTailor:
The bug is present @ hyper-v + generic/ubuntu2004 box environment also. Same situation, same error.

ALERTua:
please fix this issue. it affects hundreds of people in my community. thank you.

can-acar:
[...]

PR #12281, opened 2021-04-06
Get folder mount name

soapy1:
fixes https://github.com/hashicorp/vagrant/issues/12277

additional tests in the vagrant-spec project https://github.com/hashicorp/vagrant-spec/blob/main/acceptance/synced_folder/smb_spec.rb

### DIFF
--- a/plugins/guests/linux/cap/mount_smb_shared_folder.rb
+++ b/plugins/guests/linux/cap/mount_smb_shared_folder.rb
@@ -20,7 +20,7 @@ module VagrantPlugins
             :shell_expand_guest_path, guestpath)
           options[:smb_id] ||= name
 
-          mount_device = options[:plugin].capability(:mount_name, options)
+          mount_device = options[:plugin].capability(:mount_name, name, options)
           mount_options, _, _ = options[:plugin].capability(
             :mount_options, name, expanded_guest_path, options)
           mount_type = options[:plugin].capability(:mount_type)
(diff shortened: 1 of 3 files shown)


chrisroberts (changes requested):
We should also update the test for this by updating the allow mock to only match on the correct number of passed arguments.

DanSibbernsen:
Hi! I'm relatively new to the Vagrant world. Would this fix also need to be incorporated into the Vagrant-Parallels provider's use of mount? Example [here](https://github.com/Parallels/vagrant-parallels/blob/8ac0b2bae6428bfb485f25b59f224444fe317014/lib/vagrant-parallels/cap/mount_options.rb#L33)?

If so, I can open a change request over there (I'm stuck on this issue too)

soapy1:
@DanSibbernsen yep, I think you are right. That `mount_name` bit should look something like [this smb capability](https://github.com/hashicorp/vagrant/blob/main/plugins/synced_folders/smb/cap/mount_options.rb#L48). You can double check that this change is required by running vagrant 2.2.15 with the parallels plugin.

chrisroberts (approved):

legal90:
Thank you, @soapy1.  
It's a bit unfortunate that the positional arguments passed to `mount_name` capability were changed in Vagrant 2.2.15 (by https://github.com/hashicorp/vagrant/pull/12281). That was a breaking change for 3rd party plugins like `vagrant-parallels`, implementing this capability on their side. :(

We have to use some smelly hacks in order to work it around and keep the backward compatibility: https://github.com/Parallels/vagrant-parallels/pull/386

Issue #12270, opened 2021-04-04
ssh for shell provisioner expecting reboot happen too fast


avoidik:
### Vagrant version

Run `vagrant -v` to show the version. If you are not running the latest version
of Vagrant, please upgrade before submitting an issue.

```
Vagrant 2.2.14
```

### Host operating system

This is the operating system that you run locally.

```
Microsoft Windows [Version 10.0.19042.868]
```

### Guest operating system

This is the operating system you run in the virtual machine.

```
bento/ubuntu-20.04
```

### Vagrantfile

```ruby
$kernel_config = <<-SCRIPT
sed -ie 's/GRUB_CMDLINE_LINUX_DEFAULT=\"\([^"]*\)\"/GRUB_CMDLINE_LINUX_DEFAULT=\"\1 cgroup_enable=cpuset cgroup_enable=memory cgroup_memory=1 swapaccount=1 ipv6.disable=1\"/g' /etc/default/grub
update-grub
SCRIPT

Vagrant.configure('2') do |config|
  config.vm.define 'test-vm' do |nodeconfig|
    nodeconfig.vm.box      = 'bento/ubuntu-20.04'
    nodeconfig.vm.hostname = 'test-vm'
    nodeconfig.vm.provider 'virtualbox' do |vb|
      vb.name       = 'test-vm'
      vb.memory     = 1024
      vb.cpus       = 1
      vb.customize ['modifyvm', :id, '--audio', 'none']
    end
    nodeconfig.vm.provision 'shell', inline: $kernel_config, reboot: true
    nodeconfig.vm.network 'private_network', ip: '192.168.50.10'
    nodeconfig.vm.synced_folder '.', '/vagrant', disabled: true
  end
end
```

### Debug output

<details>
  <summary>Click to expand</summary>

```
2021-04-04 16:40:58 [DEBUG] Uploading file C:/temp/vagrant-shell20210404-2460-fmie16 to remote /tmp/vagrant-shell
2021-04-04 16:40:58 [ INFO] detail: Running: inline script
2021-04-04 16:40:58 [ INFO] detail:     test-vm: Running: inline script
    test-vm: Running: inline script
2021-04-04 16:40:58 [DEBUG] Re-using SSH connection.
2021-04-04 16:40:58 [ INFO] Execute: chmod +x '/tmp/vagrant-shell' && /tmp/vagrant-shell (sudo=true)
[...]

Issue #10776, opened 2019-04-04
add sleep period before retrying the ssh connection


rgl:
while using vagrant 2.2.4 I'm trying to reboot an ubuntu vm with:

```ruby
config.vm.provision :shell, path: 'reboot.sh'
```

with `reboot.sh` being:

```bash
nohup bash -c "ps -eo pid,comm | awk '/sshd/{print \$1}' | xargs kill; sync; reboot"
```

while the vm is rebooting the ssh communicator keeps (re)trying to connect to ssh, but that will fail because the connection is refused by the vm while its booting / not yet ready... and the communicator gives up too quickly.

while following the code, this ended up being because the retry logic of the ssh communicator at https://github.com/hashicorp/vagrant/blob/b1d8b952bb4da7e18782f6e3422cfe5e99014690/plugins/communicators/ssh/communicator.rb#L431 is not doing sleeping a bit between retries. this need to be changed to add the `sleep` argument, e.g.:

```ruby
connection = retryable(tries: opts[:retries], on: SSH_RETRY_EXCEPTIONS, sleep: timeout) do
```

maybe that `timeout` should trickle down from the Vagrantfile provision line (like `opts[:retries]`), e.g.: with the `sleep`argument:

```ruby
config.vm.provision :shell, path: 'reboot.sh', sleep: 120
```

nmaludy:
I ran into this problem as well. In my case my guests are RHEL VMs where they are pingable before the SSH daemon starts up. The behavior i see with `vagrant up --debug` is:
- The guest becomes pingable
- An SSH connection is initiated
- SSH returns CONNECTIONREFUSED because the SSH daemon on the guest isn't up yet (taking a little while to boot)
- Vagrant does NOT retry and simply exits with an error

I found the same section of code that @rgl did and manually set the `opts[:retries]` to 5 (seems to be set to 1 when the function is called) and then added in a `sleep`. This allowed the SSH connection to be retried and communication to the guest works great.

I'm thinking a good solution would be to expose something like:
```
config.ssh.retries = 5
config.ssh.retry_sleep_interval = 10
```
These options would allow the user to control the number SSH retries and the sleep time between retries.

Thoughts?

avoidik:
hello all, the very same issue I had reported [here](https://github.com/hashicorp/vagrant/issues/12270) is still relevant, if someone is also affected here is a workaround

```bash
cd vagrant/embedded/gems/*/gems/vagrant-*/plugins/communicators/ssh
[...]

PR #12292, opened 2021-04-10
SSH communicator - Reset retries counter, sleep between connection attempts

avoidik:
Relevant issue - #12270

Rework of https://github.com/hashicorp/vagrant/pull/12272

### DIFF
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -40,7 +40,8 @@ module VagrantPlugins
         Errno::EHOSTUNREACH,
         Errno::EPIPE,
         Net::SSH::Disconnect,
-        Timeout::Error
+        Timeout::Error,
+        Net::SSH::Exception
       ]
 
       include Vagrant::Util::ANSIEscapeCodeRemover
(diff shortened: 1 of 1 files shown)


soapy1:
Hi there,

Cheers!

avoidik:
Hello over there,

There's still no gap between connection attempts, and SSH exceptions aren't considered as retryable in the latest version, but of course it's up to you to decide is this a bug or feature.

Cheers.

soapy1:
Hey there,

>There's still no gap between connection attempts

Yup, that issue is open at https://github.com/hashicorp/vagrant/issues/10776. I think it would be nice if the solution had user configurable retry delay, similar to the [winrm plugin](https://developer.hashicorp.com/vagrant/docs/vagrantfile/winrm_settings#config-winrm-retry_delay). This is addressed in an existing PR https://github.com/hashicorp/vagrant/pull/13047. 

Thanks for the contribution!

avoidik:
Hey @soapy1 

Worth noting that in the current implementation there is no retries at all due to this line, which effectively limits retries to 1.

https://github.com/hashicorp/vagrant/blob/main/plugins/communicators/ssh/communicator.rb#L87

Could you please finally decide in which PR you would like to see this fix to be applied without removing my comments? Thank you.

soapy1:
> Worth noting that in the current implementation there is no retries at all due to this line, which effectively limits retries to 1.

yup, I see you also noted this as a [comment](https://github.com/hashicorp/vagrant/pull/13047) on the above mentioned PR. So, I think it's appropriate to resolve it in PR#13047. Thanks for your contribution!

PR #12304, opened 2021-04-14
implement --info flag for package cmd #12191

### LINKED ISSUE #12191: The info.json is included in the wrong directory
The `info.json` file is included in the `include` directory not in the root of the package. So the `box list -i` does not work.

Either the `package --include` is the wrong option (missing documentation) or there is a mismatch between the include path and the path for the `box list -i` (bug).

### Vagrant version

2.2.14

### Host operating system

Windows 10

### Guest operating system

Ubuntu

### Vagrantfile

```ruby
  config.vm.box = "hashicorp/bionic64"
end
```

### Expected behavior

The `box list -i` should show the information from the `info.json` file

### Actual behavior

No additional information in the output.

### Steps to reproduce

1. Create an `info.json` file as per https://www.vagrantup.com/docs/boxes/info page.
2. Included the file using the `vagrant package --include info.json` command.
3. The file is stored in the `include/info.json` path in the package.
4. The `vagrant box list -i` does not show the additional info.
5. Copy the `~/.vagrant.d/boxes/test-box/0/virtualbox/include/info.json` to the `~/.vagrant.d/boxes/test-box/0/virtualbox/info.json` location then the `vagrant box list -i` shows the desired output.

grahamhub:
Assuming that `info.json` belongs in the `include` directory, this fixes #12191 by pre-pending `include` as the first arg when joining the pathname.

### DIFF
--- a/lib/vagrant/action/general/package.rb
+++ b/lib/vagrant/action/general/package.rb
@@ -13,6 +13,7 @@ module Vagrant
       #
       #   * package.output - The filename of the outputted package.
       #   * package.include - An array of files to include in the package.
+      #   * package.info - Path of desired info.json file to include
       #   * package.directory - The directory which contains the contents to
       #       compress into the package.
       #
@@ -63,6 +64,7 @@ module Vagrant
           @app = app
 
           env["package.files"]  ||= {}
+          env["package.info"]   ||= ""
           env["package.output"] ||= "package.box"
 
           @fullpath = self.class.fullpath(env["package.output"])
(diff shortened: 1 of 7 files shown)


grahamhub:
@chrisroberts Since you've merged the versioning fix, is there a way to re-run these checks without pushing another commit?

chrisroberts:
Hi @grahamhub. I can re-run them but the modification in this PR won't properly fix the underlying problem of the linked issue. The location of the info.json file in the box contents which the list command is using is correct. It is the `package` command that is including the files that needs to be fixed. Include files defined get placed into an `./include` directory within the box (so the files showing up in that directory isn't a problem). What most likely needs to be done is introducing a new flag to the package command for using a custom info.json file, and updating the package action to support it.

Cheers!

grahamhub:
@chrisroberts Alright, makes sense! Didn't want to implement a new flag if that wasn't the design decision, but will do 👍🏻

grahamhub:
@chrisroberts alrighty, I think I updated all the files that I needed to for a command flag. If I missed something let me know!

Currently it accepts a relative path pointing to an `info.json` file and raises an error if it doesn't exist/isn't explicitly titled `info.json`, then copies it to the top directory instead of `include/`.

chrisroberts (approved):
Thank you!

Issue #12349, opened 2021-05-05
This is just regarding installation with brew. The instruction on your website no longer works. On Brew website there is a new command for this. brew install --cask vagrant


cemceliks:
Please note that the Vagrant issue tracker is in priority reserved for bug reports and enhancements. For general usage questions, please use
HashiCorp Discuss: https://discuss.hashicorp.com/c/vagrant/24 Thank you!

When submitting a bug report, please provide the minimal configuration and required information necessary to reliably reproduce the issue. It
should include a basic Vagrantfile that only contains settings to reproduce the described behavior.

**Tip:** Before submitting your issue, don't hesitate to remove the above introductory text, possible empty sections (e.g. References), and this tip.

### Vagrant version

Run `vagrant -v` to show the version. If you are not running the latest version
of Vagrant, please upgrade before submitting an issue.

### Host operating system

This is the operating system that you run locally.

### Guest operating system

This is the operating system you run in the virtual machine.

### Vagrantfile

```ruby
# Copy-paste your Vagrantfile here (but don't include sensitive information such as passwords, authentication tokens, or email addresses)
```

Please ensure the Vagrantfile provided is a minimal Vagrantfile which contains
only the required configuration to reproduce the behavior. Please note that if
your Vagrantfile contains an excess of configuration unrelated the the reported
issue, or is in a different format, we may be unable to assist with your issue.
Always start with a minimal Vagrantfile and include only the relevant configuration
to reproduce the reported behavior.

### Debug output

Provide a link to a GitHub Gist containing the complete debug output:
https://www.vagrantup.com/docs/other/debugging.html. The debug output should
be very long. Do NOT paste the debug output in the issue, just paste the
link to the Gist.

### Expected behavior

What should have happened?

### Actual behavior

What actually happened?

### Steps to reproduce

1.
2.
3.

### References

Are there any other GitHub issues (open or closed) that should be linked here?
For example:
- GH-1234
- ...

Issue #12354, opened 2021-05-08
Homebrew installation instructions do not work


carlosefr:
The instructions at https://www.vagrantup.com/downloads to install vagrant on macOS via homebrew do not work:
```
brew tap hashicorp/tap
brew install hashicorp/tap/vagrant
```

The tap is correctly registered with brew, but the second command errors out:
```
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
Error: No available formula or cask with the name "hashicorp/tap/vagrant".
==> Searching for a previously deleted formula (in the last month)...
Error: No previously deleted formula found.
```

### Host operating system

macOS 10.15.7 (Catalina)

lkysow:
From https://github.com/hashicorp/homebrew-tap#what-packages-are-available

> Note: Vagrant is available as a cask from Homebrew core and not provided by this tap.

So you should use `brew install vagrant`. But also the downloads page needs to be fixed.

chrisroberts:
Closing as fixed via #12358

PR #12358, opened 2021-05-10
Add package manager override for brew install instructions

soapy1:
ref:
https://github.com/hashicorp/homebrew-tap#what-packages-are-available

fixes: https://github.com/hashicorp/vagrant/issues/12349 and https://github.com/hashicorp/vagrant/issues/12354

### DIFF
--- a/website/pages/downloads/index.jsx
+++ b/website/pages/downloads/index.jsx
@@ -46,6 +46,24 @@ export default function DownloadsPage(staticProps) {
           <a>&raquo; Download VMware Utility</a>
         </Link>
       }
+      packageManagerOverrides={[
+        {
+          label: 'Homebrew',
+          commands: [
+            `brew tap hashicorp/tap`,
+            `brew install ${productSlug}`,
+          ],
+          os: 'darwin',
+        },
+        {
+          label: 'Homebrew',
+          commands: [
+              `brew tap hashicorp/tap`,
+              `brew install ${productSlug}`,
+          ],
+          os: 'linux',
+      },
+      ]}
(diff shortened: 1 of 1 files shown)


zchsh:
Looks good to me, thanks for fixing this 🙇  

I'm not very familiar with Homebrew - would `brew install {product}` be a preferable default for all products? If so, maybe this is something we should change in the component, ie [here](https://github.com/hashicorp/react-components/blob/d5f6c48a8893a0ddcf145738bb089b14373e1eae/packages/product-download-page/package-managers.ts#L11) and [here](https://github.com/hashicorp/react-components/blob/d5f6c48a8893a0ddcf145738bb089b14373e1eae/packages/product-download-page/package-managers.ts#L55). CC @jescalan 

Also want to flag that the Linux / Homebrew tab is still showing `brew install hashicorp/tap/vagrant`:

CleanShot 2021-05-10 at 13 08 22@2x

If that needs to be changed as well, you could do something like:

```
packageManagerOverrides={[
    {
        label: 'Homebrew',
        commands: [
            `brew tap hashicorp/tap`,
            `brew install ${productSlug}`,
        ],
        os: 'darwin',
    },
    {
        label: 'Homebrew',
        commands: [
            `brew tap hashicorp/tap`,
            `brew install ${productSlug}`,
        ],
        os: 'linux',
    },
]}
```

zchsh (approved):

soapy1:
Ahh! good catch will need to update the Linux homebrew too.

chrisroberts (approved):

PR #12361, opened 2021-05-12
fix(website/ansible_intro): Update deprecated Ansible variable names

fhemberger:
`ansible_ssh_host`, `ansible_ssh_port` and `ansible_ssh_user` are no longer used in Ansible and are called `ansible_host`, `ansible_port` and `ansible_user` instead.


soapy1 (approved):
thanks!

PR #12374, opened 2021-05-17
Adds AlertBanner to promote HashiConf EU

BrandonRomano:
[🔍 Preview](https://vagrant-qwqrcc8gb-hashicorp.vercel.app/)

Adding an AlertBanner component to promote HashiConf Europe.

This includes adding the actual AlertBanner component, similar to the way we have this set up for the other .io sites.

The significant diff size the package.lock file here is due to a differing in versions in NPM.  With our previous brand refresh updates we upgraded to npm v7.  We've since discovered issues, and this is rolling back the package.lock file to be driven via npm v6.


jmfury (approved):

BrandonRomano:
Just cherry-picked this onto stable-website.

PR #12377, opened 2021-05-19
Hyperv: add --force flag support for stop VM

megrail:
It's needed to use  -TurnOff parameter if we want to stop VM immediately  according with https://docs.microsoft.com/en-us/powershell/module/hyper-v/stop-vm

### DIFF
--- a/test/unit/plugins/providers/hyperv/action/halt_instance_test.rb
+++ b/test/unit/plugins/providers/hyperv/action/halt_instance_test.rb
@@ -1,0 +1,29 @@
+require_relative "../../../../base"
+
+require Vagrant.source_root.join("plugins/providers/hyperv/action/halt_instance")
+
+describe VagrantPlugins::HyperV::Action::HaltInstance do
+  let(:app){ double("app") }
+  let(:env){ {ui: ui, machine: machine} }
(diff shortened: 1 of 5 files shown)


chrisroberts (changes requested):
Changes look good. Needs some test coverage added before merging.

chrisroberts:
@megrail are you able to sign the CLA to allow merging?

megrail:
Yes I signed the CLA

megrail:
@chrisroberts Do I need to add tests to commit?

chrisroberts:
Adding test coverage would be very helpful

megrail:
@chrisroberts Some tests were added. Would it be enough?

megrail:
@chrisroberts could you review?

megrail:
@chrisroberts would it possible to review this pull request?

test/unit/plugins/providers/hyperv/action/halt_instance_test.rb
```diff
@@ -0,0 +1,29 @@
+require_relative "../../../../base"
```
megrail, line 1: add tests

PR #12388, opened 2021-05-24
Update "Default Router" shell script in publc_network.mdx documentation

JonTheNiceGuy:
In more recent linux versions, if you receive a DHCP address from your `public_network` interface, it will "simply" add a new gateway to your list of possible routes. You can remove this using the `ip route del` command, as seen in this PR.


chrisroberts (approved):
thanks

PR #12391, opened 2021-05-25
detects new versions of ansible-4.0.0+

yodaaut:
Ansible has switched dependency between version 3.4.0 (ansible-base) and 4.0.0 (ansible-core), which now outputs new version scheme of "ansible [core 2.11.0]" instead of "ansible 2.10.9". This fix keeps old and new version detection working.

### DIFF
--- a/plugins/provisioners/ansible/provisioner/base.rb
+++ b/plugins/provisioners/ansible/provisioner/base.rb
@@ -389,7 +389,7 @@ gathered version stdout version:
             if ansible_version_pattern
               _, @gathered_version, _ = ansible_version_pattern.captures
               if @gathered_version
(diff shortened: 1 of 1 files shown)


soapy1 (changes requested):
Thanks for the contribution, these changes look good! However, this should get a test. Probably somewhere in https://github.com/hashicorp/vagrant/blob/main/test/unit/plugins/provisioners/ansible/provisioner_test.rb.
If you could add that, it would be awesome 🎉
Cheers!

yodaaut:
Hi, thanks for your review and example. I hope this was the right way to go. =)

soapy1 (approved):

yodaaut:
thank you :tada: :smiley:

bcoca:
not something we are sure we want, but jic https://github.com/ansible/ansible/pull/74878

PR #12395, opened 2021-05-26
Migrate from GitHub HashiBot to GitHub Actions

bflad:
Replaces the only enabled behavior with an equivalent workflow.


chrisroberts (approved):
thanks!

PR #12412, opened 2021-06-07
Allow newer net-scp version

pvalena:
I've tested this in Fedora with net-scp 3.0 and it works fine, incl. the test suite. Are there any corner-cases I should test, specifically?

Test suite [*]: https://download.copr.fedorainfracloud.org/results/pvalena/vagrant/fedora-rawhide-x86_64/02225998-vagrant/builder-live.log.gz

Additional integration testing: https://git.io/JGMFw

[*] _Run as a part of Fedora build process, note that some tests are still failing after upgrading to Ruby 3.0._


chrisroberts:
Closing due to net-scp constraint being updated in this commit: https://github.com/hashicorp/vagrant/commit/24a00ee3b3472946d4c9235ff55f1d9416ab4577

PR #12413, opened 2021-06-09
Use more conservative TLS settings to work on older .NET versions

max-arnold:
A follow-up fix for https://github.com/hashicorp/vagrant/pull/12127

Tls13 is only available on .NET Framework 4.8, so the settings should be more conservative to work on older Windows versions. Based on testing reports from other collaborators:

https://github.com/hashicorp/vagrant/pull/12127#issuecomment-855910966
https://github.com/hashicorp/vagrant/pull/12127#issuecomment-856883383

Thanks to @leon-kleyn, @dafyddj and msmith in Salt Slack community.


dafyddj (approved):

max-arnold:
@soapy1 Sorry for pinging you again, but this is a small follow-up fix #12127. It would be nice to get it merged before the 2.2.17 release to ensure that older windows versions are supported as well

soapy1 (approved):
thanks!

PR #12414, opened 2021-06-09
VMware content update

chrisroberts:
Update VMware related content for open sourcing


soapy1 (dismissed):
🥳

BrandonRomano (dismissed):
Looks good from my end.  Redirects are set up properly here which is the only thing in this PR that could have caused any problems.

PR #12415, opened 2021-06-11
Flag RSA SHA1 deprecation when loading keys

### LINKED ISSUE #12344: Vagrant net-ssh is assuming that any RSA key is SHA1, resulting in a connection error
### Vagrant version

2.2.16

### Host operating system

Ubuntu 20.04

### Guest operating system

Ubuntu 20.04

### Debug output

D, [2021-05-01T15:09:40.735420 #92523] DEBUG -- net.ssh.authentication.session[2876340]: allowed methods: publickey
D, [2021-05-01T15:09:40.735468 #92523] DEBUG -- net.ssh.authentication.methods.none[28849a4]: none failed
D, [2021-05-01T15:09:40.735509 #92523] DEBUG -- net.ssh.authentication.session[2876340]: trying publickey
D, [2021-05-01T15:09:40.735739 #92523] DEBUG -- net.ssh.authentication.agent[289266c]: connecting to ssh-agent
E, [2021-05-01T15:09:40.735808 #92523] ERROR -- net.ssh.authentication.agent[289266c]: could not connect to ssh-agent: Agent not configured
D, [2021-05-01T15:09:40.735930 #92523] DEBUG -- net.ssh.authentication.methods.publickey[2892c20]: trying publickey (a9:7a:e7:aa:0b:89:72:91:86:81:19:4f:e0:d5:15:35)
D, [2021-05-01T15:09:40.736079 #92523] DEBUG -- net.ssh.authentication.methods.publickey[2892c20]: public key has been marked for deprecated ssh-rsa SHA1 behavior
D, [2021-05-01T15:09:40.736150 #92523] DEBUG -- socket[2811e90]: using encrypt-then-mac
D, [2021-05-01T15:09:40.736196 #92523] DEBUG -- socket[2811e90]: queueing packet nr 5 type 50 len 480
D, [2021-05-01T15:09:40.736238 #92523] DEBUG -- socket[2811e90]: sent 548 bytes
D, [2021-05-01T15:09:40.739143 #92523] DEBUG -- socket[2811e90]: read 516 bytes
D, [2021-05-01T15:09:40.739245 #92523] DEBUG -- socket[2811e90]: received packet nr 5 type 60 len 448
D, [2021-05-01T15:09:40.739287 #92523] DEBUG -- net.ssh.authentication.methods.publickey[2892c20]: public key has been marked for deprecated ssh-rsa SHA1 behavior
D, [2021-05-01T15:09:40.745181 #92523] DEBUG -- net.ssh.authentication.methods.publickey[2892c20]: public key has been marked for deprecated ssh-rsa SHA1 behavior
D, [2021-05-01T15:09:40.745208 #92523] DEBUG -- net.ssh.authentication.methods.publickey[2892c20]: private key has been marked for deprecated ssh-rsa SHA1 behavior
D, [2021-05-01T15:09:40.745260 #92523] DEBUG -- socket[2811e90]: using encrypt-then-mac
D, [2021-05-01T15:09:40.745307 #92523] DEBUG -- socket[2811e90]: queueing packet nr 6 type 50 len 896
D, [2021-05-01T15:09:40.745351 #92523] DEBUG -- socket[2811e90]: sent 964 bytes
D, [2021-05-01T15:09:40.752308 #92523] DEBUG -- socket[2811e90]: read 100 bytes
D, [2021-05-01T15:09:40.752410 #92523] DEBUG -- socket[2811e90]: received packet nr 6 type 51 len 32
D, [2021-05-01T15:09:40.752456 #92523] DEBUG -- net.ssh.authentication.session[2876340]: allowed methods: publickey
D, [2021-05-01T15:09:40.752484 #92523] DEBUG -- net.ssh.authentication.methods.publickey[2892c20]: publickey failed (a9:7a:e7:aa:0b:89:72:91:86:81:19:4f:e0:d5:15:35)
E, [2021-05-01T15:09:40.752507 #92523] ERROR -- net.ssh.authentication.session[2876340]: all authorization methods failed (tried none, publickey)

### Expected behavior

It should connect without any problem, because the is not actually refused by the server.

### Actual behavior

The ssh client (Vagrant, with specific patch https://github.com/hashicorp/vagrant/blob/main/lib/vagrant/patches/net-ssh.rb) is assuming that, because the ssh key is RSA, it's necessary SHA1, and it fails i.e. in "vagrant provision", despite the fact that "vagrant ssh" works well.
If you generate a ssh-key with "ssh-keygen -t rsa-sha2-512", it specifically uses the SHA2 algorithm, but this patch assumes that it can only be SHA1, because it's RSA:

"if s_ver >= Gem::Version.new("7.2") && pub_key.is_a?(OpenSSL::PKey::RSA)"

I'm working with OpenStack, but I guess it won't change in another situation. Reverting to Vagrant 2.2.15 solves (temporarily) the issue.

### LINKED ISSUE #12381: WinSSH Communicator errors out with Vagrant 2.2.16
### Vagrant version

2.2.16

### Host operating system

Windows 10 x64

### Guest operating system

Windows 10 x64

### Debug output

```
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
...
C:/HashiCorp/Vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/patches/net-ssh.rb:10:in `rsa_compat_build_request': undefined method `[]' for nil:NilClass (NoMethodError)
        from C:/HashiCorp/Vagrant/embedded/gems/2.2.16/gems/net-ssh-6.1.0/lib/net/ssh/authentication/methods/publickey.rb:41:in `send_request'
        from C:/HashiCorp/Vagrant/embedded/gems/2.2.16/gems/net-ssh-6.1.0/lib/net/ssh/authentication/methods/publickey.rb:51:in `authenticate_with'
        from C:/HashiCorp/Vagrant/embedded/gems/2.2.16/gems/net-ssh-6.1.0/lib/net/ssh/authentication/methods/publickey.rb:20:in `block in authenticate'
```

### Expected behavior

Vagrant should continue bringing up the machine.

### Actual behavior

Vagrant errors out with above stacktrace.

This is because the regex in `vagrant/lib/vagrant/patches/net-ssh.rb` for matching the OpenSSH version is not compatible with the version string of the build in OpenSSH server in Windows. 'OpenSSH_for_Windows_8.1p1, LibreSSL 3.0.2'

Changing the regex to something like `/OpenSSH_.*(?<version>\d+\.\d+).*/` makes vagrant work as expected.

### LINKED ISSUE #12408: `vagrant up` crashes with `rsa_compat_build_request': undefined method '[]' for nil:NilClass (NoMethodError)`
### Vagrant version

Vagrant 2.2.16

### Host operating system

CentOS Linux release 8.3.2011

### Guest operating system

Cisco IOS-XE 16.09.07

### Vagrantfile

```ruby
    vb.gui = true
    vb.customize ["modifyvm", :id, "--graphicscontroller", "vmsvga"]
  end
  config.vm.define "router1" do |router1|
    router1.vm.box = "cisco/csr1000v"
    # Turn off shared folders
    router1.vm.synced_folder ".", "/vagrant", id: "vagrant-root", disabled: true
    # Do not try to insert new SSH key
    router1.ssh.insert_key = false
    # Give the VM time to boot as Vagrant cannot tell when it is booted
    router1.vm.boot_timeout = 900
    router1.vm.network :private_network, virtualbox__intnet: "link1", ip: "192.168.50.21", auto_config: false, nic_type: "virtio"
    router1.vm.network "forwarded_port", guest: 22, host: 20001
    router1.trigger.after :up do |trigger|
      trigger.info = "Configuring interfaces"
      trigger.run = {inline: "./configure_ips.py -P 20001 127.0.0.1"}
    end
  end
```

### Debug output

https://gist.github.com/ggiesen/7d97219467a4ebb3d82899262963eb9a

The Gist contains 3 files:

1. vagrant_up_router1-debug-1.txt - The first time `vagrant up router1` is run
2. vagrant_up_router1-debug-2.txt - The second time `vagrant up router1` is run
3. vagrant_ssh_router1-debug.txt - Then `vagrant ssh router1` is run

### Expected behavior

Bring up box

### Actual behavior

The first time I run `vagrant up router1`, vagrant provisions the VM, and then crashes with the following error:

`/opt/vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/patches/net-ssh.rb:10:in `rsa_compat_build_request': undefined method `[]' for nil:NilClass (NoMethodError)`

The VM is left running, however, and I can run `vagrant up router1` a second time and finishes running the trigger. I also encounter no issues using `vagrant ssh router1`

If I use password-based authentication, vagrant does not crash (but I have to log into all devices manually).

### Steps to reproduce

1. Use above Vagrantfile with Cisco CSR1000v box built using https://github.com/hpreston/vagrant_net_prog
2. Run `vagrant up router1`

chrisroberts:
Set flag on RSA keys of deprecated RSA SHA1 support when loading
keys based on server version of the transport. This ensures keys
are properly flagged. Flag name has been updated to provide context
on usage.

Version matching on the OpenSSH server version has also been updated
to handle customized naming in the version string (as seen in the
Windows port) and to properly handle when no match is found.

Fixes #12344 
Fixes #12408 
Fixes #12381

### DIFF
--- a/lib/vagrant/patches/net-ssh.rb
+++ b/lib/vagrant/patches/net-ssh.rb
@@ -4,31 +4,121 @@ require "net/ssh"
 # these patches pull 6.1.0 up to the as of now
 # current 6.2.0 beta
 if Net::SSH::Version::STRING == "6.1.0"
-  require "net/ssh/authentication/methods/publickey"
-  Net::SSH::Authentication::Methods::Publickey.class_eval do
-    def rsa_compat_build_request(pub_key, *args)
-      s_ver_str = session.transport.server_version.version.match(/OpenSSH_(?<version>\d+\.\d+)/)[:version]
-      begin
-        s_ver = Gem::Version.new(s_ver_str)
-        if s_ver >= Gem::Version.new("7.2") && pub_key.is_a?(OpenSSL::PKey::RSA)
-          pub_key.deprecated_ssh_rsa = true
-          debug { "public key has been marked for deprecated ssh-rsa SHA1 behavior" }
-          info = key_manager.known_identities[pub_key]
-          if info && info[:key]
-            info[:key].deprecated_ssh_rsa = true
-            debug { "private key has been marked for deprecated ssh-rsa SHA1 behavior" }
-          else
-            warn { "cannot deprecate ssh rsa on private key, not loaded (#{info[:file]})" }
+  module DeprecatedRsaSha1
+    module KeyManager
+      def initialize(logger, options={})
+        @deprecated_rsa_sha1 = options.delete(:deprecated_rsa_sha1)
+        super
+      end
+
+      def sign(identity, data)
+        info = known_identities[identity] or raise Net::SSH::Authentication::KeyManager::KeyManagerError, "the given identity is unknown to the key manager"
+
+        if info[:key].nil? && info[:from] == :file
+          begin
+            info[:key] = Net::SSH::KeyFactory.load_private_key(info[:file], options[:passphrase], !options[:non_interactive], options[:password_prompt])
+            if @deprecated_rsa_sha1 && info[:key].respond_to?(:deprecated_rsa_sha1=)
+              info[:key].deprecated_rsa_sha1 = true
+              Vagrant.global_logger.debug("set RSA SHA1 deprecation on private key: #{info[:key].fingerprint}")
+            end
+          rescue OpenSSL::OpenSSLError, Exception => e
+            raise Net::SSH::Authentication::KeyManager::KeyManagerError, "the given identity is known, but the private key could not be loaded: #{e.class} (#{e.message})"
           end
         end
-      rescue ArgumentError
-        warn { "failed to parse OpenSSH version (raw: #{session.transport.server_version.version} attempted: #{s_ver_str}" }
+
+        if info[:key]
+          return Net::SSH::Buffer.from(:string, identity.ssh_signature_type,
+            :mstring, info[:key].ssh_do_sign(data.to_s)).to_s
+        end
+
+        if info[:from] == :agent
+          raise Net::SSH::Authentication::KeyManager::KeyManagerError, "the agent is no longer available" unless agent
+          return agent.sign(info[:identity], data.to_s)
+        end
+
+        raise Net::SSH::Authentication::KeyManager::KeyManagerError, "[BUG] can't determine identity origin (#{info.inspect})"
       end
-      _raw_build_request(pub_key, *args)
+
+      def load_identities(identities, ask_passphrase, ignore_decryption_errors)
+        identities.map do |identity|
+          begin
+            case identity[:load_from]
+            when :pubkey_file
+              key = Net::SSH::KeyFactory.load_public_key(identity[:pubkey_file])
+              if @deprecated_rsa_sha1 && key.respond_to?(:deprecated_rsa_sha1=)
+                key.deprecated_rsa_sha1 = true
+                Vagrant.global_logger.debug("set RSA SHA1 deprecation on public key: #{key.fingerprint}")
+              end
+              { public_key: key, from: :file, file: identity[:privkey_file] }
+            when :privkey_file
+              private_key = Net::SSH::KeyFactory.load_private_key(
+                identity[:privkey_file], options[:passphrase], ask_passphrase, options[:password_prompt]
+              )
+              key = private_key.send(:public_key)
+              if @deprecated_rsa_sha1 && key.respond_to?(:deprecated_rsa_sha1=)
+                key.deprecated_rsa_sha1 = true
+                private_key.deprecated_rsa_sha1 = true
+                Vagrant.global_logger.debug("set RSA SHA1 deprecation on public key: #{key.fingerprint}")
+                Vagrant.global_logger.debug("set RSA SHA1 deprecation on private key: #{private_key.fingerprint}")
+              end
+              { public_key: key, from: :file, file: identity[:privkey_file], key: private_key }
+            when :data
+              private_key = Net::SSH::KeyFactory.load_data_private_key(
+                identity[:data], options[:passphrase], ask_passphrase, "<key in memory>", options[:password_prompt]
+              )
+              key = private_key.send(:public_key)
+              if @deprecated_rsa_sha1 && key.respond_to?(:deprecated_rsa_sha1=)
+                key.deprecated_rsa_sha1 = true
+                private_key.deprecated_rsa_sha1 = true
+                Vagrant.global_logger.debug("set RSA SHA1 deprecation on public key: #{key.fingerprint}")
+                Vagrant.global_logger.debug("set RSA SHA1 deprecation on private key: #{private_key.fingerprint}")
+              end
+              { public_key: key, from: :key_data, data: identity[:data], key: private_key }
+            else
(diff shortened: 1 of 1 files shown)


soapy1 (approved):

berlin4apk:
@chrisroberts sorry but I think this is the rong way,
As it will only work with openssh and miss all the other ssh server Denons some are listed here https://en.m.wikipedia.org/wiki/Comparison_of_SSH_servers

This maybe will help
http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/src/usr.bin/ssh/PROTOCOL.certkeys?content-type=text/plain
"These RSA/SHA-2 types should not appear in keys at rest or transmitted
on the wire, but do appear in a SSH_MSG_KEXINIT's host-key algorithms
field or in the "public key algorithm name" field of a "publickey"
SSH_USERAUTH_REQUEST to indicate that the signature will use the
specified algorithm."

berlin4apk:
Since RSA keys are not dependent on the choice of hash function, the
  new public key algorithms reuse the "ssh-rsa" public key format as
  defined in [RFC4253]:

    string    "ssh-rsa"
    mpint     e
    mpint     n

  All aspects of the "ssh-rsa" format are kept, including the encoded
  string "ssh-rsa". This allows existing RSA keys to be used with the
  new public key algorithms, without requiring re-encoding, or affecting
  already trusted key fingerprints.

  Signing and verifying using these algorithms is performed according to
  the RSASSA-PKCS1-v1_5 scheme in [RFC8017] using SHA-2 [SHS] as hash.

  For the algorithm "rsa-sha2-256", the hash used is SHA-256.
  For the algorithm "rsa-sha2-512", the hash used is SHA-512.

  The resulting signature is encoded as follows:

    string    "rsa-sha2-256" / "rsa-sha2-512"
    string    rsa_signature_blob

  The value for 'rsa_signature_blob' is encoded as a string containing
  S - an octet string which is the output of RSASSA-PKCS1-v1_5, of
  length equal to the length in octets of the RSA modulus.
https://datatracker.ietf.org/doc/html/draft-ietf-curdle-rsa-sha2-12

chrisroberts:
@berlin4apk Aside from issues with other ssh server implementations (have any others fully deprecated the ssh-rsa signature?) I'm not clear on why you think this is the wrong direction. The copied sections you pasted don't really provide much context on your objection. Could you state more clearly what you think is incorrect? Thanks.

berlin4apk:
The way via the Server Version string will break to easyly
If we can we should get the supported ALG out of the season Infos form the season Initiation
Or if it's not possible to try it out one after the other at connection time not at the time of the key loading

See also https://github.com/net-ssh/net-ssh/pull/811/commits

chrisroberts:
Currently I'm unaware of any other server implementations affected outside of OpenSSH (and its derivatives). These modifications are not meant to be a long term fix, rather they are simply patches for things to work until support is available upstream (this is why the patches are _only_ applied if a specific version of the ssh library is loaded, which is the currently released version right now). Detecting based on the session is ideal, however, from the point of view of monkey patching the library after it has been loaded, is much more difficult. The patch in its current state will allow things to continue working where it is currently know to fail (openssh versions 8.3 and greater) and will no longer be applicable once the library can be upgraded.

PR #12422, opened 2021-06-17
Fixing broken reference to ansible website

nitbx:
Some of these URL were pointing to http://docs.ansible.com/ansible/[...] that results of a 404 not found.

I updated all links to https://docs.ansible.com/ansible/latest/[...].

Thank you


soapy1 (approved):
thanks! 📜

PR #12431, opened 2021-06-23
Update tests to use real UI instance

chrisroberts:
Replaces use of UI doubles within tests to use actual UI instance
    to ensure calls are passing parameters correctly.


soapy1 (approved):
🧪

PR #12439, opened 2021-07-01
Use systemctl poweroff in the background instead of shutdown

dcermak:
We have started seeing occasional shutdown failures on openSUSE Tumbleweed with
Virtualbox inside a qemu virtual machine, where `shutdown -h now` would return
nil. While the machine is successfully turned off, the command fails and vagrant
reports an error.
This commit changes the shutdown command to launch in the background which
also triggers a shutdown, but always succeeds.

### DIFF
--- a/plugins/guests/suse/cap/halt.rb
+++ b/plugins/guests/suse/cap/halt.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
       class Halt
         def self.halt(machine)
           begin
-            machine.communicate.sudo("/sbin/shutdown -h now")
+            machine.communicate.sudo("/usr/bin/systemctl poweroff &")
           rescue IOError, Vagrant::Errors::SSHDisconnected
             # Do nothing, because it probably means the machine shut down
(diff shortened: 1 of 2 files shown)


soapy1 (approved):

TheKrizz:
Won't work on SLES11.
SLES11 has no systemctl

dcermak:
> Won't work on SLES11.
> SLES11 has no systemctl

Are you really using SLES11 with vagrant? There aren't even any supported SLES 11 vagrant boxes…

TheKrizz:
Yes, we are really using vagrant with SLES11. 
We have this OS limitation by using an old version of emc(opentext) documentum.
So we need Vagrant with SLES11 to test changes locally.

chrisroberts:
@TheKrizz Would you mind opening a new issue? We can look at adding some detection and a fallback to allow continued functionality on older versions.

TheKrizz:
Created following issue https://github.com/hashicorp/vagrant/issues/12487

PR #12442, opened 2021-07-06
Remove no longer required gem webmock

dcermak:
This gem was added a long time ago and probably used to test the functionality
that is now provided by the vagrant_cloud gem


chrisroberts (approved):
:+1: thanks!

PR #12460, opened 2021-07-16
Add Astra Linux platforms guest detection and support

astralnx:
These changes are necessary for correct work with Astra Linux images hosted on Vagrantcloud.

### DIFF
--- a/plugins/guests/astra/guest.rb
+++ b/plugins/guests/astra/guest.rb
@@ -1,0 +1,10 @@
+require_relative '../linux/guest'
+
+module VagrantPlugins
+  module GuestAstra
+    class Guest < VagrantPlugins::GuestLinux::Guest
+      # Name used for guest detection
+      GUEST_DETECTION_NAME = "astra".freeze
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


soapy1:
Hey there,
We appreciate your interest in building a Vagrant plugin! However, we currently do not want to pursue adding Astra Linux as a built in plugin to Vagrant. 

While this means that the guest will not work out of the box with Vagrant, you are still able to build a third party plugin. There are plenty of popular third party Vagrant plugins, this is a totally fine way to use Vagrant.

Here is a guide for how to go about building a plugin https://www.vagrantup.com/docs/plugins/development-basics.
You might also want to take some inspiration from other popular plugins like:
- https://github.com/vagrant-libvirt/vagrant-libvirt
- https://github.com/dotless-de/vagrant-vbguest

Once you have a gem of your plugin you can then install it using the vagrant plugin command `vagrant plugin install my-plugin`

I'm going to go ahead and close this out now. If you have any more questions, or are looking for guidance for plugin development I suggest you post a question in our discuss forum https://discuss.hashicorp.com/c/vagrant/24.
Cheers

PR #12463, opened 2021-07-20
WinSSH: Send eof to let server know we're done

cliffe:
Just adds an EOF when done. Without this, the SSH connection can just persist and eventually time out. The fix matches the behaviour in the ssh communicator.


cliffe:
The way Powershell return codes are used is complicated, and this additional change fixes an error where the return code was not propagating back from WinSSH correctly.

The two above changes fixes WinSSH in our environment

cliffe:
2 years later, and this patch is still required to get WinSSH working in our environment (oVirt and Proxmox with Vagrant)

PR #12470, opened 2021-07-25
Upgrade global styles

zchsh:


zchsh:
Note for review: the `product-downloads-page` title looks a bit off, since "Vagrant VMWare Utility" isn't really a product. Have made an [Asana task]() to capture proposed changes to the component. Not sure if the dashed title in this PR is bad enough that we should block it or break out the downloads page work while we make updates to the component - open to thoughts 🙇 

Here's what [this PR's VMWare Utility downloads page](https://vagrant-git-fork-zchsh-zsupgrade-global-styles-hashicorp.vercel.app/vmware/downloads) looks like (my best effort given the title limitation):

CleanShot 2021-08-07 at 11 19 46@2x

zchsh:
Update: have made some tweaks to `product-downloads-page` via https://github.com/hashicorp/react-components/pull/281, so the VMware utility page now has a nicer title:

CleanShot 2021-08-12 at 15 30 05@2x

website/pages/404.module.css
```diff
@@ -13,24 +13,11 @@
...
-  }
-
   & a {
     color: var(--highlight-color);
   }
 }
+
+.heading {
+  composes: g-type-display-3 from global;
+}
```
jescalan, line 23: 👏

website/pages/style.css
```diff
@@ -1,7 +1,5 @@
 /* Global Component Styles */
 @import '~@hashicorp/mktg-global-styles/style.css';
-@import '~@hashicorp/mktg-global-styles/_temporary-to-remove/layout.css';
-@import '~@hashicorp/mktg-global-styles/_temporary-to-remove/tables.css';
```
jescalan, line 4: 👏

website/pages/vmware/downloads/index.jsx
```diff
@@ -1,42 +1,79 @@
...
+            label: 'Fedora',
+            os: 'NONE--IGNORE',
+          },
+          {
+            label: 'Ubuntu/Debian',
+            os: 'NONE--IGNORE',
+          },
+          {
+            label: 'CentOS/RHEL',
+            os: 'NONE--IGNORE',
```
jescalan, line 60: What's going on with this?

jescalan (approved):
This is great!

website/pages/vmware/downloads/index.jsx
zchsh: This was the only way I could find to stop the package managers from showing up. Open to other ways to configure just the binary downloads to show up, if there is one 🙇

zchsh: Added a comment here to clarify the `NONE--IGNORE` part - definitely open to iterating on this if we find a better way.

zchsh: Update here based on recent sync: @jescalan is looking at making a change to allow us to not do this, as it's certainly an unexpected use of the interface. If feasible, we might update the ProductDownloadsPage component so that "package managers" can be disabled entirely (that's the use case here, as vagrant-vmware-utility support in package managers is unknown). Alternately, if not feasible, we'll proceed with this pattern (for now).

Issue #12487, opened 2021-08-05
Fallback Solution to issue #12439 (Use systemctl poweroff in the background instead of shutdown)


TheKrizz:
Since Vagrant Version 2.2.17  improvement guest/suse: Use systemctl poweroff in the background instead of shutdown [GH-12439] was merged, we stuck on updating our vagrant version because we need to use SLES11.

Usage of SLES11 is currently mandatory for us (Dependency to Documentum Software)

machine.communicate.sudo("/usr/bin/systemctl poweroff &")  won't work on SLES11 because it has no systemctl

At this point there should be a fallback.
If os has no systemctl it should you the "old way" (machine.communicate.sudo("/sbin/shutdown -h now"))

### Guest operating system

SLES11

### Expected behavior

Clean shutdown/halt of a Vagrant Box with SLES11

### Actual behavior

No clean shutdown/halt of a Vagrant Box with SLES11

### Steps to reproduce

vagrant halt (with a SLES11 Box)

### References

https://github.com/hashicorp/vagrant/pull/12439
https://github.com/hashicorp/vagrant/issues/11983

PR #12489, opened 2021-08-06
Add fallback for SLE 11 guest machines

dcermak:
SLE 11 does not ship systemd and then using `systemctl poweroff` does not work. Therefore we fall back to using `/sbin/shutdown` for machines without systemd.
This fixes https://github.com/hashicorp/vagrant/issues/12487


chrisroberts (approved):

chrisroberts:
thanks @dcermak!

PR #12510, opened 2021-09-07
[OpenBSD rsync] Do not rely on hardcoded PKG_PATH URI but rather on the default behav…

sk4nz:
…ior of pkg_add(1) which selects the mirror based on the content of '/etc/installurl' to install rsync on OpenBSD targets.

This behaviour is present since OpenBSD 6.1 (released April 11, 2017).

The hardcoded URI prevents the use of custom OpenBSD mirrors (written down to '/etc/installurl' by the installer) while using the rsync directive in Vagrantfiles.

This prevents 'Installing rsync to the VM...' failure when provisioning a VM configured with a custom mirror.

https://man.openbsd.org/pkg_add.1
https://man.openbsd.org/installurl.5

### DIFF
--- a/plugins/guests/openbsd/cap/rsync.rb
+++ b/plugins/guests/openbsd/cap/rsync.rb
@@ -8,9 +8,7 @@ module VagrantPlugins
 
         def self.rsync_install(machine)
           install_output = {:stderr => '', :stdout => ''}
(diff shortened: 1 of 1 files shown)


soapy1:
Going ahead and closing this as per https://github.com/hashicorp/vagrant/pull/12656#issuecomment-1320453983

PR #12511, opened 2021-09-07
use nfsd update for macos

uwej711:
When using udp as nfs protocol restarting nfs on macos results in udp port still being bound and nfsd being unabled to bind to the udp port after restart. nfsd update just tells the nfs server to reload the configuration and exports which is enough after the exports file has been updated.
As nfs via tcp no longer seems to work reliably under big sur, this has some urgency.


uwej711:
Need to revisit this again.

PR #12529, opened 2021-10-01
Fix syntax error in trigger configuration docs

jayay:
Hi!

I've spotted a syntax error in the docs. I haven't checked if this part of the documentation is correct and makes sense as a whole, though.


chrisroberts (approved):
thanks!

PR #12538, opened 2021-10-14
Fix for :dvd example in disks configuration docs.

catalincezarene:
Option passed to :dvd disk type should be `file` not `path`.

```

vm:
* The following settings shouldn't exist: path
* A 'file' option is required when defining a disk of type ':dvd' for guest 'default'.
```


chrisroberts (approved):
Thanks!

Issue #12346, opened 2021-05-03
chef_solo provision wrong detection of chef existance for Windows guest


111andre111:
### Vagrant version
2.2.15

### Host operating system
OS Name : "Microsoft Windows 10 Pro 20H2"
OS Version: "10.0.19042 Build 19042"

### Guest operating system
OS Name : "Microsoft Windows Server 2019 Datacenter Evaluation"
OS Version : "10.0.17763 Build 17763"

### Vagrantfile

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :
# All Vagrant configuration is done below. The "2" in Vagrant.configure
# configures the configuration version (we support older styles for
# backwards compatibility). Please don't change it unless you know what
# you're doing.
Vagrant.configure("2") do |config|
  config.vm.box = "StefanScherer/windows_2019"
  config.vm.communicator = "winrm"
  config.vm.network "forwarded_port", host: 2200, guest: 3389
  config.vm.provider "virtualbox" do |v|
    v.cpus = 2
    v.memory = 2048
  end
  config.winrm.timeout =   1800 # 30 minutes
  config.vm.boot_timeout = 1800 # 30 minutes
  config.winrm.max_tries = 600
  config.winrm.retry_delay = 5
  config.vm.provision "chef_solo" do |chef|
    # chef.cookbooks_path = ['C:\vagrant']
    chef.add_recipe "example-windowsserver2019"
    chef.arguments = "--chef-license accept"
    chef.run_list = [
      "recipe[chef-default]"
    ]
    # chef.install = false
  end
end
```

Please ensure the Vagrantfile provided is a minimal Vagrantfile which contains
only the required configuration to reproduce the behavior. Please note that if
your Vagrantfile contains an excess of configuration unrelated the the reported
issue, or is in a different format, we may be unable to assist with your issue.
Always start with a minimal Vagrantfile and include only the relevant configuration
to reproduce the reported behavior.

### Debug output
https://gist.github.com/111andre111/73d31bf50f0f7576b0d07a9d32323262

### Expected behavior
Using `chef_solo` provisioned `chef-solo` should be found to be installed, but it doesn't.

### Actual behavior
At this point in time, the chef_solo provisioned results in following error:
```
DEBUG winrmshell: [WinRM] remote shell created with shell_id: 91B40CC4-9705-4413-9E6A-B50A68FE59AB
[...]

Issue #12337, opened 2021-04-29
Vagrant with chef solo fails to install in guest OS


rferguson10:
### Vagrant version

Vagrant 2.2.15

### Host operating system

Windows 10 Home

### Guest operating system

Centos 8 and 8.3

### Vagrantfile

```ruby
Vagrant.configure(2) do |config|
  
  config.vm.box = "bento/centos-8.3"
  config.vm.hostname = 'wordpress'

  config.vm.network :private_network, ip: '172.16.231.130'
  config.vm.network :forwarded_port, guest: 80, host: 8080
  config.vm.network :forwarded_port, guest: 9200, host: 9201

  if Vagrant.has_plugin?("vagrant-omnibus")
    config.omnibus.chef_version = 'latest'
  end
 
  config.ssh.insert_key = false
  config.vm.synced_folder ".", "/vagrant", type: "nfs", mount_options:['nolock,vers=3,udp,noatime,actimeo=1']

  config.vm.provider "virtualbox" do |v|
    v.memory = 2048
  end

  config.vbguest.auto_update = false

  config.vm.provision :shell, :inline => "ulimit -n 4048"

  config.vm.provision "chef_solo" do |chef|
    chef.arguments = "--chef-license accept"
    chef.run_list = [
      "recipe[wordpress]"
    ]
  end

end
```

### Debug output

https://gist.github.com/rferguson10/a130b436c1106c1e8a170d88d3d65b88

### Expected behavior

Chef should install into the guest OS and continue to provision using the provided recipe

### Actual behavior

I'm presented with the following output:

Vagrant could not detect Chef on the guest! Even after Vagrant
attempted to install Chef, it could still not find Chef on the system.
Please make sure you are connected to the Internet and can access
Chef's package distribution servers. If you already have Chef
installed on this guest, you can disable the automatic Chef detection
by setting the 'install' option in the Chef configuration section of
your Vagrantfile:

    chef.install = false

### Steps to reproduce

1. vagrant up, watch for output

ralfbetazi:
I find myself experiencing same problem as @rferguson10 but in a different environment as of this morning (~7:30 AM Pacific time).

### Vagrant Version

2.2.15

### Host operating system

Mac OS Big Sur 11.3

### Guest operating system

Ubuntu 20.04

### Vagrantfile

```
Vagrant.configure("2") do |config|
  config.vm.box = "bento/ubuntu-20.04"
  config.vm.define('demo') do |demo|
[...]

PR #12555, opened 2021-10-24
Updates install checks for Chef provisioners

collinmcneese:
## Issues Resolved

- #12346 
- #12337

## Changes

- Fixes `chef_installed` detection to be compatible with Chef Infra Client 17+.  Chef Infra Client 17 removed the `knife` gem from the omnibus installation so existing detection of the install status is broken.  Updates the `chef_installed` checks to look for the existence of the `chef-client` binary instead of `knife` and performs version checking with 'chef-client'
- Updates the `chef_installed` detection to use the `chef` binary if `chef-workstation` is configured as the `chef.product`. Updates to use the Chef Workstation version (via `chef --version`) for comparison if `chef.version` is specified along with `chef.product = 'chef-workstation'`.


chrisroberts (approved):
:+1: thanks!

PR #12565, opened 2021-10-29
Add blurb about detection and use of Systemd

adam-vessey:
Detection/use added in https://github.com/hashicorp/vagrant/pull/11482, which has been included since version 2.2.8

Did a quick look around for similar documentation regarding rebooting, which might need to be updated as per https://github.com/hashicorp/vagrant/pull/11483; however, didn't find anything... of course possible that there's something I've missed.


chrisroberts (approved):
Thanks

Issue #12521, opened 2021-09-15
Vagrant failed to initialize at a very early stage: Json plugin Failing


brokencode64:
### Vagrant version
Vagrant 2.2.16

### Host operating system
Fedora 34

### Guest operating system
Debian 10.10

### Environment Variables
```
VAGRANT_DEFAULT_PROVIDER=libvirt
VAGRANT_HOME=/storage/vagrant
```

### Vagrant Plugins Installed
- pkg-config (1.4.6, global)
- vagrant-libvirt (0.5.3, global)
- vagrant-mutate (1.2.0, global)

### Vagrantfile
```
Vagrant.configure("2") do |config|

  config.vm.define "jenkinsserver" do |jenkinsserver|
    jenkinsserver.vm.box_download_insecure = true
    jenkinsserver.vm.box = "brokencode/debian-10.10"
    jenkinsserver.vm.box_version = "1.0.0"
    jenkinsserver.vm.network "forwarded_port", guest: 8080, host: 8080
    jenkinsserver.vm.network "forwarded_port", guest: 8081, host: 8081
    jenkinsserver.vm.network "forwarded_port", guest: 9090, host: 9090
    jenkinsserver.vm.network "private_network", ip: "172.28.128.1", auto_config: false
    jenkinsserver.vm.hostname = "jenkinsserver"
    jenkinsserver.vm.provider "libvirt" do |v|
      v.uri = 'qemu+unix:///system'
      v.host = "jenkinsserver"
      v.memory = 2048
      v.cpus = 2
    end
  end
end

```

### Debug output
https://gist.github.com/brokencode64/3c7695e413aa13d73ed18aaa87122154

### Expected behavior
Vagrant should start up, then be able to `vagrant ssh`.

### Actual behavior
`Vagrant up` will work the first time, but after that any vagrant command will give the following output. The server itself works, I can login via virt-manager.

```
Ignoring ffi-1.13.1 because its extensions are not built. Try: gem pristine ffi --version 1.13.1
Ignoring json-2.5.1 because its extensions are not built. Try: gem pristine json --version 2.5.1
Ignoring unf_ext-0.0.7.7 because its extensions are not built. Try: gem pristine unf_ext --version 0.0.7.7
Vagrant failed to initialize at a very early stage:

The plugins failed to initialize correctly. This may be due to manual
modifications made within the Vagrant home directory. Vagrant can
attempt to automatically correct this issue by running:

  vagrant plugin repair

[...]

Issue #12345, opened 2021-05-03
Updating vagrant-libvirt from 0.3 to 0.4.1 with Vagrant 2.2.16 fails because new version cannot be "activated"


zimmski:
### Expected behavior

Installation works like it did with 0.3.0

### Actual behavior

Installation fails with `/opt/vagrant/embedded/lib/ruby/2.6.0/rubygems/specification.rb:2292:in `check_version_conflict': can't activate vagrant-libvirt-0.4.1, already activated vagrant-libvirt-0.3.0 (Gem::LoadError)`

### Steps to reproduce

1. Install a VM on openSUSE 15.2 with Vagrant 2.2.14 and vagrant-libvirt 0.3.0
2. Update the versions Vagrant to 2.2.16
3. Run `vagrant plugin expunge --local-only --force`
3. Update vagrant-libvirt to 0.4.1 using `vagrant plugin install --local`
3. See the following error

```
Installing the 'vagrant-libvirt --version '0.4.1'' plugin. This can take a few minutes...
/opt/vagrant/embedded/lib/ruby/2.6.0/rubygems/specification.rb:2292:in `check_version_conflict': can't activate vagrant-libvirt-0.4.1, already activated vagrant-libvirt-0.3.0 (Gem::LoadError)
        from /opt/vagrant/embedded/lib/ruby/2.6.0/rubygems/specification.rb:1414:in `activate'
        from /opt/vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/bundler.rb:751:in `block in activate_solution'
        from /opt/vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/bundler.rb:748:in `each'
        from /opt/vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/bundler.rb:748:in `activate_solution'
        from /opt/vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/bundler.rb:588:in `internal_install'
        from /opt/vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/bundler.rb:350:in `install'
        from /opt/vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/plugin/manager.rb:138:in `block in install_plugin'
[...]

Issue #10826, opened 2019-05-05
Vagrant 2.2.4 fails to install vagrant-vbguest plugin


dsofeir:
### Bug description
Plugin 'vagrant-vbguest' won't install after fresh install of vagrant

### Vagrant version
2.2.4

### Host operating system
OpenSUSE  Tumbleweed5.0.9-1-default

### Vagrantfile
```
# -*- mode: ruby -*-
# vi: set ft=ruby :

# All Vagrant configuration is done below. The "2" in Vagrant.configure
# configures the configuration version (we support older styles for
# backwards compatibility). Please don't change it unless you know what
# you're doing.
Vagrant.configure("2") do |config|
  # The most common configuration options are documented and commented below.
  # For a complete reference, please see the online documentation at
  # https://docs.vagrantup.com.

  # Every Vagrant development environment requires a box. You can search for
  # boxes at https://vagrantcloud.com/search.
  config.vm.box = "centos/7"

  # Disable automatic box update checking. If you disable this, then
  # boxes will only be checked for updates when the user runs
  # `vagrant box outdated`. This is not recommended.
  # config.vm.box_check_update = false

  # Create a forwarded port mapping which allows access to a specific port
  # within the machine from a port on the host machine. In the example below,
  # accessing "localhost:8080" will access port 80 on the guest machine.
  # NOTE: This will enable public access to the opened port
  # config.vm.network "forwarded_port", guest: 80, host: 8080

  # Create a forwarded port mapping which allows access to a specific port
  # within the machine from a port on the host machine and only allow access
  # via 127.0.0.1 to disable public access
  # config.vm.network "forwarded_port", guest: 80, host: 8080, host_ip: "127.0.0.1"

  # Create a private network, which allows host-only access to the machine
  # using a specific IP.
  # config.vm.network "private_network", ip: "192.168.33.10"

  # Create a public network, which generally matched to bridged network.
  # Bridged networks make the machine appear as another physical device on
  # your network.
  # config.vm.network "public_network"

  # Share an additional folder to the guest VM. The first argument is
  # the path on the host to the actual folder. The second argument is
  # the path on the guest to mount the folder. And the optional third
  # argument is a set of non-required options.
  # config.vm.synced_folder "../data", "/vagrant_data"

  # Provider-specific configuration so you can fine-tune various
  # backing providers for Vagrant. These expose provider-specific options.
  # Example for VirtualBox:
  #
  # config.vm.provider "virtualbox" do |vb|
[...]

PR #12567, opened 2021-11-01
Improve Gem spec selection when resolving

electrofelix:
When computing the solution set, if a gem is already loaded, make sure
to use the specification of the loaded one instead of the first
available as otherwise there is a risk that when multiple matches are
available the specification for the wrong version may be picked.

When this happens an error message will be triggered that looks like

  can't activate json-2.3.0, already activated json-2.5.1

This can occur for distribution packaged vagrants as well as installs
for development purposes where the ruby install may contain a default
gem spec of an older version than is needed.

Fixes: #12521
Fixes: vagrant-libvirt/vagrant-libvirt#1390
Relates: #10826
Relates: #12345

### DIFF
--- a/lib/vagrant/bundler.rb
+++ b/lib/vagrant/bundler.rb
@@ -258,7 +258,12 @@ module Vagrant
       if solution_file&.valid?
         @logger.debug("loading cached solution set")
         solution = solution_file.dependency_list.map do |dep|
-          spec = composed_set.find_all(dep).first
+          spec = composed_set.find_all(dep).select do |dep_spec|
+            next(true) unless Gem.loaded_specs.has_key?(dep_spec.name)
+
+            Gem.loaded_specs[dep_spec.name].version.eql?(dep_spec.version)
+          end.first
+
           if !spec
             @logger.warn("failed to locate specification for dependency - #{dep}")
             @logger.warn("invalidating solution file - #{solution_file}")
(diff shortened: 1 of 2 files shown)


electrofelix:
I need to check with my employer regards signing the CLA, hence leaving this in draft for now.

ranguli:
I'm currently experiencing this bug right now, would be great if it could be merged!

RockyBukake:
Bump, this is still an issue, can someone get this merged ...

electrofelix:
@chrisroberts Are any PRs being accepted for vagrant at the moment? If not, I might look into seeing if the more popular distros will pick up this PR as a patch for their packaging of vagrant as this does tend to bite users occasionally. Unfortunately that won't help when people bump into it for plugin development, but I think most devs will be able to apply this locally as needed.

chrisroberts:
@electrofelix Yes, PRs are being accepted. We're just getting the last of our build pipelines setup and will be merging in our changeset once that's done. After that I'm going to be running through current PRs to get high priority/value ones merged for this release.

electrofelix:
Thanks, I wasn't sure if the focus was on releasing v3 soon making this obsolete and need to be carried by distros staying with v2. I'll wait rather than poking some of the maintainers to carry. I think this impacts users of vagrant-libvirt more than others, so slightly vested interest here.

phinze (approved):
I was unable to reproduce any gem mismatch errors as described in the linked issues, but I'm betting this is because I was using just-provisioned hosts with pristine Rubygems environments. If anybody has from-scratch repro steps feel free to chime in!

Nevertheless, studying this change it looks to be safe enough to land on behalf of the folks who are seeing this problem. If we start seeing lots of bundler errors we can always back it out.

phinze:
Thanks for the contribution @electrofelix! This change should come out with the Vagrant 2.3 release soon.

PR #12574, opened 2021-11-04
show error in documentation

GitarPlayer:
I tried the above setup and I could not make it work because the private keys are not copied to the vagrant vms and it's not possible to include them. I even tried: 
```
  config.vm.synced_folder ".vagrant", "/.vagrant", type: "rsync",
    rsync__exclude: ".git/"
end
```
To include it but the default exclusion of .vagrant makes it not included in the end.

### DIFF
--- a/website/content/docs/provisioning/ansible_local.mdx
+++ b/website/content/docs/provisioning/ansible_local.mdx
@@ -187,9 +187,9 @@ Vagrant.configure(2) do |config|
 end
 ```
 
-### Ansible Parallel Execution from a Guest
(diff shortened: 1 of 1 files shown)


soapy1:
Hey there, thanks for opening up a PR. I'm not quite sure what the issue is, but if you are still having trouble a good place to bring it up is in the discuss forum https://discuss.hashicorp.com/c/vagrant/24. Please be sure to clearly state what you are trying to achieve, what you tried and what you expected the outcome to be. Since this is more of a question than a change that should be made to the Vagrant docs I'm going to go ahead and close this PR.

PR #12585, opened 2021-11-10
Added smb_username and smb_password to example

mikeschinkel:
Adding those to the example because I had to Google for this question and answer, and it took a bit to find it:

https://stackoverflow.com/questions/44394725/how-do-i-set-the-smb-username-and-password

Clearly, I wasn't the only one with the question, so I assume updating the docs will help the next person who otherwise would have had to track down that same Q&A?


soapy1:
Thank you for trying to make the Vagrant docs better! We appreciate the time you took to look at our documentation and submit this. At this time we don't think this change improves the docs so I'm going to go ahead and close this PR.

PR #12607, opened 2021-12-06
Only clean up /etc/fstab if it really contains Vagrant data

emanuelsteen:
This change, together with `allow_fstab_modification = false`, allows to run
a guest without root/sudo privileges which been broken since Vagrant 2.2.10.

### DIFF
--- a/plugins/guests/linux/cap/persist_mount_shared_folder.rb
+++ b/plugins/guests/linux/cap/persist_mount_shared_folder.rb
@@ -61,9 +61,17 @@ module VagrantPlugins
           machine.communicate.test("test -f /etc/fstab")
         end
 
+        def self.contains_vagrant_data?(machine)
+          machine.communicate.test("grep '#VAGRANT-BEGIN' /etc/fstab")
+        end
+
         def self.remove_vagrant_managed_fstab(machine)
           if fstab_exists?(machine)
-            machine.communicate.sudo("sed -i '/\#VAGRANT-BEGIN/,/\#VAGRANT-END/d' /etc/fstab")
+            if contains_vagrant_data?(machine)
+                machine.communicate.sudo("sed -i '/\#VAGRANT-BEGIN/,/\#VAGRANT-END/d' /etc/fstab")
+            else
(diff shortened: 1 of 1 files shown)


emanuelsteen:
FYI, initially I planned to modify https://github.com/hashicorp/vagrant/blob/main/lib/vagrant/action/builtin/synced_folders.rb#L140 to never even call `persist_mount_shared_folder` (https://github.com/hashicorp/vagrant/blob/78bc0243338cd3b271226cefd9ae21f0ed985065/plugins/guests/linux/cap/persist_mount_shared_folder.rb#L19). But there are explicit tests to validate that this function is called with a `nil` argument (https://github.com/hashicorp/vagrant/blob/78bc0243338cd3b271226cefd9ae21f0ed985065/test/unit/vagrant/action/builtin/synced_folders_test.rb#L250). I find that logic pretty odd but I left it as is in case there are use cases relying on it. But, IMHO, if you set `allow_fstab_modification = false` we should never even attempt to modify `/etc/fstab` and not even attempt to clean it up.

Zotteljedi:
I struggle with the same issue, Vagrant trying to remove its magic from a fstab that doesn't contain any markers in the first place. What is the status of this pull request, what needs to be done to get this fixed? Or is there a known workaround by now?

stan-quinn:
Can I give this an up vote.
My box does have an /etc/fstab, but my root file system is read only, so vagrant can't change it.
And even with "allow_fstab_modification = false", it insists on trying to change it, then failing hard.

emanuelsteen:
Looks like it needs a review approval to be merged. Personally, I will run a custom build of Vagrant with this patch.

stan-quinn:
It looks like 2.3.7 is released and this change didn't make it into that.
This is still an issue for me, with my read-only root file system.
Is there anything I can do to help get this into the next release?

wian007:
What is the status of this issue? I like to see it merged

taru-garg-2000:
Hi @emanuelsteen,

Thank you very much for your submission! We've merged f94e931 with #13681, which includes unit tests for the `remove_vagrant_managed_fstab` method.

Cheers!

PR #12651, opened 2022-01-14
fix(website): enable error page for 500s

zchsh:
👀 500 Error page [Preview](https://vagrant-git-zsaddress-500-error-page-hashicorp.vercel.app/docs/v2.2.10/synced-folders/nfs) (vs [live](https://www.vagrantup.com/docs/v2.2.10/synced-folders/nfs))
👀 404 Error page [Preview](https://vagrant-git-zsaddress-500-error-page-hashicorp.vercel.app/nope-doesnt-exist) (vs [live](https://www.vagrantup.com/nope-doesnt-exists))

[ WIP ]

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201534814309539


zchsh:
@BRKalow tagging ya here cause I've started abstracting out `components/error-view`, as a candidate to be moved into [react-components](https://github.com/hashicorp/react-components).

As I was doing this, ended up pulling out a `useErrorPageAnalytics` hook, which feels a bit more re-usable than the `error-view` components (which is essentially just that hook bundled with some layout-specific UI).

Seeing at least a few paths forward here, wanted to get your take on this before diving in further / opening PRs on other repos:

1. Move `components/error-view` into `react-components`, publish as `@hashicorp/react-error-view`, don't expose `useErrorPageAnalytics` hooks, since we don't need it (yet).
2. Do the above, but expose `useErrorPageAnalytics` as a named export, idea being that it seems likely want to re-use the `error-view` analytics logic, but not the layout-specific styling, on other projects, namely `dev-portal`.
3. Move `useErrorPageAnalytics` into [`@hashicorp/platform-analytics`](https://github.com/hashicorp/web-platform-packages/tree/main/packages/analytics) instead of bundled int he `react-component`. Would still make `@hashicorp/react-error-view`, would have an underlying dep on `platform-analytics`. Idea here being `platform-analytics` seems intended as the home for all analytics stuff. But maybe this should apply to top-level analytics stuff only (eg initializing `window.analytics`), and not specific calls (`useErrorPageAnalytics` just calls `window.analytics`, expects it to be set-up beforehand)?
4. Something else?

Very open to input here! 🙇‍♂️

BRKalow:
@zchsh My first thought was to place it in the analytics package, but if that feels like too many layers of abstraction I think it'd be fine to place it in `react-error-view` for now and just not export it 👍

dstaley:
Definitely a fan of providing the `useErrorPageAnalytics` hook _somehow_. No strong feelings whether it's a named export, from a package in `react-components`, or whether it's provided by `platform-analytics`. I'm leaning towards agreeing with Bryce that it makes the most sense with the other analytics stuff, but I can also see dependencies being hairy to manage when we're trying to coordinate a breaking change to the API.

PR #12653, opened 2022-01-16
Allow link-local IPv6 address on host-only network

rubenk:
Since VirtualBox 6.1.28, if you want to create a host-only network,
virtualbox checks if the ip is in a certain range.

According to the manual (https://www.virtualbox.org/manual/ch06.html),
ipv6 link-local addresses are allowed so add the IPv6 link-local range.


rubenk:
ping?

rubenk:
ping again

chrisroberts (approved):
Thanks!

PR #12656, opened 2022-01-18
openbsd: rsync switch to nluug.nl

zhanhb:
to support more versions


chrisroberts (changes requested):
I don't think that changing from the openbsd site for this download is the best here. I think a better solution would be to surface source as a configurable value which defaults to the current openbsd site.

chrisroberts:
Related: #12510

phinze:
Going to close this PR, but a different one that made this value configurable would still be welcome.

PR #12669, opened 2022-02-01
update docs: config.ssh.private_key_path if use ssh-agent will failed

linroex:
if user use config.ssh.private_key_path and also use ssh-agent (ex. macos maxgoedjen/secretive), user need disable config.ssh.keys_only param also, for prevent IdentitiesOnly influence.


soapy1:
Thank you for trying to make the Vagrant docs better! We appreciate the time you took to look at our documentation and submit this. At this time we don't think this change improves the docs since the `keys_only` setting is [already documented](https://developer.hashicorp.com/vagrant/docs/vagrantfile/ssh_settings#config-ssh-keys_only). So, I'm going to go ahead and close this PR.

PR #12763, opened 2022-05-12
Add Uos Linux guest support

manas-rust:
Signed-off-by: wangqing <wangqing@uniontech.com>

### DIFF
--- a/plugins/guests/uos/cap/flavor.rb
+++ b/plugins/guests/uos/cap/flavor.rb
@@ -1,0 +1,23 @@
+module VagrantPlugins
+  module GuestUos
+    module Cap
+      class Flavor
+        def self.flavor(machine)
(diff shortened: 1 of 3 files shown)


manas-rust:
@topfunky

soapy1:
Hi there,
thanks for opening up a PR for adding support for your linux guests! This change set looks great. But as we move towards [Vagrant 3.0](https://www.hashicorp.com/blog/toward-vagrant-3-0) we are looking to externalize plugins that aren't critical to Vagrant's core functions. So, would you consider building this plugin as an external plugin? This should be really easy to do (you already did all the hard work of implementing the required capabilities) using the [template repo](https://github.com/soapy1/vagrant-guest-plugin-skeleton) for building guest plugins.

PR #12796, opened 2022-06-22
Sanitize Docker-Compose provider project name

Venthe:
Docker compose project name does not allow upper case characters.
By default, Vagrant uses the current working directory as project name,
and as such there can exist upper-case characters.

As an alternative, one can use `COMPOSE_PROJECT_NAME` environment variable
to override Vagrant behaviour.

See:
* https://docs.docker.com/compose/reference/envvars/#compose_project_name

### DIFF
--- a/plugins/providers/docker/driver/compose.rb
+++ b/plugins/providers/docker/driver/compose.rb
@@ -220,7 +220,7 @@ module VagrantPlugins
         def compose_execute(*cmd, **opts, &block)
           synchronized do
             execute("docker-compose", "-f", composition_path.to_s,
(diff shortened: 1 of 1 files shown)


Venthe:
Please note, I am not a Ruby developer - this is a suggestion how to solve this issue in a way that is transparent for the end-user. As such, I have no knowledge how to proceed with writing the tests for it, though the setup should be simple:


```ruby
# Pseudocode
# given
env.cwd = "Sample"
#  when
command = compose_execute(env)
# then
assert(command).switchValue("-p").isEqualTo("sample)
```

Alternatively, we could expand the documentation about the project name env variable

PR #12813, opened 2022-07-21
Update nfs_client.rb for rocky 9

sni:
Starting a rocky 9 box (having nfs shares) fails with: `No match for argument: nfs-utils-lib`.
Apparently Rocky 9 does not have a `nfs-utils-lib` but mounting nfs shares without nfs-utils-lib works fine.
So updating the nfs client install to only install `nfs-utils-lib` if the package is available. That way older rhel clones
should continue to work as well.


chrisroberts (approved):

phinze (approved):
Tested NFS is broken on main w/ `generic/rocky9` and works with this patch. Thanks for the contribution! This should be released with Vagrant 2.3 soon.

bocekm:
I've bumped into the very same issue with Oracle Linux 8.6 and CentOS Linux 8.5. Apparently these systems also don't have the `nfs-utils-lib` package available.

Issue #12817, opened 2022-07-28
OPNsense hostname support


DonGiovanni83:
### Vagrant version
2.2.19

### Host operating system
Ubuntu 20.04 LTS

### Guest operating system

OPNsense 21.7.1 and OPNsense 22.1.2

### Vagrantfile
For both OPNsense versions tested as stated above, these are the two Vagrantfiles for reproduction:
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "fbertagna/opnsense"
  config.vm.box_version = "22.1.2"
  config.vm.hostname = "notworking"
end
```
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "fbertagna/opnsense"
  config.vm.box_version = "21.7.1"
  config.vm.hostname = "notworking"
end
```
### Debug output

https://gist.github.com/DonGiovanni83/c38cc7ae566942f702c2aa048374f966

### Expected behavior
`vagrant up` should set the guest hostname according to the configured value in the Vagrantfile.

### Actual behavior

Error while setting hostname, because `/etc/rc.conf` is not present on OPNsense boxes:
```
 $ vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'fbertagna/opnsense'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'fbertagna/opnsense' version '21.7.1' is up to date...
==> default: Setting the name of the VM: opnsense-vagrant-plugin_default_1658913880425_18394
==> default: Clearing any previously set forwarded ports...
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
    default: Adapter 2: intnet
    default: Adapter 3: intnet
    default: Adapter 4: intnet
==> default: Forwarding ports...
    default: 443 (guest) => 10443 (host) (adapter 1)
    default: 22 (guest) => 2222 (host) (adapter 1)
==> default: Running 'pre-boot' VM customizations...
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: password
    default: 
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
[...]

PR #12818, opened 2022-07-28
OPNsense guest plugin

DonGiovanni83:
## Description 
Even though OPNsense is based on FreeBSD, hostname settings are not managed via the `/etc/rc.conf`. To correctly configure  hostname (and domainname)  in OPNSense the `/conf/config.xml` has to be updated with the desired values. After that the Firewall has to be reloaded with the `/usr/local/etc/rc.reload_all` command. This ensures that the changes in the config file are applyed through the whole system.

### Example
1. Vagrantfile:
    ```
      config.vm.box = "fbertagna/opnsense"
      config.vm.box_version = "22.1.2"
      config.vm.hostname = "opnsense.vagrant.test"
    end
    ```

2. `vagrant up`
3.  Firewall is configured:
Screenshot from 2022-07-28 15-40-39

See #12817

### DIFF
--- a/plugins/guests/opnsense/cap/change_host_name.rb
+++ b/plugins/guests/opnsense/cap/change_host_name.rb
@@ -1,0 +1,22 @@
+module VagrantPlugins
+  module GuestOPNsense
+    module Cap
+      class ChangeHostName
+        extend Vagrant::Util::GuestHosts::BSD
+
+        def self.change_host_name(machine, name)
+          comm = machine.communicate
+
+          unless comm.test("hostname -f | grep '^#{name}$'", sudo: false, shell: "sh")
+            command = <<-EOH.gsub(/^ {14}/, '')
+              # Set the hostname and reload firewall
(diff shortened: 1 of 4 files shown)


soapy1:
Hi there,
thanks for opening up a PR for adding support for OPNsense guests! This change set looks great. But as we move towards [Vagrant 3.0](https://www.hashicorp.com/blog/toward-vagrant-3-0) we are looking to externalize plugins that aren't critical to Vagrant's core functions. So, would you consider building this plugin as an external plugin? This should be really easy to do (you already did all the hard work of implementing the required capabilities) using the [template repo](https://github.com/soapy1/vagrant-guest-plugin-skeleton) for building guest plugins.

DonGiovanni83:
Thanks for the feedback. The guest plugin has been externalized and published to rubygems.org.
  - repo: https://github.com/puzzle/vagrant-guest-opnsense
  - gem: https://rubygems.org/gems/vagrant-guest-opnsense
  
I think with that done, this MR can be closed.

Tanks for your time.

PR #12833, opened 2022-08-08
Relax overly strict grpc-tools version requirement

dcermak:
`"~> 1.41.1"` limits us to 1.41 which is already a year old now, relaxing this to `"~> 1.41"` allows us to pick up fresher and maintained releases.

@phinze you introduced this strict pin in a5f0064a86222ba55f0e162e16581ce1af8aaa69, any specific reason for going with this release?


chrisroberts (approved):
:+1:

Issue #12872, opened 2022-08-30
Enhancement Request: Support Devuan


spedepekka:
[Devuan distro](https://www.devuan.org/) is not supported right now. It is very close to Debian, but without systemd.

PR #12873, opened 2022-08-30
Support Devuan #12872

spedepekka:
Support [Devuan](https://www.devuan.org/) #12872

I took Debian as a baseline, because it should work out of the box. I had to make minor change to ignore nonexisting network device just like somebody has done before with Ubuntu/Debian.

Devuan doesn't have systemd, so it could be an option to remove those parts of the code, but this works for me already.


spedepekka:
I won't sign the agreement.

chrisroberts:
Thank you for the contribution. Unfortunately, without the CLA we won't be able to merge this.

Cheers!

Issue #12870, opened 2022-08-29
Vagrant does not respect the `--provider` flag


chrisroberts:
When the --provider flag is specified for a vagrant up command, Vagrant should use the specified provider instead of trying to get the default provider. If the specified provider is unavailable, Vagrant should return an error. For example, currently when running vagrant up --provider hyperv on a non Windows system the result is:

```
$ ./vagrant up one --provider hyperv

  Bringing machine 'one' up with 'virtualbox' provider...
  ...
```

Vagrant should not revert to bringing up the machine with virtualbox, instead it should show an error. 

```
$ ./vagrant up one --provider hyperv   

The provider 'hyperv' that was requested to back the machine
'one' is reporting that it isn't usable on this system. The
reason is shown below:

The Hyper-V provider only works on Windows. Please try to
use another provider.
```

PR #12879, opened 2022-09-06
Provider flag

soapy1:
Fixes: https://github.com/hashicorp/vagrant/issues/12870 
Depends on: https://github.com/hashicorp/vagrant-plugin-sdk/pull/189

- Adds provider validation to the Vagrantfile core module
- Sets the target provider when provided to the Vagrantfile service
- Makes the provider in the Ruby service  always raise errors when running a `usable?` check

Now when running Vagrant:
```
# Invalid provider
 % ./vagrant up --provider sadg

! Running of task up failed unexpectedly
  
! Error: rpc error: code = Unknown desc = failed to locate plugin `sadg`

# Unusable provider
% ./vagrant up --provider hyperv

! The Hyper-V provider only works on Windows. Please try to
  use another provider.

# Default provider (in this case virtualbox)
% ./vagrant up                  

  Bringing machine 'one' up with 'virtualbox' provider...
```

### DIFF
--- a/internal/core/project.go
+++ b/internal/core/project.go
@@ -392,10 +392,13 @@ func (p *Project) DefaultProvider(opts *core.DefaultProviderOptions) (string, er
 		if opts.CheckUsable {
 			logger.Debug("Checking usable on provider", "provider", pp.Name)
 			pluginImpl := plug.Plugin.(core.Provider)
-			usable, _ := pluginImpl.Usable()
+			usable, err := pluginImpl.Usable()
 			if !usable {
 				logger.Debug("Skipping unusable provider", "provider", pp.Name)
 				continue
+			}
+			if err != nil {
+				return "", err
 			}
 		}
 


chrisroberts (dismissed):

internal/core/project.go
```diff
@@ -392,10 +392,7 @@ func (p *Project) DefaultProvider(opts *core.DefaultProviderOptions) (string, er
 		if opts.CheckUsable {
 			logger.Debug("Checking usable on provider", "provider", pp.Name)
 			pluginImpl := plug.Plugin.(core.Provider)
-			usable, err := pluginImpl.Usable()
-			if err != nil {
-				return "", err
-			}
```
chrisroberts, line 398: I don't think we want to ignore this error here given that it will indicate there is something wrong with the plugin and we'll want to bail and report the error if that is the case.

chrisroberts (changes requested):
Looks great, just a note about the error handing.

chrisroberts (approved):

PR #12890, opened 2022-09-13
Add new config

kaitlincart:
This will link to the forum to encourage users to post questions there instead of GitHub. This was borrowed from Nomad.


chrisroberts (approved):
:+1:

Issue #12830, opened 2022-08-03
Use comment syntax in Github issue templates in order to remove clutter.


Flightkick:
Not much of a problem but FYI.
In Github issue templates, text like:

```
Please note that the Vagrant issue tracker is in priority reserved for bug reports and enhancements. For general usage questions, please use
HashiCorp Discuss: https://discuss.hashicorp.com/c/vagrant/24 Thank you!

When submitting a bug report, please provide the minimal configuration and required information necessary to reliably reproduce the issue. It
should include a basic Vagrantfile that only contains settings to reproduce the described behavior.

**Tip:** Before submitting your issue, don't hesitate to remove the above introductory text, possible empty sections (e.g. References), and this tip.
```

Can be surrounded with comment syntax, e.g.:
```
<!---
lorem ipsum
--->
```

So issue reporters won't have to remove clutter in order to keep things [clean.](https://github.com/hashicorp/vagrant/issues?q=is%3Aissue+%22Before+submitting+your+issue%2C+don%27t+hesitate+to+remove+the+above+introductory+text%22+)

PR #12896, opened 2022-09-14
Bug Report Template Updates

kaitlincart:
My goal was to make the template easier to complete. I again borrowed some ideas from Nomad. Hopefully, bug reporst will now be automatically labeled. 

Fixes: #12830

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -16,33 +16,45 @@ Thank you!
 
 ### Debug output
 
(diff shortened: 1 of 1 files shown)


chrisroberts (changes requested):
This looks great. One thing that would be nice to include is wrapping all the instruction content like:

```
### Expected behavior

<!--
What should have happened
-->
```
to keep them from being persisted in the new issue if the user doesn't remove them.

chrisroberts (approved):

soapy1 (approved):

PR #12912, opened 2022-09-21
Devdot updates

kaitlincart:
At @tunzor's request, I've removed a link and updated some headings so that they will appear in the side nav.


soapy1 (approved):

chrisroberts (approved):
This change is going to make the navigation in the docs on vagrantup a little weird once you're in the vagrantcloud docs, but I guess that'll be alright for the transition time.

Issue #12864, opened 2022-08-29
Uninstalling legacy plugins raises errors


chrisroberts:
```
% ./vagrant plugin install vagrant-vbguest

» Running plugin install... 
  Installing the 'vagrant-vbguest' plugin. This can take a few minutes...
Installed the plugin 'vagrant-vbguest (0.30.0)'!
(vagrant-agogo) sophia@x86_64-apple-darwin13 [13:39:26] [~/project/vagrant-ruby] [main]
-> % ./vagrant plugin list

» Running plugin list... 
  vagrant-vbguest (0.30.0, global)
(vagrant-agogo) sophia@x86_64-apple-darwin13 [13:41:55] [~/project/vagrant-ruby] [main]
-> % ./vagrant plugin uninstall vagrant-vbguest

» Running plugin uninstall... 
  Uninstalling the 'vagrant-vbguest' plugin...
! Running of task plugin uninstall failed unexpectedly
  
! Error: rpc error: code = Unknown desc = No such file or directory @ rb_check_realpath_internal - /Users/sophia/Library/Application
Support/vagrant/default/project/vagrant-ruby/plugins/gems/2.7.2
  /Users/sophia/project/vagrant-ruby/lib/vagrant/bundler.rb:441:in `realpath'
  /Users/sophia/project/vagrant-ruby/lib/vagrant/bundler.rb:441:in `realpath'
  /Users/sophia/project/vagrant-ruby/lib/vagrant/bundler.rb:441:in `block in clean'
  /Users/sophia/project/vagrant-ruby/lib/vagrant/bundler.rb:440:in `delete_if'
  /Users/sophia/project/vagrant-ruby/lib/vagrant/bundler.rb:440:in `clean'
  /Users/sophia/project/vagrant-ruby/lib/vagrant/plugin/manager.rb:229:in `uninstall_plugin'
  /Users/sophia/project/vagrant-ruby/plugins/commands/plugin/action/uninstall_plugin.rb:18:in `call'
  /Users/sophia/project/vagrant-ruby/lib/vagrant/action/warden.rb:48:in `call'
  /Users/sophia/project/vagrant-ruby/plugins/commands/plugin/action/plugin_exists_check.rb:20:in `call'
  /Users/sophia/project/vagrant-ruby/lib/vagrant/action/warden.rb:48:in `call'
  /Users/sophia/project/vagrant-ruby/lib/vagrant/action/builder.rb:180:in `call'
[...]

PR #12920, opened 2022-09-26
Vagrant legacy plugin uninstall

soapy1:
This PR aims to allow vagrant-go users to uninstall legacy Ruby plugins.
* Allow plugins to be loaded by project

Fixes: https://github.com/hashicorp/vagrant/issues/12864


internal/runner/runner.go
```diff
@@ -156,6 +156,13 @@ func New(opts ...Option) (*Runner, error) {
 	return runner, nil
 }
 
+func (r *Runner) LoadLocalProjectPlugins(path string) error {
+	// TODO: should this check if the runner is local?
+	err := r.plugins.LoadLocalLegacyPlugins(
```
soapy1, line 161: I'm not too sure about this approach. If the runner is local then it should be able to load local project plugins. Not sure how to resolve the situation where the runner is remote?

PR #12948, opened 2022-10-11
Enabling ext4 as supported for synced folders within WSL

ja0n:
Since WSL2 allows for mounting proper Linux file systems, this PR is allowing shared folders from ext4 mounts
Inspired by #12056


ja0n:
I also noticed that the tests for VMConfig were failing and fixed it, let me know if there's any problem or if it's better to submit it in another PR

PR #12952, opened 2022-10-13
introduce support for VirtualBox 7.0

scoter-oracle:


scoter-oracle:
Opened to just sign CLA

PR #12950, opened 2022-10-11
Add Virtualbox 7.0 support

T2L:
Just incorporating awesome job by @scoter-oracle. See https://blogs.oracle.com/scoter/post/getting-vagrant-23-working-with-oracle-vm-virtualbox-70-beta

Fixes #12947

PR #12953, opened 2022-10-14
Add support for VirtualBox 7.0

### LINKED ISSUE #12947: Vagrant not working with Virtualbox 7.0
<!--
Now with Virtualbox 7.0 released [ https://www.virtualbox.org/ ], Vagrant 2.3.1 does *not* recognize 7.0 and does not work. [Using Windows 11 Pro]
It says

```
C:\b>vagrant up
No usable default provider could be found for your system.

Vagrant relies on interactions with 3rd party systems, known as
"providers", to provide Vagrant with resources to run development
environments. Examples are VirtualBox, VMware, Hyper-V.

The easiest solution to this message is to install VirtualBox, which
is available for free on all major platforms.

If you believe you already have a provider available, make sure it
is properly installed and configured. You can see more details about
why a particular provider isn't working by forcing usage with
`vagrant up --provider=PROVIDER`, which should give you a more specific
error message for that particular provider.
```

I tried overriding it to run but it didn't work
```
C:\b>vagrant up --provider=virtualbox
The provider 'virtualbox' that was requested to back the machine
'default' is reporting that it isn't usable on this system. The
reason is shown below:

Vagrant has detected that you have a version of VirtualBox installed
that is not supported by this version of Vagrant. Please install one of
the supported versions listed below to use Vagrant:

4.0, 4.1, 4.2, 4.3, 5.0, 5.1, 5.2, 6.0, 6.1

A Vagrant update may also be available that adds support for the version
you specified. Please check www.vagrantup.com/downloads.html to download
the latest version.
```
### Expected behavior

It should run as normal

#### Vagrant version

```
Vargrant 2.3.1
```
#### Host operating system

Windows 11 Pro 22H2

#### Guest operating system

Rocky Linux 9

#### Steps to reproduce

1. Upgrade from Virtualbox 6.xx to Virtualbox 7.xx
2. Run vagrant up

chrisroberts:
Adds support for VirtualBox 7.0. The 7.0 driver includes a
port forward fix to account for missing port forward information
in the `showvminfo` output when more than one port forward has
been defined on a vm. This is been isolated to just the 7.0.0
VirtualBox release. It will revert to the original behavior on
the VirtualBox 7.0.1 release.

The VirtualBox extension pack was required to prevent the
`VBoxManage` executable from hanging on multiple executions. The
underlying cause of this has still not been determined and is
being investigated.

Fixes #12947 #12950

### DIFF
--- a/plugins/providers/virtualbox/driver/version_7_0.rb
+++ b/plugins/providers/virtualbox/driver/version_7_0.rb
@@ -1,0 +1,67 @@
+require "rexml"
+require File.expand_path("../version_6_1", __FILE__)
+
+module VagrantPlugins
+  module ProviderVirtualBox
+    module Driver
+      # Driver for VirtualBox 7.0.x
+      class Version_7_0 < Version_6_1
+        def initialize(uuid)
+          super
+
+          @logger = Log4r::Logger.new("vagrant::provider::virtualbox_7_0")
+        end
+
+        # The initial VirtualBox 7.0 release has an issue with displaying port
+        # forward information. When a single port forward is defined, the forwarding
+        # information can be found in the `showvminfo` output. Once more than a
+        # single port forward is defined, no forwarding information is provided
+        # in the `showvminfo` output. To work around this we grab the VM configuration
+        # file from the `showvminfo` output and extract the port forward information
+        # from there instead.
+        def read_forwarded_ports(uuid=nil, active_only=false)
+          @version ||= Meta.new.version
+
+          # Only use this override for the 7.0.0 release. If it is still broken
+          # on the 7.0.1 release we can modify the version check.
+          return super if @version != "7.0.0"
+
+          uuid ||= @uuid
+
+          @logger.debug("read_forward_ports: uuid=#{uuid} active_only=#{active_only}")
+
+          results = []
+
+          info = execute("showvminfo", uuid, "--machinereadable", retryable: true)
+          result = info.match(/CfgFile="(?<path>.+?)"/)
+          if result.nil?
+            raise Vagrant::Errors::VirtualBoxConfigNotFound,
+                  uuid: uuid
+          end
+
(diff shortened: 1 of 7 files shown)


soapy1 (approved):

plugins/providers/virtualbox/driver/version_7_0.rb
```diff
@@ -0,0 +1,67 @@
...
+          return super if @version != "7.0.0"
+
+          uuid ||= @uuid
+
+          @logger.debug("read_forward_ports: uuid=#{uuid} active_only=#{active_only}")
+
+          results = []
+
+          info = execute("showvminfo", uuid, "--machinereadable", retryable: true)
+          result = info.match(/CfgFile="(?<path>.+?)"/)
```
g0t4, line 36: Virtualbox 7 added support for full encryption of a VM including its config files. Just a heads up... I'm not sure if that's something you can pull off with a VM created with vagrant, I haven't tried this myself.

g0t4: I wouldn't hold anything up for this possibility! Just something to look into if issues arise later.

chrisroberts: That's a great point. This fix is targeted at the 7.0.0 release specifically with the assumption it won't be needed in the next release. I'll throw something together to check if the file is encrypted/unreadable and customize the error with a warning about encrypted VMs. Should be able to get it slipped in before we cut the release in the morning. Thanks!

chrisroberts: Just a note on this. I ended up not including any extras for encrypted VMs as I found it not functional on all the platforms I tested on. There's also the issue of encrypted storage devices which complicates the execution flow with password requests and such. So I've decided to punt that handling into the 2.3.3 release so we weren't holding up getting support for vbox 7 out.

Issue #12967, opened 2022-10-21
Vagrant 2.3.2 and Virtualbox 7.0.2 - Private Network Interface


Yagoor:
### Debug output

https://gist.github.com/Yagoor/66ca6861f57e2c2b09f34c0821bbb0e7

### Expected behavior

The machine is started with a private network interface

### Actual behavior

The vagrant up fails

### Reproduction information 

Reproducible with the Vagrantfile and vagrant up

#### Provider version

Virtual box: Version 7.0.2 r154219 (Qt5.15.2)

#### Vagrant version

Vagrant 2.3.2

#### Host operating system

macOS 11.7

#### Guest operating system

bento/ubuntu-20.04

#### Steps to reproduce

1. Create Vagrantfiles as follows
2. vagrant up

#### Vagrantfile

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "bento/ubuntu-20.04"

  config.vm.network "private_network", ip: "192.168.50.4"
end
```

identifysun:
I have same question,

# Vagrant and Virtualbox Version

```shell
$ vagrant --version
Vagrant 2.3.2

$ vboxmanage --version
7.0.2r154219
```

I guess running the following command is caused by a failure

```shell
$ VBoxManage hostonlyif create
0%...NS_ERROR_FAILURE
VBoxManage: error: Failed to create the host-only adapter
VBoxManage: error: VBoxNetAdpCtl: Error while adding new interface: failed to open /dev/vboxnetctl: No such file or directory
VBoxManage: error: Details: code NS_ERROR_FAILURE (0x80004005), component HostNetworkInterfaceWrap, interface IHostNetworkInterface
VBoxManage: error: Context: "RTEXITCODE handleCreate(HandlerArg *)" at line 105 of file VBoxManageHostonly.cpp
```

Refer to the workaround on the internet, but I can't find the following file or command

-  /Library/StartupItems/VirtualBox/VirtualBox restart
- /Library/Application\ Support/VirtualBox/LaunchDaemons/VirtualBoxStartup.sh restart

identifysun:
I tried reinstalling Vagrant and Virtualbox, but it still doesn't fix

```shell
$ brew reinstall virtualbox vagrant
```

jav-12:
Same issue here. I downgraded VirtualBox to v6.1 problem is gone. For now, I will keep it this way till 7 gets fixed.

f3l1x:
Same issue here. 

```
[...]

PR #12974, opened 2022-10-25
Update version_7_0.rb hostonlyifs to hostonlynets

### LINKED ISSUE #12959: VM with Host-Only Networks setting does not boot in VirtualBox7
### Debug output

https://gist.github.com/KentarouTakeda/93b575bcd048ce41a5c88f3499334593

The end of the log contains the following message: 

```
Command: ["startvm", "a1a2b62c-e192-4ba0-ba2d-1b7c6fdf965b", "--type", "headless"]

Stderr: VBoxManage: error: Host-only adapters are no longer supported!
VBoxManage: error: For your convenience a host-only network named 'Legacy vboxnet11 Network' has been created with network mask '255.255.255.0' and IP address range '192.168.33.1' - '192.168.33.254'.
VBoxManage: error: To fix this problem, switch to 'Host-only Network' attachment type in the VM settings.
VBoxManage: error:  (VERR_NOT_FOUND)
VBoxManage: error: Details: code NS_ERROR_FAILURE (0x80004005), component ConsoleWrap, interface IConsole
```

### Expected behavior

Running `vagrant up` should create and start a VM.

### Actual behavior

The VM is created and VirtualBox identifies it, but fails to initialize after boot and exits with an error. And vagrant is not aware of that VM.

#### ViirtualBox version

**VirtualBox 7.0.0** r153978 (Qt5.15.2)

#### Vagrant version

Vagrant 2.3.2

#### Host operating system

macOS 12.6

#### Guest operating system

[Amazon Linux 2](https://app.vagrantup.com/gbailey/boxes/amzn2)

#### Steps to reproduce

1. Create Vagrantfiles as follows
2. `vagrant up`

#### Vagrantfile

```ruby
  config.vm.box = "gbailey/amzn2"

  config.vm.network "private_network", ip: "192.168.33.10"
end
```

MacCracken:
attempt to convert hostonlyif command to hostonlynets as host-only adapter is depreciated in 7.x; attempting to resolve #12959 & #12967

### DIFF
--- a/plugins/providers/virtualbox/driver/version_7_0.rb
+++ b/plugins/providers/virtualbox/driver/version_7_0.rb
@@ -61,6 +61,110 @@ module VagrantPlugins
 
           results
         end
+        
+        # The initial VirtualBox 7.0 release with depreciation to Host Only Adapter.
+        # Converting to Host Only Network
+        def create_host_only_network(options)
+          # Create the interface
+          execute("hostonlynets", "create", retryable: true) =~ /^Interface '(.+?)' was successfully created$/
+          name = $1.to_s
+
+          # Get the IP so we can determine v4 vs v6
+          ip = IPAddr.new(options[:adapter_ip])
+
+          # Configure
+          if ip.ipv4?
+            execute("hostonlynets", "ipconfig", name,
+                    "--ip", options[:adapter_ip],
+                    "--netmask", options[:netmask],
+                    retryable: true)
+          elsif ip.ipv6?
+            execute("hostonlynets", "ipconfig", name,
+                    "--ipv6", options[:adapter_ip],
+                    "--netmasklengthv6", options[:netmask].to_s,
+                    retryable: true)
+          else
+            raise "BUG: Unknown IP type: #{ip.inspect}"
+          end
+
+          # Return the details
+          return {
+            name: name,
+            ip:   options[:adapter_ip],
+            netmask: options[:netmask],
+            dhcp: nil
+          }
+        end
+        
+        def delete_unused_host_only_networks
+          networks = []
+          execute("list", "hostonlynets", retryable: true).split("\n").each do |line|
+            networks << $1.to_s if line =~ /^Name:\s+(.+?)$/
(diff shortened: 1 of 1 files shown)


plugins/providers/virtualbox/driver/version_7_0.rb
```diff
@@ -61,6 +61,110 @@ def read_forwarded_ports(uuid=nil, active_only=false)
...
+        end
+        
+        def read_host_only_interfaces
+          execute("list", "hostonlynets", retryable: true).split("\n\n").collect do |block|
+            info = {}
+
+            block.split("\n").each do |line|
+              if line =~ /^Name:\s+(.+?)$/
+                info[:name] = $1.to_s
+              elsif line =~ /^IPAddress:\s+(.+?)$/
```
LukeCarrier, line 141: This condition will never match, since this line doesn't exist in the output, at least for beta 3:

```console
❯ VBoxManage list hostonlynets
Name:            Legacy vboxnet0 Network
GUID:            50c1b470-b0e7-4900-bd56-26d9c4f9f5e2

State:           Enabled
NetworkMask:     255.255.255.0
LowerIP:         192.168.56.10
UpperIP:         192.168.56.254
VBoxNetworkName: hostonly-Legacy vboxnet0 Network
```

MacCracken: was hoping it would be as simple as changing out the function from the previous command; need to dig a little deeper to get the network behavior over single ip.  Thanks for pointing that out

LukeCarrier: Yeah I'm in a similar boat and need to dig into this further. I think the semantics have changed from listing network interfaces to listing networks, so we might need to make some structural changes here (maybe similar to the way `virtualbox__intnet` configures internal-only networks).

The VirtualBox driver feels so version-specific and fragile :/

MacCracken:
as stated above this only just tried to replace the command.  switch from adapter to network is needed.

PR #12980, opened 2022-11-01
Fix CI

dcermak:
Fixes due to the Ubuntu version bump introduced in https://github.com/hashicorp/vagrant/pull/12977


chrisroberts:
Thanks @dcermak !

PR #13017, opened 2022-12-05
Add workaround for Debian 10 and older

### LINKED ISSUE #13016: ansible_local provisioner tries to install non-existed package
The `ansible_local` provisioner fails on debian versions prior to bullseye (11) due to the attempted installation of the package `python-dev-is-python3`. However this package does not exist in the official repositories for versions prior to bullseye.

### Expected behavior

Errorless setup of pip and ansible and a subsequent run of ansible

### Actual behavior

`vagrant up` fails with the following error.
```
==> default: Running provisioner: ansible_local...
    default: Installing Ansible...

DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --option "Dpkg::Options::=--force-confold" build-essential curl git libssl-dev libffi-dev python-dev-is-python3

Stdout from the command:



Stderr from the command:

E: Unable to locate package python-dev-is-python3
```

#### Vagrant version

vagrant 2.3.3

#### Host operating system

archlinux

#### Guest operating system

Debian 10 (image: debian/contrib-buster64)

#### Steps to reproduce

1. Setup Vagrantfile and playbook.yaml as shown below
2. run `vagrant up`

#### Vagrantfile

```ruby
  config.vm.box = "debian/contrib-buster64"
  config.vm.provision :shell, inline: <<-EOF
    sudo apt-get update
    sudo apt-get install -y python3-distutils # The pip install itself will fail is this package is not installed
  EOF
  config.vm.provision :ansible_local do |ansible|
    ansible.playbook = "/vagrant/playbook.yaml"
    ansible.install_mode = "pip"
    ansible.version = "latest"
    ansible.pip_install_cmd = "curl -s https://bootstrap.pypa.io/get-pip.py | sudo python3"
  end
end

```

#### Ansible Playbook
Just an empty playbook, so ansible does not fail itself.
```yaml
---

- hosts: all
  tasks: []

```

msltwtf:
This is the somewhat native and probably not the most elegant solution to fix #13016.
But using the guide [Installing Vagrant from Source](https://developer.hashicorp.com/vagrant/docs/installation/source), I tested my changes and vagrant is able to install pip and ansible without any errors and run ansible subsequently.

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/debian/ansible_install.rb
+++ b/plugins/provisioners/ansible/cap/guest/debian/ansible_install.rb
@@ -37,12 +37,11 @@ INLINE_CRIPT
 
             def self.pip_setup(machine, pip_install_cmd = "")
               machine.communicate.sudo "apt-get update -y -qq"
-              debian_version = machine.communicate.execute "lsb_release -rs", error_check: false
-              if debian_version > 10
-                machine.communicate.sudo "DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --option \"Dpkg::Options::=--force-confold\" build-essential curl git libssl-dev libffi-dev python-dev-is-python3"
-              else
-                machine.communicate.sudo "DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --option \"Dpkg::Options::=--force-confold\" build-essential curl git libssl-dev libffi-dev"
+              python_dev_pkg = "python-dev"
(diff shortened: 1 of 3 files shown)


chrisroberts:
Hi @msltwtf. Thanks for this PR! I've made a small adjustment that runs a test for the package prior to install instead of basing it off the version to make it a bit more portable.

chrisroberts (approved):

PR #13047, opened 2023-01-05
Added config.ssh.retries and config.ssh.retry_interval to allow for SSH retries

### LINKED ISSUE #10776: add sleep period before retrying the ssh connection

nmaludy:
Closes #10776 

When provisioning RHEL VMs, i noticed that the hosts become pingable before the SSH daemon is available. Watching the vagrant logs with `--debug` i noticed that the SSH connection was not being retried properly. Digging into the code, i also discovered that even with a properly setup retry counter, the retries were happening as fast as possible, not allowing time for the SSH daemon to become available.

To mitigate these problems, i added `config.ssh.retries` to allow for multiple SSH connection retry attempts before failing (default=5). I also added `config.ssh.retry_interval` to insert a sleep time (seconds) between each connection attempt, allowing time for the SSH daemon to come up.

This is my first PR to Vagrant. I tried to do the right things, but i'm sure i'm missing some stuff. If you all could help guide me through the contribution process, i would very much appreciate it!

### DIFF
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -400,7 +400,16 @@ module VagrantPlugins
         raise Vagrant::Errors::SSHNotReady if ssh_info.nil?
 
         # Default some options
-        opts[:retries] = 5 if !opts.key?(:retries)
+        opts[:retries] = if ssh_info.key?(:retries)
+                           ssh_info[:retries]
+                         elsif !opts.key?(:retries)
+                           5
+                         end
+        opts[:retry_interval] = if ssh_info.key?(:retry_interval)
+                               ssh_info[:retry_interval]
+                             else
+                               10
+                             end
 
         # Set some valid auth methods. We disable the auth methods that
         # we're not using if we don't have the right auth info.
@@ -429,7 +438,7 @@ module VagrantPlugins
           timeout = 60
 
           @logger.info("Attempting SSH connection...")
-          connection = retryable(tries: opts[:retries], on: SSH_RETRY_EXCEPTIONS) do
+          connection = retryable(tries: opts[:retries], on: SSH_RETRY_EXCEPTIONS, sleep: opts[:retry_interval]) do
             Timeout.timeout(timeout) do
               begin
                 # This logger will get the Net-SSH log data for us.
--- a/plugins/kernel_v2/config/ssh_connect.rb
+++ b/plugins/kernel_v2/config/ssh_connect.rb
@@ -2,6 +2,8 @@ module VagrantPlugins
   module Kernel_V2
     class SSHConnectConfig < Vagrant.plugin("2", :config)
       DEFAULT_SSH_CONNECT_TIMEOUT = 15
+      DEFAULT_SSH_RETRIES         = 5
+      DEFAULT_SSH_RETRY_INTERVAL  = 10
 
       attr_accessor :host
       attr_accessor :port
@@ -18,6 +20,8 @@ module VagrantPlugins
       attr_accessor :dsa_authentication
       attr_accessor :extra_args
       attr_accessor :remote_user
+      attr_accessor :retries
+      attr_accessor :retry_interval
 
       def initialize
         @host             = UNSET_VALUE
@@ -35,6 +39,8 @@ module VagrantPlugins
         @dsa_authentication = UNSET_VALUE
         @extra_args       = UNSET_VALUE
         @remote_user      = UNSET_VALUE
+        @retries          = UNSET_VALUE
+        @retry_interval   = UNSET_VALUE
       end
 
       def finalize!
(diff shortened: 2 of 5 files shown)


plugins/communicators/ssh/communicator.rb
```diff
@@ -400,7 +400,16 @@ def connect(**opts)
         raise Vagrant::Errors::SSHNotReady if ssh_info.nil?
 
         # Default some options
-        opts[:retries] = 5 if !opts.key?(:retries)
+        opts[:retries] = if ssh_info.key?(:retries)
```
soapy1, line 403: the `ssh_info` should probably already have the default value for the retries (coming from the config). So, there is no need to default this again. Instead, this can be  `opts[:retries] = ssh_info[:retries] if !opts.key?(:retries)` (same for the retry_interval below). This also implies that the internal caller of this method can override the user defined config. 

This also means that the [`ssh_info` method in the `Machine`](https://github.com/hashicorp/vagrant/blob/620df8fd0fdb74d087f051feba988e4131248f3b/lib/vagrant/machine.rb#L448) class needs to be updated to account for this new config.

plugins/kernel_v2/config/ssh_connect.rb
```diff
@@ -134,6 +142,26 @@ def validate(machine)
             given: @connect_timeout.to_s)
         end
 
+        if !@retries.is_a?(Integer)
```
soapy1, line 145: Take a look at the [`validate` method for the winrm communicator](https://github.com/hashicorp/vagrant/blob/main/plugins/communicators/winrm/config.rb#L51). It has some better validations steps. For example, it's ok to not have any retries or to have a retry interval between 0 and 1.

plugins/kernel_v2/config/ssh_connect.rb
```diff
@@ -2,6 +2,8 @@ module VagrantPlugins
   module Kernel_V2
     class SSHConnectConfig < Vagrant.plugin("2", :config)
       DEFAULT_SSH_CONNECT_TIMEOUT = 15
+      DEFAULT_SSH_RETRIES         = 5
+      DEFAULT_SSH_RETRY_INTERVAL  = 10
```
soapy1, line 6: This number can probably come down some. The [winrm communicator](https://github.com/hashicorp/vagrant/blob/main/plugins/communicators/winrm/config.rb#L43) uses 2.

soapy1 (changes requested):
Thanks for the contribution!

It looks like the `winrm` communicator plugin has similar functionality with [max_tries](https://developer.hashicorp.com/vagrant/docs/vagrantfile/winrm_settings#config-winrm-max_tries) and [retry-delay](https://developer.hashicorp.com/vagrant/docs/vagrantfile/winrm_settings#config-winrm-retry_delay). It would be nice to have consistency between the names of these options, how they are implemented, and their default values.

This also needs unit tests. They are in the [tests subdir](https://github.com/hashicorp/vagrant/tree/main/test).

avoidik:
Worth noting that the current implementation has no retries at all due to this line (it is always 1), probably you'll be keen to address this quite old observation in your work.

https://github.com/hashicorp/vagrant/blob/main/plugins/communicators/ssh/communicator.rb#L87

ashleemboyer:
Hi @nmaludy! I've just merged in https://github.com/hashicorp/vagrant/pull/13048, which modified 120 MDX files and added a new workflow that would be good to have run on this PR. Since this is from a fork, I'm not sure I have permissions to rebase on `main` for you, so I wanted to send this quick message to let you know to update with `main`. Thanks!

avoidik:
hello all,

could you please suggest what can we do to finally close this issue?

Issue #13008, opened 2022-11-24
Vagrant not using password even if password authentication is set


ssasso:
Hello, I have the following Vagrantfile configuration (relevant parts only):

```
Vagrant.configure("2") do |config|

  config.vm.define "vy1" do |vy1|
    vy1.vm.provider :libvirt do |domain|
      domain.management_network_mac = "08-4F-A9-00-00-01"
      domain.qemu_use_session = false
    end
    vy1.vm.box = "vyos/current"
    vy1.vm.guest = :linux
    vy1.vm.synced_folder ".", "/vagrant", id: "vagrant-root", disabled: true

    vy1.vm.provider :libvirt do |domain|
      domain.cpus = 2
      domain.memory = 1024
    end

    vy1.ssh.username="vyos"
    vy1.ssh.password="vyos"
    vy1.ssh.keys_only=false
    vy1.ssh.insert_key=false
    vy1.ssh.shell="bash"

  end
end
```

When running `vagrant up`, vagrant is stuck at: 
```
vy1: Warning: Authentication failure. Retrying...
```

As you can see from the debug output below, it seems vagrant does not even try to use password authentication, even if instructed to do so.

The password is right, since if I perform `vagrant ssh vy1` I am able to login after putting it on the prompt.

### Debug output

```
 INFO wait_till_up: Got IP address 192.168.121.250
 INFO wait_till_up: Time for getting IP: 63.879722595214844
 INFO warden: Calling IN action: #<Vagrant::Action::Builtin::WaitForCommunicator:0x00007f1148b2d620>
 INFO interface: output: Waiting for machine to boot. This may take a few minutes...
 INFO interface: output: ==> vy1: Waiting for machine to boot. This may take a few minutes...
==> vy1: Waiting for machine to boot. This may take a few minutes...
 INFO interface: detail: SSH address: 192.168.121.250:22
 INFO interface: detail:     vy1: SSH address: 192.168.121.250:22
    vy1: SSH address: 192.168.121.250:22
 INFO interface: detail: SSH username: vyos
 INFO interface: detail:     vy1: SSH username: vyos
    vy1: SSH username: vyos
 INFO interface: detail: SSH auth method: password
 INFO interface: detail:     vy1: SSH auth method: password
    vy1: SSH auth method: password
 INFO ssh: Attempting SSH connection...
 INFO ssh: Attempting to connect to SSH...
 INFO ssh:   - Host: 192.168.121.250
 INFO ssh:   - Port: 22
[...]

PR #13052, opened 2023-01-09
Update connection settings when using a password to connect ssh

soapy1:
When connecting over ssh using net-ssh use the non_interactive argument must be set when authenticating with a password.

The "keyboard-interactive" options may be used as a default auth method, like in the net-ssh defaults (ref: https://github.com/net-ssh/net-ssh/blob/8a176a6ea0db1b59a21834df806a257a0b76e943/lib/net/ssh/config.rb#L52)

fixes https://github.com/hashicorp/vagrant/issues/13008

### DIFF
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -404,7 +404,7 @@ module VagrantPlugins
 
         # Set some valid auth methods. We disable the auth methods that
         # we're not using if we don't have the right auth info.
-        auth_methods = ["none", "hostbased"]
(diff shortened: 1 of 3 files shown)


chrisroberts (dismissed):

chrisroberts (approved):

chrisroberts:
@legal90 Hi! Sorry for the delay on circling back around to this. I'm walking through the behavior here and the behavior described in the original issue it was meant to address. The resulting behavior does not appear to be ideal and I'll be making some modifications that will be up in a new PR shortly.

chrisroberts:
PR here: #13194

PR #13054, opened 2023-01-10
Fix installing Ansible provisioner with version and pip

### LINKED ISSUE #13010: ansible_local provisioner unable to install specific version of Ansible with pip install_mode
The `ansible_local` provisioner is able to install a specific version of Ansible using the `version` attribute. You can combine this with the `install_mode` attribute, specifically using `pip` to install and manage Ansible versions.

Ansible switched their versioning logic to rather display the version of `ansible-core` instead of the actual `ansible` package (i.e. when running `ansible --version` you get `[core 2.x.x.x]` instead of the version of the `ansible` `pip` package, which is higher than than the `ansible-core` package). This leads to the issue of Ansible getting installed correctly when specifying the `version` but the version check itself failing because it detects the version of `ansible-core` instead.

Unfortunately, you also cannot use the version of `ansible-core` because the installation command line that's executed by Vagrant is `pip install --upgrade ansible==<version>`, meaning it's installing the `ansible` package, not `ansible-core`, so using the `ansible-core` version fails since that's not a compatible semantic version of the package `ansible`.

A way out of this would be to use `importlib.metadata` for checking on the Ansible version if `install_mode: 'pip'` is specified:

```console
$ python -c "import importlib.metadata; print(importlib.metadata.version('ansible'))"
7.0.0
```

or `pip` itself:

```console
$ pip show ansible | awk '/Version/ { print $2 }'
```

Either way, there needs to be a remedy for the mismatch in behavior between Vagrant trying to install the `ansible` metapackage and then checking against a version of `ansible-core`.

### Expected behavior

The package `ansible` with the requested version should have been installed using `pip`.

### Actual behavior

The package `ansible` was installed with the requested version but the version check fails because it's looking at `ansible --version` and finds the version string `[core 2.x.x.x]` for `ansible-core`.

#### Vagrant version

2.3.3

#### Host operating system

Ubuntu Kinetic Kudu (22.10)

#### Guest operating system

- Ubuntu Jammy Jellyfish (22.04)
- Ubuntu Focal Fossa (20.04)
- Rocky Linux 8
- Rocky Linux 9

#### Steps to reproduce

Simply run `vagrant up` on the example. The `playbook.yml` just needs to be an empty file in the same directory as the `Vagrantfile`.

#### Vagrantfile

```ruby
  config.vm.define 'jammy' do |bare|
    bare.vm.box = 'ubuntu/jammy64'
    bare.vm.network 'private_network', type: 'dhcp'

    bare.vm.provision 'ansible_local' do |ansible|
      ansible.version = '7.0.0'
      ansible.playbook = 'playbook.yml'
      ansible.install_mode = 'pip'
    end
  end
end
```

Output from a Vagrant run:

```console
Bringing machine 'jammy' up with 'virtualbox' provider...
==> jammy: Checking if box 'ubuntu/jammy64' version '20221124.0.0' is up to date...
==> jammy: Running provisioner: ansible_local...
    jammy: Installing Ansible...
    jammy: Installing pip... (for Ansible installation)
The requested Ansible version (7.0.0) was not found on the guest.
Please check the Ansible installation on your Vagrant guest system (currently: [core 2.14.0]),
or adapt the provisioner `version` option in your Vagrantfile.
See https://docs.vagrantup.com/v2/provisioning/ansible_common.html#version
for more information.
```

moritzheiber:
Fixes #13010 

I'm using the rather "ugly" workaround with calling `importlib.metadata` via `python` and not any of the other alternatives mentioned in the issue because it's the most compatible. I cannot really expect people to have `pip` installed when checking for the initial version of `ansible`, and `ansible-community` wasn't a part of the standard distribution until fairly recently (6'ish?). To make sure we opt for the broadest amount of compatibility with Ansible versions, `pip` and Python I've chosen the lesser evil.

### DIFF
--- a/plugins/provisioners/ansible/cap/guest/posix/ansible_installed.rb
+++ b/plugins/provisioners/ansible/cap/guest/posix/ansible_installed.rb
@@ -11,7 +11,7 @@ module VagrantPlugins
               command = 'test -x "$(command -v ansible)"'
 
               unless version.empty?
-                command << "&& [[ $(python -c \"import importlib.metadata; print(importlib.metadata.version('ansible'))\") == \"#{version}\" ]]"
+                command << "&& [[ $(python3 -c \"import importlib.metadata; print(importlib.metadata.version('ansible'))\") == \"#{version}\" ]]"
               end
 
               machine.communicate.test command, sudo: false
--- a/plugins/provisioners/ansible/cap/guest/posix/ansible_installed.rb
+++ b/plugins/provisioners/ansible/cap/guest/posix/ansible_installed.rb
@@ -10,8 +10,8 @@ module VagrantPlugins
             def self.ansible_installed(machine, version)
               command = 'test -x "$(command -v ansible)"'
 
-              if !version.empty?
-                command << "&& ansible --version | grep 'ansible #{version}'"
+              unless version.empty?
+                command << "&& [[ $(python -c \"import importlib.metadata; print(importlib.metadata.version('ansible'))\") == \"#{version}\" ]]"
               end
 
               machine.communicate.test command, sudo: false
--- a/plugins/provisioners/ansible/provisioner/guest.rb
+++ b/plugins/provisioners/ansible/provisioner/guest.rb
@@ -75,7 +75,7 @@ module VagrantPlugins
           raw_output = ""
 
           result = @machine.communicate.execute(
-            "ansible --version",
+            "python -c \"import importlib.metadata; print('ansible ' + importlib.metadata.version('ansible'))\"",
             error_class: Ansible::Errors::AnsibleNotFoundOnGuest,
             error_key: :ansible_not_found_on_guest) do |type, output|
             if type == :stdout && output.lines[0]
--- a/plugins/provisioners/ansible/provisioner/guest.rb
+++ b/plugins/provisioners/ansible/provisioner/guest.rb
@@ -75,7 +75,7 @@ module VagrantPlugins
           raw_output = ""
 
           result = @machine.communicate.execute(
-            "python -c \"import importlib.metadata; print('ansible ' + importlib.metadata.version('ansible'))\"",
+            "python3 -c \"import importlib.metadata; print('ansible ' + importlib.metadata.version('ansible'))\"",
             error_class: Ansible::Errors::AnsibleNotFoundOnGuest,
             error_key: :ansible_not_found_on_guest) do |type, output|
             if type == :stdout && output.lines[0]
--- a/plugins/provisioners/ansible/provisioner/host.rb
+++ b/plugins/provisioners/ansible/provisioner/host.rb
@@ -109,7 +109,7 @@ module VagrantPlugins
 
         def gather_ansible_version
           raw_output = ''
-          command = ['python', '-c',
+          command = ['python3', '-c',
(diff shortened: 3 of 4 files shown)


plugins/provisioners/ansible/cap/guest/posix/ansible_installed.rb
```diff
@@ -10,8 +10,8 @@ module AnsibleInstalled
             def self.ansible_installed(machine, version)
               command = 'test -x "$(command -v ansible)"'
 
-              if !version.empty?
-                command << "&& ansible --version | grep 'ansible #{version}'"
+              unless version.empty?
+                command << "&& [[ $(python -c \"import importlib.metadata; print(importlib.metadata.version('ansible'))\") == \"#{version}\" ]]"
```
soapy1, line 14: I think instead of using this method of checking if ansible is installed Vagrant should use the `ansible-community --version` check. Python is not guaranteed to be on the guest system causing this method to fail for a pretty broad range of use cases, including if users opt of the `default` install mode. For example, the following Vagrantfile:

```
  config.vm.define 'jammy' do |bare|
    bare.vm.box = 'ubuntu/jammy64'
    bare.vm.network 'private_network', type: 'dhcp'

    bare.vm.provision 'ansible_local' do |ansible|
      ansible.version = '7.0.0'
      ansible.playbook = 'playbook.yml'
      ansible.install_mode = 'default'
    end
  end
end
```

Taking a look at the ansible [release and maintenance docs](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-community-changelogs) it looks like starting from 5.x (soon to be 6.x) is no longer maintained. So I think it's ok to use the `ansible-community --version` method. I'm pretty out of the loop as far as Ansible goes, so I'm open to other suggestions here.

soapy1 (changes requested):

plugins/provisioners/ansible/cap/guest/posix/ansible_installed.rb
moritzheiber: I’m very much in favor of using `ansible-community` here since it radically simplifies the command that’s run to verify the state, however, that would mean, going forward, the provisioner has to actively mandate a minimum version of Ansible that’s supported. That’s a breaking change/change in policy as far as I can see, hence the more cautious approach here.

As for the failing example you posted above, I’m not certain that it’ll fail eventually once `ansible` has been installed since installing it would automatically pull in Python, but I’ll give your example a go once I’m back at my desk and report back.

moritzheiber: I've pushed a small addition, turning the call to `python` into `python3` (which is a common convention for all supported distributions/platforms). With it, your example works (and fails, as per the documentation):

```console
$ vagrant provision
[...]
==> jammy: Running provisioner: ansible_local...
    jammy: Installing Ansible...
The requested Ansible version (7.0.0) was not found on the guest.
Please check the Ansible installation on your Vagrant guest system (currently: 6.7.0),
or adapt the provisioner `version` option in your Vagrantfile.
See https://docs.vagrantup.com/v2/provisioning/ansible_common.html#version
for more information.
```

soapy1 (approved):

phinze (approved):

Issue #13050, opened 2023-01-07
Enhancement Request: Artix Linux Support


Penaz91:
**Is your feature request related to a problem? Please describe.**
Greetings, first of all thanks for this project, it helps a lot in my workflow.

Sadly I find a bit annoying having to manually force every VagrantFile I work to make it work correctly with my OS (Artix linux). I'm forced to set `config.vagrant.host = "gentoo"` to make it work.

**Describe the solution you'd like**
It would be nice having out-of-the-box support for Artix Linux.

**Describe alternatives you've considered**
I considered editing each and every VagrantFile I work with, but it's a bit cumbersome.

I also considered creating the plugin myself, but I have no knowledge in Ruby, thus I think the quality wouldn't be up to par.

**Additional context**
Artix linux is a systemd-free distribution that has support for the following inits:

- OpenRC (Which I use and is covered by the "gentoo" plugin)
- runit
- s6
- dinit

It used to support suite66 too, but its support has been retired mid-2022.

Thank you very much for your attention and for making such a useful product.

PR #13055, opened 2023-01-12
Artix Linux Host Support

Penaz91:
Greetings,

As expressed in the issue #13050 I've had a bit of niggling issues using Vagrant out of the box on Artix Linux.

Hopefully this pull request successfully adds support for it, or at least can be a launchpad for something coded in a better, more reliable way.

This is my first contribution to Vagrant, as well as my first contribution in Ruby, coded by reading documentation about the init systems, the code of Vagrant itself and intuition.

This is still untested, I'll try to test it as much as possible on my Artix-OpenRC machine.

Thanks for your attention.

### DIFF
--- a/plugins/hosts/arch/host.rb
+++ b/plugins/hosts/arch/host.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
   module HostArch
     class Host < Vagrant.plugin("2", :host)
       def detect?(env)
-        File.exist?("/etc/arch-release")
+        File.exist?("/etc/arch-release") and not File.exist?("/etc/artix-release")
       end
     end
   end
--- a/plugins/hosts/arch/host.rb
+++ b/plugins/hosts/arch/host.rb
@@ -4,7 +4,7 @@ module VagrantPlugins
   module HostArch
     class Host < Vagrant.plugin("2", :host)
       def detect?(env)
-        File.exist?("/etc/arch-release") and not File.exist?("/etc/artix-release")
+        File.exist?("/etc/arch-release") && !File.exist?("/etc/artix-release")
       end
     end
   end


Penaz91:
I managed to successfully test this PR on Artix Linux (OpenRC): init, destroy, up, ssh, halt work without issues (provider used: LibVirt). NFS Folders mount successfully.

Had to edit the Arch Linux plugin to avoid issues with Artix Systems that migrated from Arch (thus containing both `/etc/artix-release` and `/etc/arch-release`). I think this also helps with Artix systems that use the `artix-archlinux-support` package.

soapy1:
Hi there,
thanks for opening up a PR for adding support for Artix Linux guests! This change set looks great. But as we move towards [Vagrant 3.0](https://www.hashicorp.com/blog/toward-vagrant-3-0) we are looking to externalize plugins that aren't critical to Vagrant's core functions. So, would you consider building this plugin as an external plugin? This should be really easy to do (you already did all the hard work of implementing the required capabilities) using the [template repo](https://github.com/soapy1/vagrant-guest-plugin-skeleton) for building guest plugins.

Penaz91:
> Hi there, thanks for opening up a PR for adding support for Artix Linux guests! This change set looks great. But as we move towards [Vagrant 3.0](https://www.hashicorp.com/blog/toward-vagrant-3-0) we are looking to externalize plugins that aren't critical to Vagrant's core functions. So, would you consider building this plugin as an external plugin? This should be really easy to do (you already did all the hard work of implementing the required capabilities) using the [template repo](https://github.com/soapy1/vagrant-guest-plugin-skeleton) for building guest plugins.

Hello!
Sorry, but I'm a bit confused, since I thought my changeset would affect the host side. In fact the changeset touches the `plugins/hosts/` subfolder.

If there is a way to externalize host plugins I would be happy to give a shot at it. Thank you!

soapy1:
> I thought my changeset would affect the host side

right! That's definitely my bad. You're going to want to take a look at the host plugin template https://github.com/soapy1/vagrant-host-plugin-skeleton-ruby. Let me know if you have any issues with that and I'll try to help you out. 

I also notice you have a change to the arch host plugin. So, we'll need to address that either here or in another PR.

Penaz91:
I managed to work my way through a working plugin, it's available here: https://github.com/Penaz91/vagrant-host-artix/

I also managed to release the plugin as a gem without issues.

We can discuss the Arch host plugin change here without issue.

I had to change that because you can migrate from Arch Linux to Artix Linux by changing repositories, this may end up leaving the `/etc/arch-release` file in the system, making Vagrant mistake Artix Linux for an Arch Linux installation.

Also it is my speculation that the Artix Package `artix-archlinux-support` (which is being slowly phased out) installs that file to maintain compatibility with some packages.

Thank you for all the tips and pointers!

soapy1:
>I managed to work my way through a working plugin, it's available here: https://github.com/Penaz91/vagrant-host-artix/ 

Awesome, great to hear!

Send a ping whenever this is ready to review. I expect that the extent of the changes to Vagrant will just be in the host detection for Arch linux?

Penaz91:
> Awesome, great to hear!

Hopefully this helps out people who are having a bit of trouble with Artix. Out of pure curiosity, is there a way to write to Vagrant's log (at debug level)? It would be useful if my plugin could log which Init system was detected.

> Send a ping whenever this is ready to review. I expect that the extent of the changes to Vagrant will just be in the host detection for Arch linux?

Correct. It's a very crude check.

I rebased my branch, dropping all commits that aren't strictly related to the Archlinux host detection. I think it's ready for a review!

Thank you!

plugins/hosts/arch/host.rb
```diff
@@ -4,7 +4,7 @@ module VagrantPlugins
   module HostArch
     class Host < Vagrant.plugin("2", :host)
       def detect?(env)
-        File.exist?("/etc/arch-release")
+        File.exist?("/etc/arch-release") and not File.exist?("/etc/artix-release")
```
soapy1, line 7:
```suggestion
        File.exist?("/etc/arch-release") && !File.exist?("/etc/artix-release")
```

soapy1 (approved):
small knit pick suggestion, but lgtm!

plugins/hosts/arch/host.rb
Penaz91: Thank you!

It's only right that the commit has the same style of the rest of the code. I was split between the two ways of writing this commit and I chose the one I thought was most readable (as a mainly Python dev).

Again, thanks for your attention and patience.

phinze (approved):

Issue #13044, opened 2023-01-04
Vagrant up fails when using VM snapshots with VB 7.0 


Kalimero66:
### Debug output

> vagrant up 
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Clearing any previously set forwarded ports...
There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.
Command: ["modifyvm", "de8f03f8-3b2e-4de9-acb7-b55669b5faf7", "--natpf1", "delete", "ssh", "--natpf1", "delete", "ssh"]
Stderr: VBoxManage.exe: error: Code E_INVALIDARG (0x80070057) (extended info not available)
VBoxManage.exe: error: Context: "RemoveRedirect(Bstr(ValueUnion.psz).raw())" at line 2108 of file VBoxManageModifyVM.cpp

### Expected behavior
Vagrant up should not stop with errors when clearing forwarded ports.

### Actual behavior
When vagrant up tries to clear any forwarded ports, it lists the forwarded ports and the same port gets duplicated which causes the issue.
read_forwarded_port() output:
DEBUG virtualbox_7_0: - [1, “ssh”, 2222, 22, “127.0.0.1”]
DEBUG virtualbox_7_0: - [1, “ssh”, 2222, 22, “127.0.0.1”]

See more troubleshooting details here:  
https://discuss.hashicorp.com/t/vbox-vm-with-vagrant-stopped-working/45826/4

The issue has to do with the routine read_forwarded_port(). It needs to be revisited to account for VB VM that makes use of Snapshots not to have the same forwarded port reported multiple times.
hashicorp/vagrant/blob/8549450cff40a55ec193540a5387a09caee309b4/plugins/providers/virtualbox/driver/version_7_0.rb#L224

#### Vagrant version

vagrant 2.3.4

#### Host operating system

Windows 10 Enterprise
**VirtualBox Version 7.0.0 r153978 (Qt5.15.2)**

#### Steps to reproduce

1. Create a VM with vagrant up 
2. Vagrant halt
3. **Take a snapshot of the VM with VB gui** (--> that would add a new section in the .vbox config file that duplicates the forwarded port)
4. vagrant up

#### Vagrantfile
Here is the Vagrantfile but 

require_relative 'devenv'

if not is_valid_environment()
  exit
end

settings = load_settings('vagrant.yml')

box_disk_size_gb = settings['vm']['disksize']

[...]

PR #13056, opened 2023-01-12
Read only machine port forward info

soapy1:
In the VM config file port forward information is captured in the Snapshot defition as well as the VM definition. This causes duplicates in port forward entries if the file is queried for all available network adapters. Querying the VM config file for the the Network Adapters only part of the Machine definition resolves this and yields the most up to date port forward info.

fixes https://github.com/hashicorp/vagrant/issues/13044


Kalimero66 (approved):

Kalimero66 (approved):
Sounds good . It ensures no duplicated port-forwarding.
Hopefully VB will not change their format

chrisroberts (approved):

Issue #12880, opened 2022-09-07
race condition when two Vagrant processes try to download the same box


evgeni:
### Vagrant version

2.2.19

### Host operating system

CentOS Stream 8 (but shouldn't matter)

### Guest operating system

CentOS Stream 8 (but shouldn't matter)

### Vagrantfile

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "centos/stream8"
end
```

### Expected behavior

I have two separate projects that use the same base box (`centos/stream8` in this example) in two different folders.

When the base box is not yet available on my system, a `vagrant up` does download it from the cloud.

As I have two projects, I might call `vagrant up` in parallel, and expect the two processes not to overwrite each others files and succeed (they will probably download the box twice, but that's OK).

### Actual behavior

In reality, if I call `vagrant up` in parallel for both projects, they both will try to download the box, usually resulting in one of the runs failing, as the one process did overwrite the files of the others:

```
    /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/lib/vagrant/util/file_checksum.rb:47:in `initialize': No such file or directory @ rb_sysopen - /root/.vagrant.d/tmp/boxf4b6e814a5003f4177ccc9098d0df7e64f39b349 (Errno::ENOENT)
            from /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/lib/vagrant/util/file_checksum.rb:47:in `open'
            from /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/lib/vagrant/util/file_checksum.rb:47:in `checksum'
            from /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/lib/vagrant/action/builtin/box_add.rb:554:in `validate_checksum'
            from /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/lib/vagrant/action/builtin/box_add.rb:366:in `box_add'
            from /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/lib/vagrant/action/builtin/box_add.rb:292:in `add_from_metadata'
            from /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/lib/vagrant/action/builtin/box_add.rb:127:in `call'
            from /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/lib/vagrant/action/warden.rb:48:in `call'
[...]

PR #13057, opened 2023-01-13
Add a file mutex when downloading box files.

soapy1:
Box's are global to Vagrant. Multiple Vagrant process can all access the box directory for both downloading and extracting boxes. A file mutex will ensure that multiple Vagrant process will not trample each other if they are trying to download the same box.

Now if multiple Vagrant process are trying to download the same box the process started first will successfully download and add the box. The second will display an error message:

```
-> % vagrant up

==> default: Box 'centos/stream8' could not be found. Attempting to find and install...
    default: Box Version: >= 0
==> default: Loading metadata for box 'centos/stream8'
    default: URL: https://vagrantcloud.com/centos/stream8
==> default: Adding box 'centos/stream8' (v20210210.0) for provider: virtualbox
    default: Downloading: https://vagrantcloud.com/centos/boxes/stream8/versions/20210210.0/providers/virtualbox.box
==> default: Box download is resuming from prior download progress
Download to global Vagrant location already in progress. This
may be caused by other Vagrant processes attempting to download
a file to the same location.

Download path: /Users/sophia/.vagrant.d/tmp/box5fb30816ce1dcffcbdb0d582d356f3626d7cc595
Lock file path: /Users/sophia/.vagrant.d/tmp/box5fb30816ce1dcffcbdb0d582d356f3626d7cc595.lock
```

fixes: https://github.com/hashicorp/vagrant/issues/12880

### DIFF
--- a/lib/vagrant/action/builtin/box_add.rb
+++ b/lib/vagrant/action/builtin/box_add.rb
@@ -485,7 +485,8 @@ module Vagrant
           mutex_path = d.destination + MUTEX_SUFFIX
           if File.file?(mutex_path)
             raise Errors::DownloadAlreadyInProgress,
-              dest_path: d.destination
+              dest_path: d.destination,
+              lock_file_path: mutex_path
           end
 
           begin
--- a/lib/vagrant/action/builtin/box_add.rb
+++ b/lib/vagrant/action/builtin/box_add.rb
@@ -22,6 +22,10 @@ module Vagrant
         # file already exists.
         RESUME_DELAY = 24 * 60 * 60
 
+        # When a box file is being downloaded, identify the download is in progress
+        # by Vagrant with a mutex file named <box file destination>.<MUTEX_SUFFIX>
+        MUTEX_SUFFIX = ".lock"
+
         def initialize(app, env)
           @app    = app
           @logger = Log4r::Logger.new("vagrant::action::builtin::box_add")
@@ -478,13 +482,22 @@ module Vagrant
             end
           end
 
+          mutex_path = d.destination + MUTEX_SUFFIX
+          if File.file?(mutex_path)
+            raise Errors::DownloadAlreadyInProgress,
+              dest_path: d.destination
+          end
+
           begin
+            File.write(mutex_path, "")
             d.download!
           rescue Errors::DownloaderInterrupted
             # The downloader was interrupted, so just return, because that
             # means we were interrupted as well.
             @download_interrupted = true
             env[:ui].info(I18n.t("vagrant.actions.box.download.interrupted"))
+          ensure
+            File.delete(mutex_path)
           end
 
           Pathname.new(d.destination)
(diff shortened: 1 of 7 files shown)


lib/vagrant/action/builtin/box_add.rb
```diff
@@ -478,13 +482,23 @@ def download(url, env, **opts)
...
           begin
+            File.write(mutex_path, "")
             d.download!
           rescue Errors::DownloaderInterrupted
             # The downloader was interrupted, so just return, because that
             # means we were interrupted as well.
             @download_interrupted = true
             env[:ui].info(I18n.t("vagrant.actions.box.download.interrupted"))
+          ensure
+            File.delete(mutex_path)
```
soapy1, line 501: I think this is an ok time to delete the mutex. I can imagine there being a situation between the download finishing and the box being added to the collection where another download of the same box starts causing an error. So, maybe this file needs to be created earlier and deleted earlier to better line up with the box being added to the box collection?

lib/vagrant/util/file_mutex.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      def with_lock(&block)
+        lock
+        block.call
+      ensure
+        unlock
+      end
+
+      def lock
+        if File.file?(@mutex_path)
```
chrisroberts, line 16: There ends up being a race condition here between checking for the existence of the file, and creating the file. Instead, it would be better to open the file and then [lock it](https://ruby-doc.org/3.1.3/File.html#method-i-flock) to ensure an exclusive lock.

chrisroberts (changes requested):

test/unit/vagrant/action/builtin/box_add_test.rb
```diff
@@ -84,6 +84,25 @@ def with_web_server(path, **opts)
...
   end
 
+  context "the download location is locked" do
+    let(:box_path) { iso_env.box2_file(:virtualbox) }
+    let(:mutex_path) {  env[:tmp_path].join("box" + Digest::SHA1.hexdigest("file://" + box_path.to_s) + ".lock").to_s }
+
+    it "raises a download error" do
+      # Lock file
+      f = File.open(mutex_path, "w+") 
+      f.flock(File::LOCK_EX|File::LOCK_NB)
```
soapy1, line 94: I would have liked to put this in a before block like

```
before do
      f = File.open(mutex_path, "w+") 
      f.flock(File::LOCK_EX|File::LOCK_NB) 
    end
```

but then I don't have a good method for closing the file. Open to suggestions about a slicker way to do this 📖

chrisroberts: You could use an instance variable so it's still available for an `after` block. Something like:

```ruby
before do
    @f = File.open(mutex_path, "w+") 
    @f.flock(File::LOCK_EX|File::LOCK_NB) 
end

after { @f.close }

it ...
```

chrisroberts (dismissed):
:+1: looks great

soapy1:
Going ahead with merging since the last push was just an update to the tests. Thanks @chrisroberts for the suggestion!

PR #13059, opened 2023-01-14
[website] chore: updates generated docs

kendallstrautman:
This updates the generated docs for the new [mdx inline alert component](https://github.com/hashicorp/dev-portal/pull/1517). See [docs on dev portal here.](https://developer.hashicorp.com/swingset/components/mdxinlinealert)

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "author": "HashiCorp",
   "devDependencies": {
-    "@hashicorp/platform-cli": "^1.2.0",
+    "@hashicorp/platform-cli": "^2.5.1",
     "husky": "4.3.8",
     "next": "^12.3.1",
     "prettier": "2.2.1"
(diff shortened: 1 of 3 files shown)


zchsh (approved):

website/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "author": "HashiCorp",
   "devDependencies": {
-    "@hashicorp/platform-cli": "^1.2.0",
```
kendallstrautman, line 7: I should check if this breaking upgrade affects anything else. @BRKalow if you have quick thoughts?

kendallstrautman: Seems to me that the breaking change just upgrade the eslint version. So we should be fine.

thiskevinwang (approved):
Re: the `@hashicorp/platform-cli` major upgrade:

I'll say this is safe. (I don't remember if there is any _gotcha_ to note here, but should be mostly OK, given that it's a dev dependency)

**update:** Here's the `2.0.0` bump for Waypoint, for reference. (Nothing outstanding here)
- https://github.com/hashicorp/waypoint/pull/3170

PR #13112, opened 2023-04-01
Fix NFS exports pruning of other users. Issue 9070.

kotsis:
I have tested this locally and it works.


kotsis:
All should be ok now, I also added tests.

Issue #12430, opened 2021-06-23
Docker provider with nfs synced folders: No host IP was given to the Vagrant core NFS helper


vafschoen:
### Vagrant version
Vagrant 2.2.16

### Host operating system

Ubuntu Impish Indri 21.10
&
macOS 11.4

### Guest operating system

Debian GNU/Linux 10 (buster)

### Vagrantfile

```ruby
Vagrant.configure("2") do |config|
  config.vm.synced_folder ".", "/usr/share/nginx/html", nfs: true

  config.vm.provider "docker" do |d|
    d.image = "nginx"
  end
end
```

### Debug output

https://gist.github.com/vafschoen/d4f9885bdcc8592aef0ceaf56a918c9f

### Expected behavior

Start the container and mount the synced_folder using nfs.

### Actual behavior

Vagrant exists with
`No host IP was given to the Vagrant core NFS helper. This is an internal error that should be reported as a bug.`

### Steps to reproduce

1. vagrant up

### References

- #10693

vafhornung:
Same problem here, any update or workarounds known ?

PR #13149, opened 2023-05-12
docker: sync folders before preparing nfs settings

allisonlarson:
Re-order the `SyncedFolders` to occur before `PrepareNFSSettings`, ensuring that the host & machine ips are set appropriately. 

Fixes https://github.com/hashicorp/vagrant/issues/12430


chrisroberts (approved):
:+1:

PR #13150, opened 2023-05-12
Fix LANG value used for VirtualBox driver

### LINKED ISSUE #13145: Vagrant makes LANG=C setting which breaks VBoxManage output
Vagrant [sets LANG=C](https://github.com/hashicorp/vagrant/blob/9076445d648a221ec43f0fcf363d0581dcb51de3/plugins/providers/virtualbox/driver/base.rb#L465) to make VBoxManage output independent of locale. However, on my system (Arch Linux, VirtualBox 7.0.8) VBoxManage doesn't produce any useful output when LANG is set to C. It does, however, produce the needed output when LANG=C.UTF-8.

### Debug output

https://gist.github.com/gentoo-root/911854ad48374df6cf0fe116d606fb30

### Expected behavior

Vagrant should work, i.e. start a VM with the simple Vagrantfile below.

### Actual behavior

`vagrant up` fails, because vagrant sets LANG=C when it runs VBoxManage commands, and VBoxManage doesn't produce any useful output when LANG=C is set.

#### Vagrant version

Vagrant 2.3.4

#### Host operating system

Arch Linux

#### Guest operating system

Doesn't matter.

#### Steps to reproduce

Simple test for VirtualBox:

0. Make sure VBoxXPCOMIPCD is not running, because it caches the LANG setting. It's usually enough to wait for 5 seconds, unless something is interacting with VirtualBox.
1. Run `LANG=C VBoxManage list systemproperties` and observe empty output.
2. After VBoxXPCOMIPCD quits, run `LANG=C.UTF-8 VBoxManage list systemproperties` and observe plenty of output.

Vagrant-specific test:

1. Run `VAGRANT_LOG=info vagrant up` with the Vagrantfile below.
2. Observe the [error](https://gist.github.com/gentoo-root/911854ad48374df6cf0fe116d606fb30#file-vagrant-bad-log).
3. Wait for VBoxXPCOMIPCD to quit.
4. Run `LANG=C.UTF-8 VBoxManage list systemproperties | tail -n1; VAGRANT_LOG=info vagrant up`. The first command starts VBoxXPCOMIPCD and assigns LANG=C.UTF-8 to it, so even when vagrant uses LANG=C afterwards (within 5 seconds after the first command), vagrant's setting is ignored, and the VBoxManage commands started by vagrant run properly and produce output.
5. Observe the [good log](https://gist.github.com/gentoo-root/911854ad48374df6cf0fe116d606fb30#file-vagrant-good-log).

#### Vagrantfile

```ruby
  config.vm.box = "ubuntu/jammy64"
  config.vm.synced_folder ".", "/shared"
end
```

chrisroberts:
The VirtualBox driver sets the LANG env var to prevent localized output
being returned when executing CLI commands. If the `locale` command is
present, do a best effort lookup to determine the properly value to use
for the LANG environment variable.

Fixes #13145

### DIFF
--- a/test/unit/plugins/providers/virtualbox/driver/base.rb
+++ b/test/unit/plugins/providers/virtualbox/driver/base.rb
@@ -1,0 +1,99 @@
+require_relative "../base"
+require Vagrant.source_root.join("plugins/providers/virtualbox/driver/base")
+
+describe VagrantPlugins::ProviderVirtualBox::Driver::Base do
+  describe "#env_lang" do
+    context "when locale command is not available" do
+      before do
+        allow(Vagrant::Util::Which).to receive(:which).with("locale").and_return(false)
+      end
+
+      it "should return default value" do
+        expect(subject.send(:env_lang)).to eq({LANG: "C"})
+      end
+    end
+
+    context "when the locale command is available" do
+      let(:result) { Vagrant::Util::Subprocess::Result.new(exit_code, stdout, stderr) }
+      let(:stderr) { "" }
+      let(:stdout) { "C.default" }
+      let(:exit_code) { 0 }
+
+      before do
+        allow(Vagrant::Util::Which).to receive(:which).with("locale").and_return(true)
+        allow(Vagrant::Util::Subprocess).to receive(:execute).with("locale", "-a").and_return(result)
+      end
+
+      context "when locale command errors" do
+        let(:exit_code) { 1 }
+
+        it "should return default value" do
+          expect(subject.send(:env_lang)).to eq({LANG: "C"})
+        end
+      end
+
+      context "when locale command does not error" do
+        let(:exit_code) { 0 }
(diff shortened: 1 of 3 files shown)


test/unit/plugins/providers/virtualbox/driver/base.rb
```diff
@@ -0,0 +1,99 @@
...
+      before do
+        allow(Vagrant::Util::Which).to receive(:which).with("locale").and_return(false)
+      end
+
+      it "should return default value" do
+        expect(subject.send(:env_lang)).to eq({LANG: "C"})
+      end
+    end
+
+    context "when the locale command is avaialble" do
```
allisonlarson, line 16:
```suggestion
    context "when the locale command is available" do
```

allisonlarson (approved):

allisonlarson (approved):
👍

PR #13153, opened 2023-05-15
Docker: Attempt using docker command for bridge ip

allisonlarson:
Don't rely on `/sbin/ip` to fetch the docker bridge ip address, instead first attempt to use the docker command to fetch it. If it fails, fall back to previous behavior.


plugins/providers/docker/driver.rb
```diff
@@ -236,8 +236,20 @@ def all_containers
...
+        bridge = inspect_network("bridge")[0]
+        bridge["IPAM"]["Config"][0]["Gateway"]
+      rescue => e
+        @logger.debug("Failed to get bridge ip from docker: #{e}")
+        @logger.debug("Falling back to /sbin/ip")
+        docker_bridge_ip_from_sbin_ip
+      end
+
+      def docker_bridge_ip_from_sbin_ip
         output = execute('/sbin/ip', '-4', 'addr', 'show', 'scope', 'global', 'docker0')
```
chrisroberts, line 253: I think updating this to just `'ip'` and removing the explicit path would be appropriate.

plugins/providers/docker/driver.rb
```diff
@@ -236,8 +236,20 @@ def all_containers
         execute('docker', 'ps', '-a', '-q', '--no-trunc').to_s.split
       end
 
+      # Attempts to first use the docker-cli tool to inspect the default bridge subnet
+      # Falls back to using /sbin/ip if that fails
+      #
       # @return [String] IP address of the docker bridge
       def docker_bridge_ip
+        bridge = inspect_network("bridge")[0]
+        bridge["IPAM"]["Config"][0]["Gateway"]
```
chrisroberts, line 245: Doing some result type checks here over running into an exception would be preferred: 

```ruby 
bridge = inspect_network("bridge")&.first
if bridge
  bridge_ip = bridge.dig("IPAM", "Config", 0, "Gateway")
end
return bridge_ip if bridge_ip
docker_bridge_ip_fallback
```

plugins/providers/docker/driver.rb
```diff
@@ -236,8 +236,20 @@ def all_containers
...
+      rescue => e
+        @logger.debug("Failed to get bridge ip from docker: #{e}")
+        @logger.debug("Falling back to /sbin/ip")
+        docker_bridge_ip_from_sbin_ip
+      end
+
+      def docker_bridge_ip_from_sbin_ip
```
chrisroberts, line 252: Maybe update this method name to `docker_bridge_ip_fallback` to make it a bit more descriptive on its intention.

chrisroberts (approved):
:+1: looks great

PR #13166, opened 2023-05-19
Salt: Verify bootstrap-salt download

allisonlarson:
Verifies that the bootstrap-salt script that is downloaded during Salt provisioning time is verified against a SHA256 sum. It moves the script download to the host machine, where it is able to also download a corresponding SHA256 file, and compute and validate the two values. This removes the need for the shell scripts in the repo, but preserves the behavior of supplying your own.

### DIFF
--- a/templates/locales/en.yml
+++ b/templates/locales/en.yml
@@ -3077,6 +3077,8 @@ en:
           You must set `python_version` as an integer or string that represents an integer.
         version_type_missing: |-
           You must set the option `install_type` when specifying a `version`.
+        salt_invalid_shasum_error: |-
+          The bootstrap-salt script downloaded from '%{source}' couldn't be verified. Expected SHA256 '%{expected_sha}', but computed '%{computed_sha}'
 
     pushes:
       file:
--- a/plugins/provisioners/salt/bootstrap-salt.ps1
+++ b/plugins/provisioners/salt/bootstrap-salt.ps1
@@ -1,5 +1,0 @@
-# Define script root for PowerShell 2.0
-$ScriptRoot = Split-Path $script:MyInvocation.MyCommand.Path
-
-# Run the upstream bootstrap script with passthrough arguments
-& "${ScriptRoot}\bootstrap_salt_upstream.ps1" @args
--- a/plugins/provisioners/salt/bootstrap-salt.ps1
+++ b/plugins/provisioners/salt/bootstrap-salt.ps1
@@ -1,11 +1,5 @@
-# Powershell supports only TLS 1.0 by default. Add support for TLS 1.2
-[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]'Tls12'
-
 # Define script root for PowerShell 2.0
 $ScriptRoot = Split-Path $script:MyInvocation.MyCommand.Path
-
-# Download the upstream bootstrap script
-(New-Object System.Net.WebClient).DownloadFile('https://winbootstrap.saltproject.io', "${ScriptRoot}\bootstrap_salt_upstream.ps1")
 
 # Run the upstream bootstrap script with passthrough arguments
 & "${ScriptRoot}\bootstrap_salt_upstream.ps1" @args
--- a/plugins/provisioners/salt/bootstrap-salt.sh
+++ b/plugins/provisioners/salt/bootstrap-salt.sh
@@ -1,6 +1,0 @@
-#!/bin/sh -
-if [ -e bootstrap-salt.sh ]; then
-  sh bootstrap-salt.sh "$@"
-else
-  exit 1
-fi
--- a/plugins/provisioners/salt/bootstrap-salt.sh
+++ b/plugins/provisioners/salt/bootstrap-salt.sh
@@ -1,21 +1,4 @@
 #!/bin/sh -
-
-cd `mktemp -d`
-
-# We just download the bootstrap script by default and execute that.
-if [ -x /usr/bin/fetch ]; then
-    /usr/bin/fetch -o bootstrap-salt.sh https://bootstrap.saltproject.io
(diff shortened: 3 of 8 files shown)


plugins/provisioners/salt/default_downloader.rb
```diff
@@ -0,0 +1,93 @@
...
+
+module VagrantPlugins
+  module Salt
+    class DefaultDownloader
+      BASE_URL = "https://repo.saltproject.io/bootstrap/stable/"
+      PS1 = "bootstrap-salt.ps1"
+      SH  = "bootstrap-salt.sh"
+      SHA256_SUFFIX = "sha256"
+      SIG_SUFFIX = "asc"
+      KEY_URL = "#{BASE_URL}SALT-PROJECT-GPG-PUBKEY-2023.gpg"
```
allisonlarson, line 13: Don't love maintaining the GPG keyfile name here, can we do something clever?

plugins/provisioners/salt/default_downloader.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      def get_bootstrap_script
+        @logger.debug "Downloading bootstrap script from #{source_url}"
+        script_file = download_to_tmpfile(source_url)
+
+        if system("gpg --version")
+          @logger.debug "Verifying gpg signature"
+          verify_signature(script_file)
+        else
+          @logger.debug "Skipping gpg signature verification: gpg not found"
```
allisonlarson, line 32: I wasn't sure what to do if GPG wasn't supported here. Do we want to block the download? Or is falling back to just verifying the SHA256 sum good?

plugins/provisioners/salt/default_downloader.rb
```diff
@@ -0,0 +1,93 @@
...
+        sha256_file = download_to_tmpfile("#{source_url}.#{SHA256_SUFFIX}")
+        sha256 = extract_sha256(sha256_file.read)
+
+        @logger.debug "Computing sha256 value from script file"
+        computed_sha256 = Digest::SHA256.hexdigest(script.read)
+        script.rewind
+
+        @logger.debug "Comparing sha256 values"
+        if computed_sha256 != sha256
+          raise Salt::Errors::SaltError, :unverifiable_download
```
allisonlarson, line 52: Any other error handling tips for here?

plugins/provisioners/salt/default_downloader.rb
```diff
@@ -0,0 +1,93 @@
+require 'open-uri'
+require 'digest'
+require_relative "./errors"
+
+module VagrantPlugins
+  module Salt
+    class DefaultDownloader
```
allisonlarson, line 7: Taking name suggestions

chrisroberts: `BootstrapDownloader` seems alright since this is specifically for that.

plugins/provisioners/salt/default_downloader.rb
```diff
@@ -0,0 +1,93 @@
...
+      BASE_URL = "https://repo.saltproject.io/bootstrap/stable/"
+      PS1 = "bootstrap-salt.ps1"
+      SH  = "bootstrap-salt.sh"
+      SHA256_SUFFIX = "sha256"
+      SIG_SUFFIX = "asc"
+      KEY_URL = "#{BASE_URL}SALT-PROJECT-GPG-PUBKEY-2023.gpg"
+
+      def initialize(communicator)
```
chrisroberts, line 15: Should include passing in what the guest platform is for script determination (value in `config.vm.guest`)

plugins/provisioners/salt/default_downloader.rb
```diff
@@ -0,0 +1,93 @@
...
+        @communicator = communicator
+        @logger  = Log4r::Logger.new("vagrant::salt::default_downloader")
+      end
+
+      def source_url 
+        @communicator == :winrm ? "#{BASE_URL}#{PS1}" : "#{BASE_URL}#{SH}"
```
chrisroberts, line 21: Script should be chosen based on guest platform where if it is `:windows`, then `.ps1` is used, otherwise `.sh` is used.

plugins/provisioners/salt/default_downloader.rb
chrisroberts: Creating a new error type (subclassing `SaltError`) would be good. Something like `InvalidShasumError` or something. The remote location of the script, expected sha value, and calculated sha value should also be included so the information can be displayed in the error message.

templates/locales/en.yml
```diff
@@ -3077,6 +3077,8 @@ en:
           You must set `python_version` as an integer or string that represents an integer.
         version_type_missing: |-
           You must set the option `install_type` when specifying a `version`.
+        unverifiable_download: |-
+          The bootstrap script wasn't able to be verified
```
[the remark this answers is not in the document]
chrisroberts: Include file source, expected sha, and actual sha in the error message here

plugins/provisioners/salt/default_downloader.rb
```diff
@@ -0,0 +1,93 @@
...
+        @logger.debug "Downloading sha256 file from #{source_url}/sha256"
+        sha256_file = download_to_tmpfile("#{source_url}.#{SHA256_SUFFIX}")
+        sha256 = extract_sha256(sha256_file.read)
+
+        @logger.debug "Computing sha256 value from script file"
+        computed_sha256 = Digest::SHA256.hexdigest(script.read)
+        script.rewind
+
+        @logger.debug "Comparing sha256 values"
+        if computed_sha256 != sha256
```
chrisroberts, line 51: Include debug output here with sha mismatch (values included)

chrisroberts (approved):

PR #13171, opened 2023-05-22
Fix ssh keyboard interactive login

natpicone:
The PR https://github.com/hashicorp/vagrant/pull/13052 has introduced the "keyboard-interactive" option for SSH connections.
However, the order of preference is crucial, and when using this option, even if a valid SSH key is available, the connection still prompts for a password.

### DIFF
--- a/plugins/communicators/ssh/communicator.rb
+++ b/plugins/communicators/ssh/communicator.rb
@@ -404,7 +404,7 @@ module VagrantPlugins
 
         # Set some valid auth methods. We disable the auth methods that
         # we're not using if we don't have the right auth info.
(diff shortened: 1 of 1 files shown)


natpicone:
cc: @soapy1 @chrisroberts

chrisroberts:
@natpicone Thanks for this PR! After further investigation I don't believe the original change (in #13052) was an appropriate modification based on the [issue](#13008) it was attempting to resolve. The root cause of the issue was an improperly configured Vagrant box. #13194 removes the keyboard-interactive authentication method, which should resolve the associated issues introduced by its addition.

Thanks again!

Issue #12610, opened 2021-12-07
Enhancement Request: CPU architecture aware box handling


fatso83:
**Is your feature request related to a problem? Please describe.**
A lot of developers are now on ARM architectures. Even if they should be able to to get VMWare Fusion playing nicely with Vagrant, the issue of finding boxes that runs on the right provider is left missing. You cannot filter boxes on Vagrant Cloud by architecture, nor will you get a good error message if one does not run on your provider due to architecture mismatch.

**Describe the solution you'd like**
Two things:
- Vagrant Cloud should require an `architecture` field for new boxes and let users filter on `amd64` (`x86`?), `arm64`, etc. (from a list of enums) when searching for images.
- When specifying `bionic64` in your Vagrantfile, and that box does not exist for _your_ architecture, you should be told through an error message.

**Describe alternatives you've considered**
Clear documentation mentioning the architecture issues.

**Additional context**
- People thinking ARM on macOS is insupported (which it is): #12559
- The long and troublesome route ventured to get it working (which involved waaay too many details I never wanted to know about): https://github.com/hashicorp/vagrant-vmware-desktop/issues/22

jcsalem:
Yes, I need this feature as well.
I got Vagrant working well with Parallels but I have to create a separate box name just for the arm version.
Particularly since Vagrant runs under Rosetta on the Apple M1, it was challenging to make my Vagrantfile select the appropriate box name based on the underlying CPU architecture.

brycekahle:
> it was challenging to make my Vagrantfile select the appropriate box name based on the underlying CPU architecture.

@jcsalem Mind sharing how you did this? I'm running into the same issue.

fatso83:
@brycekahle Since the Vagrantfile is just a Ruby script, you just need to switch on architecture to set the box name.

The pre-requisite is of course that you create differently named boxes for each architecture. Like `@brycekahle/ubuntu-focal-amd64` and `@brycekahle/ubuntu-focal-arm64`. 

```ruby
architecture = do_something_to_get_some_architecture_abbreviation()

Vagrant.configure("2") do |config|
  config.vm.box = "@brycekahle/ubuntu-focal-#{architecture}" 
  # the rest
  # ...
end
```

All that is left is then to implement `do_something_to_get_some_architecture_abbreviation()` or use some built-in. I am no Ruby programmer, so calling out to shell utils like `uname` could be one way of achieving this, but there are of course native built-ins that are probably preferable :smiley:

[...]

PR #13239, opened 2023-07-28
Add architecture support

chrisroberts:
This PR adds box architecture support. It introduces a new setting to
the vm configuration within the Vagrantfile:

```ruby
  config.vm.box = "box-name"
  config.vm.box_architecture = "amd64"
end
```

In most circumstances, this setting will not be needed. By default it
is set to a special value of `:auto`. The behavior when set to `:auto` is
as follows: Vagrant will determine the architecture of the host and automatically 
request it from the box repository. If architecture information is available in 
the repository metadata and a matching architecture was not found, Vagrant will
look for a box with an architecture of "unknown" and flagged as the default
architecture. This behavior allows Vagrant to continue working correctly with
boxes currently published on Vagrant Cloud that do not include architecture
information. 

The `box_architecture` setting can also be set to `nil`, which will prevent
any architecture filtering being performed, and the box flagged as the default
architecture will be used.

The box collection will now unpack boxes based on architecture. Special handling
is done when `box_architecture` is set to `:auto` which prevents Vagrant from
not properly detecting boxes installed without architecture support. 

The `vagrant cloud` commands have been updated to include architecture support
where required. These updates are dependent on hashicorp/vagrant_cloud#82

Fixes: #12610

### DIFF
--- a/test/unit/vagrant/util/platform_test.rb
+++ b/test/unit/vagrant/util/platform_test.rb
@@ -11,6 +11,70 @@ describe Vagrant::Util::Platform do
   after { described_class.reset! }
   subject { described_class }
 
+  describe "#architecture" do
+    let(:cpu_string) { "unknown" }
+
+    before do
+      allow(RbConfig::CONFIG).
+        to receive(:[]).with("target_cpu").
+             and_return(cpu_string)
+    end
+
+    context "when cpu is x86_64" do
+      let(:cpu_string) { "x86_64" }
+
+      it "should be mapped to amd64" do
+        expect(described_class.architecture).to eq("amd64")
+      end
+    end
+
+    context "when cpu is i386" do
+      let(:cpu_string) { "i386" }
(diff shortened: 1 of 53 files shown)


test/unit/vagrant/util/platform_test.rb
```diff
@@ -8,6 +8,70 @@
...
+    before do
+      allow(RbConfig::CONFIG).
+        to receive(:[]).with("target_cpu").
+             and_return(cpu_string)
+    end
+
+    context "when cpu is x86_64" do
+      let(:cpu_string) { "x86_64" }
+
+      it "should be mapped to am64" do
```
vancluever, line 23:
```suggestion
      it "should be mapped to amd64" do
```

vancluever (dismissed):
Great work! 🎉

websafe:
Great! Thank you!

PR #13241, opened 2023-07-30
Remove duplicate appending of "6" to network type

lordgurke:
To distinct between IPv4 and IPv6 configuration, a "6" was added to the network configuration type if an IPv6 address should be configured. This is now duplicate, as with pull [13024](https://github.com/hashicorp/vagrant/pull/13024) the same thing is already done prior and more globally, thus leading to a duplicate "6" at the end of the network config type, i.e. "static66".  
  
Tested on my local system without breakage using a "private_network" config, but I was unable to extensively test this.


chrisroberts (approved):
Thanks!

Issue #13342, opened 2024-01-28
Docker provider build fails when using containerd storage backend


tnaroska:
The recently released [docker desktop 4.27.0](https://docs.docker.com/desktop/release-notes/#4270) ships with the containerd backend storage feature in GA. When activating this feature, vagrant docker provider fails building vms from Dockerfile.

The log format `docker build` has changed with the new storage backend enabled. Vagrant fails to parse the built image id from the new log format.

### Debug output

```
Bringing machine 'test' up with 'docker' provider...
==> test: Creating and configuring docker networks...
==> test: Building the container from a Dockerfile...
    test: #0 building with "desktop-linux" instance using docker driver
    test: 
    test: #1 [internal] load build definition from Dockerfile
    test: #1 transferring dockerfile: 414B done
    test: #1 DONE 0.0s
    test: 
    test: #2 [internal] load metadata for docker.io/library/busybox:latest
    test: #2 ...
    test: 
    test: #3 [auth] library/busybox:pull token for docker-hub-remote.dr.corp.adobe.com
    test: #3 DONE 0.0s
    test: 
    test: #2 [internal] load metadata for docker.io/library/busybox:latest
    test: #2 DONE 1.0s
    test: 
    test: #4 [internal] load .dockerignore
    test: #4 transferring context: 2B done
    test: #4 DONE 0.0s
    test: 
    test: #5 [1/1] FROM docker.io/library/busybox:latest@sha256:6d9ac9237a84afe1516540f40a0fafdc86859b2141954b4d643af7066d598b74
    test: #5 resolve docker.io/library/busybox:latest@sha256:6d9ac9237a84afe1516540f40a0fafdc86859b2141954b4d643af7066d598b74 0.0s done
    test: #5 CACHED
    test: 
    test: #6 exporting to image
    test: #6 exporting layers done
    test: #6 exporting manifest sha256:c01f658213f7ce452a676d83e151941180c922ddddb7372d7ca656cab1febe37 done
    test: #6 exporting config sha256:3e4fd538a9a0b729be05707cf805388be2fb701cfd5d44c6542f1988e8aef6e3 done
    test: #6 exporting attestation manifest sha256:4655203e3ebfc7a5ba35a4b4810abbdeb2567a39e240433a8529dbbf3a95ed2c done
    test: #6 exporting manifest list sha256:5ed6833079fe1dcc7b4c7573cd2dc73349e6a43d4c07e33250a8e9e8d07ac1d8 done
[...]

PR #13343, opened 2024-01-28
Fix#13342 adapt docker provider build for containerd storage

tnaroska:
Fixes https://github.com/hashicorp/vagrant/issues/13342 adding another regex for the docker log format with containerd storage enabled.

Another approach would be to use the `--iidfile` parameter of docker build to have the id written to a file. This would be less brittle than trying to keep up with different docker/podman log outputs. A working prototype is here https://github.com/tnaroska/vagrant/commit/c8073cf440e401b35ce0a4ce78bd73e8ac33b7c7 This would require more elaborate changes in the unit tests though.


chrisroberts (approved):
Thanks!

PR #13358, opened 2024-02-28
Add support for configuring amazon linux 2023 network via systemd

peckpeck:
Amazon linux 2023 is configured through systemd. However amazon linux guest in vagrant inherits from redhat, which is configured through network scripts (which was valid on amazon linux 2).

So this PR adds support for systemd networkd for AL2023 while keeping compatibility with network-script for AL2.


peckpeck:
does anyone have an idea about this ?

PR #13362, opened 2024-03-09
Adding warning for IPv6 ending with 1

chadek:
After struggling quite some time I figured out I had an issue caused by setting an IPv6 ending with :1 

The network_ip_ends_in_one warning message is use for IPv4 to warn user about this matter, I would like to add this warning to IPv6 to avoid some painful debug time to future IPv6 users


chadek:
Will this MR go further?

chrisroberts (approved):
Thanks!

PR #13383, opened 2024-04-22
NFS Folder Mounting message

ALameLlama:
When mounting lots of NFS folders it would be good to get some sort of feedback to know what it's doing.

This is the same as the vboxsf message:
https://github.com/hashicorp/vagrant/blob/main/plugins/providers/virtualbox/synced_folder.rb#L43-L5
https://github.com/hashicorp/vagrant/blob/main/templates/locales/en.yml#L2704


ALameLlama:
I was playing around with SMB and noticed it does src => dest which I think makes more sense so I've updated both to match SMB

chrisroberts (approved):
Looks great, thanks!

PR #13394, opened 2024-05-20
virtualbox version detection: handle WARNING lines

paulcacheux:
As reported in https://www.virtualbox.org/ticket/22060, starting with Virtualbox 7.0.16 running:
```
$ vboxmanage --version
WARNING: Environment variable LOGNAME or USER does not correspond to effective user id.
7.0.18r162988
```
the warning line is written to stdout which is an issue since the parsing of the version is used in the provider code to use the correct class to handle following communication with Virtualbox.

This PR improves the version detection by parsing the version in the first line that does not start with `WARNING:`.

### DIFF
--- a/plugins/providers/virtualbox/driver/meta.rb
+++ b/plugins/providers/virtualbox/driver/meta.rb
@@ -197,7 +197,11 @@ module VagrantPlugins
             end
           end
 
-          parts = output.split("_")
+          version_line = output.each_line.find do |line|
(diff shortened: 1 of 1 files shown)


paulcacheux:
@chrisroberts would you mind taking a look at this? Thanks a lot 👍

chrisroberts (approved):
Thanks!

PR #13409, opened 2024-06-10
fix openwrt nfs synched folders

rgl:
in openwrt, /sbin/init --version does not print the version, instead it calls init, which breaks the running system

this was already fixed in https://github.com/hashicorp/vagrant/pull/7813 but somehow was lost in later code changes


chrisroberts (approved):
Looks like it got missed during some refactoring. Thanks for the PR to get this fixed back up!

PR #13411, opened 2024-06-14
Make installation of NFS client on Debian non-interactive to prevent the process from being stuck

koberbe:
…


chrisroberts (approved):
Thanks!

PR #13465, opened 2024-06-24
Add entry for vagrant-registry packer post-processor

chrisroberts:


website/content/vagrant-cloud/hcp-vagrant/post-migration-guide.mdx
```diff
@@ -70,3 +70,9 @@ If all organizations have been migrated to HCP Vagrant Registry, only the HCP ac
 ```
 export VAGRANT_CLOUD_TOKEN=";$(hcp auth print-access-token)"
 ```
+
+# Packer
+
+After migrating to HCP box uploads may fail when using the [vagrant-cloud](/packer/integrations/hashicorp/vagrant/latest/components/post-processor/vagrant-cloud) post-processor. This is due to the HCP access token used in the composite token expiring prior to the post-processor being executed.
```
allisonlarson, line 76: Feel free to disregard/I'm not sure this is right, but this felt slightly confusing on first read without a comma.

```suggestion
After migrating to HCP, box uploads may fail when using the [vagrant-cloud](/packer/integrations/hashicorp/vagrant/latest/components/post-processor/vagrant-cloud) post-processor. This is due to the HCP access token used in the composite token expiring prior to the post-processor being executed.
```

allisonlarson (approved):
I think this all sounds good. My one comma suggestion isn't blocking

Issue #13470, opened 2024-07-02
Checksum type comparison is case-sensitive


calligraf0:
Not sure this is a widespred problem, or a problem at all (might be intended behavior).
I couldn't see any other open issue about it.

I worked on a possible fix, I'm opening a PR. 
Spoiler alert: I am not too familiar with ruby :) so if I use any unsafe or wrong/deprecated functions: any feedback is welcome!

### Debug output

```
vagrant up dc --provision
Bringing machine 'dc' up with 'vmware_desktop' provider...
==> dc: Box 'gusztavvargadr/windows-server-core' could not be found. Attempting to find and install...
    dc: Box Provider: vmware_desktop, vmware_fusion, vmware_workstation
    dc: Box Version: >= 0
5==> dc: Loading metadata for box 'gusztavvargadr/windows-server-core'
    dc: URL: https://vagrantcloud.com/api/v2/vagrant/gusztavvargadr/windows-server-core
==> dc: Adding box 'gusztavvargadr/windows-server-core' (v2102.0.2404) for provider: vmware_desktop (amd64)
    dc: Downloading: https://api.hashicorp.cloud/vagrant/2022-08-01/gusztavvargadr/boxes/windows-server-core/versions/2102.0.2404/providers/vmware_desktop/amd64/vagrant.box
    dc: Calculating and comparing box checksum...
The specified checksum type is not supported by Vagrant: SHA256.
Vagrant supports the following checksum types:

md5, sha1, sha256, sha384, sha512
```

### Expected behavior

Comparison should work, regardless of case, since choices are limited to supported checksums anyway.

### Actual behavior

While comparing a box checksum vagrant maps the checksum type against this map:
```ruby
CHECKSUM_MAP = {
    :md5 => Digest::MD5,
    :sha1 => Digest::SHA1,
    :sha256 => Digest::SHA256,
    :sha384 => Digest::SHA384,
    :sha512 => Digest::SHA512
  }.freeze
```

this makes any comparison with a box which specifies the checksum type in UPPERCASE fail to map.

#### Vagrant version
2.4.1 

#### Host operating system

Archlinux

#### Guest operating system

Windows, but shouldn't matter.

#### Steps to reproduce

1. Try to download the following box: `gusztavvargadr/windows-server`
2. Fails checksum regardless of type

#### Vagrantfile

```ruby
Vagrant.configure('2') do |config|
  # Vagrant allowed IP ranges: 192.168.56.0/21
[...]

PR #13471, opened 2024-07-02
possible fix for #13470

calligraf0:
As the title says.

Fix is quite simple, forces the `type` to be downcased while getting it from `CHECKSUM_MAP`.
Let me know if I need to fix anything and if it's acceptable or needs more work!


chrisroberts (approved):

chrisroberts:
Just a note that this has been fixed on the server side, however this change does provide better resiliency for issues like this. Thanks for the PR!

PR #13490, opened 2024-08-20
Use `nfsd update` command for restart 

### LINKED ISSUE #13364: synced_folder fails for NFS on Sonoma 14.4
### Debug output

https://gist.github.com/dfreudenberger/7e894870dc739cba7876fef4983c8508

### Expected behavior

synced_folder should mount the specified path using NFS.

### Actual behavior

[Apple decided](https://developer.apple.com/documentation/macos-release-notes/macos-14_4-release-notes#Known-Issues) to limit the access to kickstart which is currently used to restart the NFS daemon. Therefore the mount operation fails. 

Edit:
It seems like Vagrant doesn't actually use kickstart by itself; instead, it runs "sudo nfsd restart" which causes the error mentioned. Perhaps modifying [this line](https://github.com/hashicorp/vagrant/blob/d8fdc500b76c840cdeaa69869d0c000530b036b3/plugins/hosts/bsd/cap/nfs.rb#L163) to use `update` instead of `restart` could solve the problem.

```bash
sudo nfsd restart
Could not kickstart service "com.apple.nfsd": 1: Operation not permitted
/bin/launchctl exited with status 1
```

#### Vagrant version

Vagrant 2.4.1

#### Host operating system

MacOS Sonoma 14.4

#### Guest operating system

Fedora 

#### Steps to reproduce

1. vagrant up

#### Vagrantfile

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

    config.vm.box = "bento/fedora-latest"

    config.vm.provider "parallels" do |prl|
        prl.update_guest_tools = false
    end

    config.vm.synced_folder '~/dev', "/data", :type => 'nfs', :nfs_version => 3
end
```

allisonlarson:
Updates the `nfs_restart_command` for darwin hosts to use `update` instead of `restart`, after mac Sonoma update. Also adjusts the nfs mount log to run within the block, so it can access the opts hash. 

fixes #13364

### DIFF
--- a/plugins/hosts/bsd/cap/nfs.rb
+++ b/plugins/hosts/bsd/cap/nfs.rb
@@ -13,7 +13,7 @@ module VagrantPlugins
       class NFS
         def self.nfs_export(environment, ui, id, ips, folders)
           nfs_exports_template = environment.host.capability(:nfs_exports_template)
-          nfs_restart_command  = environment.host.capability(:nfs_restart_command)
+          nfs_update_command  = environment.host.capability(:nfs_update_command)
           logger = Log4r::Logger.new("vagrant::hosts::bsd")
 
           nfs_checkexports! if File.file?("/etc/exports")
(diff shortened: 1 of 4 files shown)


chrisroberts (changes requested):
Since the `nfs_restart_command` capability is defined, and the export wants to just do an update, it would be better to add a new `nfs_update_command` capability here and then update the `nfs_export` to call that.

chrisroberts (approved):
This looks great!

PR #13496, opened 2024-08-29
Update provider/virtualbox to allow paused state when booting vm

### LINKED ISSUE #13410: Allow Virtualbox provider to tolerate a "paused" state for brief period of time

So, I have a Ubuntu-based, virtualbox vagrant box that I am trying to successfully deploy, but have been having some intermittent issues that are cropping up.  I should note that this vagrant box is deployed headless and does not initially display the GUI for the VM, but, in order to attempt to troubleshoot why Vagrant intermittently times out when attempting to connect to the box via ssh, I will open the GUI for vm instance that vagrant is attempting to deploy to see if I can spot any notable behavior:

image

However, it is here where we run into issues.  If I want to close the GUI for the vm instance again, attempting to close the GUI will result in the below prompt: 

image

As you can see, one of the options will allow the virtualbox to continue running the vm in the background, but the vagrant deployment process has already failed at this point.  The error vagrant is producing is as follows:

```
The guest machine entered an invalid state while waiting for it to boot. Valid states are 'starting, running'. The machine is in the 'paused' state. Please verify everything is configured properly and try again.
```
The error message even goes on to ironically suggest that the user open the GUI of the deployed machine to see if any helpful error messages can be observed.  It would appear that the simple act of attempting to close the gui (which is what brings up the menu seen in the above picture) is the action that is putting the deployed vm in a paused state, and I wanted to suggest that vagrant be a little more tolerant of vms that are in a paused state (maybe only error out if the vm is detected to have been in a paused state for over 5 seconds or something) to allow user's to close the GUI to their vagrant vm instance without vagrant erroring out.

It would appear that the simple act of attempting to close the gui (which is what brings up the menu seen in the above picture) is the action that is putting the deployed vm in a paused state, and I wanted to suggest that vagrant be a little more tolerant of vms that are in a paused state (maybe only error out if the vm is detected to have been in a paused state for over 5 seconds or something) to allow user's to close the GUI to their vagrant vm instance without vagrant erroring out.

None

**Additional context**

image

allisonlarson:
When a virtualbox vm is booting initially, allow it to enter the `:paused` state and wait until a new state is entered or the boot_timeout is reached. This allows the GUI menu to be accessed and state selected before it errors, but to still error as expected if the outcome of the selection moves the vm to an invalid state. This matches how previous versions of virtualbox behaved, before the `:paused` state was introduced.

Also adds some simple tests for `WaitForCommunicator`, which came up as I was investigating different ways to address the issue

fixes #13410

### DIFF
--- a/test/unit/vagrant/action/builtin/wait_for_communicator_test.rb
+++ b/test/unit/vagrant/action/builtin/wait_for_communicator_test.rb
@@ -1,0 +1,58 @@
+require File.expand_path("../../../../base", __FILE__)
+
+describe Vagrant::Action::Builtin::WaitForCommunicator do
+  let(:app) { lambda { |env| } }
+  let(:ui) { lambda { |env| } }
+  let(:env) { { machine: machine, ui: ui } }
+
+  let(:vm) do
+    double("vm",
+      communicator: nil
+    )
+  end
+
+  # Configuration mock
+  let(:config) { double("config", vm: vm) }
+
+  # Communicate mock
+  let(:communicate) { double("communicate") }
+
+  let(:state) { double("state") }
+
+  let(:ui) { Vagrant::UI::Silent.new }
+
+  let(:machine) do
+    double("machine",
+      config: config,
+      communicate: communicate, 
+      state: state,)
+  end
+
+  before do 
+    allow(vm).to receive(:boot_timeout).and_return(1)
+    allow(communicate).to receive(:wait_for_ready).with(1).and_return(true)
+  end
+
+  it "raise an error if a bad state is encountered" do
+    allow(state).to receive(:id).and_return(:stopped)
+    
+    expect { described_class.new(app, env, [:running]).call(env) }.
+    to raise_error(Vagrant::Errors::VMBootBadState)
+  end
+
+  it "raise an error if the vm doesn't boot" do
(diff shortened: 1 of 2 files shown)


test/unit/vagrant/action/builtin/wait_for_communicator_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    expect { described_class.new(app, env, [:running]).call(env) }.
+    to raise_error(Vagrant::Errors::VMBootTimeout)
+  end
+
+  it "succeed when a valid state is encountered" do
+    allow(communicate).to receive(:wait_for_ready).and_return(true)
+    allow(state).to receive(:id).and_return(:running)
+    
+    expect { described_class.new(app, env, [:running]).call(env) }.
+    to_not raise_error()
```
chrisroberts, line 56: Just a small style nit, would you remove the parens on the `raise_error` here since there are no parameters being passed.

chrisroberts (changes requested):
This change looks great. Just one style nit in the tests, and I'll approve after since they get auto-dismissed on push.

chrisroberts (approved):
Thanks!

nurrony:
https://github.com/hashicorp/vagrant/pull/13496

PR #13499, opened 2024-09-06
Add logger and ostruct to the gemspec

hswong3i:
logger and ostruct will no longer be the default gem in the Ruby 3.5, without this patch following warning message may come from ruby 3.3+

```
/opt/vagrant/lib/vagrant/patches/net-ssh.rb:4: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
/usr/lib/ruby/3.3.0/json/common.rb:3: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
```

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -16,6 +16,8 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
 
   s.add_dependency "base64", "~> 0.2.0"
+  s.add_dependency "logger", "~> 1.6.1"
+  s.add_dependency "ostruct", "~> 0.1.0"
   s.add_dependency "bcrypt_pbkdf", "~> 1.1"
   s.add_dependency "childprocess", "~> 4.1.0"
   s.add_dependency "ed25519", "~> 1.3.0"


chrisroberts (approved):

chrisroberts:
Thanks!

vagrant.gemspec
```diff
@@ -16,6 +16,8 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
 
   s.add_dependency "base64", "~> 0.2.0"
+  s.add_dependency "logger", "~> 1.6.1"
```
brlin-tw, line 19: This line should probably be changed to `~>1.0` or `~>1.6`, refer to #13527 for more info.

hswong3i: Or say ~> 1.6.0?

brlin-tw: I don't speak Ruby so I'm not sure (shrugs), I'll leave the judgement to you.

PR #13532, opened 2024-11-04
Fix overly strict dependency requirement of the logger gem to allow plugin installation

### LINKED ISSUE #13527: Vagrant plugins cannot be installed anymore for vagrant 2.4.2 under Windows
In the **Debug output** and **Steps to reproduce** `vagrant-env` plugin is used, but that happened to other plugins I am using (`vagrant-disksize` and `vagrant-vbguest`) as well as picking some random plugins from [here](https://github.com/hashicorp/vagrant/wiki/Available-Vagrant-Plugins).

It seems that adding the dependency of logger ([this PR](https://github.com/hashicorp/vagrant/pull/13499)) breaks the installation of (probably) all plugins under Windows.

When opening a `cmd` in `C:\Program Files\Vagrant\embedded\mingw64\bin` and run `gem search -be logger`

```
C:\Program Files\Vagrant\embedded\mingw64\bin>gem search -be logger

*** LOCAL GEMS ***

logger (1.6.0)

*** REMOTE GEMS ***

logger (1.6.1)

```

### Debug output

https://gist.github.com/dib42/7113c057c40dac93760afe4b813cb763

### Expected behavior

Vagrant should install the plugin

### Actual behavior

Vagrant does not install the plugin because of a ruby dependency problem

#### Vagrant version

Vagrant 2.4.2

#### Host operating system

Windows 10 22H2

#### Guest operating system

Ubuntu 22.04

#### Steps to reproduce

1. Open powershell
2. Run `vagrant plugin install vagrant-env --verbose --debug-timestamp`

#### Vagrantfile
Just for the bug report requirements, as this is not relevant for the bug

```ruby
# Copy-paste your Vagrantfile here. Remove any sensitive information such as passwords, authentication tokens, or email addresses.
  config.vm.box = "ubuntu/jammy64"
end
```

brlin-tw:
Vagrant version 2.4.2 introduced a patch(512bb74) that sets the logger gem dependency to "~>1.6.1", which is too strict and prevents many Vagrant plugins from being installed.

This patch tones down the dependency specification to mitigate this problem.

Fixes #13527.

Refer-to: Add logger and ostruct to the gemspec by hswong3i · Pull Request #13499 · hashicorp/vagrant <https://github.com/hashicorp/vagrant/pull/13499>

I'm not acquainted with Ruby, RubyGems, and the usage of the `logger` gem in Vagrant, please review.

### DIFF
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
   s.add_dependency "i18n", "~> 1.12"
   s.add_dependency "listen", "~> 3.7"
   s.add_dependency "log4r", "~> 1.1.9", "< 1.1.11"
-  s.add_dependency "logger", "~> 1.6.1"
+  s.add_dependency "logger", "~> 1.6.0"
   s.add_dependency "mime-types", "~> 3.3"
   s.add_dependency "net-ftp", "~> 0.2"
   s.add_dependency "net-ssh", "~> 7.0"
--- a/vagrant.gemspec
+++ b/vagrant.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
   s.add_dependency "i18n", "~> 1.12"
   s.add_dependency "listen", "~> 3.7"
   s.add_dependency "log4r", "~> 1.1.9", "< 1.1.11"
-  s.add_dependency "logger", "~> 1.6.0"
+  s.add_dependency "logger", "~> 1.0"
   s.add_dependency "mime-types", "~> 3.3"
   s.add_dependency "net-ftp", "~> 0.2"
   s.add_dependency "net-ssh", "~> 7.0"


brlin-tw:
I can verify that the patched version works at my end.

pranabdas:
Just curious: on my MacOS system, I do not see all the dependencies listed in this file installed locally (with `gem list --local`). Why is the newly added `logger` looking for locally installed gems? I suppose there should be a better way to handle dependencies, either bundle them together with vagrant or run an installation script to install/upgrade gems without user interaction 🤔

chrisroberts (approved):
Thank you for this PR! Just FYI that I did further relax this constraint.

chrisroberts:
@pranabdas The gems that are installed will not be listed with the `gem` command. Vagrant uses an isolated installation area for gems (two, actually: one for Vagrant's direct dependencies that's in `/opt/vagrant/embedded/gems` and one for plugin installs that's in `$VAGRANT_HOME/gems`) so the output of `gem list` will not include them.

gdbonino:
Is there any way to install plugins with 2.4.2 installed? Or do I necessarily have to wait for the new version of vagrant to be released?

brlin-tw:
@gdbonino 

> Is there any way to install plugins with 2.4.2 installed? Or do I necessarily have to wait for the new version of vagrant to be released?

There's an environment variable that can be set to workaround the problem, check the source issue for nore information.

PR #13576, opened 2025-01-15
Update template link

ccec1t91077:
This PR update the template link.

The old link currently redirect to "https://portal.cloud.hashicorp.com/vagrant/discover/search" which is a 404 not found page, I correct the link url to "https://portal.cloud.hashicorp.com/vagrant/discover".


chrisroberts:
Thanks for this PR! The redirects have been updated so this path is valid. Thanks!

Issue #13575, opened 2025-01-13
Unable to authenticate to HCP using Vagrant


lnussbaum:
Following https://developer.hashicorp.com/vagrant/vagrant-cloud/hcp-vagrant/post-migration-guide , I do:
```
export VAGRANT_CLOUD_TOKEN=";$(hcp auth print-access-token)"
```
but then, I get:
$ vagrant cloud auth login
An unexpected error occurred: Vagrant Cloud request failed - authentication is required
```

it worked in December.

This is different from [issue 13574](https://github.com/hashicorp/vagrant/issues/13574) : I try to login to HCP, not Vagrant Cloud.

### Reproduction information 

#### Vagrant version

Vagrant 2.4.3

#### Host operating system

Linux Debian 12

**Comment by hswong3i** [comment]
I am facing a similar issue, too.

My .gitlab-ci.yml logic for pushing Vagrant box to HCP (see https://gitlab.com/alvistack/vagrant-debian/-/blob/develop/.gitlab-ci.yml?ref_type=heads#L128-177):
```
.script-vagrant-cloud-provider-upload: &script-vagrant-cloud-provider-upload
  - |
    if [[ -n "$CI_COMMIT_TAG" ]] && [[ "$CI_COMMIT_TAG" =~ ^[0-9]+\.[0-9]+\.[0-9]+ ]]; then
      export _VERSION="$CI_COMMIT_TAG"
    elif [[ -n "$CI_COMMIT_BRANCH" ]] && [[ "$CI_COMMIT_BRANCH" =~ master ]]; then
      export _VERSION="$(date +%Y%m%d).0.0"
    else
      exit 0
    fi
    sudo -E hcp auth login --client-id=$HCP_CLIENT_ID --client-secret=$HCP_CLIENT_SECRET
    export _VAGRANT_CLOUD_TOKEN=";$(sudo -E hcp auth print-access-token)"
    cd $CI_PROJECT_DIR/packer/$_BOX-$_PROVIDER
    export _ORG="$CI_PROJECT_NAMESPACE"
    export _CHECKSUM="$(sha256sum output-vagrant/package.box | awk '{ print $1 }')"
    curl -skL \
      --request POST \
      --header "Content-Type: application/json" \
      --header "Authorization: Bearer $_VAGRANT_CLOUD_TOKEN" \
      --data "{ \"box\": { \"username\": \"$_ORG\", \"name\": \"$_BOX\", \"is_private\": false } }" \
[...]

Issue #13574, opened 2025-01-13
Unable to Authenticate with vagrant cloud auth login - “Method Not Allowed” Error


warnyul:
I am unable to authenticate with Vagrant Cloud using the vagrant cloud auth login command. I tried the following methods:

1. Default command:

```bash
vagrant cloud auth login
```

2. Explicitly specifying the server URL:
```bash
VAGRANT_SERVER_URL=https://app.vagrantup.com vagrant cloud auth login
```

In both cases, the command fails with the following error:
```
An unexpected error occurred: Vagrant Cloud request failed - Method Not Allowed
```

### Debug output

[debug.txt](https://github.com/user-attachments/files/18399250/debug.txt)

### Expected behavior

The command should authenticate successfully with Vagrant Cloud.

### Actual behavior

The command fails with the “Method Not Allowed” error.

#### Vagrant version

Vagrant 2.4.3

#### Host operating system

MacOS Sequoia 15.0.1
Chip: Apple M1 Pro

or 

MacOS Sonoma 14.7
Intel Core i9

#### Guest operating system

Windows 11 arm64

#### Steps to reproduce

1. Run `vagrant cloud auth login` in the terminal.
2. Enter the credentials.
3. Answer the token description question with default option.
4. Observe the “Method Not Allowed” error.

allisonlarson:
Hello! 

Just as a bit of background, Vagrant Cloud has been moved to the HCP Vagrant Box Registry, and so the login process has changed. If you haven't already logged in to HCP or are just hearing about it now, [this page](https://developer.hashicorp.com/hcp/docs/vagrant) has more information. Apologies if this is already known!

If have an HCP account, the authentication process doesn't require the `vagrant cloud auth login` command. Instead the `VAGRANT_CLOUD_TOKEN` environment variable set to your HCP access-token will work for authentication using the vagrant cli. 

```
export VAGRANT_CLOUD_TOKEN="$(hcp auth print-access-token)"
```

There is more documentation about authentication [here](https://developer.hashicorp.com/vagrant/vagrant-cloud/hcp-vagrant/post-migration-guide#hcp-access-token), but the important part is that the authentication goes through HCP Vagrant Box Registry now.

Cheers!

Issue #13571, opened 2025-01-10
An unexpected error occurred: Vagrant Cloud request failed - Method Not Allowed


mconigliaro:
This is what happens when I try to download a box:

```
❯ vagrant up
Bringing machine 'default' up with 'vmware_desktop' provider...
==> default: Box 'gutehall/ubuntu24-10' could not be found. Attempting to find and install...
    default: Box Provider: vmware_desktop, vmware_fusion, vmware_workstation
    default: Box Version: 2024.10.11
==> default: Loading metadata for box 'gutehall/ubuntu24-10'
    default: URL: https://vagrantcloud.com/api/v2/vagrant/gutehall/ubuntu24-10
==> default: Adding box 'gutehall/ubuntu24-10' (v2024.10.11) for provider: vmware_desktop (arm64)
    default: Downloading: https://vagrantcloud.com/gutehall/boxes/ubuntu24-10/versions/2024.10.11/providers/vmware_desktop/arm64/vagrant.box
The box failed to unpackage properly. Please verify that the box
file you're trying to add is not corrupted and that enough disk space
is available and then try again.
The output from attempting to unpackage (if any):

bsdtar: Error opening archive: Unrecognized archive format
```

I haven't used Vagrant in years, but after a lot of digging, I think I need to log in to download boxes now. This is what happens when I try that:

```
❯ vagrant cloud auth login
In a moment we will ask for your username and password to HashiCorp's
Vagrant Cloud. After authenticating, we will store an access token locally on
disk. Your login details will be transmitted over a secure connection, and
are never stored on disk locally.

If you do not have an Vagrant Cloud account, sign up at
https://www.vagrantcloud.com

Vagrant Cloud username or email: <redacted>
Password (will be hidden):
Token description (Defaults to "Vagrant login from Mikes-Mac-mini.local"):
An unexpected error occurred: Vagrant Cloud request failed - Method Not Allowed
```

eugen-zwinger-protos:
I have the same issue. Tried to downgrade vagrant first, but had no luck there.

Using Apple silicon MacOS 15.2, reproduced on vagrant 2.4.3 and 2.4.1

anitvam:
I have the same issue on local machine running Manjaro Linux, 
and on Github Action workflows running Ubuntu Lastest. 

Yesterday worked just fine on both of them...

eugen-zwinger-protos:
I've been able to work around it by manually downloading the box files and importing them afterwards with
[...]

PR #13580, opened 2025-01-18
Update from app.vagrantup.com to vagrantcloud.com

hswong3i:
As <https://github.com/hashicorp/vagrant/issues/13571#issuecomment-2589249126> mentioned:

> After further debug, this looks like a change introduced in HCP
> Vagrant Registry after 2025-01-09, that broke Vagrant versions before
> 2.4.0. I could confirm that everything works fine with Vagrant 2.4.0
> (and later), but fails to work with Vagrant 2.3.7 (and earlier). The
> root cause is that Vagrant 2.3.7 uses the equivalent of curl
> --location --verbose -H "Accept: application/json"
> <https://vagrantcloud.com/debian/bookworm64> to fetch the box
> metadata, but this no longer returns a JSON document. Vagrant 2.4.0
> uses <https://vagrantcloud.com/api/v2/vagrant/debian/bookworm64>
> instead.

This PR replace most <https://developer.hashicorp.com/vagrant/vagrant-cloud/api/v2> URL from existing <https://app.vagrantup.com> to <https://vagrantcloud.com>.

Fixes <https://github.com/hashicorp/vagrant/issues/13571> 
Fixes <https://github.com/hashicorp/vagrant/issues/13574> 
Fixes <https://github.com/hashicorp/vagrant/issues/13575>

### DIFF
--- a/internal/core/box_test.go
+++ b/internal/core/box_test.go
@@ -27,7 +27,7 @@ func hashicorpBionicBoxData() *vagrant_server.Box {
 		Version:     "0.0.282",
 		Directory:   "/tmp/boxes",
 		Metadata:    testMetadata,
-		MetadataUrl: "https://app.vagrantup.com/hashicorp/boxes/bionic64.json",
+		MetadataUrl: "https://vagrantcloud.com/hashicorp/boxes/bionic64.json",
 		Name:        "hashicorp/bionic64",
 		LastUpdate:  timestamppb.Now(),
 	}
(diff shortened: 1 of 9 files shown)


Stromweld:
> I wonder if there is a known workaround to login in "cloud" and pull private box?

try `export VAGRANT_CLOUD_TOKEN="$(hcp auth print-access-token)"` and that should work.

allisonlarson:
Hello! Thanks for the PR! This issue was fixed with a different, internal change, and it should be working correctly now with the existing URL. I'll be closing this PR and the linked issues. If this isn't fixed, respond or open a new issue and I'll take a look.

Thank you for the contribution!

PR #13587, opened 2025-01-29
Add VirtioSCSI to the list of supported storage controllers for VirtualBox

### LINKED ISSUE #13586: Enhancement Request: Add VirtioSCSI storage controller to list of supported for Virtualbox
Starting from Virtualbox 7.1 support for arm hosts was officially enable. But there is a limitation on which storage controller can be used. In case of ARM, the only supported storage controller is VirtIO SCSI, but it is not supported by vagrant, making it impossible to use with dedicated boxes.

https://www.virtualbox.org/manual/topics/BasicConcepts.html#settings-storage

> VMs with an Arm architecture have VirtIO SCSI only.

The solution seems to be pretty simple: add VirtioSCSI to the list of supported SCSI controller in `VagrantPlugins::ProviderVirtualBox::Model::StorageController`. I've tested this solution by using following monkey-patch

```
class VagrantPlugins::ProviderVirtualBox::Model::StorageController
    SCSI_CONTROLLER_TYPES = ["LsiLogic", "BusLogic", "VirtioSCSI"].map(&:freeze).freeze
end
```

Note: VirtIO and VirtIO SCSI controllers are different, so this ticket is not the same as #13506

betalb:
This minor change, adds virtio-scsi storage controller to the list of supported. closes #13586

Also it changes formatting of exception

From
> Vagrant was unable to detect a disk controller with any of the
> following supported types:
> 
> **IntelAhci ,PIIX4 ,PIIX3 ,ICH6 ,LsiLogic ,BusLogic**
> 
> Please add one of the supported controller types in order to use the
> disk configuration feature.

To

> ...
> IntelAhci, PIIX4, PIIX3, ICH6, LsiLogic, BusLogic
> ...

### DIFF
--- a/plugins/providers/virtualbox/model/storage_controller.rb
+++ b/plugins/providers/virtualbox/model/storage_controller.rb
@@ -9,7 +9,7 @@ module VagrantPlugins
       class StorageController
         IDE_CONTROLLER_TYPES = ["PIIX4", "PIIX3", "ICH6"].map(&:freeze).freeze
         SATA_CONTROLLER_TYPES = ["IntelAhci"].map(&:freeze).freeze
-        SCSI_CONTROLLER_TYPES = [ "LsiLogic", "BusLogic"].map(&:freeze).freeze
+        SCSI_CONTROLLER_TYPES = ["LsiLogic", "BusLogic", "VirtioSCSI"].map(&:freeze).freeze
 
         IDE_DEVICES_PER_PORT = 2.freeze
         SATA_DEVICES_PER_PORT = 1.freeze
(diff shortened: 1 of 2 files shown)


betalb:
If some additional work is required, I would be glad to help. Initially I thought that unit test can be added, but looking at existing test it seems that it doesn't verify all potential values for controller types, so I've dropped this intent.

Also probably important note is that virtio-scsi support was added in version 6.1.0 (released December 10 2019), but I can't see any provisions in model to support versions.

chrisroberts (approved):
Thanks

PR #13618, opened 2025-03-15
[vagrant-vmware-utility.mdx] Added download link for macOS

SamuelMarks:
…

It was unclear how to get vmware working on macOS until I found this dmg link and installed it

### DIFF
--- a/website/content/docs/providers/vmware/vagrant-vmware-utility.mdx
+++ b/website/content/docs/providers/vmware/vagrant-vmware-utility.mdx
@@ -19,8 +19,15 @@ page.
 ## Manual Installation
 
 If there is no officially supported system package of the utility available,
-it may be possible to manually install utility. This applies to Linux platforms
-only. First, download the latest zip package from the releases page.
+it may be possible to manually install utility.
+
+### macOS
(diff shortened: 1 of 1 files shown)


website/content/docs/providers/vmware/vagrant-vmware-utility.mdx
```diff
@@ -19,8 +19,15 @@ page.
 ## Manual Installation
 
 If there is no officially supported system package of the utility available,
-it may be possible to manually install utility. This applies to Linux platforms
-only. First, download the latest zip package from the releases page.
+it may be possible to manually install utility.
+
+### macOS
+Choose the latest version from https://releases.hashicorp.com/vagrant-vmware-utility,
```
cafferata, line 25: Hey @SamuelMarks — thanks for adding this, very helpful!

Do you happen to use Homebrew? If so, there’s also the option to install the Vagrant VMware Utility directly via:

```shell
brew install vagrant-vmware-utility
```

Here’s the direct link to the Homebrew Cask:
https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/v/vagrant-vmware-utility.rb

Might be a convenient alternative for macOS users who manage their tooling with Homebrew.

PR #13628, opened 2025-03-21
Provide configurable retries for connect on SSH communicator

chrisroberts:
Adds two new options to the SSH connect configuration: `connect_retries`
and `connect_retry_delay`. Provides user configurable values used when
establishing the SSH connection. Previous behavior would retry generally
by the default value without a pause between attempts. Updated behavior
will retry the number of times set within the config (unless value is
provided directly to connect call) and each retry will pause based on
the delay value set in the config (unless value is provided directly
to the connect call).

Fixes #13047


briancain (approved):
♻️

PR #13670, opened 2025-05-21
Fix hyperv import failure due to lack of resources

### LINKED ISSUE #12180: vagrant up with Hyper-V fails when box was packaged from a VM that has more resources than available on current Machine.
When I package a Hyper-V Box on a machine with lots of resources (e.g. 50GB memory, 20 CPUs), I am unable to launch the instance on another machine with fewer resources, even if local resources are specified in the vagrantfile.

It seems that Vagrant packages the Hyper-V *.vcmx file in the box and uses that to launch the VM instance on the new machine. Vagrant subsequently modifies the Hyper-V resources according to the vagrantfile. This works great on the original machine where it can get past step 1, but on the new machine with fewer resources it cant get past step 1 and fails to launch the VM. I think the appropriate way to handle this would be to edit the *.vcmx file with the resource constraints before launching it. 

Workaround for now is to just build the box with those resource constraints in mind.

### Vagrant version
2.2.14

### Host operating system
Windows10 20H2

### Guest operating system
Windows10 20H2

###Vagrantfile

  ...
  config.vm.provider "hyperv" do |hyperv|
    hyperv.memory = memory * 1024
    hyperv.maxmemory = max_memory * 1024
    hyperv.cpus = cpus
  end

end

### Error
INFO interface: error: An error occurred while executing a PowerShell script. This error
is shown below. Please read the error message and see if this is
a configuration error with your system. If it is not, then please
report a bug.

Script: import_vm.ps1
Error:

[
    "Memory virtual quantity above maximum.",
    "Cannot assign the specified number of processors for virtual machine \u0027Development Environment Version 1.2.9\u0027 is out of range. The range is 1 through 4. (Virtual machine ID 420E68AF-BEBC-4762-B68F-3B141F171582)"
]

### Expected behavior

Vagrant up launches a virtual machine with the Hyper-V provider with the resources specified in the vagrantfile.

### Actual behavior

Vagrant up fails with above error message

### Steps to reproduce

1. Package vagrant box from Hyper-V VM with 50GB RAM on system with 64GB RAM.
2. Transfer box to machine with 32GB RAM and specify in vagrantfile 16GB RAM to be used.

chrisroberts:
When importing a box that was built with more resources than is
available on the current system, import would fail. If the memory
and/or cpu values are set in the Vagrantfile configuration, provide
them when importing the guest. This allows the values to be modified
prior to registering the new guest preventing a resource unavailability
error.

Fixes #12180

### DIFF
--- a/plugins/providers/hyperv/action/import.rb
+++ b/plugins/providers/hyperv/action/import.rb
@@ -70,8 +70,10 @@ module VagrantPlugins
             "LinkedClone" => !!env[:machine].provider_config.linked_clone,
             "SourcePath" => Vagrant::Util::Platform.wsl_to_windows_path(image_path).gsub("/", "\\"),
             "VMName" => env[:machine].provider_config.vmname,
+            "Memory" => env[:machine].provider_config.memory,
+            "MaxMemory" => env[:machine].provider_config.maxmemory,
+            "Processors" => env[:machine].provider_config.cpus,
           }
-
 
           env[:ui].detail("Creating and registering the VM...")
           server = env[:machine].provider.driver.import(options)
--- a/plugins/providers/hyperv/scripts/import_vm.ps1
+++ b/plugins/providers/hyperv/scripts/import_vm.ps1
@@ -15,6 +15,12 @@ param(
     [parameter (Mandatory=$false)]
     [switch] $LinkedClone,
     [parameter (Mandatory=$false)]
+    [int] $Memory = $null,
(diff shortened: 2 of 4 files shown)


zhindes:
@chrisroberts - this seems to have broken hyperv xml import because `New-VagrantVM` forwards to `New-VagrantVMXML` using `@PSBoundParameters`, but your new parameters (`$Memory` etc) weren't added to the XML-specific version of the function.

PR #13674, opened 2025-05-22
Add resource parameters for hyperv xml import

chrisroberts:
The generic import cmdlet passes all received arguments to
the specialized import cmdlets. The resource parameters for
providing memory and cpu values were added to the vmcx based
cmdlet but not the xml based cmdlet. This adds the to the
xml based cmdlet and includes configuration adjustment if set.

Thanks to @zhindes for [surfacing this oversight](https://github.com/hashicorp/vagrant/pull/13670#issuecomment-2902276401)

### DIFF
--- a/plugins/providers/hyperv/scripts/utils/VagrantVM/VagrantVM.psm1
+++ b/plugins/providers/hyperv/scripts/utils/VagrantVM/VagrantVM.psm1
@@ -224,6 +224,12 @@ function New-VagrantVMXML {
         [string] $SourcePath,
         [parameter (Mandatory=$false)]
         [bool] $LinkedClone = $false,
+        [parameter (Mandatory=$false)]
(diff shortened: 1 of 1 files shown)


zhindes:
> The generic import cmdlet passes all received arguments to the specialized import cmdlets. The resource parameters for providing memory and cpu values were added to the vmcx based cmdlet but not the xml based cmdlet. This adds the to the xml based cmdlet and includes configuration adjustment if set.
> 
> Thanks to @zhindes for [surfacing this oversight](https://github.com/hashicorp/vagrant/pull/13670#issuecomment-2902276401)

No, thank you! I was in a hurry and didn't have much time to figure out the right way to engage the community, create an issue, or much less create my own fix. I was hopeful that a quick ping would suffice since it was a very recent regression. Very quick turnaround, nice job :)

PR #13678, opened 2025-05-29
Ensure auto-assigned container name starts with valid characters

### LINKED ISSUE #13484: Invalid auto-generated container name using Docker provider and a folder that starts with an underscore
Running `vagrant up` in a folder that begins with an underscore (eg, `_foo`) using the `docker` provider that does not explicitely set the `name` attribute results in an `Invalid container name` error.

### Debug output

It's easy to see the problem so I don't think the debug output is needed since the error lies in the auto container name code of the docker provider.

### Expected behavior

It was expected for `vagrant up` to succeed and bring up the docker container with an auto-generated name.

### Actual behavior

```
A Docker command executed by Vagrant didn't complete successfully!
The command run along with the output from the command is shown
below.

Command: ["docker", "run", "--name", "_foo_default_1722722989", "-d", "-v", ".../_foo:/vagrant", "57ed1512823d6c273d82fe84bf974a3f7b833c2065787e655ae9746719f99866", {:notify=>[:stdout, :stderr]}]

Stderr: docker: Error response from daemon: Invalid container name (_foo_default_1722722989), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed.
See 'docker run --help'.
```

#### Vagrant version

```
$ vagrant -v
Vagrant 2.4.1
```

#### Host operating system

macOs Sonoma 14.4.1

#### Guest operating system

Ubuntu 20.04

#### Steps to reproduce

1. Create a folder that starts with an underscore (eg `_foo`)
2. Create a minimal Dockerfile in that directory (eg `_foo/Dockerfile`:
    ```
    FROM ubuntu:20.04
    CMD ["sleep", "infinity"]
    ```
4. Create a minimal Vagrantfile in that directory (eg `_foo/Vagrantfile`) (see the Vagrantfile section for the contents)

You can fix the error and successfully bring up the docker container by adding to the Vagrantfile's `docker` config block:

```
d.name = "foo"
```

Which bypasses the auto name generating code.

#### Vagrantfile

```ruby

  config.vm.provider :docker do |d|
    d.build_dir = "."
    d.remains_running = true
  end

end
```

taru-garg-hashicorp:
Many container engines, including Docker and Podman, require container names to begin with an alphanumeric character (`[a-zA-Z0-9]`). In the current Docker provider implementation, if a container name is not explicitly provided in the user configuration, Vagrant automatically generates one based on the directory path.

However, if the directory name starts with an invalid character such as `_`, `-` , or any non-alphanumeric symbol, the generated container name becomes invalid (characters other than `-` `_` are currently dropped though), causing vagrant up to fail due to container engine constraints.

This change addresses the issue by dropping any leading `_`, `-` from the generated name when it would otherwise start with an invalid character.

Fixes #13484

### DIFF
--- a/test/unit/plugins/providers/docker/action/create_test.rb
+++ b/test/unit/plugins/providers/docker/action/create_test.rb
@@ -16,7 +16,6 @@ describe VagrantPlugins::DockerProvider::Action::Create do
     sandbox.create_vagrant_env
   end
 
-  let(:machine_name) {"test-machine"}
   let(:machine) do
     iso_env.machine(iso_env.machine_names[0], :virtualbox).tap do |m|
       allow(m.provider).to receive(:driver).and_return(driver)


plugins/providers/docker/action/create.rb
```diff
@@ -91,9 +92,7 @@ def call(env)
         def create_params
           container_name = @provider_config.name
           if !container_name
-            container_name = "#{@env[:root_path].basename.to_s}_#{@machine.name}"
-            container_name.gsub!(/[^-a-z0-9_]/i, "")
-            container_name << "_#{Time.now.to_i}"
+            container_name = container_name()
```
chrisroberts, line 95: Method call with no arguments should not include parenthesis.

plugins/providers/docker/action/create.rb
```diff
@@ -156,6 +155,16 @@ def forwarded_ports(include_ssh=false)
 
           result
         end
+
+        def container_name
```
chrisroberts, line 159: Provide a bit more information in the method name, for example: `generate_container_name`. This will have the added benefit in this changeset of not masking the method name above where it is used.

plugins/providers/docker/action/create.rb
```diff
@@ -156,6 +155,16 @@ def forwarded_ports(include_ssh=false)
...
+          base = @env[:root_path].basename.to_s
+          base = base.sub(/^[^a-zA-Z0-9]/, VAGRANTCONTAINER)
```
chrisroberts, line 161: Considering that the timestamp is appended to the container name below, collisions won't be a concern here. This means that instead of replacing an invalid first character, it can just be removed (the same as the invalid characters being removed from the full container name in the `gsub!` below).

chrisroberts (changes requested):
This looks good. I left a few notes. Since the generated name has a timestamp appended and isn't just the string, there's less worry of collisions so the replacement on the first character isn't really needed. 

Some test coverage should be added as well. Looking at the current tests for this action, they are very small. But the new method should include tests at least.

test/unit/plugins/providers/docker/action/create_test.rb
```diff
@@ -55,4 +55,16 @@
...
+  describe "#generate_container_name" do
+    it "generates a unique container name" do
+      subject.instance_variable_set(:@env, { root_path: Pathname.new("/path/to/__--root-test-1_1--") })
+      subject.instance_variable_set(:@machine, machine)
+
+      name = subject.generate_container_name
+
+      expect(name).to match(/^root-test-1_1--_.*$/)
+    end
+  end
```
chrisroberts, line 68: The tests here should include variations of names that include a name with no characters removed, a name with prefix characters removed, a name with non-prefix characters removed, and a name with prefix and non-prefix characters removed. The tests should exercise the various different ways the method is expected to modify the generated name. This can be done with some setup helpers and look something like this:

```ruby
describe "#generate_container_name" do
  let(:machine_name) { "test-machine" }
  let(:env) { {root_path: root_path}
  let(:root_path) { Pathname.new("/path/to/project") }

  before do
    subject.instance_variable_set(:@env, env)
    allow(machine).to receive(:name).and_return(machine_name)
  end

  it "should not remove any characters" do 
    expect(subject.generate_container_name).to start_with("project_test-machine")
  end

  context "when root path starts with underscores characters" do
    let(:root_path) { Pathname.new("/path/to/__project") }

    it "should remove underscores" do
      expect(subject.generate_container_name).to start_with("project_test-machine")
    end
  end

  context "when root path starts with dashes" do
    let(:root_path) { Pathname.new("/path/to/--project") }

    it "should remove dashes" do 
      expect(subject.generate_container_name).to start_with("project_test-machine")
    end
  end

  ...
end
```

chrisroberts (changes requested):
This change looks great. Left a note on the test. The original test looks good but some additional tests to exercise the expected behavior should be added..

test/unit/plugins/providers/docker/action/create_test.rb
taru-garg-hashicorp: Added tests and changed the sanitization logic

chrisroberts (changes requested):

chrisroberts (approved):
Looks great!

PR #13690, opened 2025-07-03
Provision action trigger ordering with provision action

anshulsharma-hashicorp:


website/content/docs/triggers/usage.mdx
```diff
@@ -196,3 +196,46 @@ config.trigger.before :"Vagrant::Action::Builtin::GracefulHalt", type: :action d
...
 ```
+
+#### Provision action
+
+The provision stack works little different than the other action stacks. The Vagrant
+stack is made up of middlewares that are run in a specific order. generally it tend to
+to perform their operation and call to the next item but in some cases will continue to
+do things once the next item completes. In the Provision action, the next item in the
+stack is executed [here](https://github.com/hashicorp/vagrant/blob/main/lib/vagrant/action/builtin/provision.rb#L82-L83).but the actual provisioning happens at the end [here](https://github.com/hashicorp/vagrant/blob/main/lib/vagrant/action/builtin/provision.rb#L129-L133), after the rest of the stack
+has executed and exited.
```
chrisroberts, line 207: This needs to be updated for grammar and clarity as it's mixing terms.

> The provision action behaves slightly different than other actions. When Vagrant executes its stack, each action performs some function and then calls the next action in the stack. In some cases (like the provision action), the action will perform some function, call the next action, and wait for the it to return to perform some extra functionality or cleanup. This behavior can result in seemingly unexpected results for after triggers since the targeted action will complete after the rest of the actions in the stack have executed.

website/content/docs/triggers/usage.mdx
```diff
@@ -196,3 +196,46 @@ config.trigger.before :"Vagrant::Action::Builtin::GracefulHalt", type: :action d
...
+  config.trigger.after Vagrant::Action::Builtin::Provision, type: :action,
+    name: "ACTION TRIGGER AFTER Vagrant::Action::Builtin::Provision",
+    info: "INSIDE AFTER Vagrant::Action::Builtin::Provision TRIGGER"
+
+  config.trigger.after :machine_action_provision, type: :hook,
+    name: "HOOK TRIGGER AFTER machine_action_provision",
+    info: "INSIDE AFTER machine_action_provision HOOK TRIGGER"
+
+end
+```
```
chrisroberts, line 237: Output of this should be included to display the behavior.

website/content/docs/triggers/usage.mdx
```diff
@@ -196,3 +196,46 @@ config.trigger.before :"Vagrant::Action::Builtin::GracefulHalt", type: :action d
...
+
+When running `vagrant provision` or `vagrant up`, the before and after triggers
+will run before the provisioning of the box happens.
```
chrisroberts, line 240: Is this true?

chrisroberts (changes requested):

Issue #13698, opened 2025-07-30
vagrant box for vmware fusion doesn't come up and has an error


Stromweld:
### Vagrant version

2.4.7

### Vagrant VMware plugin version

3.0.5

### Vagrant VMware utility version

1.0.24

### Host operating system

MacOS 15.5 Apple Silicon M3

### Guest operating system

almalinux-9 arm64

### Vagrantfile

```ruby
Vagrant.configure("2") do |c|
  c.berkshelf.enabled = false if Vagrant.has_plugin?("vagrant-berkshelf")
  c.vm.box = "bento-almalinux-9.6"
  c.vm.box_url = "file:///Users/corey.hemminger/github/personal/chef/cookbooks/testing/bento/builds/almalinux-9.6-aarch64.vmware.box"
  c.vm.box_architecture = "arm64"
  c.vm.hostname = "bento-test"
  c.vm.synced_folder ".", "/vagrant", disabled: true
  c.vm.synced_folder "/Users/corey.hemminger/github/personal/chef/cookbooks/testing/bento/builds/test-kitchen", "/vagrant", disabled: false
  c.vm.provider :vmware_desktop do |p|
  end
end
```

### Debug output

https://gist.github.com/Stromweld/a1992ebe3d9f3847ee860b28b885d0f7

### Expected behavior

box would start up in virtualbox

### Actual behavior

nothing but an error

### Steps to reproduce

1. install bento gem from https://github.com/chef/bento
2. install test-kitchen ruby gem
3. run `bento build -o vmware-iso.vm os_pkrvars/almalinux/almalinux-9-aarch64.pkr.hcl`
4. run `bento test`

taru-garg-2000:
Hi @Stromweld,

From the error it looks like it might be an issue with the box and not Vagrant, I have little knowledge of `bento` so don't know if the steps listed above are complete to have a correct `boxfile` generated. That said I tested the box provided by the bento project for the OS you mentioned on Vagrant Public Registry, for reference [here](https://portal.cloud.hashicorp.com/vagrant/discover/bento/almalinux-9), and it seems to come up properly. Further, the `bento test` command documentation seems to suggest it will run if a correct `boxfile` has been generated, which again leads me to think it might be an issue with the custom box itself.

Cheers!

Stromweld:
[...]

Issue #13697, opened 2025-07-30
vagrant box fails to come up with an error for virtualbox


Stromweld:
### Debug output

https://gist.github.com/Stromweld/45917d24bdaccf12c928db215b515501

### Expected behavior

box comes up normally in virtualbox

### Actual behavior

nothing but an error

#### Vagrant version

2.4.7

#### Host operating system

macos 15.6 apple silicon m3

#### Guest operating system

almalinux-9 arm64

#### Steps to reproduce

1. install bento gem from https://github.com/chef/bento
2. install test-kitchen gem
3. run `bento build -o virtualbox-iso.vm os_pkrvars/almalinux/almalinux-9-aarch64.pkrvars.hcl`
4. run `bento test`

#### Vagrantfile

```ruby
Vagrant.configure("2") do |c|
  c.berkshelf.enabled = false if Vagrant.has_plugin?("vagrant-berkshelf")
  c.vm.box = "bento-almalinux-9.6"
  c.vm.box_url = "file:///Users/corey.hemminger/github/personal/chef/cookbooks/testing/bento/builds/almalinux-9.6-aarch64.virtualbox.box"
  c.vm.box_architecture = "arm64"
  c.vm.hostname = "bento-test"
  c.vm.synced_folder ".", "/vagrant", disabled: true
  c.vm.synced_folder "/Users/corey.hemminger/github/personal/chef/cookbooks/testing/bento/builds/test-kitchen", "/vagrant", disabled: false
  c.vm.provider :virtualbox do |p|
    p.name = "kitchen-test-kitchen-default-almalinux-96-arm64-virtualbox-5f9fafb5-cb3c-4dca-93c9-cd827b88f907"
    p.customize ["modifyvm", :id, "--audio", "none"]
  end
end
```

taru-garg-2000:
This issue should also be fixed via #13699

Stromweld:
tested patch and it fixes this issue

PR #13699, opened 2025-07-31
Update box add action to pass arch correctly

taru-garg-hashicorp:
Currently, during the box add action `env[:architecture]` gets passed down to `add_direct`, however, there is no `architecture` in env, the correct arch which is set in `Vagrantfile` is `box_architecture`, this PR changes the action to pass down the correct parameter

Closes: #13698, #13697


chrisroberts (approved):
:+1:

PR #13704, opened 2025-08-12
fix: retry download without Authorization header on 401 for public boxes

### LINKED ISSUE #13703: Vagrant fails to download public boxes or get metadata when auth token has expired
### Debug output

none

### Expected behavior

fetching of box and bringing up vm regardless of auth status for public boxes

### Actual behavior

fails to fetch metadata or download a box with 401 error

#### Vagrant version

2.4.8

#### Host operating system

MacOS 15.6 M3 MBP

#### Guest operating system

bento/almalinux-9

#### Steps to reproduce

1. Set env variable VAGRANT_CLOUD_TOKEN with expired session auth token
2. Try to bring up any vagrant box not already found on the workstation

#### Vagrantfile

```ruby
Vagrant.configure("2") do |c|
  c.berkshelf.enabled = false if Vagrant.has_plugin?("vagrant-berkshelf")
  c.vm.box = "bento/almalinux-9"
  c.vm.hostname = "default-almalinux-9.vagrantup.com"
  c.vm.synced_folder ".", "/vagrant", disabled: true
  c.vm.synced_folder "/Users/corey.hemminger/.kitchen/cache", "/tmp/omnibus/cache", create: true
  c.vm.provider :virtualbox do |p|
    p.name = "kitchen-test-default-almalinux-9-8f421b51-925f-42ce-a4da-05f5f4dff419"
    p.customize ["modifyvm", :id, "--audio", "none"]
  end
end
```

damacus:
fixes #13703

### DIFF
--- a/lib/vagrant/util/downloader.rb
+++ b/lib/vagrant/util/downloader.rb
@@ -104,6 +104,7 @@ module Vagrant
         @logger.info("  -- Destination: #{@destination}")
 
         retried = false
+        retried_auth = false
         begin
           # Get the command line args and the subprocess opts based
           # on our downloader settings.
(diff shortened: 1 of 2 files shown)


Stromweld:
@taru-garg-hashicorp can we get this merged for the next release?

taru-garg-2000:
Hi there,

Apologies for the delayed response—I’ve been considering the best way to address this issue. Currently, when an `auth` token has expired, `vagrant` continues to send it as a header, which leads to authentication failures. Ideally, public boxes shouldn’t require authentication, but since the expired token is still being sent, the registry attempts to validate it.

There are two main issues here:
- `vagrant` is sending an auth token even when it’s not necessary.
- The vagrant public registry is trying to authenticate every request, even for public boxes.

While this fix will address the immediate problem, I’m concerned it might create new issues for users who logged in previously and now have expired tokens. In the case of public boxes, this could mean two requests per download, which would cause users to hit rate limits more quickly.

One idea I’ve been considering is introducing something like a `--no-auth` flag to the CLI. Most users downloading public boxes will likely know if authentication is required, but adding this flag could disrupt current workflows.

I will discuss these options with the team to determine the best solution. As for the release timeline, there’s no set date for `2.4.10` yet, but if we resolve this beforehand, I’ll make sure it’s included in that release.

Thanks!

Stromweld:
can the vagrant public registry be updated to ignore auth for public boxes?

PR #13709, opened 2025-08-15
Add Vbox 7.2 support

### LINKED ISSUE #13707: Enhancement Request: Add support for VirtualBox 7.2.x
There is no support for VirtualBox 7.2.x, which just came out recently https://www.virtualbox.org/wiki/Changelog-7.2

It looks like the versions of VirtualBox are hardcoded in https://github.com/hashicorp/vagrant/blob/main/plugins/providers/virtualbox/plugin.rb . At a minimum, 7.2 needs to be added to the plugin file and a new 7.2 driver file added as well. There might be more stuff needed, but I'm not too familiar with the codebase.

no alternatives

**Additional context**
none

Stromweld:
Adds virtualbox 7.2 support. Tested on MacOS 15.6 M3 MBP with Vagrant 2.4.8(patched with this update) and Virtualbox 7.2.0.

Fixes #13707

### DIFF
--- a/plugins/providers/virtualbox/driver/version_7_2.rb
+++ b/plugins/providers/virtualbox/driver/version_7_2.rb
@@ -6,7 +6,7 @@ require File.expand_path("../version_7_0", __FILE__)
 module VagrantPlugins
   module ProviderVirtualBox
     module Driver
-      # Driver for VirtualBox 7.1.x
+      # Driver for VirtualBox 7.2.x
       class Version_7_2 < Version_7_1
         def initialize(uuid)
           super


localhorst-org:
@Stromweld,

i checked your patch "manualy" create or update the affected files under
Windows 11 Version 24H2 (Build 26100.4946)
VirtualBox 7.2.0
Vagrant 2.4.8

Worked for me, thx

plugins/providers/virtualbox/driver/version_7_2.rb
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+require File.expand_path("../version_7_0", __FILE__)
+
+module VagrantPlugins
+  module ProviderVirtualBox
+    module Driver
+      # Driver for VirtualBox 7.1.x
```
taru-garg-hashicorp, line 9: [nit]: Can you please update the comment here to say its the driver for version 7.2.x

taru-garg-hashicorp (changes requested):
Hi @Stromweld,

Thanks a lot for the PR, this looks great! I haven't tested Vagrant with the latest release of Virtualbox yet, I will try out with this patch and see if things are looking ok.

Cheers!

plugins/providers/virtualbox/driver/version_7_2.rb
Stromweld: done

Stromweld:
Thanks, I just simply copied the changes from the previous PR for the 7.1 release so I can't take all the credit.

gcrum-bc:
Apologies if this is a stupid question.  This is entirely an opinion not a rant.
Why do we need to jump through all these hoops for a minor version update.  Wouldn't it be easier to have vagrant support 7.x?  This isn't a major update.  Meanwhile people using brew to manage their copy of virtualbox are stuck with removing the brew version (if they updated to 7.2) and installing 7.1 from virtualbox.org, which isn't a big deal.  But it seems like this is just generating cycles that developers really don't need to spend their time on.  Once again I apologize if this is missing the point here.  I do appreciate all the work everyone does here.

taru-garg-2000:
> Apologies if this is a stupid question. This is entirely an opinion not a rant. Why do we need to jump through all these hoops for a minor version update. Wouldn't it be easier to have vagrant support 7.x? This isn't a major update. Meanwhile people using brew to manage their copy of virtualbox are stuck with removing the brew version (if they updated to 7.2) and installing 7.1 from virtualbox.org, which isn't a big deal. But it seems like this is just generating cycles that developers really don't need to spend their time on. Once again I apologize if this is missing the point here. I do appreciate all the work everyone does here.

Hi @gcrum-bc,

I agree with your point—it would definitely be advantageous for vagrant to support 7.x directly. This would eliminate much of the extra work and mean we wouldn’t need to release a new version of vagrant with every VirtualBox update. I’ll need to look into this further to make sure I’m not overlooking anything, so I can’t commit to a timeline for if or when this will be available.

Thanks!

PR #13733, opened 2025-10-18
Replace the deprecated Ansible inventory file option

cavcrosby:
When testing an upgrade to the `ansible-core` package (version `2.19`) via Vagrant, the following warning was noticed:

```shell
INFO| [DEPRECATION WARNING]: The '--inventory-file' argument is deprecated. This feature will be removed from ansible-core version 2.23.
```

A drop-in to the `--inventory-file` option would be the `--inventory` variant, as mentioned [in Ansible's documentation](https://docs.ansible.com/ansible/latest/cli/ansible-playbook.html#cmdoption-ansible-playbook-i).

PR #13740, opened 2025-11-05
Fix tests for #13733

taru-garg-hashicorp:
Fixing test failures in #13733

### DIFF
--- a/plugins/provisioners/ansible/provisioner/base.rb
+++ b/plugins/provisioners/ansible/provisioner/base.rb
@@ -68,10 +68,9 @@ module VagrantPlugins
           begin
             set_gathered_ansible_package_version("ansible-core", gather_ansible_version("ansible-core"))
           rescue StandardError => e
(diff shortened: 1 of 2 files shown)


chrisroberts (dismissed):

plugins/provisioners/ansible/provisioner/base.rb
```diff
@@ -175,6 +176,15 @@ def prepare_common_command_arguments
           @command_arguments << "--start-at-task=#{config.start_at_task}" if config.start_at_task
         end
 
+        def get_inventory_command(inventory_path)
+          # print gathered version
+          if !@gathered_version_major.nil?  && !@gathered_version_minor.nil? && @gathered_version_major.to_i >= 2 && @gathered_version_minor.to_i >= 19
```
chrisroberts: you can drop the not nil checks here since the `.to_i` will force an integer for the check (with `nil.to_i` being `0`).

Issue #13720, opened 2025-09-03
Second 'vagrant up' destroys primary disk on hyper-v if it was resized in Vagrantfile


efussi:
### Expected behavior

Second `vagrant up` succeeds with a resized primary disk.

### Actual behavior

Got a stack trace:
```
Bringing machine 'default' up with 'hyperv' provider...
==> default: Verifying Hyper-V is enabled...
==> default: Verifying Hyper-V is accessible...
    default: Configuring the VM...
    default: Setting VM Enhanced session transport type to disabled/default (VMBus)
==> default: Disk 'vagrant_primary' no longer exists in Vagrant config. Removing and closing medium from guest...
==> default: Configuring storage mediums...
C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/plugins/providers/hyperv/cap/configure_disks.rb:70:in `get_current_disk': undefined method `[]' for nil (NoMethodError)
            current_disk = machine.provider.driver.get_disk(current_disk["Path"])
                                                                        ^^^^^^^^
    from C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/plugins/providers/hyperv/cap/configure_disks.rb:90:in `handle_configure_disk'
    from C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/plugins/providers/hyperv/cap/configure_disks.rb:29:in `block in configure_disks'
    from C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/plugins/providers/hyperv/cap/configure_disks.rb:27:in `each'
    from C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/plugins/providers/hyperv/cap/configure_disks.rb:27:in `configure_disks'
    from C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/lib/vagrant/capability_host.rb:114:in `call'
    from C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/lib/vagrant/capability_host.rb:114:in `capability'
    from C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/lib/vagrant/action/builtin/disk.rb:23:in `call'
    from C:/Program Files/Vagrant/embedded/gems/gems/vagrant-2.4.9/lib/vagrant/action/warden.rb:38:in `call'
[...]

PR #13748, opened 2025-11-22
Retain 'primary' attribute when resizing the Hyper-V primary disk

efussi:
Fix https://github.com/hashicorp/vagrant/issues/13720

The `primary` attribute is set when not resizing the current disk (`else` branch), but it is not set when resizing the current disk:
https://github.com/hashicorp/vagrant/blob/162b7c0d700a0fb244d1549ae8523006f4892eb0/plugins/providers/hyperv/cap/configure_disks.rb#L96-L102

Due to the missing `primary` attribute, a resized primary disk is deleted here upon the next `vagrant up`:
https://github.com/hashicorp/vagrant/blob/162b7c0d700a0fb244d1549ae8523006f4892eb0/plugins/providers/hyperv/cap/cleanup_disks.rb#L37-L48

### DIFF
--- a/plugins/providers/hyperv/cap/configure_disks.rb
+++ b/plugins/providers/hyperv/cap/configure_disks.rb
@@ -239,6 +239,9 @@ module VagrantPlugins
 
           # Store updated metadata
           disk_metadata = {UUID: disk_info["DiskIdentifier"], Name: disk_config.name, Path: disk_info["Path"]}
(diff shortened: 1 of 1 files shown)


efussi:
@chrisroberts @taru-garg-hashicorp apologies for tagging you directly, but can I please get a review for this small fix that helps prevent inadvertent data loss?

PR #13797, opened 2026-03-25
Reload systemd daemon after changing fstab

giner:
Calling `mount` may fail after modifying fstab unless systemd daemon is reloaded.

Relevant error message:
```
mount: (hint) your fstab has been modified, but systemd still uses
       the old version; use 'systemctl daemon-reload' to reload.
```

Refs:
- https://github.com/util-linux/util-linux/commit/1db0715169954a8f3898f7ca9d3902cd6c27084d


giner:
@taru-garg-hashicorp could you help here please?

giner:
@taru-garg-hashicorp friendly ping

giner:
@taru-garg-hashicorp would you take a look at this PR?

